PR #1808, opened 2011-10-27
coffeescript shorthand object notation

paulmars:
salary = 50
name = "Joe"
dept = "Accounting"
employee = {salary, name, dept}
output = "#{employee.name} works in #{employee.dept}"

added to the docs

### DIFF
--- a/documentation/coffee/objects_shorthand.coffee
+++ b/documentation/coffee/objects_shorthand.coffee
@@ -1,0 +1,5 @@
+salary = 50
+name = "Joe"
+dept = "Accounting"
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Has this PR really been sitting open for **five years** just because no one can come up with a more exciting example than Joe in accounting? This PR is only for documentation, not a new feature. This object creation shorthand has actually been adopted by ES2015, and people probably have no idea that we pioneered it, because this never made it into the documentation.

Okay folks now’s your chance. The floodgates are open. Best example wins.

Issue #1494, opened 2011-07-08
Flag to Disable "__" Helper Functions


lucaswoj:
I'm working on a codebase which has to compile many CoffeeScript files on the fly and sends them to the browser. As it stands, I have no choice but to include `function __extends() {...}` in every single file. Surely not a huge performance hit, but I'd love to just have a single file of all CoffeeScript helpers and include that instead.

PR #1820, opened 2011-11-03
Flag to disable helpers during compilation (#1494)

TrevorBurnham:
As discussed at #1494, this pull request allows you to tell `CoffeeScript.compile` that you don't want helpers/utilities like `__hasProp` to show up in the compiled output. That way, if you have a large number of modules that you're compiling for front-end deployment, you can just put the helpers in one file and eliminate them in the other. This was discussed briefly on Twitter today:

http://twitter.com/#!/ryanflorence/status/132122428913614849
http://twitter.com/#!/CoffeeScript/status/132156500364894208
http://twitter.com/#!/jashkenas/status/132161592451268608

Example:

```
coffee> CoffeeScript.compile 'x in arr', bare: true
'var __indexOf = Array.prototype.indexOf || function(item) {\n  for (var i = 0, l = this.length; i < l; i++) {\n    if (this[i] === item) return i;\n  }\n  return -1;\n};\n__indexOf.call(arr, x) >= 0;'
coffee> CoffeeScript.compile 'x in arr', bare: true, utilities: false
'__indexOf.call(arr, x) >= 0;'
```

This does not affect the `coffee` command. It's purely to give other compilation tools more flexibility.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -121,7 +121,7 @@ task 'build:utilities', 'rebuild the utilities file for use with compile(utiliti
     "__#{utility} = #{func()}"
   ).join(',\n') + ';'
   fs.writeFileSync 'extras/utilities.js', code
-  consoole.log "built utilities.js"
+  console.log "built utilities.js"
 
 task 'doc:site', 'watch and continually rebuild the documentation for the website', ->
   exec 'rake doc', (err) ->


samuelhorwitz:
@lydell the size isn't necessarily the issue. With code coverage reporters, for example, you end up with branches that are marked as uncovered despite the Coffeescript not showing any branches. You can see the hack I have in place to solve this here https://github.com/Constellation/ibrik/pull/36

This isn't only about splitting hairs about file size.

DinisCruz:
@lydell  I'm having the same problem has @samuelhorwitz , the inclusion of these extra methods (like the check for indexOf) it creating code paths that my code coverage checks cant reach

image

lydell:
@samuelhorwitz @DinisCruz The use case of this PR (as well as in a few other issues) is to reduce repetition of helper declarations, in order to save bytes when sending compiled JS from server to client. However, my tests have proven that this is unnecessary thanks to gzip.

Thanks for bringing up the issue with code coverage tools, though, because I've not heard and thought about such issues before.

I don't see how this PR would help in the code coverage case, though. Even if the helpers are not duplicated, there can still be unused branches is them. Removing _all_ helpers won't cut it either, since then your program won't run at all -- unless you were planning to inject custom versions of them on your own? Feels pretty hacky.

It sounds like you'd rather be looking for some way to extend or override the helpers, more like what is done in the PR @samuelhorwitz linked to. Perhaps we could add an option to pass in a function that would pre-process the [helper strings](https://github.com/jashkenas/coffeescript/blob/2d1a6fa6ec76b4e721da7e5e4a9c6b90172309ae/src/nodes.coffee#L2230-L2274) (that'd be done on [this line](https://github.com/jashkenas/coffeescript/blob/2d1a6fa6ec76b4e721da7e5e4a9c6b90172309ae/src/nodes.coffee#L2310)). Then you could do regex hacks on it, or parse it using one of the many JavaScript parsers, modify the AST and return in a string using one of the many AST-to-string modules.

jashkenas:
I'd tend to think that points to a fault in the capability of the code coverage tool — better to fix it at the source, instead of contorting your code, no?

samuelhorwitz:
Yeah, that's what this beautiful PR I submitted to the Ibrik team does :P https://github.com/Constellation/ibrik/pull/36/files

I'm not familiar enough with Coffeescript compiler to know what the best way to go about doing what I'm trying to do is, but I'm pretty sure my solution isn't it, though it works

PR #1959, opened 2011-12-21
fixes #1910: loop index should be mutable within a loop iteration and immutable between loop iterations

### LINKED ISSUE #1910: for .. in array.length not getting updated when I reset index
``` coffeescript
for item, i in arr
   arr.splice(0, 1)
   i = 0
  # we get undefined item errors
```

michaelficarra:


senotrusov:
Personally I found a bit odd to add an extra variable to all the loops everywhere in the CoffeScript world just to fight with the rare case when somebody want to have i+=2 somewhere in the loop body.

PR #1996, opened 2012-01-02
add simple startup file support to REPL (.coffeerc)

codelahoma:
When starting REPL, loads and runs the file `.coffeerc` if present in the current directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ raw
 presentation
 test.coffee
 parser.output
+.coffeerc
 test/fixtures/underscore
 test/*.js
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
The `-r` approach works. With `test.coffee`:

```coffee
global.foo = Date.now()
```

Then do:

```sh
$ coffee -r './test.coffee'
coffee> foo
Expression assignment to _ now disabled.
1492574452171
coffee>
```

I think this is much more flexible a solution than `.coffeerc`, and as has been noted above, one can use this to create aliases to achieve the same result.

PR #2337, opened 2012-05-16
Use strict. For real.

### LINKED ISSUE #2335: Idea: auto-add `'use strict'` if compiled with top-level wrapper
Coffee is strict already and the change will work everywhere. What do you think?

paulmillr:
This will auto-add `'use strict';` to every coffeescript file if `—bare` wasn't passed to options.

Closes #2335.

### DIFF
--- a/test/function_invocation.coffee
+++ b/test/function_invocation.coffee
@@ -306,7 +306,12 @@ test "Prefix unary assignment operators are allowed in parenless calls.", ->
 
 test "#855: execution context for `func arr...` should be `null`", ->
   contextTest = ->
-    eq this, window ? global if this?
+    if this is undefined
+      eq this, undefined
+    else if this is null
+      eq this, null
+    else
+      eq this, window ? global
   array = []
   contextTest array
   contextTest.apply undefined, array


JaKXz:
The merge conflicts on this branch are probably way out of hand to easily resolve at this point... but would it be possible to re-use this work and make this a default thing?

GeoffreyBooth:
In ES2015, strict mode is automatically enabled for ES modules or classes, so for many projects this shouldn’t be an issue anymore. In CS2, it is assumed that most people will probably be piping CoffeeScript’s output through Babel or a similar transpiler; if Babel is used, the [strict mode transform](https://babeljs.io/docs/plugins/transform-strict-mode/) can be enabled to automatically prepend `'strict mode'` when necessary. I’m not sure this needs to be a responsibility of the CoffeeScript compiler.

Closing, but willing to reopen if people can convince the maintainers that this is something CoffeeScript should handle. At this point we have very limited resources for development, so in general if there’s something that Babel can do for us, we want to offload that responsibility onto Babel (hence CS2 outputs ES2015 and leaves the ES2015-to-ES5 conversion to Babel).

PR #2860, opened 2013-03-19
Prototypes of subclasses no longer type as 'ctor' 

bkietz:
(I am using WebKit Developer Tools in Chrome,
which types objects using GetConstructorName()
[WebKit](http://www.webkit.org/building/checkout.html)/Source/WebCore/bindings/v8/custom/V8InjectedScriptHostCustom.cpp)

```
class A
class B extends A
class C extends B
```

Current:

```
x = new C
x.__proto__                     # ctor
x.__proto__.__proto__           # ctor
x.__proto__.__proto__.__proto__ # A
```

Modified:

```
x = new C
x.__proto__                     # C
x.__proto__.__proto__           # B
x.__proto__.__proto__.__proto__ # A
```

This is what the inheritance of native objects like HTMLDivElement looks like.


GeoffreyBooth:
This is fixed in CoffeeScript 2.

```coffee
class A
class B extends A
class C extends B
```
becomes
```js
var A, B, C;

A = class A {};

B = class B extends A {};

C = class C extends B {};
```
In Chrome:
image

PR #3024, opened 2013-06-08
Added combined comparison operator.

vladtsf:
Hi,

I've added combined comparison operator (<=>), which present in Ruby. It would be useful in `.sort()` method of array.

The following code...

``` coffeescript
[ 1, 3, 2 ].sort ( left, right ) -> left <=> right
```

is equivalent to...

``` coffeescript
[ 1, 3, 2 ].sort ( left, right ) ->
  return 0 if left is right
  return -1 if left < right
  1
```

but much shorter and pretty :)


Christianjuth:
Please add this

PR #3171, opened 2013-09-18
Hygienic macro proposal

vanviegen:
Here's the macro system we've been using internally. It's not very lispy, but that might be a good thing to most. :) I'd say it's pretty clean, well, for a macro system anyway...

Some examples. Optionally compiled debug info:

``` coffeescript
macro debug (args...) -> new macro.Call(new macro.Literal("debugImpl"), args) if @debugging
macro -> @debugging = true
debug "test", 123 # console.log gets generated
macro -> @debugging = false
debug "test", somethingHeavy() # nothing gets into the output js
```

The mandatory swap example. Cheating a bit by using deconstruction, but this does show some of the more interesting features. The coffeescript code "[x,y] = [y,x]" get translated to nodes, and then an identifier substitution replaces x and y by whatever nodes are the arguments.

``` coffeescript
macro swap (a,b) -> (macro.codeToNode -> [x,y] = [y,x]).subst {x:a,y:b}
swap var1, var2
swap d[123], e[f].g
```

If you want, you can go pretty crazy with this. Non-sense from the unit test:

``` coffeescript
tst = (a,b) -> a*b
tst2 = -> 4
macro calc (c1,c2,c3,c4) ->
  func = macro.codeToNode ->
    x = (a) -> tst(a,c1) * c2
    x(c3) + x(c4)
  func.subst {c1,c2,c3,c4}
eq 144, calc 6, 3, 5, 3
eq 144, calc (macro.codeToNode -> tst2()+2), 3, 5, 3
```

One thing we're using compile-time logic for, is resource versioning of web resources (so we can have clients cache them permanently, and all references to them will automatically change the (otherwise ignored) version number whenever the resource is updated.

``` coffeescript
macro -> @getFileVer = (file) -> 13 # stub
macro resUrl (name) ->
    name = macro.nodeToVal name
    macro.valToNode "/data/#{name}?v=" + @getFileVer("static/#{name}")
someImg.src = resUrl 'test.png'
```

Also one of the more interesting applications (for us), is string translations.

``` coffeescript
interpolateAndPluralize = (msg,arg) -> msg.replace("%1",arg).replace(/[\[\]]/g,'') # stub

macro -> @transDict = waterBottles: "%1 bottle[s] of water" # can be loaded from a file

macro trans (args...) ->
  args[0] = macro.valToNode @transDict[macro.nodeToId args[0]]
  new macro.Call(new macro.Literal("interpolateAndPluralize"), args)

someExample = trans waterBottles, 17
```

In this example, we're just doing the actual translation compile-time, but one could easily improve upon this by resolving the string insertions at compile time, and generating custom code for the pluralizations.
One could even accept the coffeescript string interpolation syntax. Eg:

``` coffeescript
console.log trans "I'm to be translated within #{days} day[[s]]!"
```

would translate to:

``` coffeescript
console.log "Vertaal mij binnen #{days} dag#{days==1?'':'en'}!"
```

I'm quite curious to hear any thoughts on this approach...

Regards,
Frank


OttoRobba:
This is seriously incredible. :+1:
The capabilities for file inclusion alone are seriously awesome.

GeoffreyBooth:
As awesome as this is, it doesn’t look like this is headed into the compiler anytime soon. I think an approach that might pass muster is if the compiler is updated to allow plugins somehow, and then this could be released as its own project as a plugin. (The parsing of Markdown in Literate CoffeeScript would make another excellent candidate to extract into a plugin.)

I’ve added BlackCoffee and a link to this PR to the [List of languages](https://github.com/jashkenas/coffeescript/wiki/List-of-languages-that-compile-to-JS) wiki page.

zolomatok:
Could we please please please have [blackcoffee](https://github.com/paiq/blackcoffee) merged into coffeescript? 🥺

I specifically heard Jesus mention he's not coming back until we have macros (and infix operators on custom objects) in coffeescript. For realsies.

I'll forgo all other presents if I can just have macros for christmas. 🎁

GeoffreyBooth:
> Could we please please please have blackcoffee merged into coffeescript?

It hasn't been updated in 8 years, which means it's a fork of CoffeeScript 1 rather than 2. So unless blackcoffee catches up to trunk, it can't be merged in (on a technical level, aside from whether folks agree that it's a good idea).

Charuru:
> > Could we please please please have blackcoffee merged into coffeescript?
> 
> It hasn't been updated in 8 years, which means it's a fork of CoffeeScript 1 rather than 2. So unless blackcoffee catches up to trunk, it can't be merged in (on a technical level, aside from whether folks agree that it's a good idea).

I don't think he means literally more that the macros are implemented into coffee. I really want this too, I've since started using TS but if coffee gets macros...

zolomatok:
> It hasn't been updated in 8 years, which means it's a fork of CoffeeScript 1 rather than 2. So unless blackcoffee catches up to trunk, it can't be merged in

Thanks so much for the reply!
Oh, yeah, you’re right! I wonder if it’s a lot of work to bring it up to snuff? To be honest, I’m mostly of course just longing for macros, I’m not sure if there’s a simpler way than modernising blackcoffe.

> aside from whether folks agree that it's a good idea

I really, really hope they do. From the looks of it, the reception to this issue was very warm.

zolomatok:
> I don't think he means literally more that the macros are implemented into coffee.

Yeah, I’m just wishing for those tasty macros. Especially for operators.

> I really want this too, I've since started using TS but if coffee gets macros...

Me too :D I miss CS so much though. I thought the type system would be enough of a carrot, but alas, no. I miss the expressivity and cleanliness of CS.

STRd6:
@zolomatok and @Charuru if you like CoffeeScript and TypeScript you may want to check out https://github.com/DanielXMoore/Civet. It's currently in active development and aims to merge the benefits of both CoffeeScript and TypeScript. It doesn't yet have hygenic macros but I'm not fundamentally opposed to the idea and I would be happy to explore it with your feedback.

zolomatok:
@STRd6 Very intriguing, thanks! 👏
_(sorry for the late response I couldn't respond before)_

It's awesome that this is a thing! I was happy to see the VSC extension as well!
I had some trouble with the compiler, I opened an issue.

I would love to see macros and custom infix operators in Civet 🌈
Should I open an issue on the repo for that?

PR #3240, opened 2013-11-15
using 'yield' automatically turns functions into generators

alubbe:
See discussion https://github.com/jashkenas/coffee-script/pull/3078

### DIFF
--- a/test/generators.coffee
+++ b/test/generators.coffee
@@ -7,23 +7,18 @@ test "generator as argument", ->
   ok -> yield 0
 
 test "generator definition", ->
-  x = ->
+  x = do ->
     yield 0
     yield 1
     yield 2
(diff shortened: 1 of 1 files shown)


lbeschastny:
`yield` support works great, but there is still no way to create a non-yieldable generator in CS. This feature was actively discussed here, but discussion stopped after PR was merged.

So, any plans to support it?

I think, there should be some way to explicitly create a generator function. Otherwise, generators support seems incomplete.

lydell:
It is already supported. http://coffeescript.org/#try:nonYieldingGenerator%20%3D%20-%3E%0A%20%20doSomething()%0A%20%20yield%20return

lbeschastny:
@lydell thanx!

PR #3258, opened 2013-11-25
Fix #2489 again: preserve bound methods length property

### LINKED ISSUE #2489: Function.length in fat arrow class method

layerssss:
Hi, this is yet another approach trying to fix #2489 .

I read through the discussion of #2489 #2872 , and I came up with this lighter solution. Since `__bind` is not a public part, patching it will not break anything.

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -1415,12 +1415,12 @@
     };
 
     Class.prototype.addBoundFunctions = function(o) {
-      var bvar, lhs, _i, _len, _ref2;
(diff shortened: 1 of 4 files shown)


GeoffreyBooth:

Issue #3143, opened 2013-08-29
Fat arrow leaks generated `_this` reference to other closures


joshma:
Using a fat arrow generates a `_this` variable for use in the current closure, but due to how the JS engines work it gets leaked to all closures in the function scope:

``` coffee
@foo = 42

printFoo = => # This creates a "_this" variable
  console.log @foo

bar = ->
  debugger # When Chrome's debugger stops here, the "_this" variable is visible under the "Closure Scope" variables
  # console.log (eval '_this').foo # If this is executed, it'll frigging work!

bar()
```

(Thanks to @epidemian for the example, see end of #3052)

In the example above, `printFoo` uses a fat arrow to create a closure on `_this`, but `bar` accidentally gets a reference through the Closure Scope as well. Even if all references to `printFoo` are discarded, a reference to `bar` will indirectly hold a reference to `_this` and prevent it from being garbage collected.

Given that the use of a closure is an implementation detail and not part of the fat-arrow abstraction, this seems like an undesirable side-effect. As I suggested in #3052, an IIFE approach should provide the same abstraction without this issue:

``` coffee
bar = => console.log(@)
// ...translates to
bar = (function(_this) {
  return function() {
    console.log(_this);
  }
})(this)
```

(Both @epidemian and @michaelficarra seem to somewhat agree.)

Issue #1819, opened 2011-11-03
replacing bound methods gives wrong results


abelousov:
If you replace class prototype method which is bound with fat arrow, it isn't replaced in any previously created instance.

Jasmine spec:

```
describe 'Coffeescript: class methods binding (fat arrow)', ->
    it 'should allow method replacement for both future and existing instances', ->
        class Foo
            replaced: => false

        existing = new Foo
        Foo::replaced = -> true
        future = new Foo

        expect(future.replaced()).toBeTruthy()   #OK
        expect(existing.replaced()).toBeTruthy() #fail: Expected undefined to be truthy

```

It happens because __bind takes final, computed reference to prototype method, so when prototype method changes, bound method references old prototype method.

The solution is to use specific method to bind prototype methods, giving it not with method itself, but its name:

```
var __bindPrototypeMethods = function(methodName, me){ 
    return function(){ return me.constructor.prototype[methodName].apply(me, arguments); }; };
```

complete javascript for previous code would be:

```
var __bindPrototypeMethods = function(methodName, me){ 
    return function(){ return me.constructor.prototype[methodName].apply(me, arguments); }; };

describe('Coffeescript: class methods binding (fat arrow)', function() {
  return it('should allow method replacement for both future and existing instances', function() {
    var Foo, existing, future;
    Foo = (function() {
      function Foo() {
        this.replaced = __bindPrototypeMethods('replaced', this);
      }
      Foo.prototype.replaced = function() {
        return false;
      };
      return Foo;
    })();
    existing = new Foo;
    Foo.prototype.replaced = function() {
      return true;
    };
    future = new Foo;
    expect(future.replaced()).toBeTruthy();
    return expect(existing.replaced()).toBeTruthy();
  });
});
```

GeoffreyBooth:
CS2 has removed class bound methods. #4530.

PR #3260, opened 2013-11-26
Fix #2489 once more -- preserve bound methods arity

### LINKED ISSUE #2489: Function.length in fat arrow class method

marchaefner:
Since compiler hacking is way more fun than CLI fixing, here's my version of "Fix #2489 again"
- Adds a "private" `_bindMethods` function wherein all binding takes place.
  This preserves the readability of the compiled constructor and avoids any potential memory leaks (related: #3143).
- Wrappers clone the parameters of the wrapped method and therefore have the same number of parameters and, for the most part (see example below), the same parameter names.
- Superclass constructors do neither override previously bound methods (as in redux) nor mindlessly wrap already wrapped methods (as is currently happening)

This comes with a hefty dose of ugly. Though I doubt the compilation could get much nicer. Compare #3258 which fixes this issue in a different way (prettier classes but more helpers).

Based upon prior art by @epidemian and @michaelficarra. Feel free to bash...
#### Side effect

This PR unintentionally "fixes" #1819 (the same way redux does). This  breaks other fun stuff:

``` coffee
class Nobody then expect: => console.log 'fluffy bunny'
callback = (new Nobody).expect
# ...
setTimeout callback, 1000
# ...
Nobody::expect = -> console.log 'spanish inquisition'
```

I can't really assess the severity of this change since I can't imagine why anybody would ever do anything like that (or #1819 for that matter).
#### Example compilation

``` coffee
class Foo extends Bar
  baz: (@this=1, that, [which, what]) =>
  qux: (args...) =>
```

``` js
/* ... */

Foo = (function(_super) {
  var _bindMethods;

  __extends(Foo, _super);

  _bindMethods = function(_this) {
    if (Foo.prototype.baz === _this.baz) {
      _this.baz = function(_this1, that, _arg) {
        return Foo.prototype.baz.apply(_this, arguments);
      };
    }
    if (Foo.prototype.qux === _this.qux) {
      _this.qux = function() {
        var args;
        return Foo.prototype.qux.apply(_this, arguments);
      };
    }
  };

  function Foo() {
    _bindMethods(this);
    return Foo.__super__.constructor.apply(this, arguments);
  }

  Foo.prototype.baz = function(_this, that, _arg) {
    /* ... */
  };

  Foo.prototype.qux = function() {
    /* ... */
  };

  return Foo;

})(Bar);
```

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -1414,14 +1414,36 @@
       });
     };
 
-    Class.prototype.addBoundFunctions = function(o) {
-      var bvar, lhs, _i, _len, _ref2;
+    Class.prototype.addBoundFunctions = function(name, o) {
+      var bindMethods, bindMethodsRef, bvar, context, func, instFunc, p, protoFunc, wrapper, _i, _len, _ref2, _ref3;
+      if (!this.boundFuncs.length) {
+        return;
+      }
+      context = new Literal('_this');
+      bindMethods = new Code([new Param(context)]);
+      bindMethods.noReturn = true;
       _ref2 = this.boundFuncs;
       for (_i = 0, _len = _ref2.length; _i < _len; _i++) {
-        bvar = _ref2[_i];
-        lhs = (new Value(new Literal("this"), [new Access(bvar)])).compile(o);
-        this.ctor.body.unshift(new Literal("" + lhs + " = " + (utility('bind')) + "(" + lhs + ", this)"));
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:

PR #3342, opened 2014-01-30
Add tab key usage to try-it editor

colelawrence:
Use the tab character in the try-it editor's textbox


srgraham:
Are there any plans to implement something like this? I usually go to the coffeescript site when testing random quick things, and I usually have a moment of confusion when tab doesn't do what I expect it to do.

PR #3383, opened 2014-02-21
[RFC] Optionally emit comments from the lexer and the parser.

eventualbuddha:
The lexer emits comment tokens inline since they don't interfere semantically. In the generated AST they are added to the root node as a "comments" property because they cannot (easily) be put directly into the nodes where valid comments might appear. This is similar to how Esprima handles parsed comments in the generated AST:

``` js
> parse('1 + // foo\n1', { comment: true })
{ type: 'Program',
  body:
   [ { type: 'ExpressionStatement',
       expression: [Object] } ],
  comments: [ { type: 'Line', value: ' foo' } ] }
```

I tried searching for something similar to this in the coffee-script issues but never found one. There were a bunch of discussions about how hard it would be to support attaching comment nodes to their nearest node, and I agree that is hard.
#### Goals
- allow extracting the single-line comments from CoffeeScript source, for example to facilitate a [recast](https://www.npmjs.org/package/recast)-style re-writer for CoffeeScript
#### Non-Goals
- emit single-line comments in the generated JavaScript
- include comments in the generated AST or token list by default

I probably did something wrong in this PR and am mostly looking for comments to improve the approach I've taken. I don't especially care that this is merged, though it would make the project I'm working on (the single example from "Goals" above) easier to realize as an open source project rather than an internal tool at my company.


GeoffreyBooth:
Regarding your goal:

> allow extracting the single-line comments from CoffeeScript source, for example to facilitate a recast-style re-writer for CoffeeScript

Doesn’t Decaffeinate do this?

As for one of your non-goals:

> emit single-line comments in the generated JavaScript

_This_ I can see being desired in CoffeeScript, and if your PR did this I would encourage its speedy completion.

eventualbuddha:
Wow, blast from the past. This was part of my initial attempt to use the official CoffeeScript parser to back decaffeinate. Since then, I wrote my own lexer (coffee-lex) and parser wrapper (decaffeinate-parser).

GeoffreyBooth:
If in that work you found a way that can preserve inline comments, and you feel like backporting that solution, I would appreciate it 😄

GeoffreyBooth:
See #4572

Issue #3399, opened 2014-03-07
getting rlwrap to work with coffeescript's repl


novodinia:
How would you utilize rlwrap with coffeescript's builtin repl?

I can do it with the node repl like this:
    NODE_NO_READLINE=1 rlwrap coffee

GeoffreyBooth:
This isn’t really a bug, as we’re not advertising this feature. If you want to submit a PR that enables support for this but doesn’t harm any current functionality, I would consider it.

PR #3433, opened 2014-03-21
Let the REPL be rlwrap-friendly by passing NODE_NO_READLINE=1

epidemian:
This is a quick fix for #3399.

Basically, it copies the `node` command and [passes `terminal: false` to the REPL](https://github.com/joyent/node/blob/6cbfcdad46d733bb04332063727e304e449dc86b/src/node.js#L137) if `NODE_NO_READLINE` env var is set.

Now users that prefer to have the nice readline features like <kbd>ctrl</kbd>+<kbd>R</kbd> (history search) can do:

``` bash
alias coffee="env NODE_NO_READLINE=1 rlwrap coffee"
```

Side note: may i add a comment with that Bash alias to the changelog so more users can see it? Is there some problem with doing that alias?

Should we also copy the NODE_DISABLE_COLORS (what's up with Node and negated boolean names?) [configuration](https://github.com/joyent/node/blob/6cbfcdad46d733bb04332063727e304e449dc86b/src/node.js#L140)?

### DIFF
--- a/lib/coffee-script/command.js
+++ b/lib/coffee-script/command.js
@@ -57,6 +57,9 @@
     replCliOpts = {
       useGlobal: true
     };
+    if (parseInt(process.env.NODE_NO_READLINE, 10)) {
+      replCliOpts.terminal = false;
+    }
     if (opts.nodejs) {
       return forkNode();
     }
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
I tested this, and @calebcase’s [comment](https://github.com/jashkenas/coffeescript/issues/3399#issuecomment-48385424) still stands:

> Unfortunately this also disables the colorized output and CTRL-V multiline support. rlwrap’s multi-line support with CTRL-^ is insufficient due to line continuation issues. Trying to trick it with -m’^M’ doesn’t work either (still gets interpreted as a single line).

PR #3438, opened 2014-03-31
Strict mode commandline option.

codepunkt:
I want TypeErrors when trying to add properties to sealed objects - without 'use strict' statement, these fail silently.

Therefore i'd like to have an option that generates 'use strict' statements in the top-level function safety wrapper.

See #2337.


GeoffreyBooth:
Using `coffee` to run scripts at the command line is supported, but not encouraged. There are many things we could do to expand its command-line functionality, but we don’t want to encourage its use for this purpose.

PR #3479, opened 2014-05-15
Issue spawning child process of `coffee` with `--nodejs` argument

ming-codes:
I ran into a strange issue with coffee script executable. Enabling `--harmony` seems to mess up IPC with node's `child_process.fork`

``` coffee

# parent.coffee

spawn = require('child_process').spawn

child = spawn('/usr/local/bin/coffee', [ '--nodejs', '--harmony-generators', './child.coffee' ], {
    stdio: [0, 1, 2, 'ipc']
})

child.on('message', console.log.bind(console, 'from child'))
child.send({ hello: 'child' })
```

``` coffee

# child.coffee

# comment
process.on('message', console.log.bind(console, 'from parent'))
process.send({ hello: 'parent' })

gen = `function *() {
}`

gen()
```

In the above code, if you remove `'--nodejs', '--harmony-generators'`, IPC works correctly but generators won't be enabled. Once you add the params in, IPC breaks (process.send is  undefined), but generators are enabled.
#3450 May be related to this.

node@0.11.12
coffee-script@1.7.1

### DIFF
--- a/lib/coffee-script/command.js
+++ b/lib/coffee-script/command.js
@@ -545,15 +545,24 @@
   };
 
   forkNode = function() {
-    var args, nodeArgs, p;
+    var args, customFds, nodeArgs, p;
     nodeArgs = opts.nodejs.split(/\s+/);
     args = process.argv.slice(1);
     args.splice(args.indexOf('--nodejs'), 2);
+    customFds = [0, 1, 2];
+    if (process.send) {
+      customFds = customFds.concat('ipc');
+    }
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
We don’t encourage the use of `coffee` for running scripts, aside from testing. We expect people to use `coffee` to compile their scripts to .js and then run them via `node`.

This is also much less of an issue today with Node 7’s unflagged support for just about everything aside from modules.

Issue #3498, opened 2014-05-27
Script.createContext is undefined in node 0.11.9


chrisyip:
I noticed that node had move `Script.createContext` to `vm.createContext` (http://nodejs.org/api/vm.html#vm_vm_createcontext_initsandbox).

I wonder if it's acceptable to use something like

``` js
var createContext = Script.createContext || vm.createContext;
```

to make it compatible with older and newer versions of node.

balupton:
Because of this bug `require('coffee-script').eval` is broken in Node 0.12. This is not good.

jashkenas:
@mbrio Is there a new pull we can take a look at?

mbrio:
@jashkenas I'll look into it right now, we've been using this fix for months now and haven't had any issues with it.

mbrio:
@jashkenas I just submitted a PR, the browser file didn't match the output of the compiled 1.9.0 source code. When I compiled it and overwrote the browser file supplied by coffee-script-source ruby gem, the error no longer occurred.

GeoffreyBooth:
CoffeeScript 2 supports only Node 7+. If this issue only occurs in old versions of Node, I think we can stop worrying about it. Please comment if I’m wrong and I’ll reopen.

PR #3527, opened 2014-06-30
Re: Script.createContext is undefined in node 0.11.9 #3498

mbrio:
This should fix the issue specified in issue #3498.

There seems to be another issue in Node 0.11.9+ whereby REPL doesn't end after an exception occurs. I felt it easier for you to follow my changes if I submitted issue #3498 changes before working on REPL fix, I'll post another issue for that if one doesn't exist.


GeoffreyBooth:
Seems to have been abandoned. Closing out.

PR #3552, opened 2014-07-24
Update dependency on mkdirp

flowlo:


lydell:
Why isn’t this merged?

michaelficarra:
Does it fix any open issues?

flowlo:
My idea was to fix this outdated dependency:

```bash
$ npm install coffee-script
coffee-script@1.8.0 node_modules/coffee-script
└── mkdirp@0.3.5

$ npm outdated
Package  Current  Wanted  Latest  Location
mkdirp     0.3.5   0.3.5   0.5.0  coffee-script > mkdirp
```

michaelficarra:
That doesn't fix a bug or add a new feature. :-1:

lydell:
Latest isn’t always best. I just checked through the commit history of mkdirp since 0.3.5. The only changes are:

- Updated docs
- Added `opts.fs` optio
- Added CLI

So the only effect of updating would be adding minimist (which mkdirp uses for its CLI) as a sub-dependency.

PR #3564, opened 2014-08-01
Fixing links to project pages formerly hosted on RubyForge

nickveys:
RubyForge is gone, so several links broke.


GeoffreyBooth:
I guess since this was submitted, RubyForge has returned from the dead.

PR #3577, opened 2014-08-08
Simplify do.coffee

bfab:
This version is directly executable, and illustrates more directly the concept. One can comment out line #3 and run again to see the difference.


GeoffreyBooth:
I think the current example is actually more appropriate. A variable name like `err` is exactly the type of variable you want to wrap a closure around, to limit in scope. Also this example uses a for loop as the right hand of an assignment, which is an unusual pattern that could confuse people.

PR #3601, opened 2014-08-21
AST adds precise details for nodes and positions.

dabbler0:
This change annotates nodes with types, and in a couple cases where rewrites
result in lost token position information, it adds references to
the original tokens.

It is used by the visual block editor at [pencilcode.net](http://pencilcode.net/edit/intro). It shouldn't affect other users of CoffeeScript.


davidbau:
Any chance this change might be able to be merged?  Having a readable AST is likely useful for others; and since we depend on it, merging would also save us from maintaining the fork.

yjerem:
Having the `nodeType` on each node class would make working with the AST much easier. Right now the only way to identify each node type is to use `.constructor.name`, which breaks when the coffeescript module is minified. My current workaround to make my library work with minified coffeescript is to abuse `CoffeeScript.nodes()` like so:

```coffeescript
nodeTypes.Return = CoffeeScript.nodes("return").expressions[0].constructor
#...
if currentNode instanceof nodeTypes.Return
  #...
```

Please consider adding `nodeType`, it would make it much simpler to walk/manipulate the AST.

usirin:
I am trying to create a PR to [fkling/astexplorer](https://github.com/fkling/astexplorer), I am done with the code changes for there, but right now I had to use my own hosted version of browser version of coffeescript; because of the fact that minification process changes the constructor name, and there is no other way to understand node types via `http://coffeescript.org/extras/coffee-script.js`.

I wouldn't ideally want to bump up an old thread but since this is still open, I didn't create a new issue. 

Being able to visually see the AST, and explore it, would give us the possibility to debug new token types easier.

alangpierce:
@usirin FYI the [decaffeinate repl](http://decaffeinate-project.org/repl/#?evaluate=true&stage=coffeescript-parser&code=%23%23%23%0A%23%20Examples%20taken%20from%20coffeescript.org%0A%23%23%23%0A%0A%23%20Assignment%3A%0Anumber%20%20%20%3D%2042%0Aopposite%20%3D%20true%0A%0A%23%20Conditions%3A%0Anumber%20%3D%20-42%20if%20opposite%0A%0A%23%20Functions%3A%0Asquare%20%3D%20(x)%20-%3E%20x%20*%20x%0A%0A%23%20Arrays%3A%0Alist%20%3D%20%5B1%2C%202%2C%203%2C%204%2C%205%5D%0A%0A%23%20Objects%3A%0Amath%20%3D%0A%20%20root%3A%20%20%20Math.sqrt%0A%20%20square%3A%20square%0A%20%20cube%3A%20%20%20(x)%20-%3E%20x%20*%20square%20x%0A%0A%23%20Splats%3A%0Arace%20%3D%20(winner%2C%20runners...)%20-%3E%0A%20%20print%20winner%2C%20runners%0A%0A%23%20Existence%3A%0Aalert%20%22I%20knew%20it!%22%20if%20elvis%3F%0A%0A%23%20Array%20comprehensions%3A%0Acubes%20%3D%20(math.cube%20num%20for%20num%20in%20list)%0A) has the ability to show intermediate results, including the CoffeeScript lexer tokens and the CoffeeScript parser AST, so you may find that useful. Just pick "CoffeeScript Lexer" or "CoffeeScript Parser" from the dropdown. Currently it's based off the 1.10.0 codebase, not master. It's not as nice as astexplorer, but I've still found it really useful to get a better understanding of the token and AST formats.

Here's the formatting code for AST nodes: https://github.com/decaffeinate/decaffeinate/blob/master/src/utils/formatCoffeeScriptAst.js

usirin:
@alangpierce After your comment I somehow ended up myself implementing `decaffeniate` parser for astexplorer :) Right now i couldn't find a browser version for it, so i used wzrd.in to pull a browserified version of the `decaffeniate-parser`.

I really want to contribute to coffee-script tooling; as today es6 tooling as way beyond where `coffee-script` is. Because to me, and for all the other engineers in my company, es6 is not a replacement for coffee-script, it's still javascript and it still has its quirks. What coffee-script added to the table is more than just arrow functions, or classes. 

Being able to explore the results of parsers are beneficial for any case. Increasing the discoverability enables more people to face with coffee-script's internals, and in return we could have better tools to support how we write coffee-script.

GeoffreyBooth:
@usirin we would welcome your contributions 😄  We’re organizing the next steps for CoffeeScript over at https://github.com/coffeescript6/discuss if you’d like to join the discussion there.

GeoffreyBooth:
@alangpierce or @usirin is there still value in this PR? In CoffeeScript 2 `nodes.coffee` now compiles to native ES classes; can you see the class name or other info you need from those?

If this PR makes life easier for AST Explorer or js2coffee or (gasp) Decaffeinate, then I’m happy to merge it in if someone wants to update it for CS2. I would be especially curious if this could someday be used to pass an AST to Babel for Babel to generate the output from.

alangpierce:
@GeoffreyBooth I'm pretty sure `node.constructor.name` works fine both in the existing code and when it's compiled to ES classes. Adding a `nodeType` field and having the extra location data both seem mildly useful, but I likely wouldn't end up actually using the changes.

Going from CoffeeScript to a Babel-compatible JavaScript AST would probably require a whole AST rewriting step, so I don't really expect that this PR would make that any easier.

GeoffreyBooth:
@alangpierce thanks for the feedback. Closing then until we have a need for this.

PR #3661, opened 2014-10-04
Use stdio option instead of customFds

dtaniwaki:
The customFds option in spawn is deprecated.
http://nodejs.org/docs/latest/api/child_process.html#child_process_child_process_spawn_command_args_options

Although it's actually converted for now, I don't want to see the annoying deprecated message.
https://github.com/joyent/node/blob/master/lib/child_process.js#L808

Could you consider to merge this?

[Review on Reviewable](https://reviewable.io/reviews/jashkenas/coffeescript/3661)

### DIFF
--- a/lib/coffee-script/command.js
+++ b/lib/coffee-script/command.js
@@ -554,7 +554,7 @@
     p = spawn(process.execPath, nodeArgs.concat(args), {
       cwd: process.cwd(),
       env: process.env,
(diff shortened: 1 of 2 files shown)


rmg:
:+1: x :100: - this pops up all the time for users of iojs and node-v0.12.

In case it matters, it was deprecated over 3 years ago (node-v0.5.11), so this has been a long time coming.

PR #3696, opened 2014-10-27
Add the required code to launch the proper nodejs module to get a good backtrace

xapantu:
Using this module https://github.com/evanw/node-source-map-support

bin/coffee -c -m --nodejsmap test.cs
nodejs test.js

auiua()
^
ReferenceError: auiua is not defined
    at Object.<anonymous> (/home/xapantu/coffeescript/test.cs:1:1)
    at Object.<anonymous> (/home/xapantu/coffeescript/test.cs:1:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:902:3


GeoffreyBooth:
That module has instructions for its use with CoffeeScript right in its readme. I’m not sure why it needs to be integrated into the compiler.

We’re also not trying to encourage people to run scripts via the `coffee` command.

Issue #3678, opened 2014-10-13
put source maps into a datauri


akloster:
Currently the coffeescript command line compiler only offers source maps as separate files. Incidentally this means that streaming compilation is impossible and any asset manager has to use and organize temporary files, then either mangle the javascript output to include the source map or make sure the source map is someplace it can be found by the browser.

The alternative is to inline sourcemaps as base64 encoded datauris at the end of the file. Granted, this isn't exactly "readable javascript", but putting it at the end of the file and only turned on by a command line flag would be great for streaming compilation.

Here is a description of the approach: http://kybernetikos.github.io/jsSandbox/srcmaps/dynamic.html

PR #3697, opened 2014-10-27
add CLI opt to put sourcemap into .js file

bwin:
--inline-map
Puts a base64-encoded (v3)sourcemap at the end of a .js file
see #3678
Works in Chrome, doesn't work in FF.
Had to add a new dependency: btoa (for base64 encoding).


DylanPiercey:
@jashkenas @michaelficarra @lydell @satyr 

Something like this would allow `node-inspector` to work with coffeescript using inline sourcemaps as far as I know and would be useful. This could also be added to `#register()` for better tooling support.

Curious if any of you are using CS with node-inspector currently?

bwin:
This feature was implemented by #4111.

PR #3706, opened 2014-10-31
Optionally to read data-src out from <script>

imcotton:
By using `data-src` instead, this would avoid double script loading.


imcotton:
Any further thoughts guys? :ear:

michaelficarra:
LGTM, but you need to submit the compiled files as well.

imcotton:
Thanks for reminding, I've had rebase against `master` branch and including compiled `browser.js`.

Issue #3738, opened 2014-11-23
if and ? causes wrong evaluation


raymond-h:
``` coffee
test = no
console.log if (test ? no) then 'Yes' else 'No'

test = yes
console.log if (test ? no) then 'Yes' else 'No'

test = null
console.log if (test ? no) then 'Yes' else 'No'
```

...compiles to...

``` js
var test;

test = false;

console.log(test != null ? test : false ? 'Yes' : 'No');

test = true;

console.log(test != null ? test : false ? 'Yes' : 'No');

test = null;

console.log(test != null ? test : false ? 'Yes' : 'No');
```

The output ends up as:

```
false
true
No
```

...instead of the expected:

```
No
Yes
No
```

Instead of the entire first `?:`, only the `false` literal ends up as the conditional for the second `?:`, i.e. the final expression is equal to `test != null ? test : (false ? 'Yes' : 'No')` instead of the expected `(test != null ? test : false) ? 'Yes' : 'No'`. Maybe a set of parentheses would fix this issue?

Issue #3309, opened 2014-01-13
Incorrect code generation with existential operator inside brackets


Diggsey:
The following code will alert the value "true", when clearly the only two possible outputs of the "if" are the string or undefined. It seems like coffeescript doesn't properly understand the precedence rules for the javascript ternary operator. Removing the brackets from the condition will miraculously fix it...

``` coffeescript
f = -> true

result = if (f?()) then "THIS SHOULD BE ALERTED!"

alert result
```

PR #3766, opened 2014-12-30
Fix parenthesized conditions in if-else assignment

lumeet:
This addresses at least #3309 and #3738 and has to do with parenthesized existential operator. After turning it back and forth several times, I couldn't find anything more elegant than this. I'm not sure if it requires something more fundamental to make the check with some more beauty. Additionally, in what case it could possibly generate extra parentheses now?


GeoffreyBooth:
Moved to #4519

Issue #3560, opened 2014-07-29
compilation error with attribute access


loveencounterflow:
i'm feeling a little bit stupid here but i really tried hard to make sure i'm not reporting a ghost. thing is, when i write 

```
d[ 'a' ][ 'b' ][ 'c' ]
```
i get (using coffeescript 1.7.1 and `coffee -c foo.coffee`)
```
error: unexpected 'c'
d[ 'a' ][ 'b' ][ 'c' ]
                 ^^^
```
when i do the same on http://www.compileonline.com/try_coffeescript_online.php i get
```
main.coffee:2:18: error: unexpected STRING
d[ 'a' ][ 'b' ][ 'c' ]
                 ^^^
```
even worse,
```
x = d[ 'a' ][ 'b' ]
x[ 'c' ]
```
gives the same error. what's going on here?

lydell:
Fixed by #3774.

loveencounterflow:
@Delapouite definitely helps, i found out using syntax hiliting in the editor.

Issue #1718, opened 2011-09-20
Invalid unicode identifiers


satyr:
```
$ coffee -bpe '♪ = 1'
var ♪;
♪ = 1;

$ coffee -e '♪ = 1'

.:2
  var ♪;
       ^
SyntaxError: Unexpected token ILLEGAL
```

GeoffreyBooth:
The consensus above seems to be that this isn’t worth the performance hit to the compiler for the compiler to catch such errors, especially since the runtime will do so. If someone wants to submit a pull request that achieves such validation in a performant way, it will be welcomed; but in the meantime I think we’re choosing to take no action here.

PR #3774, opened 2015-01-06
Fix #2516, #3560: Unicode space handling

### LINKED ISSUE #2516: Bug: u00A0 Character not treated as whitespace.
Check out the sample code here and see how the same code generates different output

http://bit.ly/U6NmeD

Basically the same code generates different output.  

      view = new App.Views.Portal.Index(
        wtf: true
      )

Will generate 

view = new App.Views.Portal.Index({
  el: "wtf"
});

and 

view =  new(App.Views.Portal.Index({
  el: "wtf"
}));

Is this an encoding issue?

lydell:
It is possible to match only valid JavaScript identifiers with a really long
regex (like coco and CoffeeScriptRedux does), but CoffeeScript uses a much
simpler one, which allows a bit too much.

Quoting jashkenas/coffeescript#1718 #issuecomment-2152464 @jashkenas:

> But it still seems very much across the "worth it" line. You'll get the
> SyntaxError as soon as it hits JS, and performance aside -- even the increase
> in filesize for our browser coffee-script.js lib seems too much, considering
> this is something no one ever does, apart from experimentation.

In short, CoffeeScript treats any non-ASCII character as part of an identifier.
However, unicode spaces should be excluded since having blank characters as part
of a _word_ is very confusing. This commit does so, while still keeping the
regex really simple.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -815,7 +815,7 @@
 
   BOM = 65279;
 
-  IDENTIFIER = /^([$A-Za-z_\x7f-\uffff][$\w\x7f-\uffff]*)([^\n\S]*:(?!:))?/;
+  IDENTIFIER = /^(?!\d)((?:(?!\s)[$\w\x7f-\uffff])+)([^\n\S]*:(?!:))?/;
 
   NUMBER = /^0b[01]+|^0o[0-7]+|^0x[\da-f]+|^\d*\.?\d+(?:e[+-]?\d+)?/i;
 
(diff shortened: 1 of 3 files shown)


jashkenas:
Sounds alright. I guess you're imagining some really bizarre copy-and-paste scenarios?

lydell:
Yes. Most importantly the “I accidentally pressed AltGr+Space which inserted a non-breaking space” scenario, though.

michaelficarra:
:+1:

loveencounterflow:
+1.

PR #3777, opened 2015-01-08
Fix #3598: Make unary + and - generate _refs

### LINKED ISSUE #3598: Unary `+` and `-` do not generate `_ref`s, which coerces the operand more than once (inconsistent)
```
~/coffee-script [master|✔] 
20:31 $ bin/coffee -e 'b = {valueOf: -> console.log "!"}; for a in [0] by ~b then'
!
~/coffee-script [master|✔] 
20:32 $ bin/coffee -e 'b = {valueOf: -> console.log "!"}; for a in [0] by -b then'
!
!
!
!
~/coffee-script [master|✔] 
20:32 $ bin/coffee -e 'b = {valueOf: -> console.log "!"}; for a in [0] by +b then'
!
!
!
!
~/coffee-script [master|✔] 
20:32 $ bin/coffee -e 'b = {valueOf: -> console.log "!"}; for a in [0] by Number(b) then'
!
```

A `_ref` should be generated for `+b` and `-b`, just as for `~b` etc.

(Inspired by #3361.)

lydell:
Before commit c056c93e `Op::isComplex()` used to return true always. As far as I
understand, that commit attempts to exclude code such as `+1` and `-2` from
being marked as complex (and thus getting cached into `_ref` variables
sometimes). CoffeeScript is supposed to generate readable output so that choice
is understandable. However, it also excludes code such as `+a` (by mistake I
believe), which can cause `a` to be coerced multiple times. This commit fixes
this by only excluding unary + and - ops followed by a number.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1639,7 +1639,8 @@ exports.Op = class Op extends Base
     not @second
 
   isComplex: ->
-    not (@isUnary() and @operator in ['+', '-']) or @first.isComplex()
+    not (@isUnary() and @operator in ['+', '-'] and
+         @first instanceof Value and @first.isSimpleNumber())
 
   # Am I capable of
   # [Python-style comparison chaining](http://docs.python.org/reference/expressions.html#notin)?
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -2309,7 +2309,7 @@
 
     Op.prototype.isComplex = function() {
       var _ref2;
-      return !(this.isUnary() && ((_ref2 = this.operator) === '+' || _ref2 === '-')) || this.first.isComplex();
+      return !(this.isUnary() && ((_ref2 = this.operator) === '+' || _ref2 === '-') && this.first instanceof Value && this.first.isSimpleNumber());
     };
 
     Op.prototype.isChainable = function() {
--- a/test/operators.coffee
+++ b/test/operators.coffee
@@ -370,3 +370,65 @@ test "#3363: Modulo operator coercing order", ->
   b = valueOf: -> count += 1
   eq 4, a %% b
   eq 5, count
+
+test "#3598: Unary + and - coerce the operand once when it is an identifier", ->
+  # Unary + and - do not generate `_ref`s when the operand is a number, for
+  # readability. To make sure that they do when the operand is an identifier,
+  # test that they are consistent with another unary operator as well as another
+  # complex expression.
+  # Tip: Making one of the tests temporarily fail lets you easily inspect the
+  # compiled JavaScript.
+
+  assertOneCoercion = (fn) ->
+    count = 0
+    value = valueOf: -> count++; 1
+    fn value
+    eq 1, count
+
+  eq 1, 1 ? 0
+  eq 1, +1 ? 0
+  eq -1, -1 ? 0
+  assertOneCoercion (a) ->
+    eq 1, +a ? 0
+  assertOneCoercion (a) ->
+    eq -1, -a ? 0
+  assertOneCoercion (a) ->
+    eq -2, ~a ? 0
+  assertOneCoercion (a) ->
+    eq 0.5, a / 2 ? 0
+
+  ok -2 <= 1 < 2
+  ok -2 <= +1 < 2
+  ok -2 <= -1 < 2
+  assertOneCoercion (a) ->
+    ok -2 <= +a < 2
+  assertOneCoercion (a) ->
+    ok -2 <= -a < 2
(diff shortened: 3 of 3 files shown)


lydell:
I’m not sure how to test this :(

epidemian:
> I’m not sure how to test this :(

I'm not sure if i understand what this PR does but, could something like counting how many times the value is coerced work?

```coffeescript
count = 0
a = valueOf: -> count++
b = +a ? c
assertEqual 1, count # 2 on CoffeeScirpt 1.8.0
```

lydell:
Yes, that’s a test that could be added. That’s the easy part. But what should the test be called? “Unary operators should only coerce its operands once” and then test a couple of unary operators in a few cases (such as your `+a ? c`, or `for i in [0..10] by +a then`). Or should I just test + and -? And can I test somehow that `+1` isn’t turned into a `_ref`? And should there be tests for other “complex” things than unary operators? Moreover, I couldn’t find any unary operator tests in test/operator.coffee.

epidemian:
Hmm... i see.

Yes, the naming part is probably the hardest one, i'll leave that to you :p 

About checking that `+1` is not `_ref`ined, this is ugly, but it could work:

```coffeescript
fn = -> +1 ? ignored
occurrences = fn.toString().match(/\+1/g).length
assertEqual 2, occurrences

fn = -> +foo ? ignored
occurrences = fn.toString().match(/\+foo/g).length
assertEqual 1, occurrences
```

This is something that could more elegantly be checked if the compiler emitted some intermediate representation before JS code, like a JS AST.

A variant of this could be asking the compiler itself to compile something and checking its output:

```coffeescript
js = CoffeeScript.compile 'a = +1 ? b'
assert js.indexOf('a = +1 != null ? +1 : b;') isnt -1

js = CoffeeScript.compile 'a = +b ? c'
assert js.indexOf('a = (_ref = +b) != null ? _ref : c;') isnt -1
```

I think i prefer the later one.

lydell:
Thanks, that helps! I’ll think about this and will continue tomorrow.

bjmiller:
A test using `CoffeeScript.compile` in this way strikes me as a little fragile.  You're testing whether one thing happens in an entire line of compiled code.  But a change to something unrelated might break the test by modifying a different part of the line, even though the thing you really want to look at here hasn't changed.

Just food for thought.

epidemian:
> A test using CoffeeScript.compile in this way strikes me as a little fragile.

Yes, it is completely fragile. Sorry if i didn't make that clear before. It's just that, given the current design of the compiler, i don't know how one could check if a `_ref` was used or not in a non hacky way.

bjmiller:
I guess the best way to minimize that would be to search for the smallest possible thing that confirms what you're expecting.
`assert js.indexOf('_ref) isnt -1`, perhaps?

michaelficarra:
We have a policy against making assertions about the generated code. Tests should only test semantics. So the `valueOf` tests are the only ones needed.

src/nodes.coffee
```diff
@@ -1639,7 +1639,8 @@ exports.Op = class Op extends Base
     not @second
 
   isComplex: ->
-    not (@isUnary() and @operator in ['+', '-']) or @first.isComplex()
+    not (@isUnary() and @operator in ['+', '-'] and
+         @first instanceof Value and @first.isSimpleNumber())
```
michaelficarra: Just change [`Value::isComplex`](https://github.com/lydell/coffee-script/blob/24398774fc1b2254d083f73745faee49ad166e7b/src/nodes.coffee#L490) instead.

lydell: I’m not sure I follow here ...

michaelficarra: You're calling `Value::isSimpleNumber`. I say you don't do an `instanceof` test and instead use `Value::isComplex`. If `Value::isComplex` is returning `yes` for simple numbers, change its implementation.

lydell: I’ve tried different things back and forward, but every time my new test cases break as well as some other tests.

michaelficarra: That's not a very good reason to do an `instanceof` test when we have `isComplex` specifically for this. Which test cases are breaking? Why?

lydell: If you mean that I should change the line to `not (@isUnary() and @operator in ['+', '-'] and not @first.isComplex()`, then that is a problem because for example `a` is considered non-complex and the whole point of this PR is that for example `+a` should _be_ considered complex, so that a `_ref` is generated for it.

michaelficarra: Okay, that explanation helped. I was confused for a bit. This LGTM. Merging.

PR #3784, opened 2015-01-10
Unique generated vars

lydell:


lydell:
Includes #3754.

src/nodes.coffee
```diff
@@ -1431,9 +1430,8 @@ exports.Param = class Param extends Base
     return @reference if @reference
     node = @name
     if node.this
-      node = node.properties[0].name
-      if node.value.reserved
-        node = new Literal o.scope.freeVariable node.value
+      name = "at_#{node.properties[0].name.value}"
```
michaelficarra: No need to add `at_`.

lydell: First I didn’t. But then I ran into this issue: `(@_slice)->`. What should we do then?

michaelficarra: Are you saying that would be a problem because it would start with double underscores?

lydell: Yes, there’s a risk that it’ll get the same name as one of the helper functions such as `__slice`. If it has been used before there’s no worries: `o.scope.freeVariable` will rename the argument to `__slice1`, but if `__slice` has not been used before there will be a “reserved word __slice” error. There’s [a test for this](https://github.com/jashkenas/coffeescript/pull/3784/files#diff-e01140821ab246a31253bf5bcc14c4a3R439).

lydell: Moreover, it breaks the consistency that all generated variables start with a single `_`, and all helper functions start with two `__`.

michaelficarra: Names of helper functions should be counted as used if they are defined, so the conflict won't be a problem. But I can understand wanting to keep generated names starting with a single underscore for consistency, so we can leave it like this.

lydell: > Names of helper functions should be counted as used if they are defined, so the conflict won't be a problem.

Yes they _should_. But they don’t. Does this mean that there’s a bug somewhere?

src/nodes.coffee
```diff
@@ -1451,9 +1449,7 @@ exports.Param = class Param extends Base
   # `name` is the name of the parameter and `node` is the AST node corresponding
   # to that name.
   eachName: (iterator, name = @name)->
-    atParam = (obj) ->
-      node = obj.properties[0].name
-      iterator node.value, node unless node.value.reserved
+    atParam = (obj) -> iterator "@#{obj.properties[0].name.value}", obj
```
michaelficarra: Can't we just no-op here? Would a consumer of this function expect to get `"@name"` for an at-param or just have it omitted?

src/nodes.coffee
michaelficarra: Seems like it. The [`utility` function](https://github.com/lydell/coffee-script/blob/8ab15d737225acf3d0c2e9d575225df6a470971a/src/nodes.coffee#L2262) should register it.

lydell: Ok, the "issue" is that "_slice" is getting validated and passes the validation. Then, after the validation, `Param::asReference` changes "_slice" into "__slice". At that point no more validation is done, so it slips through. If the __slice helper hasn't been used yet there are no scope checks to stop it either. So the bug (if any) is that `Scope::freeVariable` does iterate until the var name is also not a reserved word.

lydell: I’ve fixed it now. `Scope::freeVariable` can never return a reserved word, because the names it returns always start with an underscore and as of the latest commit there are no reserved words starting with an underscore anymore.

Still, I prefer `@param`s to still compile to `_at_param`, because:

- It is consistent with that all generated variables start with a single `_`, and all helper functions start with two `__` (as said before).
- The not at all unlikely parameters `@base`, `@name`, `@ref` and `@arg` would otherwise be compiled into `_base`, `_name`, `_ref` and `_arg`, which might confuse a reader.
- An extra parameter in `Scope::freeVariable` is needed to force `@a` to compile to `_a1` instead of `_b` if `_a` was already used.
- It is in my opinion more readable, and CoffeeScript should have readable output. I read both `@param` and `_at_param` as “at param”.

michaelficarra:
Beautiful.

Issue #3597, opened 2014-08-20
Allow dynamic keys for object (hash) definition


aurium:
This feature is available in Javascript, but not in Coffee.
```coffee
opts = { s('param'): 'some value' }
# unexpected (
```

```coffee
replacements =
  "#{prefix}keyname1": 'some value 1'
  "#{prefix}keyname2": 'some value 2'
# unexpected string interpolation
```

lydell:
I’d vote for `"a#{b}c": value` to compile to `{["a" + b + "c"]: value}`. Just like `yield`, this would be a feature that works only in supporting environments; it’s up to the developer not to use it if support for an older environment is needed.

michaelficarra:
@lydell: I like that.

epidemian:
@lydell, but, unlike `yield`, this is a feature that can be easily implemented with ES3. 

I'd vote for using the same syntax and a similar implementation than [LiveScript](http://livescript.net/#literals) (see "Dynamic Keys"), where:

```livescript
myObj =
  x: 1
  "#{name}#{i}": 'some value' # Using string interpolation.
  (name + j): 'other value' # Using a parenthesised expression.
  somekey: someValue + myObj.x
```

Compiles to:

```javascript
// Generated by LiveScript 1.3.1
var myObj, ref$;
myObj = (ref$ = {
  x: 1
}, ref$[name + "" + i] = 'some value', ref$[name + j] = 'other value', ref$.somekey = someValue + myObj.x, ref$);
```

(the output probably needs better whitespace handling)

I like the LS syntax because dynamic string literals *look like* strings, which is the benefit of string interpolation in the first place; and because it also allows using any value as a key (using the parentheses syntax), which in turn would be useful if we want to support [Symbol properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) at some point:

```coffeescript
countdown = 
  (Symbol.iterator): *->
    yield 3
    yield 2
    yield 1

# Assuming this compiles to JS's `for ... of`
for n of countdown
  console.log n
```

lydell:
We must not forget #786 and #1731, though.

lydell:
Good point about `Symbol` keys. That means that we need to support an alternate syntax in addition to string interpolation. Both old CoffeeScript and @epidemian used `()`. An argument for this syntax is ease of implementation, as said in https://github.com/jashkenas/coffeescript/issues/786#issuecomment-484874:

[...]

Issue #3039, opened 2013-06-19
Interpolated dynamic keys are allowed if the interpolation string is empty


aaylward:
As of https://github.com/jashkenas/coffee-script/commit/4cab45c759e1699822ea3508355b43d1ecd0b5a1 , dynamic keys are not allowed.

```
➜  coffee-script git:(master) ✗ coffee -e 'console.log "#{1}":1'
[stdin]:1:19: error: unexpected {
console.log "#{1}":1
                  ^
```

This is as it should be, but

```
➜  coffee-script git:(master) ✗ coffee -e 'console.log "#{}":1'
{ '': 1 }
```

As @michaelficarra noted in #3038 , this should be a syntax error.

lydell:
I’d vote for closing this and implement https://github.com/jashkenas/coffeescript/issues/3597#issuecomment-68623619 instead.

jashkenas:
Nah -- If we bring back dynamic keys, we should implement them reliably.

That said, we would need a patch that brings back dynamic keys and also allows you to call `super` (somehow) from a method defined with a dynamic key as the property name.

vendethiel:
This works in coco:

```livescript
class A
  (b + "a"): -> super
```

which compiles to

```js
var key$, A;
A = (function(){
  A.displayName = 'A';
  var key$, prototype = A.prototype, constructor = A;
  prototype[key$ = b + "a"] = function(){
    return superclass.prototype[key$];
  };
  function A(){}
  return A;
}());
```

jashkenas:
Looks like a fine solution. I guess there wouldn't be any strange situations where the value of the super reference needs to be known somewhere that isn't in lexical scope for the generated name...

lydell:
Does that mean that

```coffee
a::[x]=->super
```

should compile to:

```js
var _key;
a.prototype[_key = x] = function() {
  return a.__super__[_key].apply(this, arguments);
};
```

?

jashkenas:
Nah -- In general, we the current error message is fine for those. We're never going to be able to detect at compile-time when you're dynamically assigning prototype methods. For example, even if you "fixed" the above, that wouldn't fix this:

    proto = a.prototype

    proto[key] = ->
      super

lydell:
I see, thanks.

lydell:
But still:

```coffee
class A
  x: -> super # currently valid
  @::x = -> super # current valid
  "x": -> super # currently valid
  "#{x}": -> super # decided to be ok

[...]

PR #3785, opened 2015-01-11
Allow super in methods with dynamic names

### LINKED ISSUE #1392: calling `super` in methods defined on namespaced classes
`C::m = -> super` compiles correctly into

    C.prototype.m = function() {
      return C.__super__.m.apply(this, arguments);
    };

`namespaced.C::m = -> super"` compiles incorrectly into

    namespaced.C.prototype.m = function() {
      return m.__super__.constructor.apply(this, arguments);
    };

Proper compilation would have been

    var _ref;
    _ref = namespaced.C;
    _ref.prototype.m = function() {
      return _ref.__super__.constructor.apply(this, arguments);
    };

The error originates from [the `METHOD_DEF` regex](https://github.com/jashkenas/coffee-script/blob/085874d5f35835bc8e61/src/nodes.coffee#L1807) being used by [`Assign::compileNode`](https://github.com/jashkenas/coffee-script/blob/085874d5f35835bc8e61/src/nodes.coffee#L935) matching only identifiers. [`Call::superReference`](https://github.com/jashkenas/coffee-script/blob/085874d5f35835bc8e61/src/nodes.coffee#L466) is then compiling incorrectly.

lydell:
As discussed in https://github.com/jashkenas/coffeescript/issues/3039#issuecomment-68916918.
This is the first step to implement dynamic object literal keys (see #3597).

This also fixes #1392.

In short, `super` is now allowed:

    # in class definitions:
    class A
      instanceMethod: -> super
      @staticMethod: -> super
      @staticMethod2 = -> super

    # in assignment where the next to last property is 'prototype':
    A::m = -> super
    A['prototype'].m = -> super
    a.b()[5]::m = -> super
    A::[x()] = -> super
    class B
      @::m = -> super


lydell:
Since

```coffee
class A
  @a = -> super
```

is allowed, and the static method compiles to

```js
  A.a = function() {
    return A.__super__.constructor.a.apply(this, arguments);
  };
```

shouldn’t

```coffee
A.a = -> super
```

be allowed too, and compile to the same thing? This would allow:

```coffee
# class style
class A extends B
  @a: -> super
  a: -> super

# classic style
A = ->
A extends B
A.a = -> super # currently an error
A::a = -> super

File: src/nodes.coffee [hunk 1]
```diff
@@ -2296,3 +2303,10 @@ unfoldSoak = (o, parent, name) ->
   parent[name] = ifn.body
   ifn.body = new Value parent
   ifn
+
+# Check if `.prop` or `["prop"]` is equal to `property`.
```
lydell: Oops, this comment is wrong.

lydell:
I’ve fixed the bad comment now.

lydell:
My proposal above might not be a good idea, because then

```coffee
A['prototype'].a = -> super
```

and

```coffee
prototype = 'prototype'
A[prototype].a = -> super
```

would compile to different things, which might be confusing. An error message in the latter case might be preferable.

lydell:
Fixed merge conflict.

lydell:
I’ve removed the code to allow `A['prototype'].m=->super`; see #3796. This effectively turns this PR into a code cleanup plus implementation of agreed-upon features:

- https://github.com/jashkenas/coffeescript/issues/3039#issuecomment-68916918
- #3597 

In other words, nothing controversial. This should make it easier to merge, I think.

jashkenas:
Rebase for a clean merge, and we'll merge?

lydell:
Rebased.

jashkenas:
Merged.

jashkenas:
Bravo on killing off that nasty METHOD_DEF, by the way.

PR #3786, opened 2015-01-11
Fix #3778: Make for loops more consistent

### LINKED ISSUE #3778: Sometimes possible to alter how many times a for loop is run, sometimes not
```
$ coffee -e 'a=2; for i in [1..10] by a then a=i; console.log i'
1
2
4
8
$ coffee -e 'a=2; for i in [1..10] by a+0 then a=i; console.log i'
1
3
5
7
9
```

If the by expression is `_ref`ed you cannot change the number of times the loop runs, otherwise you can.

`for i in [a..b]`: Same thing, unless `a` or `b` is `_ref`ed you can alter how many times the loop is run. I guess that should not be possible, just like modifying `array.length` inside a `for item in array` loop does not alter the loop.

The solution might be to _always_ cache range starts and ends as well as the step in for loops if they reference a variable (even if `.isComplex()` returns false).

Extracted from https://github.com/jashkenas/coffeescript/issues/3598#issuecomment-69221755.

lydell:
The following two lines might seem equivalent:

    for n in [1, 2, 3] by  a then a = 4; n
    for n in [1, 2, 3] by +a then a = 4; n

But they used not to be, because `+a` was cached into a `_ref`, while the plain
`a` wasn’t. Now even simple identifiers are cached, making the two lines
equivalent as expected.


lydell:
Simple bug, simple fix, simple merge? Is there anything you want me to do before merging?

vendethiel:
LGTM

jashkenas:
Unless I'm misreading the diff. we should never have an `unless...else`. `if` would be better.

I don't understand how caching a simple identifier could change the observed behavior here. Can you enlighten me?

vendethiel:
The juicy bits are in #3778.

jashkenas:
Ah. I'd think that just would just generally fall under the category of something you're not expected to do. So whatever.

lydell:
> we should never have an unless...else.

It was that way before, and I let it stay that way to reduce diff noise. But sure I can change it.

> So whatever.

Please either merge or close.

jashkenas:
Kill the unless-else -- and I'll be happy to merge.

lydell:
Done.

PR #3770, opened 2015-01-03
Refactor interpolation (and string and regex) handling in lexer

lydell:
See the commit messages for details.

I’m sorry about the size of the first commit, but I don’t think I could have split it up.

Fixes #3621, #3301, #3394, #3348, #2388, #2321.

PR #3787, opened 2015-01-12
Fix #1316: Interpolate interpolations safely

### LINKED ISSUE #1316: Single token misinterpolation
$ coffee -e 'console.log "#{+}-"'
    NaN

lydell:
Instead of compiling to `"" + + (+"-");`, `"#{+}-"'` now gives an appropriate
error message:

    [stdin]:1:5: error: unexpected end of interpolation
    "#{+}-"
        ^

This is done by _always_ (instead of just sometimes) wrapping the interpolations
in parentheses in the lexer. Unnecessary parentheses won't be output anyway.

I got tired of updating the tests in test/location.coffee (which I had enough of
in #3770), which relies on implementation details (the exact amount of tokens
generated for a given string of code) to do their testing, so I refactored them
to be less fragile.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -570,7 +570,7 @@
     };
 
     Lexer.prototype.matchWithInterpolations = function(str, regex, end, offsetInChunk) {
-      var column, index, line, nested, strPart, tokens, _ref2, _ref3, _ref4;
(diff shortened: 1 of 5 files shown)


lydell:
I’ve clarified the commit message now. Before it sounded like `"a#{foo}b"` would compile to `"a" + (foo) + "b"`, which is not the case. That still compiles to `"a" + foo + "b"`, just like we’re used to.

PR #3790, opened 2015-01-13
Get rid of `Scope.root` hack

lydell:
Using the static property `Scope.root` for the top-level scope of a file is a
hack, which makes it impossible to have several independent `Scope` instances
at the same time (should we ever need that).

This commit makes every instance have a reference to its root instead.


vendethiel:
I thought we used this in the CLI somewhere, but we don't seem to be.

michaelficarra:
Much better.

jashkenas:
A lot more `o` flying around, but hey...

PR #3791, opened 2015-01-13
Fix #3502: Define param variables when expansion

### LINKED ISSUE #3502: Variable definitions missing when (unnamed) splat arguments are used
The following lines
```coffee
a = b = 1
funct1 = (a, ..., b) ->
  a = 2
  b = 2
funct2 = (a..., b) ->
  b = 3
```
compile to
```javascript
var a, b, funct1, funct2,  __slice = [].slice;
a = b = 1;
funct1 = function() {
  a = arguments[0], b = arguments[arguments.length - 1];
  a = 2;
  return b = 2;
};
funct2 = function() {
  var a, b, _i;
  a = 2 <= arguments.length ? __slice.call(arguments, 0, _i = arguments.length - 1) : (_i = 0, []), b = arguments[_i++];
  return b = 3;
};
```
`funct1` is missing the variable declarations for a and b within the function scope, which obviously leads to surprising side effects. `funct2` declares all arguments as expected.

lydell:


src/nodes.coffee
```diff
@@ -1360,8 +1360,8 @@ exports.Code = class Code extends Base
     for param in @params when param not instanceof Expansion
       o.scope.parameter param.asReference o
     for param in @params when param.splat or param instanceof Expansion
-      for {name: p} in @params when param not instanceof Expansion
-        if p.value then o.scope.add p.value, 'var', yes
+      for p in @params when p not instanceof Expansion
+        if p.name.value then o.scope.add p.name.value, 'var', yes
```
vendethiel: seems weird to have a `when` and a single `if` as body

```coffee
for p in @params
  if p not instanceof Expansion and p.name.value
    o.scope.add p.name.value, 'var', yes
```

?

lydell: Or perhaps:

```coffee
for p in @params
  o.scope.add p.name.value, 'var', yes if p.name?.value
```

vendethiel: I'd go for the first (or use some kind of `any`)

lydell: I already pushed a new version before you commented ...

PR #3792, opened 2015-01-14
Fix #3194: Make strings always uncallable

### LINKED ISSUE #3194: String interpolation in array
```coffeescript
a = [ "something-#{id}"
      key: value
    ]
```

Produces

```javascript
var a;

a = [
  ("something-" + id)({
    key: value
  })
];
```

I expect something like below

```javascript
var a;

a = [
  "something-" + id, {
    key: value
  }
];
```

lydell:
No matter if they have interpolations or not.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -301,7 +301,7 @@
           index = regex.length;
           prev = last(this.tokens);
           if (prev) {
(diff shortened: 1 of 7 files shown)


vendethiel:
So, that's supposedly valid in ES5(/ES6?) (but it'll always throw at runtime), so I'd vote for disallowing them :). Thanks @lydell

lydell:
I just realized regexes suffer from the same problem, so I added a commit for them too.

michaelficarra:
:+1:

PR #3800, opened 2015-01-21
Decouple "mkdirp".

ysmood:
Make coffee completely independent.
As such a great project, coffee should keep as light as possible. Let's decouple the last one~

### DIFF
--- a/test/mkdirp.coffee
+++ b/test/mkdirp.coffee
@@ -1,0 +1,30 @@
+command = require '../lib/coffee-script/command'
+path = require 'path'
+fs = require 'fs'
+
+test "mkdirp", ->
+  deepDir = 'test/mkdirp/deep/dir'
+  process.argv = [
+    'node', 'coffee'
+    '-o', deepDir
+    '-c', __filename
+  ]
+  command.run()
+
+  compiledPath = path.join(
+    deepDir
+    path.basename(__filename, '.coffee') + '.js'
+  )
+
+  # Coffee's test suit doesn't support async task.
+  # Here some hack.
+  setTimeout ->
+    ok fs.existsSync(compiledPath)
+  , 1000
+
+  process.on 'exit', ->
+    fs.unlinkSync compiledPath
+    fs.rmdirSync 'test/mkdirp/deep/dir'
+    fs.rmdirSync 'test/mkdirp/deep'
+    fs.rmdirSync 'test/mkdirp'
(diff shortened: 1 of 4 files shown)


ysmood:
@michaelficarra Will this PR be merged?

michaelficarra:
I'd feel more comfortable if there were tests related to this.

ysmood:
@michaelficarra Well, I'll add one.

ysmood:
@michaelficarra A simple test has been added. How you test command line functions of coffee? I cannot find any.

test/mkdirp.coffee
```diff
@@ -0,0 +1,30 @@
...
+  process.argv = [
+    'node', 'coffee'
+    '-o', deepDir
+    '-c', __filename
+  ]
+  command.run()
+
+  compiledPath = path.join(
+    deepDir
+    path.basename(__filename, '.coffee') + '.js'
```
michaelficarra: Don't use + for string concatenation.

test/mkdirp.coffee
```diff
@@ -0,0 +1,30 @@
...
+  )
+
+  # Coffee's test suit doesn't support async task.
+  # Here some hack.
+  setTimeout ->
+    ok fs.existsSync(compiledPath)
```
michaelficarra: Just move the opening paren before the arrow and we don't have to do this gross line continuation.

jashkenas:
I'm not sure that we should be adding tests that muck around with the file system, and rely on process.exit not being killed or interrupted early.

But otherwise, let's merge this! Wonderful to get rid of the only dependency.

test/mkdirp.coffee
```diff
@@ -0,0 +1,30 @@
...
+  , 1000
```
epidemian: I don't see why this setTimeout of 1000 ms is necessary. Can't the same assertion be run on the `'exit'` callback of the process?

ysmood: @epidemian Good idea, I slightly altered the test, now it's more comfortable.

ysmood:
Removed the `setTimeout` in the test. Still very tricky though.

ysmood:
@jashkenas Removed the `setTimeout` in the test. Still very tricky though.

jashkenas:
I'm going to merge, and then tweak, or maybe kill the test.

Issue #3735, opened 2014-11-22
Lexer: Multiline comment in an array does not compile


seiyria:
This code will not compile:

``` coffee
class LightsFromTheStars extends Spell
  name: "Lights from the Stars"
  @element = LightsFromTheStars::element = Spell::Element.buff
  @tiers = LightsFromTheStars::tiers = [
    ###
      * This spell buffs the int and wis of your allies.
      *
      * @name Lights from the Stars
      * @requirement {class} Bard
      * @requirement {mp} 300
      * @requirement {level} 10
      * @effect +[caster.int/4] int
      * @effect +[caster.wis/4] wis
      * @duration 3 rounds
      * @category Bard
      * @package Spells
    ###
    {name: "Lights from the Stars", spellPower: 1, cost: 300, class: "Bard", level: 10}
  ]
```

Is this intentional?

GeoffreyBooth:
It only works if you remove the braces because then the comment is interpreted as _inside_ the object, i.e.:

```js
tiers = [
  {

    /*
      * This spell buffs the int and wis of your allies.
      *
      * @name Lights from the Stars
      * @requirement {class} Bard
      * @requirement {mp} 300
      * @requirement {level} 10
      * @effect +[caster.int/4] int
      * @effect +[caster.wis/4] wis
      * @duration 3 rounds
      * @category Bard
      * @package Spells
     */
    name: "Lights from the Stars",
    spellPower: 1,
    cost: 300,
    "class": "Bard",
    level: 10
  }
];
```

And multiline comments within objects are already supported.

Minimal example:

```coffee
arr = [
  ###
    Comment
  ###
  1
]
```
___
```bash
[stdin]:4:6: error: unexpected newline
  ###
     ^
```

GeoffreyBooth:
Fixed via #4572.

PR #3802, opened 2015-01-24
Allow multiline comment at end of an object definition [Fixes #3761]

### LINKED ISSUE #3761: Multiline comment  in object parse problem
failure.coffee file containing:

    AnObject =
       someFunction: (x) -> x
     
       ###
          A multiline comment
       ###
 
result:

    $ coffee failure.coffee
    failure.coffee:2:26: error: unexpected newline
       someFunction: (x) -> x
                             ^

If additional members of AnObject are added after the multiline comment, the error goes away.

mapmeld:
#3761 appears to only be an issue when the multiline comment is indented, and it appears at the end of an object definition

includes test

Issue #3735 appears to be unaffected by this patch - that may not be caused by rewriter.coffee at all.

### DIFF
--- a/src/rewriter.coffee
+++ b/src/rewriter.coffee
@@ -305,6 +305,7 @@ class exports.Rewriter
           # the continuation of an object.
           else if inImplicitObject() and tag is 'TERMINATOR' and prevTag isnt ',' and
                   not (startsLine and @looksObjectish(i + 1))
+            return forward 1 if nextTag is 'HERECOMMENT'
             endImplicitObject()
           else
             break
--- a/test/comments.coffee
+++ b/test/comments.coffee
@@ -418,3 +418,12 @@ test "#3638: Demand a whitespace after # symbol", ->
   """
 
   eq CoffeeScript.compile(input, bare: on), result
+
+test "#3761: Multiline comment at end of an object", ->
+  anObject =
+    x: 3
+    ###
+    #Comment
+    ###
+
+  ok anObject.x is 3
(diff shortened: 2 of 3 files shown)


mapmeld:
@jashkenas any feedback on whether this would work? I could add additional tests if it seems weird to change the rewriter

test/comments.coffee
```diff
@@ -418,3 +418,12 @@ test "#3638: Demand a whitespace after # symbol", ->
   """
 
   eq CoffeeScript.compile(input, bare: on), result
+
+test "#3761: Multiline comment inside object", ->
```
lydell: inside -> at the end of

would be more accurate

src/rewriter.coffee
```diff
@@ -305,7 +305,10 @@ class exports.Rewriter
           # the continuation of an object.
           else if inImplicitObject() and tag is 'TERMINATOR' and prevTag isnt ',' and
                   not (startsLine and @looksObjectish(i + 1))
-            endImplicitObject()
+            if nextTag && nextTag == 'HERECOMMENT'
```
lydell: `if nextTag is 'HERECOMMENT'`

lydell:
Apart from my comments, LGTM. +1 for merging.

jashkenas:
@mapmeld Want to make those tweaks?

mapmeld:
@lydell @jashkenas thanks for the feedback - those changes should be in the PR now

lydell:
Please squash the commits

PR #3805, opened 2015-01-26
Fixes broken REPL (fixes #3804)

### LINKED ISSUE #3804: Broken REPL (throws for any source that needs `scope.freeVariable`)
Due to the new mechanism of the name collision prevention implemented by https://github.com/jashkenas/coffeescript/commit/8ab15d737225acf3d0c2e9d575225df6a470971a
now we have a broken REPL.

To repoduce this bug simply open the REPL and type in
```
coffee> 1 %% 2
```
You will get
```
TypeError: Array.prototype.indexOf called on null or undefined
  at indexOf (native)
  at Scope.exports.Scope.Scope.freeVariable (.../coffeescript/lib/coffee-script/scope.js:98:45)
```

This happens because `%%` needs `utility 'modulo'` that calls `scope.freeVariable` and last changes in its' implementation assume that there are `referencedVars` in `options` object. This is true for `compile` and `eval` from the main `coffeescript` module, but is wrong for the REPL. 
I'd propose a hotfix, to use empty array for referencedVars in REPL, but the problem is a bit deeper. Solution proposed by @lydell is based on the fact that all the source is known in advance, that is true for normal context, but not for REPL. So the question is still open.

I will publish a PR with hotfix in a few minutes.

Artazor:
This fixes #3804

### DIFF
--- a/src/repl.coffee
+++ b/src/repl.coffee
@@ -26,7 +26,7 @@ replDefaults =
       ast = new Block [
         new Assign (new Value new Literal '_'), ast, '='
       ]
-      js = ast.compile bare: yes, locals: Object.keys(context)
+      js = ast.compile bare: yes, locals: Object.keys(context), referencedVars: []
       result = if context is global
         vm.runInThisContext js, filename
       else
--- a/lib/coffee-script/repl.js
+++ b/lib/coffee-script/repl.js
@@ -28,7 +28,8 @@
         ast = new Block([new Assign(new Value(new Literal('_')), ast, '=')]);
         js = ast.compile({
           bare: true,
-          locals: Object.keys(context)
+          locals: Object.keys(context),
+          referencedVars: []
         });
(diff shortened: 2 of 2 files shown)


src/repl.coffee
```diff
@@ -26,7 +26,7 @@ replDefaults =
       ast = new Block [
         new Assign (new Value new Literal '_'), ast, '='
       ]
-      js = ast.compile bare: yes, locals: Object.keys(context)
+      js = ast.compile bare: yes, locals: Object.keys(context), referencedVars: []
```
lydell: Perhaps setting `referencedVars` to `Object.keys(context)`?

lydell: Better yet: tokenize `input`, filter out all referenced variables then pass the tokens to `CoffeeScript.nodes`.

lydell:
This can be closed now.

PR #3807, opened 2015-01-26
Fix #3804: Provide list of referenced vars to REPL

### LINKED ISSUE #3804: Broken REPL (throws for any source that needs `scope.freeVariable`)

lydell:
Supersedes #3805. Here is a comparison of master, #3805 and this commit:

    # master
    $ bin/coffee
    coffee> 1 %% 2
    TypeError: Array.prototype.indexOf called on null or undefined

    # #3805
    $ bin/coffee
    coffee> 1 %% 2
    1
    coffee> (_results = null; i) for i in [1, 2, 3]
    TypeError: Cannot call method 'push' of null

    # this commit
    $ bin/coffee
    coffee> 1 %% 2
    1
    coffee> (_results = null; i) for i in [1, 2, 3]
    [ 1, 2, 3 ]

### DIFF
--- a/lib/coffee-script/repl.js
+++ b/lib/coffee-script/repl.js
@@ -19,16 +19,29 @@
     historyFile: process.env.HOME ? path.join(process.env.HOME, '.coffee_history') : void 0,
     historyMaxInputSize: 10240,
     "eval": function(input, context, filename, cb) {
-      var Assign, Block, Literal, Value, ast, err, js, result, _ref1;
+      var Assign, Block, Literal, Value, ast, err, js, referencedVars, result, token, tokens, _ref1;
       input = input.replace(/\uFF00/g, '\n');
       input = input.replace(/^\(([\s\S]*)\n\)$/m, '$1');
       _ref1 = require('./nodes'), Block = _ref1.Block, Assign = _ref1.Assign, Value = _ref1.Value, Literal = _ref1.Literal;
       try {
-        ast = CoffeeScript.nodes(input);
+        tokens = CoffeeScript.tokens(input);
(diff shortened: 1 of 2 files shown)


epidemian:
I don't understand why this change is necessary. Is it necessary after #3808? Can't a test catch this?

lydell:
#3808 would have fixed the `1 %% 2` the same way that #3805 does—the compilation does not break because `referencedVars` is not an array. But neither of those fixes the `(_results = null; i) for i in [1, 2, 3]` case. Therefore this PR is needed.

PR #3812, opened 2015-01-28
Update index.html.js

xfq:
Node comes with npm installed and "npm" is not an abbreviation for "Node Package Manager".


ymeine:
Is grabbing the proper `node.exe` executable is a custom way considered as installing Node.js? Because in this case `npm` wouldn't necessarily come with it.

----

_(Funny how both products claim to be quite independent, while they are so much in practice, be it for the implementation/development or for the packaging/distribution.)_

Issue #3596, opened 2014-08-20
CoffeeScript 2.0: “UNIX philosophy compilation strategy”?


lydell:
I suggest that CoffeeScript should always compile to the latest “stable-enough” standard JavaScript. If the user’s environment doesn’t support that JavaScript, it is up to the user to include needed polyfills—perhaps using [autopolyfiller](http://azproduction.ru/autopolyfiller/)—and/or pipe the JavaScript into another transpiler, such as [esnext](https://github.com/esnext/esnext).

Wanting to use future JavaScript features isn’t unique to CoffeeScript. Regular JavaScript wants that too.

Examples:
- Compile spread/rest/splat/whatever into ES6 `...foo` whenever possible. [es6-spread](https://github.com/esnext/es6-spread)
- Compile classes into ES6 classes. [es6-classes](https://github.com/esnext/es6-class)
- Compile `=>` functions into ES6 `=>` functions when possible. [es6-arrow-function](https://github.com/esnext/es6-class)
- Compile `a in b` into `b.indexOf(a) >= 0`. [polyfill](https://github.com/jonathantneal/polyfill/blob/master/source/Array.prototype.indexOf.js).
- Compile `yield` into `yield`. [regenerator](http://facebook.github.io/regenerator/)
- Compile destructuring into ES6 destructuring.
- Compile `a.throw` to `a.throw` (not `a["throw"]`) etc.
- Remove the “bare” option and always compile bare. Users should use a module system, or a separete “wrap-in-IIFE”-tool instead.

Pros:
- Allows the user to choose which environments they want to support, which has been asked for numerous times.
- Simplifies CoffeeScript compilation a lot.
- Easy to support the latest ES specs (no need to worry about old browsers).
- Follows the UNIX philosophy better. CoffeeScript becomes an alternate syntax with more features, such as chained comparisons (`0 < a < 5`), more splats (`(a, b..., c) ->`), expansion (`[..., last] = foo`), etc.

What do you think?

adambiggs:
> In my opinion, CoffeeScript 2 should become ES6++. CoffeeScript 3 should be ES7++.

Amen :raised_hands:

lydell:
Just for the fun of it, I wrote up a bit about my “dream compile-to-JS” language with UNIX philosophy in mind: https://github.com/lydell/frappe. If you also like fantasizing about syntax it might be worth a read ;)

vendethiel:
[...]

PR #3813, opened 2015-01-28
async/await semantics (diamod arrows as demonstration)

Artazor:
I'm aware of #3757 but I don't think that we should speak in terms of the future ES7, let stay in the pragmatic domain. This PR is intended to start a discussion and research in this field.
I've discovered that await semantics can be achieved by a tiny utility, much more simple than that of #3757 
Here it is:

``` javascript
    function (fn) {
      return function () {
        var gen = fn.apply(this, arguments);
        try {
          return resolved();
        } catch (e) {
          return Promise.reject(e);
        }
        function resolved(res) { return next(gen.next(res)); }
        function rejected(err) { return next(gen.throw(err)); }
        function next(ret) {
          var val = ret.value;
          if (ret.done) {
            return Promise.resolve(val);
          } else try {
            return val.then(resolved, rejected);
          } catch (_) {
            throw new Error('Expected Promise/A+');
          }
        }
      };
    }
```

Also small tweaks in compiler enable async'ed functions to be used as inheritable methods of CS classes, and even more - as constructors (in both cases `super` is available), so the inheritance and extends will not cause any issues (like they do for ordinary generators when used with `co`).

I've chosen a bit ridiculous syntax for async'ed functions: `-<>` and `=<>`  and kept `yield` for resolving promises (instead of `await`). This is not critical and easily can be refactored into `await` and `await return` style as of @GabrielRatener (and keep `->` and `=>`).

You can review the test (/test/generators/async).

Also I've moved all generator-related test stuff into separate directory, and removed --harmony flag sniffing (is my way of detecting generators valid enough?) - this is separate commit.

Also in the separate commit I've extended testing framework to be capable for dealing with promises.

The main question is: is the proposed semantics useful enough to be included into language in some way?

I think that  generators+promises coding patterns will be very popular at the server side (especially with io.js)
and there are so tiny conservative changes needed from the language to enable brand new type of libraries/packages to emerge.

@GabrielRatener, @jashkenas, @alubbe, @michaelficarra and other guys - what are you thinking about it?

P.S. All discussed things are already implemented in this PR


GabrielRatener:
My personal belief is that it is best for generator and async functions to be explicit, but @jashkenas  seems to have leaned towards implicit generator syntax, so I try to stay consistent with that in #3757, by using `await` or `await return` to make functions async. In my PR having an `await` operator instead of the same yield has two advantages. First it distiguishes coroutines, and async functions syntactically, which I believe is good as they are fundamentally different. Secondly, the `await` operator has higher precedence than `yield`. Since its operand is always an object, I give it precendence over all arithmetic operators, cause arithmetic operations in JS never result in an object.

Artazor:
@GabrielRatener no, they don't. The semantics of EcmaScript allows replacing the result of the constructor [if you are using explicit return of non-primitive type inside the constructor, then the result of the `new` will be overriden]. Here result of the constructor is replaced by a promise that when fullfilled will return the instance of the appropriate class.

GabrielRatener:
My bad, that does actually seem useful.

GabrielRatener:
I still don't like the idea of excluding features based on ES7 rumors.

Artazor:
@GabrielRatener I've considered `await` as unary operator, but found that any new keyword will break backward compatibility, because it will be indistinguishable from the implicit call. I think there are tons of packages where `await` is the name of the function. I believe that if we choose the unary operator syntax, then we should use one of the existing CS keywords in the manner that is prohibited at present moment. Or even combine an identifier with keyword. I can easily imagine solutions like

```coffeescript
asyncFn1 = ->
    result =  wait for myPromise            # awaitng for the promise
    results = all of [promise1, promise2]   # shorhand for Promise.all()
    winner = race of [promise1, promice2]   # shorthand for Promise.race()
    ...

asyncFn2 = (a, b) ->
       x = wait for someOtherAsync a
       y = wait for someOtherAsync b
      return x + y
```

where `wait for`, `all of`, `race of`  are essentially compound keywords. I believe that in present syntax these constructs are illegal, that means we can add them without introducing any breaking changes, though this will introduce additional complexity into lexer/normalizer/parser.
Somebody like @jashkenas or @michaelficarra can correct me if I'm wrong here.
In this case I can't find appropriate compound keyword for modelling `await return`.
Any Ideas?

vendethiel:
"await" might be a bit too widely used (after a very quick github search, it seems moslty used as part of compound names, though..), yes; but I'm not sure we want to use another name. If we add it as a keyword, `await =` will be disallowed anyways.

GabrielRatener:
@Artazor The race and all examples are not backwards compatible because of the `of` keyword

GabrielRatener:
What about keeping `await` and having a completely backwards compatible alias, maybe `~<`?

Artazor:
@GabrielRatener yes, unfortunately `of` leads to ambiguities.

Artazor:
Let brainstorm. What about single postfix exclamation immediately after expression (without space). Looks very minimalistic.

```
x = promise!     # resolution of promise
y = promise!.promiseField!.promiseFunc(a,b)!.nonPromiseField     # nice chaining
```

 And looks that it will not produce ambiguities. Will it?

bcherny:
@Artazor that might be confusing, since in ruby `!` means the method mutates its arguments.

GabrielRatener:
what about `promise<>.chainedPromise<>`?

Artazor:
@bcherny I know about this mnemonic. I've supposed that being reminiscent of mutating modifier would be relevant for the resolution of promises in some sense, since here we deal with the special kind of mutating semantics (though the idempotent one): promise being resolved changes it's own state from `pending` to `resolved`. So I'd stay with postfix `!` (another one-character alternative is postfix `~` which is visually hard to distinguish from minus and it has less relevant connotation (negation).

@GabrielRatener, I appreciate your trial to keep diamonds in a language, but I'm afraid that  `<>` as a postfix operator is much more frustrating than the original `-<>` and `=<>`. However, I think that Jeremy will approve only implicit asynchronous functions (if he'll approve them at all). Thus, I agree that we may implement `await` feature as an unary operator, but there are still unclear for me

A. if an `await` will be a keyword — how will we ensure the backward compatibility? 

B. which forms of the `await` should be implemented (prefix, postfix, both)?

**prefix** operator — better for standalone calls without chaining. I don't like '~' in '~<' any other alternatives? I'll use `~<` in the examples below, but I don't like it:
```coffeescript
x = ~< (~< (~< a).b).c    # ugly
x = ~< fn x, y, z         # nice (except the ~< itself)
~< sleep 100              # nice
```
**postfix** operator  — better for chaining, but ugly for standalone calls:
```coffeescript
x = a!.b!.c!        # nice
x = fn(x, y, z)!    # ok
x = (fn a, y, z)!   # strange
(sleep 1000)!       # ugly      
```
 **both**?
Here I'd propose `!` as postfix and `(!)` as prefix (to keep track of the exclamation sign - a bit haskelish)

```coffeescript
x = a!.b!.c!         # nice
x = (!) fn x, y, z   # ok
(!) sleep 100        # ok
 ```

or kinda inconsistent

```coffeescript
x = a!.b!.c!         # nice
x = ~< fn x, y, z   # nice but inconsistent
~< sleep 100        # the same
 ```

Any other variants, thoughts?

**Comment by GabrielRatener** [comment]
@Artazor Yes, finding a solution to backwards compatibility is not easy. I am starting to  think that supporting bacwards compatibility of the `await` keyword would only very rarely cause issues though, based on a keyword search.

**Comment by vlad-p** [comment]
+1 for postfix `!`

**Comment by dev-rke** [comment]
@Artazor 
I am just a user of CoffeeScript and would like to see a real keyword rather than a new operator.
CoffeeScript always tried to focus on clarity of its expressions and keywords, and i dont think it would break existing applications to use new keywords like `await` and `async`. Developers using a new software version should accept new features and things that might break in favor of clarity and usability. Using operators could improve the chainability, but how often will this feature really used? 
Furthermore a single exclamation will be hard to read, it might easily beeing overlooked.
If it is hard to read and hard to understand, people won't use it. 
If it is clear and understandable, people will use it, that's why i am using CoffeeScript.

I tried the IcedCoffeeScript implementation from maxtaco, which uses just the additional keywords `await` and `defer`, maybe helpful for further discussion: https://maxtaco.github.io/coffee-script/

Some ideas for the planned implementation you are talking about:

E.g. when waiting for multiple async methods to be finished:
```coffeescript
await [
   method1: ->
   method2: ->
]
async 
  # code beeing executed after all methods have been finished
```

E.g. to chain methods
```coffeescript
currentState = getIn() await order() await getBeer() await drink() async getOut()
```
where currentState will have the result of getOut after all methods have been chained.
Another approach would be to introduce an additional keyword like `achain` (abbreviation for "asynchronous chain"):
```coffeescript
currentState = await getIn() achain order() achain getBeer() achain drink() async getOut()
```
This approach would propably keep the syntax clean to use cascaded structures like
```coffeescript
currentState = await getIn() achain (await getMenu() achain order() async pay()) achain getBeer() achain drink() async getOut()
```

This could also be expanded to
```coffeescript
await
  achain [
    getIn()
    order()
    getBeer()
  ]
async
  currentState = getOut()
```
or as alternative notation:
```coffeescript
currentState = await
  achain [
    getIn()
    order()
    getBeer()
  ]
async
   getOut() # which implicitly returns and its return value will be applied to currentState
```

For a race promise currently i have no idea without an additional keyword.

There might be some things i missed out, e.g. parameters for the async callback.
But i am not here to solve this, i am just here to introduce some ideas. :-)

I would really like to see a stable async feature in CoffeeScript, `await`ing it since i found IcedCoffeeScript. ;-)

What do you think about this?

GabrielRatener:
Lets all remember that coffeescript broke backwards compatibility when introducing implicit generator syntax. Considering @jashkenas pushed for this approach over an explicit syntax as in JS, it seems backwards compatibility is a low priority for the devs.

dev-rke:
I do not think it broke backwards compatibility. There are only a few brackets, commas and no semicolons required when writing CS code. Developers using CS are interested in writing code with less effort, using just reachable keys without wrenching fingers to type special characters. So in scope of CS everything seems to be clear, there are no underscores in variable names, no dollar signs, nor questionmarks. So introducing new keywords like `await` *might* break something (whoever calls his own variables or methods 'await'), but in prospect of usability and clarity it would be an enrichment. And developers already using an async library should be interested in getting an implementation by the language.

Artazor:
I think that `await` keyword ideally suits for the mentioned role. The only thing I'm afraid of is creating serious difficulties for that guys who already created libs with the same intention, like async/await on top fibers, and those who created their systems in top of such libraries. Ok, maybe we should sacrifice them. 

Other thing that I've started to think of is the performance of the proposed semantics, since @jashkenas blamed me for being "not smart" in usage of the new JS features. The point within current discussion is the fact that code inside generator functions is not optimized at all (in v8). As well it is not optimized when the function contains the `try/catch` block. Maybe `await` can be implemented on top of the pure promises and very straigtforward code transformation. There are couple of patterns (loop, branch, sequence) that allow you rewrite your async/await code with promises in that way that the resulting code will have the same visual structure and will be familiar to the original author (we can even keep the original line numbers fore every original statement/exoression). This approach is conservative (it is not the way of @maxtaco), and keeps the proposed semantics: every function that uses `await` will return a promise.

@jashkenas, what are you thinking of the generator-less awaits? They will work on the client-side as well without any 6to5 postprocessing...

alubbe:
To clarify your discussion, generator support never broke backwards compatibility, "yield" was always a reserved word and could not be used in the past.

Artazor:
@alubbe exactly!

GabrielRatener:
@Artazor @alubbe 

```
var a = function(){
	var yield = 5;
	return yield;
}

a() // -> 5
```

Is valid JS. Since when has `yield` been a reserved word in CS?

alubbe:
since CS 1.3.0 https://github.com/jashkenas/coffeescript/blob/de511e034882056481833cca77ad896add5825f7/src/lexer.coffee#L584

Artazor:
Guys, let the concrete syntax for the later stages of the discussion.

What I want to discuss here too is the possibility to implement `await` on top of the promises only. I know that @maxtaco's transformations were rejected by @jashkenas and @maxtaco created its own language. But there was a serious semantic improvement + runtime + generated code was surely generated (no one manually writes in the style of the javascript generated by IcedCoffee). Here I propose the style that is already used by many developers when they deal with promises without generators. 

Here is an example how `await` operator can be desugared into Promise-based code without ES6 generators. 
The only assumption is that there is `Promise.resolve` (to create an empty promise for the loop finalization) among the globals. 

Examples are written with "throw early" policy adoption:
if promisified code has a **synchronous** preamble that throws we prefer to throw instead of creating rejected Promise.

Note that patterns below don't cover control flow breaks like `return`, `break`, `continue` and `throw` (for these breaks we should use a bit more complex structures).

#### Asynchronous sequence
```javascript
function() {
  var b, e;
  a();
  b = await c;
  d();
  e = (await f[3].f()).x.y.z;
  g();
  return h;
}
```
when desugared into Promises becomes

```javascript
function() {
  var b, e;
  a();
  return c.then(function(_ref){ 
    b = _ref;
    d();
    return f[3].f().then(function(_ref){
      e = _ref.x.y.z;
      g();
      return h;
    });
  });
}
```
Note, that we can reformat the code to make it line-to-line prefect: 

![Sequence line-to-line comparison]
(https://s3.amazonaws.com/f.cl.ly/items/470K1D3c382b0c2O3G3b/Screen%20Shot%202015-02-12%20at%2013.26.01.png)

#### Asynchronous selection

```javascript
function() {
  var d, n;
  a()
  if (x < y) {
    b();
    d = await e();
    f();
  } else {
    m();
    n = await p();
    q();
  }
  r();
  return x;
}
```

when desugared to promises becomes:

```javascript
function() {
  var d, n;
  a();
  function _if() { if (x < y) {
    b();
    return e().then(function(_ref){ 
      d = _ref;
      f();
    });
  } else {
    m()
    return p().then(function(_ref){ 
      n = _ref;
      q();
    });
  }
  return _if().then(function(){
    r();
    return x;
  });
}
```
Line-to-line:
![Selection line-to-line comparison]
(https://s3.amazonaws.com/f.cl.ly/items/122t2M1Z2m0U2V3E3k1G/Screen%20Shot%202015-02-12%20at%2013.30.08.png)


#### Asynchronous repetition

```javascript
function() {
  var i, c;
  a();
  i = 0;
  while (i < 10) {
    b();
    c = await d();
    e();
    i++;
  }
  f();
  return g;
}
```
when desugared to promises becomes

```javascript
// Repetition
function() {
  var i, c;
  a();
  i = 0;
  function _while() { 
    if(i < 10) {
      b();
      return d().then(function(_ref){
        c = _ref;
        e();
        i++; 
        return _while();
      });
    } else { 
      return Promise.resolve();
    }
  } 
  return _while().then(function(){
    f();
    return g; 
  });
}
```
Line-to-line comparison:
![Repetition line-to-line comparison]
(https://s3.amazonaws.com/f.cl.ly/items/0b3k2h252F0U2h2i233k/Screen%20Shot%202015-02-12%20at%2013.38.14.png)

@jashkenas, is it possible for CS to generate such structures? Or you don't like this?

benzen:
I would be happy to see async await land in cs.
I didn't see how we could polyfill the Promise object in your PR, is this intentional ?

Artazor:
@benzen, yes, it is intentional, you should polyfill the Promise by yourself, as it is already landed in io.js and soon will be a standard feature. Also you can assign an alternative implementation into global Promise (e.g. Bluebird)

benzen:
I remembered seing a Talking from Jérémy about cs where he had a strong opinion about supporting env where only es 3 is awailable.
Does this mindset has changed?
If not, could we Provide a simple polyfill??

Envoyé de mon iPhone

GabrielRatener:
Support for async generators, seems like a logical feature to support if generators and async functions are both supported, am I right?

Artazor:
@benzen I think yes, it implicitly has changed since `yield` was added to the language. I think that if asynchronous functions will be implemented on top of generators, then any generator-related technology could assume ES6 environment (or at least the most important parts of it). 

On the other hand, as @GabrielRatener pointed out, it would be nice to have asynchronous generators. This way implies the strong orthogonality between the features. So if asynchronous functions will be implemented via Promises, then (if we also include a polyfill for promises into coffee)

**Pros:**

1. Async functions will be available in ES3 environment out of the box
2. Implementing asynchronous generators becomes trivial

**Cons:**

1. CoffeeScript should incorporate relatively fat polyfill in **every** source code that refers to mentioned features

This can be avoided if we allow the end user to add the Promise polyfill manually (factor out the Promises polyfill). I think it should be an end user responsibility to ensure an apropriate environment (i.e. emulate missing features) as we honor the **"no flags"** policy explicitly expressed by Jeremy.

alubbe:
I personally believe that Promises, async/await and their respective polyfills should not be a part of coffeescript. CS should "just" be JavaScript.

That being said, async/await seems to be really happening and CS should have a plan of how to handle it. Here is what I think that should look like:

```coffee
a = ->
  console.log await User.find id:123
```
should just give back
```js
var a;
a = async function() {
  console.log(await User.find({id:123}));
}
```
You as the consumer of CS should then handle this code - polyfill, babel, regex to coroutines & yield, ...

@jashkenas what are your thoughts?

michaelficarra:
> That being said, async/await seems to be really happening

A stage 1 ES7 proposal is "really happening" now? Look how far ES4 made it before being trashed.

GabrielRatener:
@alubbe The transformation of async functions to ES6 code is pretty clean. If you choose to target ES7, then you are really just targeting Traceur at the moment, cause there is no JS runtime which supports this feature currently. Traceur in my experience has been unreliable with this feature, and additionally generates copious amounts of code when it does work.

alubbe:
@michaelficarra From what you can read on the members' blogs, this is indeed happening.

@GabrielRatener I think any transformation should live in the "userland", so you can quickly improve speed/memory usage or add new features, akin to the many Promise implementations. If you dislike traceur, you can try babel or build your own transformation. 

To give you a clear example, I would want different solutions for the back and frontend. On the backend, the way to go is with bluebird's Promise.coroutine instead of any transformation/polyfill, since that gives you faster performance (bluebird can use internal resolve/reject methods instead of going through the .then api) and access to the longStackTraces features.

GabrielRatener:
@alubbe Regardless of the specific ES7 transpiler implementations. It seems to me better to implement this feature in ES6, since the generated ES6 will be able to run natively on many runtimes. Generated ES6 code can then be transpiled into ES3 code with tools like Traceur and Babel if necessary.

If you target ES7 then you unnecessarily force everyone using this feature to transpile the generated code.

lydell:
See also #3596.

dev-rke:
Just for the log: Dart has await/async now:
http://news.dartlang.org/2015/03/dart-19-release-youve-been-await-ing-for.html

GabrielRatener:
@dev-rke yup, looks like Dart shattered the ceiling on this one. Support for async/await, generators, async generators, and special for loops to iterate over them. Coffeescript should really add this stuff ASAP.

luisfontes:
:+1:

dev-rke:
@GabrielRatener it would be great to have a simple and stable implementation. I do not need any fancy stuff, i just want to keep code small, simple and readable. This is the reason why i like CS. 
And i also think that async / await could massively simplify huge asynchronous code structures.

So why the discussion stopped? 
Is it too complicated? 
Or uninteresting?

GabrielRatener:
@dev-rke I would love to add asynchrony support, but the devs seem unenthusiastic. I have an experimental [fork](https://github.com/GabrielRatener/coffeescript-next) that implements a rough equivalent of all 4 of Dart's function types. Note that it is not backwards compatible.

Artazor:
@GabrielRatener I think that discussion struggled on choosing an apropriate syntax. I suppose that we are waiting for the final word from @jashkenas.

FYI: async constructors for now can not be correctly typed in TypeScript (I will create an issue in MS repo, but chances are little that it will be supported in foreseeable future)

GabrielRatener:
@Artazor One can pretty easily create a base class that would work like async constructors. For example:

```Coffeescript
class AsyncBase
	constructor: (args...) ->
		resolve = () =>
			return this
		reject = (err) ->
			throw err

		@_promise = @async_constructor(args...)
		.then(resolve, reject)

	async_constructor: () ->
		await some_random_promise
		await some_other_promise

	then: (args...) ->
		return @_promise.then(args...)

class SomeAsyncClass extends AsyncBase
	async_constructor: (vars...) ->
		await super(vars...)
		@prop1 = await some_promise
		@prop2 = await some_dumb_promise
	method1: () ->
		do_stuff()
	method2: () ->
		do_other_stuff()

do ->
	 some_async_object = await new SomeAsyncClass()
	 some_async_object.method2()
	 some_async_object.method1()
```

so is it really necessary to add asynchrony to constructors natively?

Artazor:
@GabrielRatener yes. I just wanted to have a clean signature separation between an instance sign and a promise-for-instance. Anyway we can use static factories to perform any async parts of the constructors and let the constructors to be a simple initializers (marked with `api private`)

davidbau:
I desperately want a clean "await" implementation to use with my students (with penclicode.net).

I like the idea of getting it to work on ES3 environments.  However, would this approach let `await` appear inside arbitrary expressions?

```coffee
if (await readchoice('ready?')) then output [await readnum('A'), b(), await readnum('C')]
```

To get this to run in the right order would require unrolling all the subexpressions.  Is that beyond what we'd want the Coffeescript compiler to do?  My understanding is that traceur (and Iced Coffeescript) both avoid this problem by limiting "yield" (and await) so that it cannot appear inside arbitrary expressions.

To me #3757 seems like a simpler, more robust implementation for promise-based await.

I think the right place to implement an ES3-compatible promise-based Await is probably as an extension to Iced Coffeescript, along these lines: http://ideas.pencilcode.net/home/await.html .  Within ICS, it would have some limitations (no arbitrary expressions), but it would be something that could be deployed on old JS engines and could also be compatible with old ICS programs.  And it would provide a migration path to allow code and libraries to use promise-based await.

tony-kerz:
+1 to @dev-rke on:
> i just want to keep code small, simple and readable. This is the reason why i like CS. 
And i also think that async / await could massively simplify huge asynchronous code structures.

and barring devil-in-detail, i think what @alubbe says makes a ton of sense:
> I personally believe that Promises, async/await and their respective polyfills should not be a part of coffeescript. CS should "just" be JavaScript.

fwiw, i have no issue adding something like `babel` as a build step.

GabrielRatener:
@alubbe @tony-kerz First off, promises are standard in ES6. I think the basic idea behind "Its just JavaScript" is when programming in CS, the programmer should have a pretty good idea of what the generated JS will look like. @Artazor 's implementation of async functions just compiles them to JS generator functions wrapped in function calls, thats it. No state machines or any fancy AST transforms are used. CS already has polyfills it invokes to implement classes and other features (to work in ES3), so how is async/await any different?

tony-kerz:
@GabrielRatener i don't have the background to have a strong opinion on the actual implementation, i was basically just trying to resonate that:

1. even with es* closing the gap with c/s in terms of function, i still prefer the terse syntax of c/s
2. imho, in order for c/s to continue to be viable, there needs to be a clean way to access newer es* features like `async/await`
3. i, personally, would be willing to add transpilation steps (e.g. `babel`), if it eased the effort of implementation

DylanPiercey:
+1 to simply outputting es7 code with a compilation step.

dev-rke:
Management Summary:
We are all discussing since January about this async/await feature.
There are several implementations out there, we have an conflicted pull request here, discussed several improvements, but we don't know if this feature will be available in the next CS Release and we did not hear anything from @jashkenas .

What can we do / do we need to do, to get this useful feature into the next CS Release?

Mtpc:
ETA on this ?

phaux:
Please just make it so that `async ->` (or whatever syntax for async functions we choose) compiles to `async function(){}`. The `await` keyword works fine as it is.

The only problem with the current behaviour is that `async ->` compiles incorrectly to `async(function(){})`.

DylanPiercey:
I agree with @phaux. This would be trivial to implement, be closest to the spec and also be much clearer that the function returns a promise.

It would be annoying to have to scan through functions for the await keyword in order to tell if it is a async function (promise).

phaux:
It's also worth noting that the self-invoked functions inserted by coffee would need to be asyncified as well:

```coffee
updatePage = async ->
  # for-loop compiles to a function when used as expression
  items = ( await getJson "/item/#{i}" for i in [1..3] )
  doSomethingWith items
```

Should compile to

```js
updatePage = async function() {
  var i, items;
  // note the await/async keywords here
  items = await (async function() {
    var j, results;
    results = [];
    for (i = j = 1; j <= 3; i = ++j) {
      results.push(await getJson("/item/" + i));
    }
    return results;
  })();
  return doSomethingWith(items);
};
```

phaux:
I've found some use cases for the postfix `!` operator. It would work well with the current Fetch API:

Current:
```coffee
fetch '/gallery.html'
.then (res) -> res.text()
.then (html) -> html.match /<img src="(.*?)"/
.then ([all, url]) -> download url
```

Using `await`:
```coffee
html = await fetch '/gallery.html'
url = await html.text()
download url.match(/<img src="(.*?)"/)[1]
```

Using postfix `!`:
```coffee
download fetch('/gallery.html')!.text()!.match(/<img src="(.*?)"/)[1]
```

---

Another unrelated thought: since generators are marked with `*` and async functions used to be marked with `^`, why not use syntax like:

```coffee
*-> # generator
^-> # async function
^*-> # async generator
```

GabrielRatener:
+1 for a postfix await operator.

On Wed, Sep 23, 2015 at 10:57 AM, Nathan Stefaniak <notifications@github.com

DylanPiercey:
-1 for postfix. Hard to skim for promises.

```javascript
await (await fetch(...)).json()
```

Seems fine.

drewhamlett:
@michaelficarra, async / await is moving right a long.  Currently at Stage 3.  Stage 4 ("Finished") by the end of November

http://tc39.github.io/ecmascript-asyncawait/

brandonaaskov:
@drewhamlett thanks for the update! Can't wait :+1:

davibe:
Coffeescript should support await and compile to js code containing async/await just like it supports yield and compiles to generator*/yield. Then if you want to run it on an engine that does not support ES7 you can transpile the js code as a second stage. It would be strange if coffee compiler tried to polyfill new features.

That being said in the mean time we can do something very similar to await with things that are already supported with little syntax overhead like this https://github.com/davibe/async-js-coffee/blob/master/src/4_generators_promises.litcoffee . Then once await will be supported its going to be simple to adapt things

GabrielRatener:
I would like to see a backwards compatible syntax introduced to CS < 2.0. Essentially an alternative to the `await` keyword for awaiting a promise, such as a postfix `!` perhaps. This could compile to generators which are now widely supported. Then, when CS 2.0 is released the implementation could switch to using JS async functions, and the `await` keyword would be added to CS as well.

???

barretron:
I like the idea of tapping into CoffeeScript's functional ancestry and going with a reflected "Monad sequencing operator" from Haskell.

```
text = ->
  (JSON.parse << http.get "/service.json").text
```
Resulting in:
```
var text;

text = async function() {
  return JSON.parse(await http.get("/service.json")).text;
};
```
The nice touch here is that assignment to a variable is also a reflection of the Haskell monad sequencing with assignment (`>>=`) operator, e.g.:

```
textA = ->
  service =<< http.get "/service.json"
  (JSON.parse service).text
```

mikeyhew:
We should start compiling async/await pretty soon - at the time of writing (December 2015), it has been implemented in one browser, Microsoft Edge, and I'm guessing we'll see implementations in more of the major javascript engines by the end of 2016.

I don't think we need to do anything fancy, like compiling async and await to es6 code à la iced-coffee-script, or by automatically restructuring code to use promises as @Artazor proposed. While I like the idea, and was actually trying to figure out how to do that earlier today before I found this discussion, I think it belongs in a fork of coffeescript. Instead just compile straight to ES7's async function syntax. From a practical perspective, there's already tools out there, i believe, that will [regenerate](https://github.com/facebook/regenerator) the compiled javascript code.

- I'm a bit iffy about using a new symbol for async functions instead of just `async ->`, because it seems like it would be confusing to read code with it. But it might actually be pretty easy to get used to, and the payoff of not having to type the extra characters every time might be worth it. I think a good candidate would be `a->`, since the letter `a` is much more meaningful than any special character could be. 
- The postfix `!` for `await` as @phaux mentioned above seems interesting, but confusing, and I don't know that there's very many use cases where you'll be chaining method calls on await expressions without saving the result to a local variable, like in the above example. Plus, an exclamation point isn't nearly as clear as the actual english word "await".
- What definitely *wouldn't* be a good idea, in my opinion, is *implicit* async functions, which are magically changed from normal functions by the existence of an await expression within the function. That just seems like a new source of confusion for anyone reading the code.

@drewhamlett thanks for the update and the link.

Food for though: since coffeescript's `class`es are themselves functions, is there any useful concept of async classes?

davibe:
about your third point: We already have "implicit generators" when the function contains `yield` tho

dustinrjo:
I agree with @phaux 

> Please just make it so that async -> (or whatever syntax for async functions we choose) compiles to async function(){}. The await keyword works fine as it is.

> The only problem with the current behaviour is that async -> compiles incorrectly to async(function(){}).

Using more symbols for different kinds of functions in CS when a keyword is used in JS will just get confusing for people to learn CS and ultimately make it harder to read.  For generators this was different because the * is used in JS as well, for better or worse.  

Staying as close to JS as possible is really important to sell people on using CS for projects.  I would only trade words for symbols if any bum off the street could look at the change and say "now that is super helpful and I can remember that".  Otherwise we are going backwards.  There is a reason `||` became `or` and `&&` became `and` and `async` should be `async`.  Let's not lose that.

mpseidel:
+1 for any solution that allows nice async control flow with coffee. I would love to use coffee for all my tests insteaf of typescript but I need a clean way to await stuff for the tests to stay clean and easy to read.

taoeffect:
+1 for `!` postfix operator. It fits nicely with CS's existing succinctness and has precedent with the succinctifying nature of the `?` postfix operator.

AJeezy9:
+1-ing in hopes that CS can somehow be resuscitated to implement new ES6/7 features. 

Increasingly prevalent use of async/await & decorators are starting to force my hand to make the switch, but I much prefer CS.  

ES6 does more to obfuscate than improve javascript, and its grammatical conventions remain far more tedious than Coffeescript.

DylanPiercey:
@AJeezy9 totally agree, was sad to move away from CoffeeScript. Personally I would love something like "babel-lite" where a huge amount of features are forcibly removed. The only thing that I always really miss from CS is the "one-page" docs (much less syntax than js) and the "everything is an expression" which works so well with JSX and functional programming.

phaux:
@dustinrjo 

>Staying as close to JS as possible is really important to sell people on using CS for projects.

I don't necessarily agree on that. I'd prefer Coffee to stay one step ahead of JS and make things prettier and more concise. `^->` and `*->` look cool to me. `async ->` might be confused with a function taking single `async` argument.

@DylanPiercey 

>-1 for postfix. Hard to skim for promises.

Then how about using a different sigil? (`~^` as an example)
```coffee
download fetch('/gallery.html')~^.text()~^.match(/<img src="(.*?)"/)[1]
```

kirly-af:
Concerning the existence of a postfix syntax in general, it's not that I'm against the idea but I don't feel the need for this. Though, I didn't find any of the suggested operators appealing.

I personally feel `!` as a terrible operator choice of its usual meaning in most programming languages (ie: not). Considering await is a new feature and potentially not widely used for a while, it would bring even more confusion for people discovering ES* features or to CoffeeScript new comers (not even mentioning people in both cases). It might look to be a simple detail but I'm sure I'm not the only one who faced 

 Besides, to me all the pseudo-arrow usage result in ugly syntaxes. I would hate to write something like: `fetch('/gallery.html')~^.text()` (very ugly) or whatever similar to `fetch('/gallery.html')*->.text()` (quite ugly and too close to classic arrows).

I find way more elegant the idea of an operator such as  `<-` or `<<` (suggested by @barretron), but in the end I think the simple use of the `await` keyword is completely fine.

I'm new to this discussion, and I'm a bit afraid by the time it's taking to add a feature that is said to be ready (forgive me if I missed a post saying the opposite, the thread got quite long now, so please be indulgent if it is the case).

I love CoffeeScript for its elegant syntax like most people on this thread, but what I primarily like a language for is the features it provides. Wouldn't it be more logical to add the feature with the simple `await` keyword so we can start playing with it, while keep thinking about syntactic sugar that could be added later on ? Many people share this point of view but after more than a year it looks like we're still not getting anywhere. Appart from that, has @jashkenas ever shown any interest for this feature ?

DomVinyard:
@mikeyhew said:

> What definitely wouldn't be a good idea, in my opinion, is implicit async functions, which are magically changed from normal functions by the existence of an await expression within the function. That just seems like a new source of confusion for anyone reading the code.

I disagree, where would the confusion arise? Implicit async definitely seems like the way to go - if it's a function, it's a function. If you see 'await' then you know that line of code is 'awaiting'... It seems like the only elegant solution proposed.

GabrielRatener:
@DomVinyard having both implicit generator functions with `yield`, and implicit async functions is challenging. For generators without yields, there are `yield return` statements, for implicit async functions there may be an `await return` analog. When async generator functions are implemented, what would be its analog?

Even Python 3.5 went with explicit syntax for async functions.

mikeyhew:
@DomVinyard I've mostly changed my mind since I made that comment. At this point, I don't think it would be the end of the world if a function returns a promise because it contains an `await` expression. Sure, it might make things a bit more confusing at first, but it would be better to be consistent with generators, and there's something to be said for using the same `->` for all types of functions and not having to write `async ->` or invent another syntax for async functions. And the improvement in readability that arises when you switch from explicit promises to using `await` really trumps any regressions that this could cause.

DomVinyard:
@GabrielRatener No mixing `yield` with `async`. Compile time error. For some edge cases it's problematic and would require coding around, granted. But, ultimately, async is too powerful to not bake into default coffeescript, it's too natural an evolutionary language feature.

mikeyhew:
@GabrielRatener said:
>Even Python 3.5 went with explicit syntax for async functions.

 yes, python 3.5 uses `async def` for coroutines but just `def` for generators. But why is that? I thought it was just because they added coroutines later on.

DomVinyard:
@mikeyhew And it would reduce the friction for people just starting out with async, they can throw an await call into an existing production function without the mental overhead of what some would consider "refactoring the whole function". @jashkenas Count me +1

GabrielRatener:
@mikeyhew The reasons for adding explicit syntax in Python are described [here](https://www.python.org/dev/peps/pep-0492/#importance-of-async-keyword). I think both explicit and implicit syntax would be better than nothing, but explicit syntax really indicates that the functions are fundamentally different.

@DomVinyard Async generators will be in JS eventually, so shouldn't accommodations be made for the future syntax? Also, implicit syntax can cause friction and confusion for people who understand that JS is single threaded. They see rogue `await` statements that magically appear to make functions block. Both syntaxes can be confusing to different people.

maxtaco:
FWIW, @zapu and I recently rewrote IcedCoffeeScript to use ES6 generators. The compiler and transpiled code are significantly simpler, but the semantics remain the same. Code is available in our [iced3 branch](https://github.com/maxtaco/coffee-script/tree/iced3) or via npm at [iced-coffee-script-3](https://www.npmjs.com/package/iced-coffee-script-3).

DomVinyard:
@GabrielRatener 

> They see rogue await statements that magically appear to make functions block. Both syntaxes can be confusing to different people.

I totally agree, I'm making my case based on the idea that coffeescript is appealing to the very people who prefer the magic over the explicit, fans of abstraction. Although I do take your point that it would be alienating to some, I think it's absolutely worth the conceit.

In the same way that - if an explicit form won out - I think there are a group of people that would use the async function form by default 'just in case'. Like many do with the fat arrow.

kirly-af:
@DomVinyard
> if an explicit form won out - I think there are a group of people that would use the async function form by default 'just in case'. Like many do with the fat arrow.

Such cases are easily handled with a linter. I don't see this point as a valid reason.

I have a preference for the implicit syntax (without the async keyword). As many said, it keeps the logic of implicit generators. However I have no idea if it could be a problem to later bring async generators to coffeescript as some were discussing.

DomVinyard:
@mikeyhew 

The link to the Python discussion was interesting, thanks for posting up

> Importance of "async" keyword
> While it is possible to just implement await expression and treat all functions with at least one await as coroutines, this approach makes APIs design, code refactoring and its long time support harder.
> Let's pretend that Python only has await keyword:
> If [a] function is refactored and someone removes all await expressions from it, it would become a regular python function, and all code that depends on it, [...] would be broken. To mitigate this issue a decorator similar to @asyncio.coroutine has to be introduced.

Am I right in saying that this limitation doesn't apply to ES7 because

`var three = await 3;
console.log(three === 3) // true
`

GeoffreyBooth:
Closed in favor of #3757.

PR #3814, opened 2015-01-28
Update README.md

xfq:
"npm" is not an abbreviation for "Node Package Manager". It is a recursive bacronymic abbreviation for "npm is not an acronym". See https://docs.npmjs.com/misc/faq for details.


michaelficarra:
It's a descriptor, not an expansion of an implied abbreviation.

PR #3821, opened 2015-01-30
Name generated variables without leading underscore

### LINKED ISSUE #3816: In 1.9.0, Coffeescript breaks Angular dependency injection
If I had the following code:

```coffee
beforeEach ->
  module 'foo'
  inject (@$httpBackend, @$templateCache, @foo) =>
```

**1.8.0** compiles it to:

```js
beforeEach(function() {
  module('foo');
  return inject((function(_this) {
    return function($httpBackend, $templateCache, foo) {
      _this.$httpBackend = $httpBackend;
      _this.$templateCache = $templateCache;
      _this.foo = foo;
    };
  })(this));
});
```

While **1.9.0** produces:

```js
beforeEach(function() {
  module('foo');
  return inject((function(_this) {
    return function(_at_$httpBackend, _at_$templateCache, _at_foo) {
      _this.$httpBackend = _at_$httpBackend;
      _this.$templateCache = _at_$templateCache;
      _this.foo = _at_foo;
    };
  })(this));
});
```

The issue is that Angular parses the function parameters to figure out which dependencies to inject into the test. `$httpBackend` is a valid dependency, but `_at_$httpBackend` is not, breaking the test.

lydell:
For example, `ref` not `_ref`. It's cleaner.

This also fixes #3816.


jashkenas:
Sweet!

xixixao:
Imho this will make CS generated JS hell to debug, thanks to variables I didn't declare but are indistinguishable from mine (plus it's a bad hack fix for the Angular issue).

:-1:

vendethiel:
> Imho this will make CS generated JS hell to debug, thanks to variables I didn't declare but are indistinguishable from mine (plus it's a bad hack fix for the Angular issue).

Shouldn't source maps handle this? Otherwise I agree, it's useful to be able to differentiate compiler-generated variables and the others.

xixixao:
@vendethiel Of course I am talking about the case where you are not debugging via source maps. Let's not kid ourselves, that will still happen :)

Issue #1334, opened 2011-05-07
Proposal: Dot notation for numerical property names


jed:
Just an idea, but is there any reason why

```
first = arguments.0
```

(c|sh)ouldn't be converted to

```
first = arguments[0];
```

?

It always bugged me that only numbers need to use the bracket syntax despite a lack of ambiguity... Or am I missing something?

utkarshkukreti:
@GeoffreyBooth would you accept a PR implementing this feature?

utkarshkukreti:
Just saw #3828. I can try updating that PR to work with CoffeeScript 2 if you'll accept it!

vendethiel:
@utkarshkukreti That'd be welcomed. https://github.com/jashkenas/coffeescript/pull/3828#issuecomment-298298122

GeoffreyBooth:
@utkarshkukreti yes that would be great. My only word of caution is that a new PR can’t introduce any breaking changes. There was some discussion elsewhere of no longer allowing floats with leading decimal points, e.g. `.5` instead of `0.5`; we need to avoid breaking changes like that.

GeoffreyBooth:
Closing per https://github.com/jashkenas/coffeescript/pull/5006#issuecomment-559665873

Issue #1982, opened 2011-12-28
Variable name with any character and whitespace


ghost:
These features kinda have something in common so I just put them in the same request:

Feature 1:

One feature I would like to see is the ability to use whatever character as name eg.

```
ƒ = (name) -> ...

ƒ "load models for view"
```

or

```
π = 3.14
```

This is already possible (also means that it's simple to do in JS) if you are doing like this:

```
o =
  "ƒ": (name) -> ...

o["ƒ"] "load models for view"
```

If this was a built in feature we wouldn't have to declare it as an object method and also don't have to call it with brackets.

Why should $ be allowed and not ß Ω £ π and others.

I can't think of any drawback with this feature since it's already possible, with more code. CS would just make this simpler so the code gets prettier.

Feature 2:

Another thing I would like to see is acceptance of whitespace in variable names eg:

```
`create a new app` = (app) -> ...

`create a new app` "test"

o =
  "create a new app": (app) ->

o.`create a new app` app
```

These are just suggestions borrowed from Ruby where it runs a shell command:

```
`echo hello`
```

Also Neo4j is using them to allow attributes with whitespace:

```
person.`first name`
```

Why I think it's a good suggestion.

I know it's an old habit to use camelCased variable names but I just don't know why we still have to do that. It collides with literate programming. Also being forced to hit SHIFT for each word just doesn't make sense, not to mention that this just makes the whole program _much_ more fun to read. Just read it like you read a book. Why suddenly reading sentences in camelCase or preventing variable names to be too long (thus unreadable)?

I could only imagine that this would open a new door for literate programming style in CS. No need to declare a function and have a comment above it explaining what it does. The code itself should be the documentation.

And as mentioned you can already have whitespace but a lot of programmers don't want to do something like:

```
o["create a new app"] app

o["load view for page"] = (page) -> ...
```

why everyone still prefers camelCasing the name.

Javascript is very flexible under the hood but it's not smart enough to allow users to use the cool features with ease. "Smell but don't taste".

For me this is something CS could take care of very easily and would make everyone love to have descriptive names thus indirectly making the whole app more of a literate program, for free.

Issue #383, opened 2010-05-18
Numbers that start with . not recognized


TrevorBurnham:
Is it an intentional language design decision that numbers like .75 aren't recognized by the parser? You have to write 0.75 instead.

I guess you could justify this on "Only One Way To Do It" grounds, but given the uninformativeness of the parser error messages, I think it would greatly reduce confusion if .75 were seen for what it is.

PR #3828, opened 2015-02-03
[WIP] Allow foo.'bar-bat' and foo.1 to access object members

### LINKED ISSUE #3520: Cleaner syntax for accessing "illegal" object members: object.'some-value'
This comes up a lot when working with compiled templates in SPAs and I imagine in other places as well. We end up with template objects that have a lot of members who's names are not legal JavaScript identifiers because they are based on the original file names of the templates.  Needing to wrap brackets and quotes around the member names seems like more syntax that is needed.

Would it be possible to make:

```
templates.'my-template-name'
#and
templates."another-template-name"
```

valid syntax equivalent to:

```
templates.['my-template-name']
#and
templates.["another-template-name"]
```

pscollins:
This change addresses  #3520. I wasn't sure if there is a consensus on that issue, but I think that it's a simple change that could clean up code in some places.


epidemian:
:+1: 

This makes for a nice symmetry with literal object notation.

```coffeescript
obj = 
  # Can use barewords for keys that look like identifiers.
  foo: 42
  # But need enclosing commas for non-identifier-like keys.
  'wordy key': 43

# Can access identifier-like keys by a bareword after the dot.
obj.foo # => 42
# But need enclosing commas for accessing non-identifier-like keys.
obj.'wordy key' # => 43
```

jashkenas:
I'm not sure really what to think about this. There is a certain symmetry, but there's also a certain awkwardness, especially when it comes to more complicated accesses:

    configuration.'ajax-get-request'(url, options)

... that's just odd. Any staunch defenders of this idea?

vendethiel:
It works in coco/LS, but I admit I don't use it that much. It's only interesting if you have to deal with "strange" json schemas...

lydell:
_If_ this is going to be supported, numbers should be as well. `a.1`

michaelficarra:
@lydell: Agreed. I'm :+1: for both of them.

pscollins:
@lydell maybe it would be easiest to do something like

    PropertyIdentifier ::= Identifier | AlphaNumeric 

and then replacing the existing uses of `Identifier` with `PropertyIdentifier` in `Accessor`?

pscollins:
@lydell maybe it would be easiest to do something like

    PropertyIdentifier ::= Identifier | AlphaNumeric 

and then replacing the existing uses of `Identifier` with `PropertyIdentifier` in `Accessor`?

lydell:
> especially when it comes to more complicated accesses:
> 
>     configuration.'ajax-get-request'(url, options)
> 
> ... that's just odd.

I always try to avoid the `a[b](c)`, because IMO it is not very readable. `configuration.'ajax-get-request'(url, options)` isn’t more odd than `configuration['ajax-get-request'](url, options)` or the syntactically valid JS `"string"()`.

jashkenas:
Alright then. I'm persuaded. Let's add numbers to this patch, and some more robust tests, and merge.

michaelficarra:
:grinning:

lydell:
@pscollins Please do what you suggested and add tests for all the access variants (not just `.`) and tests that you can call such accesses as well as access deeper properties.

pscollins:
@lydell `::` and `?::` as well, or just `.` and `?.`? My guess is no, since things accessible through `::` won't have been written as object literals, and so shouldn't have crazy names.

michaelficarra:
@pscollins: Of course those should be allowed.

```coffee
a::' ' = ->
a.prototype = ' ': ->
```

jashkenas:
Yes. If it's for any accessor, it should be for all accessors. And tested ;)

pscollins:
I've got it working and tested for every form except 

     obj.NUMBER 

It's listed in the testcases I just pushed, commeted out. Uncommenting it yields:

    test/objects.coffee:474:9: error: unexpected .42
    eq obj.42, 3

I assume that what's happening is `Lexer#numberToken` is consuming the `.` and lexing it as `IDENTIFIER NUMBER` rather than `IDENTIFIER . NUMBER`. 

I'm not sure how this could be dealt with in the parser. The best case I can imagine would be adding in a special `IDENTIFIER . [0-9]+` case in the lexer that comes before `numberToken` is called, but that seems a little hacky. 

It's also worth noting that something like 

    obj = 
      4.1: 1
      4:
        1: 2

will be ambiguous. 

What's the best way to approach it? Make a special case for numeric accessors, and forbid non-integers? Require that numeric accessors go in parens?

jashkenas:
Re: the parser error, I don't know — but as to your second question...

    obj = 
      4.1: 1

... would simply be:

    obj = 
      "4.1": 1

no? I don't see the ambiguity.

pscollins:
@jashkenas Sorry, I didn't mean that defining the object would be ambiguous -- I meant accessing it would be ambiguous. 

My worry is that `obj.4.2` can be interpreted either as `obj[4][2]` or `obj[4.2]`. If your object looks like `obj = {4.2: "foo"}`, you're going to be confused if `obj.4.2` throws an error for trying to access the property `2` on the undefined `obj[4]`. If your object looks like `obj = {4: {2: "foo"}}`, then you're going to be confused if `obj.4.2` gives you back `undefined` for trying to access `obj[4.2]`. 

Or you might have 

    obj = 
      4:
        2: 1
      4.2: 2

Now, what should `obj.4.2` be? 1, or 2? 

It seems to me like it would be easy for someone new to the language to get tripped up by floating point numbers behaving differently from other kinds of numbers, but maybe I'm overreacting?

michaelficarra:
It should be a decimal integer, not floats, not hex, not scientific notation, etc. So my answer is it would be `obj[4][2]`.

src/grammar.coffee
```diff
@@ -258,13 +258,20 @@ grammar =
...
   # or by array index or slice.
   Accessor: [
-    o '.  Identifier',                          -> new Access $2
-    o '?. Identifier',                          -> new Access $2, 'soak'
-    o ':: Identifier',                          -> [LOC(1)(new Access new Literal('prototype')), LOC(2)(new Access $2)]
-    o '?:: Identifier',                         -> [LOC(1)(new Access new Literal('prototype'), 'soak'), LOC(2)(new Access $2)]
+    o '.  AccessorIdentifier',                  -> new Access $2
+    o '?. AccessorIdentifier',                  -> new Access $2, 'soak'
+    o ':: AccessorIdentifier',                  -> [LOC(1)(new Access new Literal('prototype')), LOC(2)(new Access $2)]
+    o '?:: AccessorIdentifierIdentifier',       -> [LOC(1)(new Access new Literal('prototype'), 'soak'), LOC(2)(new Access $2)]
```
lydell: Typo?

lydell: It's correct in the compiled JS though. Remember to rebuild everything.

lydell:
I think `a . b` is valid, so there should be tests for `a . 'abc'` too

lydell:
> and tests that you can call such accesses as well as access deeper properties

You forgot this.,

src/grammar.coffee
```diff
@@ -258,13 +258,20 @@ grammar =
     o 'This'
   ]
 
+  # The symbols that can be used to access an object member:
+  # identifiers, string literals, and numbers.
+  AccessorIdentifier: [
+    o 'Identifier'
+    o 'AlphaNumeric'
```
lydell: A possible solution for the number problem could be to replace this line with:

```coffee
o 'STRING'
o 'INTEGER'
```

Then, in `AlphaNumeric` replace `o 'NUMBER'` with `o 'Number'` and add:

```coffee
Number: [
  o 'NUMBER'
  o 'INTEGER'
]
```

and adjust the `NUMBER` regex in the lexer to match integers separately from other decimal numbers and make `Lexer::numberToken` tag integers as such.

michaelficarra:
I forgot we even allowed leading dots in number literals. Who thought that was a good idea?

lydell:
> I forgot we even allowed leading dots in number literals.

Me too. That’s really making things more complicated here.

lydell:
I think the best way to solve it is to add `return 0 if number.charAt(0) is '.' and @tag() in INDEXABLE` in `Lexer::numberToken`.

jashkenas:
> I forgot we even allowed leading dots in number literals. Who thought that was a good idea?

It's not a good idea ;) We shouldn't have let that sneak in.

lydell:
See #383.

jashkenas:
> It should be a decimal integer, not floats, not hex, not scientific notation, etc. So my answer is it would be obj[4][2].

Agreed.

lydell:
> We shouldn't have let that sneak in.

That makes it sound like they were never intended but there were no tests for it and during some refactor they were allowed by mistake. Since that's not the case, do you think that we should remove them (which makes this PR a lot simpler to finish, but likely results in 10 duplicate issues the same day it is released) or let them stay?

jashkenas:
Ha ha. There's a certain class of feature (of which there are dozens of examples in CoffeeScript) that I intentionally didn't include at the beginning because they're not my personal preferred style ... but that managed to sneak in over time because people ask for them, and you say "why not?". 

This was one of those.

I don't think we can remove them at this point. But perhaps we could tag them specially in the lexer (or something) to make this change easier.

michaelficarra:
Numbers with a leading `.` instead of `0.` are exceptionally uncommon in JS, and I bet even more so in CoffeeScript. Same goes with numbers with trailing `.`. I think we can safely remove them with a simple comment in the changelog.

jashkenas:
I'd love it if that were really the case. My anecdotal sense is that lots of folks prefer to write `speed + 0.5` as `speed + .5`.

Can we employ Github's code search to find out if folks are using it in CoffeeScript?

lydell:
At least in CSS I’ve seen lots of `.5`s. (Don’t know about JS though.)

lydell:
https://github.com/jashkenas/coffeescript/issues/1982#issuecomment-3314044

lydell:
Just found that this would close #1334 as well.

shreeve:
Any status on this PR? Seems like the issue was related to a possible breaking change if some people used numbers like this: ```x = y + .25```. Is breaking this use case the reason for this PR not being accepted?

lydell:
1. It seems like it was decided that disallowing `.5` etc. is too much of a breaking change to be a real option.
2. Some trickery is required to get `foo.1.2` working.
3. There are lots of merge conflicts.

pscollins:
@shreeve IIRC I had trouble getting it to work in the `.5` case and I got busy with other projects --- feel free to revive this/close this PR for a new one/etc.

GeoffreyBooth:
FYI we should probably put this on hold until [tagged template literals](https://github.com/coffeescript6/discuss/issues/28) are supported, unless we’re certain that there won’t be any ambiguity between this proposed syntax and the syntax we’ll need for that. I would think tagged template literals should have a much higher priority if we’re forced to choose between them.

GeoffreyBooth:
@pscollins or anyone on this thread, I’d love to get this updated and targeted against the `2` branch. It seems like it was rather close?

GeoffreyBooth:
What’s going on with this? Is this something we’re still interested in getting finished, or should I close this as an abandoned effort? If we are trying to finish it, are there any breaking changes that would result from merging this in?

michaelficarra:
I can't speak to the quality of this PR, but this is a great feature and should be pursued.

GeoffreyBooth:
At this point the codebase has changed so much since this PR was started that a new PR will be necessary, so I’m closing this one. I’m leaving issue #1334 open as a reminder that we’re interested in pursuing this; that issue has a link to this thread.

PR #3830, opened 2015-02-05
Fix incorrect token representation

swang:
The third element in a token should just be an object containing line number and column info. This PR fixes the problem with one of the tokens being set incorrectly.


lydell:
Nice catch! Thanks.

PR #3833, opened 2015-02-05
Fix #3795: Never generate invalid strings and regexes

### LINKED ISSUE #3795: Invalid generated string literals
It is currently possible to write CoffeeScript that results in invalid JavaScript when it comes to strings literals:

- Both the U+2028 and the U+2029 characters are passed through into the JavaScript output. However, in JavaScript they are line terminators (just like \n and \r), and string literals may not contain (unescaped) line terminators. Redux handles this case by converting those characters to the escape sequences `\u2028` and `\u2029`. Another way of handling it could be to just like JavaScript allow them to be line terminators, and normalize them into \n characters in `Lexer::clean`, similarly to how \r\n is handled.

- Invalid escape sequences beginning with `\u` or `\x` are also passed through into the JavaScript output, but that is syntactically invalid. Redux handles this by turning the `\u` into `u` and `\x` into `x`, making them work like most other characters (such as `\a`, `\+`, etc.). An argument against that could be that if a string literal contains `a\u002b` the author probably meant to type `a\u0020b` and might prefer an error (just like in JavaScript). This could be checked together with octal literals in `Lexer::makeString`.

lydell:
- Invalid `\x` and `\u` escapes now throw errors.
- U+2028 and U+2029 (which JavaScript treats as newline characters) are now
  escaped to `\u2028` and `\u2029`, respectively.
- Octal escapes are now forbidden not only in strings, but in regexes as well.
- `\0` escapes are now escaped if needed (so that they do not form an octal
  literal by mistake). Note that `\01` is an octal escape in a regex, while `\1`
  is a backreference. (Added a test for backreferences while at it.)
- Fixed a bug where newlines in strings weren't removed if preceded by an
  escaped character.


jashkenas:
Dude. Lovely stuff.

I particularly enjoy how heregexes are useful for clarity while working on regex escaping.

PR #3834, opened 2015-02-06
Improve lexer error messages

lydell:
- Erraneous tokens are now fully underlined with ^:s.
- The error messages are now a bit more consistent.

### DIFF
--- a/test/error_messages.coffee
+++ b/test/error_messages.coffee
@@ -247,14 +247,14 @@ test "unclosed strings", ->
   """, """
     [stdin]:1:1: error: missing '''
     '''
-    ^
+    ^^^
   """
   assertErrorFormat '''
     """
   ''', '''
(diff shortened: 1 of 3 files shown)


test/error_messages.coffee
```diff
@@ -378,7 +378,7 @@ test "octal escapes", ->
   ''', '''
     [stdin]:1:10: error: octal escape sequences are not allowed \\07
     "a\\0\\tb\\\\\\07c"
-      \  \   \ \ ^
+      \  \   \ \ ^\^^
```
michaelficarra: What's with these backslashes?

jashkenas:
Fantastic! Yeoman's work!

test/error_messages.coffee
lydell: Yep.

PR #3840, opened 2015-02-07
Fix #3597: Allow interpolations in object keys

### LINKED ISSUE #1131: Clarify error messages for attempted key interpolation
$ coffee -bpe '{"#{key}": val}'
    Error: Parse error on line 1: Unexpected '('

and in a weird converse

    $ coffee -bpe "(foo): bar"
    Error: Parse error on line 1: Unexpected '{'

I understand why string interpolation in hash keys is (no longer) supported. But as the CoffeeScript compiler matures, I'd love to see more specific, helpful error messages—or at least ones that don't refer to punctuation that isn't even in the original input.

For instance, I'd love to see more compiler errors like

    $ coffee -bpe "i for x, i in [1..3]"
    SyntaxError: indexes do not apply to range loops

It doesn't *fully* explain why `for x, i in [1..3]` is invalid while `for x, i in [1,2,3]` is valid—no short error message could—but it at least gives a newcomer to the language a starting point for answering that question.

### LINKED ISSUE #3039: Interpolated dynamic keys are allowed if the interpolation string is empty

### LINKED ISSUE #3597: Allow dynamic keys for object (hash) definition

lydell:
The following is now allowed:

    o =
      a: 1
      b: 2
      "#{'c'}": 3
      "#{'d'}": 4
      e: 5
      "#{'f'}": 6
      g: 7

It compiles to:

    o = (
      obj = {
        a: 1,
        b: 2
      },
      obj["" + 'c'] = 3,
      obj["" + 'd'] = 4,
      obj.e = 5,
      obj["" + 'f'] = 6,
      obj.g = 7,
      obj
    );

- Closes #3039. Empty interpolations in object keys are now _supposed_ to be
  allowed.
- Closes #1131. No need to improve error messages for attempted key
  interpolation anymore.
- Implementing this required fixing the following bug: `("" + a): 1` used to
  error out on the colon, saying "unexpected colon". But really, it is the
  attempted object key that is unexpected. Now the error is on the opening
  parenthesis instead.
- However, the above fix broke some error message tests for regexes. The easiest
  way to fix this was to make a seemingly unrelated change: The error messages
  for unexpected identifiers, numbers, strings and regexes now say for example
  'unexpected string' instead of 'unexpected """some #{really long} string"""'.
  In other words, the tag _name_ is used instead of the tag _value_.
  This was way easier to implement, and is more helpful to the user. Using the
  tag value is good for operators, reserved words and the like, but not for
  tokens which can contain any text. For example, 'unexpected identifier' is
  better than 'unexpected expected' (if a variable called 'expected' was used
  erraneously).
- While writing tests for the above point I found a few minor bugs with string
  locations which have been fixed.


lydell:
I forgot destructuring tests.

src/rewriter.coffee
```diff
@@ -93,24 +93,31 @@ class exports.Rewriter
...
+  # Returns `yes` if standing in front of something looking like
+  # `@<x>:`, `<x>:` or `<EXPRESSION_START><x>...<EXPRESSION_END>:`,
+  # skipping over 'HERECOMMENT's.
   looksObjectish: (j) ->
-    @matchTags(j, '@', null, ':') or @matchTags(j, null, ':')
-
-  # yes iff current line of tokens contain an element of tags on same
+    return yes if @indexOfTag(j, '@', null, ':') > -1 or @indexOfTag(j, null, ':') > -1
+    index = @indexOfTag(j, EXPRESSION_START)
+    if index > 0
```
lydell: I should probably compare to `-1` here. Wonder if it makes a difference.

lydell:
Fixed the issues I found.

michaelficarra:
LGTM.

lydell:
Fixed merge conflict.

jashkenas:
Fantastic!

Issue #3829, opened 2015-02-05
Cannot set property 'root' of null


asafbrukarz:
My apologies if this is the wrong place to put this, but I can't find the git repo for coffee-script-source.

We have a Rails 4.1.9 app and a strange error has started to occur after upgrading our dependencies. 

With the latest version of `gem 'coffee-script-source', '=1.9.0'` we receive the following error when trying to precompile our assets.

We have confirmed that rolling back to `v1.7.1` fixes the issue.

I believe that this was caused by #3498.

We have looked at the offending file `tags.js.coffee` and could find no obvious errors. When we remove that file, the same error occurs on another file.

```
> rm -rf public/assets/ && rm -rf tmp/cache/
> RAILS_GROUP=assets RAILS_ENV=production bundle exec rake assets:precompile
...
...
...
rake aborted!
ExecJS::ProgramError: TypeError: Cannot set property 'root' of null
  (in /Users/a/Development/pc/app/assets/javascripts/widgets/tags.js.coffee)
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:58:in `rescue in block in call'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:52:in `block in call'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:87:in `block in lock'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:85:in `call'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:85:in `Locker'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:85:in `lock'
/Users/a/.rvm/gems/ruby-2.1.4/gems/execjs-2.3.0/lib/execjs/ruby_racer_runtime.rb:51:in `call'
/Users/a/.rvm/gems/ruby-2.1.4/gems/coffee-script-2.3.0/lib/coffee_script.rb:76:in `compile'
[...]

PR #3851, opened 2015-02-12
Rebuild browser file

mbrio:
While looking into the issues that were plaguing reactjs/react-ruby (https://github.com/jashkenas/coffeescript/issues/3829#issuecomment-74134197) I found that the browser JS file did not match the compiled output from the 1.9.0 source code. I believe this is why the error was occurring.


michaelficarra:
I did a `git checkout 1.9.0` and ran `cake build:browser`. It caused no changes to `extras/coffee-script.js`, and is different than the file in this PR. Closing as I cannot reproduce this issue or the submitted file in this PR.

mbrio:
Admittedly, I didn't git checkout 1.9.0 I think I was using master. The error above was definitely fixed when I recompiled the latest code's browser file. So either the error has been fixed and not released or could an update to a third party package fix it. Unfortunately I'm not in front of a computer, when the JS is compiled is it injecting JS from other libraries?

I only mention this because the error is occurring when assigning a root property on a null object. In the coffee src, root only appears in one place, but in the compiled JS root appears numerous times.

If you checkout 1.9.0 and re-install your node_modules, is the output still the same? This would mean that the error actually originated from a third-party module.

michaelficarra:
I did re-install my node_modules, and it built an identical file to what is in 1.9.0. It is possible this has been fixed somehow on master, but I do not recall anything.

mbrio:
@michaelficarra, I just responded on a previous thread that corresponds with this error (https://github.com/jashkenas/coffeescript/issues/3829#issuecomment-74134197) and described what I had found in the time I had trying to figure out what the problem was.

It's interesting to note that @koron states (https://github.com/jashkenas/coffeescript/issues/3829#issuecomment-75706760) that node 0.12 fixes the problem.

Issue #3837, opened 2015-02-07
"yield return" from a conditional does not return if followed by a statement


ide:
If you have `yield return` in a conditional that has a statement after it, it doesn't actually return. This is a serious bug.

```coffeescript
example = ->
  if a
    yield return b
  return c
```
compiles to
```javascript
var example;

example = function*() {
  if (a) {
    b;
  }
  return c;
};
```

nfour:
Woah, I'm surprised I haven't encountered code breakage from this yet.

jashkenas:
Paging @alubbe.

If things like this are lingering, it would also be advisable to further beef up the `yield` test suite.

alubbe:
@nfour 
That's because it is unusual to use yield return anywhere but on your last line of code.

@ide 
Use this in the mean time:
```
example = ->
  if a
    return b
  yield return c
```

@jashkenas 
I'm on it, hope to fix it by the weekend. How close are we to publishing a 1.9.1, to get this fix in?

ide:
We replaced all yield returns with return yields as an easy fix (also easy to change back). It incurs an extra pass of generator delegation but that was fine for our workload.

alubbe:
Or you could just use ```return``` as always and use ```yield return``` only for the last case.

In any case, it'll get fixed.

jashkenas:
We can publish one soon, if you like.

alubbe:
fixed with https://github.com/jashkenas/coffeescript/pull/3854

georgy7:
I don't know if I right or wrong, but I think that is also an incorrect behaviour.

    fact = (n) ->
      n * fact(n-1) if n >= 1
      1
    
    console.log fact 10

Result:

    var fact;
    
    fact = function(n) {
      if (n >= 1) {
        n * fact(n - 1);  ///// return n * fact(n - 1); expected
      }
      return 1;
    };
    
    console.log(fact(10));

(Generated right now on coffeescript.org)

alubbe:
hi @georgy7, the behavior you are observing is intended. What you need to do is to make your ```else``` explicit 
```coffee
fact = (n) -> if n >= 1 then n * fact n - 1 else 1
```
or to ```return``` on your if condition
```coffee
fact = (n) ->
  return n * fact n - 1 if n >= 1
  1
```

georgy7:
@alubbe, thanks to your response.
In that case I would prefer this.

    fact = (n) ->
      if n >= 1
[...]

PR #3854, opened 2015-02-15
fixed yield return producing incorrect output when used outside of the last line

alubbe:
Fixes https://github.com/jashkenas/coffeescript/issues/3837

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -652,13 +652,18 @@
     };
 
     Return.prototype.compileNode = function(o) {
-      var answer;
+      var answer, exprIsYieldReturn, ref3;
       answer = [];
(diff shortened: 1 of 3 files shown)


alubbe:
rebased.

alubbe:
any feedback @jashkenas @michaelficarra ?

jashkenas:
I don't really understand why `yield return`s need to be handled specially, but ...

I think this PR could use a more robust test, where `yield return`s existed in various different branches of a function body. Maybe a complicated `if`? Maybe with a loop in it?

vendethiel:
It's not a JS construct. It's a coffee construct.

michaelficarra:
I think `yield return` anywhere but the last position in the body should be an error.

alubbe:
made the test a bit more complicated and added an error into the last (unreachable) line

PR #3861, opened 2015-02-18
Fix single-line heredocs starting with "undefined"

lydell:


michaelficarra:
I don't see how this has anything to do with undefined. It's just a heredoc with no leading indentation.

lydell:
1. `indent = null`
2. If there’s no leading indentation, `indent` stays `null`.
3. If `indent` is falsey (which `null` is), `indentRegex` is never assigned (and therefore is `undefined`).
4. `'string'.replace(undefined, '')` is the same as `'string'.replace(/undefined/, '')`

jashkenas:
Egad!

PR #3867, opened 2015-02-19
add -r/--require command line option

sgentle:
Hi,

I've added a Ruby-style `-r` option, which allows you to require a module before execution with `--eval` or `--interactive`:

```
$ touch mylib.coffee
$ coffee -r ./mylib
coffee> mylib
{}
```
```
$ coffee -r crypto -e 'console.log crypto.createHash("sha1").update("hello").digest("hex")'
aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d
```
```
$ # You need to have underscore in ./node_modules or ~/.node_modules
$ coffee -r _=underscore -e 'console.log _.intersection [1,2,3], [3,4,5]'
[ 3 ]
```

The first example has been particularly useful for me when I want to use the repl to play around with a module I'm writing. Having to type `mylib = require './mylib'` every time is a pain, and it's usually too far back in the command history to find easily.

The second and third examples go to a vision of the coffee command being suitable for quick shell one-liners, and if that sounds good I'd also like to suggest adding the global `node_modules` to the search path when you use `-e` and adding a `-E` eval+print option like node's `-p` so you don't have to type `console.log` so much. (happy to do pull requests for those too)

If you're not a fan of this approach, I could alternatively solve my main pain point by allowing `-e` and `-i` to coexist, in which case example 1 would look like: `coffee -e 'mylib = require "./mylib"' -i` which I think is less nice, but also less code.

I'm happy to add tests and clean up any style issues, but I wanted to get opinions before I went any further.


jashkenas:
We've talked about this before ... but it looks good to me.

PR #3871, opened 2015-02-20
Replace references to jQuery's 'bind' with 'on'

jcrben:
"As of jQuery 1.7, the .on() method provides all functionality required for attaching event handlers" per http://api.jquery.com/on/ and I expect .bind will be deprecated at some point. Plus, the multiple meanings of "bind" in this section could be a bit confusing for newcomers.


jashkenas:
Good call!

PR #3873, opened 2015-02-22
Fix getting of character at index

Taritsyn:
In Internet Explorer version 8 and below is not supported getting of character from a string by using indexer. Instead of indexer need to use the `charAt` method.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -208,7 +208,7 @@
       heredoc = quote.length === 3;
       ref2 = this.matchWithInterpolations(regex, quote), tokens = ref2.tokens, end = ref2.index;
       $ = tokens.length - 1;
-      delimiter = quote[0];
+      delimiter = quote.charAt(0);
       if (heredoc) {
         indent = null;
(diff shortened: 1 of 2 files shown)


jashkenas:
Thank you!

Is that the only problem you ran into, trying to use the CoffeeScript compiler within IE8?

Taritsyn:
[BundleTransformer.CoffeeScript](http://www.nuget.org/packages/BundleTransformer.CoffeeScript/) can use the [MSIE JavaScript Engine](https://github.com/Taritsyn/MsieJavaScriptEngine) (in `ChakraActiveScript` mode) to run CoffeeScript-compiler. In the MSIE JavaScript Engine, most compatibility problems are solved by using a [polyfill](https://github.com/Taritsyn/MsieJavaScriptEngine/blob/master/MsieJavaScriptEngine/Resources/ES5.js).

jashkenas:
Without the polyfill, do you notice any errors?

Taritsyn:
I have not tried to use CoffeeScript-compiler without polyfill.

PR #3881, opened 2015-02-25
Literate ".coffee.md" files incorrect line numbers

wdavidw:
Filenames made of two extensions are not handled inside the "getSourceMap" function. The previous implementation using "path.extname" only return the first extension.
I check and couldn't find any test checking filenumber while importing ".litcoffee" or ".coffee.md" so I'm not including any test either. Let me know if you wish to add one and i'll send you a complete push request.


lydell:
Duplicate of #4204. (One could argue that it is the other way around, but whatever ;) )

PR #3885, opened 2015-02-26
Escape literal [ in regexp

josh:
Regexps parsing regexps always causing trouble.

It'd be nice if we could escape the literal `[` in within the character class here to improve compatibility with "stricter ES5 regexp engines".

For context, I'm trying to get the CoffeeScript parser running under the [Duktape interpreter](http://duktape.org). It has a stricter regexp parser https://github.com/svaarala/duktape/issues/86 https://github.com/svaarala/duktape/issues/74

/cc @sstephenson @jashkenas


jashkenas:
Huh. I thought that `/[[]/` was legit in JavaScript:

    (/[[]/).test("one two [ three")
    => true

"escaping" these won't have any adverse consequences?

josh:
>  I thought that /[[]/ was legit in JavaScript

"SourceCharacter but not one of \ or ] or -" from http://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1 the regular expression grammar seems to exclude it.

Most other language's require it to be escaped too. `/[[]/` is a syntax error in Ruby and Python. But seems to be accepted by Perl. My guess is most browsers have been just using a PCRE backed engine so its mostly kinda just worked.

> "escaping" these won't have any adverse consequences?

Seems to still match any literal `[` usage.

``` js
(/[\[]/).test("one two [ three")
// true
```

michaelficarra:
> SourceCharacter but not one of \ or ] or -

@josh: That allows `[`. There's no harm in escaping it, but there's also no reason to do so, even back to ES3. You should tell Duktape to fix its parser or RegExp engine.

jashkenas:
@michaelficarra GitHub garbled that. It actually disallows `[` as well.

michaelficarra:
No, I don't think it does. I looked at the production in the spec.

jashkenas:
Ah, you're right. I was confused by the "PatternCharacter" bit.

This is actually the relevant rule: 

    ClassAtomNoDash ::
        SourceCharacter but not one of \ or ] or -
        \ ClassEscape

I'll revert.

josh:
Sorry, my mistake. The character class stuff actually seems fine, L840 doesn't need to be changed. The parse error is actually on the closing `]` L844

Thats a literal `]` thats unescaped. That seems to be excluded in the `PatternCharacter` set.

Would you guys reconsider just changing L844? or am I still wrong about the grammar :wink:

michaelficarra:
That sounds right. Please open a new PR.

josh:
See #3893.

PR #3892, opened 2015-03-08
Added checking on process.stdout to be compatible with browserify

matthieubulte:
Currently, browserify adds a shim of `process` to the global scope. The problem is that the shim created doesn't properly implement the process api, in particular it doesn't implement the `stdin` field. This leads to a `TypeError: Cannot read property 'isTTY' of undefined` in `helpers.js#syntaxErrorToString`.

The best solution is probably to also patch browserify but in this case the fix is quite small on the coffeescript side and makes the compiler usable in more environments.

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -173,7 +173,7 @@ syntaxErrorToString = ->
 
   # Check to see if we're running on a color-enabled TTY.
   if process?
-    colorsEnabled = process.stdout.isTTY and not process.env.NODE_DISABLE_COLORS
+    colorsEnabled = process.stdout?.isTTY and not process.env?.NODE_DISABLE_COLORS
 
   if @colorful ? colorsEnabled
     colorize = (str) -> "\x1B[1;31m#{str}\x1B[0m"
--- a/lib/coffee-script/helpers.js
+++ b/lib/coffee-script/helpers.js
@@ -201,7 +201,7 @@
   };
 
   syntaxErrorToString = function() {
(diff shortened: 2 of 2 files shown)


src/helpers.coffee
```diff
@@ -172,7 +172,7 @@ syntaxErrorToString = ->
   marker   = codeLine[...start].replace(/[^\s]/g, ' ') + repeat('^', end - start)
 
   # Check to see if we're running on a color-enabled TTY.
-  if process?
+  if process? and process.stdout? and process.env?
```
bjmiller: This can be shortened to: `if process?.stdout? and process?.env?`

matthieubulte:
Just pushed the change, thanks!

michaelficarra:
Please compile and submit the built files along with the PR.

matthieubulte:
Sorry, I didn't check the Cakefile properly. I just amended my commit to include the built files.

src/helpers.coffee
```diff
@@ -172,7 +172,7 @@ syntaxErrorToString = ->
   marker   = codeLine[...start].replace(/[^\s]/g, ' ') + repeat('^', end - start)
 
   # Check to see if we're running on a color-enabled TTY.
-  if process?
+  if process?.stdout? and process?.env?
```
lydell: `if process?
  colorsEnabled = process.stdout?.isTTY and not process.env?.NODE_DISABLE_COLORS`?

matthieubulte: yes, looks even better.

jashkenas:
Lovely use case for the ol' existential operator.

PR #3893, opened 2015-03-09
Escape literal ] in regexp

josh:
Reattempt #3885.

Just escaping the `]` this time.

/cc @sstephenson @jashkenas @michaelficarra


michaelficarra:
:ok_hand:

PR #3911, opened 2015-03-23
Fix syntaxErrorToString helper

kerbyfc:
To avoid uncaught error "Cannot read property 'isTTY' of undefined" with browserify and ect


goto-bus-stop:
The `lib/` directory contains generated code. You should apply changes in the `src/` directory: https://github.com/jashkenas/coffeescript/blob/master/src/helpers.coffee#L175

GeoffreyBooth:
This is more properly addressed in #4501

PR #3922, opened 2015-03-30
Optimises class creation when new arguments contain splats

### LINKED ISSUE #3236: ctor appearing in the nodejs heap instead of the classname
I recently discovered that a call like this 

```
new SomeClassName(args...) 
```

gets referenced as ctor in nodejs heap snapshots and not as SomeClassName. 
The bummer about this is if you have multiple classes using the syntax above then you end up with a mixed bag of classes called ctor in the heap.

Here is a screen shot of a heap output I'm seeing 

coffee-nodejs-heap-ctor

The line of code that is causing this is [here](https://github.com/jashkenas/coffee-script/blob/efb9809d3b4c67cfb483dcdc4563e8c1548f6cb1/src/nodes.coffee#L694)

I've pondered what could be done when this happens and I came up with a possible solution. We could rename the variable "ctor" to  use the class name as the variable instead. 

So as an example the javascript output ends up like this for the class called "SomeClassName"

```
(function(func, args, SomeClassName) {
    SomeClassName.prototype = func.prototype;
    var child = new SomeClassName, result = func.apply(child, args);
    return Object(result) === result ? result : child;
})(SomeClassName, args, function(){}))
```

I've tried this javascript code and the list of ctor disappear and the expected class names appear as expected in the snapshot heap.

Does that make sense, would it be possible or is there a caveat I'm not aware of?

pflannery:
Comparison of 100000 iterations heap memory ->

Before:
image

After:
image

Script source used and other comparisons [here](https://github.com/jashkenas/coffeescript/issues/3236#issuecomment-85102994)

This also results in being 5 times faster than it's predecessor.

closes #3236

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -700,8 +700,8 @@ exports.Call = class Call extends Base
 
   # If you call a function with a splat, it's converted into a JavaScript
   # `.apply()` call to allow an array of arguments to be passed.
-  # If it's a constructor, then things get real tricky. We have to inject an
-  # inner constructor in order to be able to pass the varargs.
+  # If it's a class constructor then we create a new Object
+  # and apply the class constructor to the Object instance
   #
   # splatArgs is an array of CodeFragments to put into the 'apply'.
   compileSplat: (o, splatArgs) ->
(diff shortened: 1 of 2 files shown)


pflannery:
@jashkenas @michaelficarra PTAL thanks

michaelficarra:
Looks fine. Are you really doing this that often?

src/nodes.coffee
```diff
@@ -2245,7 +2240,14 @@ UTILITIES =
       return child;
     }
   "
-
+  # Creates and applies arguments to a class constructor.
+  applyCtor: (ctor, args) -> "
+    function(ctor, args) {
+      var child = Object.create(ctor.prototype);
```
lydell: `Object.create` isn't available in older browsers.

pflannery:
@michaelficarra  https://github.com/bevry/taskgroup is a good example we've seen a big gain from this.

src/nodes.coffee
pflannery: your right but would be a shame to be held back by obsolete browsers - http://kangax.github.io/compat-table/es5/

vendethiel: Sadly, coffee targets es3

PR #3931, opened 2015-04-05
Commandline improvements

sgentle:
This is a followup to #3867 adding -E (for eval + print) and making the -r option able to require global modules.

At last, I can do this:

``` shell
$ coffee -E '1 + 1'
2
```

And this:

``` shell
$ coffee -r _=underscore -r fs -E "_.sample(fs.readFileSync('/usr/share/dict/words', 'utf8').split('\n'), 5)"
[ 'reassume', 'beachless', 'tourette', 'troublesome', 'unjostled' ]
```

Happy for feedback. In particular I think the global require hack is pretty gross, but it's the best thing I could find.


coderofsalvation:
In case this doesn't get merged, there's also https://www.npmjs.com/package/coffeepipe-bash

GeoffreyBooth:
I’m not sure why we need `-E` when we have `-ep`.

Global modules seem to work now, in 1.12.5:
```bash
$ coffee -r 'path' -e 'console.log path.sep'
/
```

sgentle:
`-p` doesn't do what you think it does:
```
  -p, --print        print out the compiled JavaScript
```

And `path` isn't a global module, it's part of the [Node standard library](https://nodejs.org/dist/latest-v6.x/docs/api/path.html).

You can install a global module with `npm install -g`:

```
$ npm install -g underscore
/usr/local/lib
└── underscore@1.8.3

$ coffee -r 'underscore' -e 'console.log underscore'
Error: Cannot find module 'underscore'
```

GeoffreyBooth:
Fair enough. At the moment though I don’t see any appetite for expanding the `coffee` command’s use as a shell scripting tool.

jashkenas:
Should `--eval` by default always print the value of the last expression? What do other CLIs do?

PR #3946, opened 2015-04-16
[CS2] Add #! support for executable scripts on Linux.

### LINKED ISSUE #1752: process.argv behaves differently to nodejs
In a file (both valid coffee script and vanilla JS):

```
console.log(process.argv)
```

then call with:

```
$ node argv.js -myargs
[ 'node', '/Users/Chris/Desktop/argv.js', '-myargs' ]
```

and

```
$ coffee argv.js -myargs
[ 'coffee',
  '/Users/Chris/Desktop/argv.js',
  '-m',
  '-y',
  '-a',
  '-r',
  '-g',
  '-s' ]
```

CoffeScript still interprets flags after the script name.

cosmicexplorer:
Pass arguments to executable script unchanged if using "#!/usr/bin/env
coffee". (Previously, "./test.coffee -abck" would be turned into "-a -b -c -k",
for example.)

Fixes #1752.


dbushong:
:+1: to this.  The current behavior also has it so that you can't pass `--` through to your coffee command-line program:
```
% echo 'console.log process.argv' > foo.coffee
% coffee -c foo.coffee
% coffee foo.coffee -a -- -b
[ 'coffee', '/Users/dbushong/foo.coffee', '-a', '-b' ]
% node foo.js -a -- -b
['/path/to/node', '/Users/dbushong/foo.js', '-a', '--', '-b']
```

While it works correctly if you say `coffee foo.coffee -- -a -- -b`, that's not on option on the linux shebang line.

src/optparse.coffee
```diff
@@ -25,6 +25,14 @@ exports.OptionParser = class OptionParser
   # parsers that allow you to attach callback actions for every flag. Instead,
   # you're responsible for interpreting the options object.
   parse: (args) ->
+    # Pass all arguments to script unchanged if first argument is the script to
+    # be run; assume no options are for the coffeescript compiler. This allows
+    # the use of '#!/usr/bin/env coffee' to run executable scripts on Linux
+    # systems, which do not parse the '--' argument in the first line correctly.
+    if (args.indexOf '--' is -1) and
+       (args.length > 0) and
+       (isCoffee args[0])
```
GeoffreyBooth: I would rewrite this is as
```coffee
if args.length isnt 0 and '--' in args and isCoffee(args[0])
```

GeoffreyBooth:
Is there any chance that this could be a breaking change? If so please target the `2` branch.

cosmicexplorer:
I think so, for linux hosts; the situation I described at the top won't matter if the script follows standard command line conventions (where `-abck` is equivalent to `-a -b -c -k`), but it'll mysteriously break for users who weren't aware of this. I can fix this to work off the 2 branch sometime this week.

GeoffreyBooth:
How would it break? Would `-a -b -c -k` no longer be possible?

cosmicexplorer:
Sorry, the other way around; `-abck` is always split to `-a -b -c -k` (not just these letters, others too) unless you use a workaround. There is no way to make a standalone script executable on Linux using an `env` shebang line so that it can accept multiple arguments; some discussion is [here](https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-with-a-shebang-i-e).

Invoking a script either like `./script.coffee -- -abck` or  `coffee script.coffee -- -abck` both have an `argv` of `[ 'coffee', '/path/to/script.coffee', '-abck']`, so it is possible, but requires a wrapper script. This is a problem because these arguments are consumed by the script, but still processed by the coffeescript executable's own argument parsing. On non-linux boxes you can have the line `#!/usr/bin/env coffee --`, but in linux that argv is parsed by the kernel as a two-element array `['/usr/bin/env', 'coffee --']`, which obviously fails unless you add a symlink at `"/usr/bin/coffee --"` or something else ridiculous.

GeoffreyBooth:
How will this affect traditional users running `coffee` from the command line?

cosmicexplorer:
`isCoffee` checks if the script to be evaluated was invoked with a shebang (as the first argument) or as an argument to the `coffee` executable. Anyone using a command line like `coffee [args...]` will be **unaffected**. Users who may be affected currently invoke a script as `argv[0]`, e.g. `./test.coffee [args...]`, where `test.coffee` has a shebang line and is executable. If `test.coffee` has:

``` coffee
#!<SHEBANG LINE>

console.log process.argv
```
then there are about two and a half cases it differs:

1. Scripts with the shebang `#!/usr/bin/env coffee --` (which does not work on Linux) and are invoked as `./test.coffee -abck` will see a `--` after the script path: `[ 'coffee', '/path/to/test.coffee', '--', '-abck' ]` (where before there was no `--`). Users who invoke the script with shebang `#!/usr/bin/env coffee --` (not on Linux) as `./test.coffee -- -abck` intentionally to pass `--` as an argument to their own script will see the `--` argument disappear: `[ 'coffee', '/home/cosmicexplorer/test.coffee', '-abck' ]`.

2. Users on any platform (probably on Linux because it was the only way it would work, although not necessarily) who had used the shebang `#!/usr/bin/env coffee` or `#!/path/to/some/executable/coffee` and invoked it as `./test.coffee -- -abck` to avoid the argument splitting into `-a -b -c -k` will see an extra `--`, just as above: `[ 'coffee', '/path/to/test.coffee', '--', '-abck' ]`, where before there was none.

The solution above won't fix this, but we **can** detect the case where the user invokes the script with `--` in the shebang line and *not* in the script invocation. I'll push my (small) change in a second which does this.

GeoffreyBooth:
We should add some tests around this. I think it’s testable within our current framework, via stuff like `execSync('./imports/cli_test.coffee -a -b -c')` or similar. There should also be tests for more traditional commands, like `execSync('coffee -v')`, so that the current behavior is covered against regressions. The tests should make sure to use the `coffee` binary in the repo, not whatever global `coffee` might be in the user’s `$PATH`.

cosmicexplorer:
I think the previous option parsing code was convoluted and made it too easy to make mistakes like the `--` processing (which this commit also fixes), especially for platform-specific stuff. With respect to the (possible) breaking changes I mentioned in my previous post, invoking the `coffee` interpreter is still the exact same. When invoking a shebang script `./test.coffee [args...]`:

1. If the script begins with a shebang line of `#!/usr/bin/env coffee --` *or* `#!/usr/bin/env coffee` (or just a full path to the executable), it will *not* modify any arguments passed on the command line to the script. **This part breaks nothing and makes scripts work on Linux.**

2. There *may* be users (most probably on Linux, if they exist) who used a shebang script of `#!/usr/bin/env coffee` and passed `--` as an argument to their script, like `./test.coffee -- [intended-args...]` to work around the previous behavior (which would have only passed `intended-args...` to their script). **These users will see an "extra" `--`, which is a breaking change, but I think easily spotted.**

I can absolutely add some tests. I do not have time today, but will try to do so tomorrow. Let me know if there are any style changes I can make, I tried to follow the rest of the code as much as possible.

src/optparse.coffee
```diff
@@ -25,37 +25,26 @@ exports.OptionParser = class OptionParser
...
-      if skippingArgument
-        skippingArgument = no
-        continue
-      if arg is '--'
-        pos = originalArgs.indexOf '--'
-        options.arguments = options.arguments.concat originalArgs[(pos + 1)..]
+    state =
+      argsLeft: args[..]
+      options: {}
+    while (arg = state.argsLeft.shift())?
```
GeoffreyBooth: I’ve been trying to avoid assignments inside conditionals. Can we just use `for arg in args`?

src/optparse.coffee
```diff
@@ -25,37 +25,26 @@ exports.OptionParser = class OptionParser
...
+      if (arg.match(LONG_FLAG) ? arg.match(SHORT_FLAG))?
+        tryMatchOptionalArgument(arg, state, @rules)
+      else if (multiMatch = arg.match(MULTI_FLAG))?
+        # Normalize arguments by expanding merged flags into multiple
+        # flags. This allows you to have `-wl` be the same as `--watch --lint`.
+        normalized = "-#{multiArg}" for multiArg in multiMatch[1].split ''
+        state.argsLeft.unshift(normalized...)
+      else
+        # the CS option parser is a little odd; options after the first
```
GeoffreyBooth: The comments become the annotated source, so please use sentence case (capitalize first word, use periods).

src/optparse.coffee
```diff
@@ -99,14 +88,22 @@ buildRule = (shortFlag, longFlag, description, options = {}) ->
...
+  options[rule.name] = if rule.isList
+    (options[rule.name] ? []).concat value
+  else value
+
+tryMatchOptionalArgument = (arg, state, rules) ->
+  for rule in rules
+    if arg in [rule.shortFlag, rule.longFlag]
+      if rule.hasArgument
+        value = state.argsLeft.shift()
+        if not value?
```
GeoffreyBooth: `unless`

src/optparse.coffee
```diff
@@ -99,14 +88,22 @@ buildRule = (shortFlag, longFlag, description, options = {}) ->
...
+          throw new Error "#{arg} requires a value, but was the last argument"
+      else
+        value = true
+
+      addArgument(rule, state.options, value)
```
GeoffreyBooth: Avoid parentheses when possible.

src/optparse.coffee
```diff
@@ -99,14 +88,22 @@ buildRule = (shortFlag, longFlag, description, options = {}) ->
...
+      return
```
GeoffreyBooth: Is there a reason to avoid the implicit return?

GeoffreyBooth:
Do you mind please submitting the “current state” tests of the CLI as it works now as a separate PR? So that we can see that they work before trying them against your refactor.

src/optparse.coffee
cosmicexplorer: Now that I think about it, I think we can just do a single pass over the arguments beforehand to 1. group arguments with values (`-o [dir]`) 2. explode combined options (e.g. `-abck`) 3. identify the first non-optional argument, then use a `for`. I'll do that.

src/optparse.coffee
cosmicexplorer: Got it, will fix. The first two lines are actually from the original source, so I wasn't sure whether I should change them.

src/optparse.coffee
cosmicexplorer: Got it, will change.

src/optparse.coffee
cosmicexplorer: Got it, will fix.

src/optparse.coffee
cosmicexplorer: If not, it would eventually leave the for loop and hit the throw statement. This was dealt with previously by keeping a boolean testing whether or not the rule was matched; I thought this was cleaner here since we don't then loop over the rest of the rules if we successfully match one.

src/optparse.coffee
GeoffreyBooth: Yeah. Never too late to fix things 😄

cosmicexplorer:
Yes! I'll make a separate PR for those sometime tomorrow. Thanks for your responsiveness.

GeoffreyBooth:
@cosmicexplorer there are some merge conflicts now. Might want to get this wrapped up soon in order to not have to keep updating it. If you'd like to submit the other PR with tests I'd be happy to review it.

cosmicexplorer:
Understood. I will make a new PR with tests first, then fix merge conflicts on this one, then ensure it passes the tests from the separate PR. I was busier with exams than I expected, but those are finished now.

GeoffreyBooth:
@cosmicexplorer did you intend to close this?

cosmicexplorer:
No I did not, I didn't realize that force pushing would have this effect. As you can see I also based it against 2, as you suggested previously. I will submit an attempt at the "current state" tests today and then continue to work on this.

GeoffreyBooth:
Thanks. Yes, please avoid force-pushing or rebasing or otherwise revising history. It’s easier to review your progress when the history is clear.

cosmicexplorer:
Understood.

cosmicexplorer:
See #4565 for tests for current cli.

GeoffreyBooth:
@cosmicexplorer is this still WIP or is it ready for review (assuming we merge in the tests from the other PR)?

cosmicexplorer:
This is now ready for review. I was thinking over your comments on `while` vs `for` for iterating over the arguments and eventually arrived at what I've just pushed (which looks a lot more like the original code). I've also added tests for everything I left a `TODO: test` on.

There was a `TODO: refactor option parser tests` I've removed from `test/option_parser.coffee` because I've refactored the option parsing code, along with the tests, and the comment wasn't specific about what needed to change. That can stay if there's more refactoring that should be done -- I could also do that refactoring in a separate PR if you can tell me what sort of refactoring to do.

src/helpers.coffee
```diff
@@ -49,8 +49,9 @@ extend = exports.extend = (object, properties) ->
 exports.flatten = flatten = (array) ->
   flattened = []
   for element in array
-    if '[object Array]' is Object::toString.call element
-      flattened = flattened.concat flatten element
+    # TODO: is this the same as using Object::toString.call?
```
GeoffreyBooth: It is the same: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray?v=control#Polyfill

Let’s not change this function unless there’s a reason to.

src/optparse.coffee
```diff
@@ -18,8 +18,8 @@ exports.OptionParser = class OptionParser
   #     [short-flag, long-flag, description]
   #
   # Along with an optional banner for the usage help.
-  constructor: (rules, @banner) ->
-    @rules = buildRules rules
+  constructor: (ruleDecls, @banner) ->
```
GeoffreyBooth: What is `ruleDecls`? Rule declarations?

src/optparse.coffee
```diff
@@ -28,36 +28,28 @@ exports.OptionParser = class OptionParser
...
-            break
-        throw new Error "unrecognized option: #{arg}" if isOption and not matchedRule
-      if seenNonOptionArg or not isOption
-        options.arguments.push arg
+    # The CS option parser is a little odd; options after the first
+    # non-option argument are treated as non-option arguments themselves.
+    # Optional arguments are normalized by expanding merged flags into multiple
+    # flags. This allows you to have `-wl` be the same as `--watch --lint`.
+    # Note that executable scripts do not need to have a `--` at the end of the
+    # shebang ("#!") line, and if they do, they won't work on Linux (see #3946).
```
GeoffreyBooth: (`` `#!` ``)

src/optparse.coffee
```diff
@@ -75,41 +67,98 @@ exports.OptionParser = class OptionParser
...
 # unspecified, leave it out by padding with `null`.
-buildRules = (rules) ->
-  for tuple in rules
+buildRules = (ruleDecls) ->
+  ruleList = for tuple in ruleDecls
     tuple.unshift null if tuple.length < 3
     buildRule tuple...
+  flagDict = {}
+  for rule in ruleList
+    # shortFlag is null if not provided in the rule.
```
GeoffreyBooth: `` `shortFlag` ``.

src/optparse.coffee
```diff
@@ -75,41 +67,98 @@ exports.OptionParser = class OptionParser
...
+    for flag in [rule.shortFlag, rule.longFlag] when flag?
+      if flagDict[flag]?
+        throw new Error "flag #{flag} for switch #{rule.name}
+          was already declared for switch #{flagDict[flag].name}"
```
GeoffreyBooth: Does this get output to the user as a friendly error, or do they see a stack trace?

src/optparse.coffee
```diff
@@ -75,41 +67,98 @@ exports.OptionParser = class OptionParser
...
+normalizeArguments = (args, flagDict) ->
+  rules = []
+  positional = []
+  needsArgOpt = null
+  for arg, argIndex in args
+    # If the previous argument given to the script was an option that uses the
+    # next command-line argument as its argument, create copy of the option's
+    # rule with an `argument` field.
+    if needsArgOpt?
+      # FIXME: use object spread when that gets merged (in #4493)
```
GeoffreyBooth: We don’t need this comment. This can stay `Object.assign` forever.

src/optparse.coffee
```diff
@@ -75,41 +67,98 @@ exports.OptionParser = class OptionParser
...
+          throw new Error "cannot use option #{flag} in multi-flag #{arg} except
+          as the last option, because it needs an argument"
+        rules.push rule
+      if lastOpt.rule.hasArgument
+        needsArgOpt = lastOpt
+      else
+        rules.push lastOpt.rule
+    else if ([LONG_FLAG, SHORT_FLAG].some (pat) -> arg.match(pat)?)
+      # TODO: do we need to check if `arg` matches regex if we already check if
+      # it's in flagDict?
```
GeoffreyBooth: ?

src/optparse.coffee
```diff
@@ -75,41 +67,98 @@ exports.OptionParser = class OptionParser
...
+        throw new Error "unrecognized option #{arg}"
+      if singleRule.hasArgument
+        needsArgOpt = {rule: singleRule, flag: arg}
+      else
+        rules.push singleRule
     else
-      result.push arg
-  result
+      # This is a positional argument.
+      # TODO: should we check these with `isCoffee`?
```
GeoffreyBooth: ?

src/helpers.coffee
cosmicexplorer: Ok, reverted.

src/optparse.coffee
cosmicexplorer: Yes. The SWITCHES array in command.coffee that `buildRules()` consumes isn't executable on its own, so I wanted to distinguish it from the result (a dictionary and array with long and short options).

src/optparse.coffee
cosmicexplorer: Fixed.

src/optparse.coffee
cosmicexplorer: Fixed.

src/optparse.coffee
cosmicexplorer: Ok, removed.

src/optparse.coffee
cosmicexplorer: Checking for multi-flags makes sense, since they have to be split into individual flags that we do immediately recognize. I was thinking it was a little redundant to check if an argument matches a regex when we do exact string matching the line right after, in `flagDict[arg]`. This comment can be removed, however, because we want to notify the user if they put in an option that doesn't exist.

src/optparse.coffee
cosmicexplorer: In a long ago version of this PR I checked positional arguments to see whether they were recognized coffeescript files (`.coffee`, `.litcoffee`) or `--`. If an argument is provided that is not such a script, it looks like `compilePath` in `command.coffee` will just silently return, which seems unhelpful. An alternative would be to show a warning if a positional argument is given which `compilePath` won't do anything to.

src/optparse.coffee
cosmicexplorer: Fixed.

src/optparse.coffee
cosmicexplorer: This produces a stack trace. This specific error indicates a bug in the flag declarations given to the `OptionParser` class, which indicates a bug in the program using these -- that might be appropriate to leave as a stack trace. The rest (errors popping up during parsing the given options) should probably be caught and handled in `command.coffee` where it calls `.parse()`. I'll do that.

GeoffreyBooth:
@cosmicexplorer Please commit the compiled output files in `lib`.

So I created a `test.coffee` with:

```coffee
#!/usr/bin/env coffee
console.log process.argv
```

And I made it executable, and ran it via `./test.coffee -abck`. With both this PR and CoffeeScript 1.12.6, I get this output:

```
[ 'coffee',
  '/Volumes/Geoffrey/Sites/coffeescript/test.coffee',
  '-a',
  '-b',
  '-c',
  '-k' ]
```

I thought from the initial post that this PR’s output should include `'-abck'`?

Also, what about a test that files starting with `#!/usr/bin/env coffee` are handled this way?

cosmicexplorer:
You are right, that should include `-abck` (that's kind of the whole point of this PR, so I'm pretty embarrassed). I will investigate why that fails and why the current tests aren't catching that (even without testing on an actual shebang script, they should still detect that failure). I was already playing with a test for the actual shebang lines, so I'll fix that up and add it as well. When I have done that, I will commit the compiled output files too.

GeoffreyBooth:
See also the `test/imports` folder. You could put a real shell script in there and run it as one of the tests, maybe with a check that bash is available or that we're not in a Windows environment (if that's necessary).

cosmicexplorer:
I tried the example you just gave, and got the output:

``` coffeescript
[ 'coffee',
  '/home/cosmicexplorer/tools/coffeescript/test/test.coffee',
  '-abck' ]
```

Can you confirm that `/usr/bin/env coffee` points to the version you just built from this PR? I had to screw around with my path a bit to make sure that worked.

cosmicexplorer:
The test I just added uses `#!/usr/bin/env coffee` and ensures the most recent executable is used by prepending the `bin/` directory to the PATH. It only checks the example you just mentioned, but we could make it so that it invokes the shebang script (on non-windows) for all the tests in `test/argument-parsing.coffee`.

test/argument-parsing.coffee
```diff
@@ -1,3 +1,6 @@
+path = require 'path'
```
GeoffreyBooth: You need to start the file with `return unless require?` so that this doesn’t break the browser tests `cake test:browser`

cosmicexplorer: Fixed.

cosmicexplorer: (by moving the invocation stuff to a separate file)

cosmicexplorer:
If you look at issue #1752 (the one that motivated this PR) or search "make coffeescript executable" you'll find people suggesting `#!/usr/bin/env coffee --`, which (as discussed in this thread) works on OSX, but not Linux (which parses it as "invoke the executable named `coffee --`"). Also, the `Usage:` line is now incorrect -- if you call `coffee path/to/script.coffee -- [args]`, the `script.coffee` now sees `--` as its first argument.

I want to make three changes so that these invocations remain backwards compatible, but users are warned about the now-preferred command-line semantics.

1. If we run a script, check if the script's first line matches `/^#!.*coffee.+$/`. If it does, there are arguments (or whitespace) after `coffee`, and the script won't run on Linux. Print a warning to explain the issue and how to fix it.
2. If `options.arguments` looks like `[ 'script.coffee', '--', ... ]`, remove the `--` argument at index 1 before proceeding. Print a warning to explain that the `--` argument is no longer needed after the script argument, and if you want to pass `--` to the script, you should use `coffee -- script.coffee -- [args]`, which passes `-- [args]` to the script.
3. Change the `Usage:` line to:

```
Usage: coffee [options] [--] path/to/script.coffee [args]
Usage: coffee [options] path/to/script.coffee -- [args] (deprecated)

If called without options, `coffee` will run your script. Previous versions of
coffeescript required a `--` after the script to run, but this convention is now
deprecated.
```

src/optparse.coffee
cosmicexplorer: I misread `compilePath()`, so this comment can be removed.

cosmicexplorer:
I've just pushed a commit which does the above. (I'd need to add tests too, to check if the warnings pop up). I put it in the same PR because it addresses backwards incompatibility introduced by the rest of these changes. I can move it to a new PR or forget about it if it's not necessary.

GeoffreyBooth:
You were correct about `$PATH`, thanks for pointing me in the right direction. Now it seems to work:
```bash
✦ cd ~/Sites/coffeescript
✦ export PATH=/Users/Geoffrey/Sites/coffeescript/bin:$PATH
✦ ./test.coffee -abck
[ 'coffee',
  '/Users/Geoffrey/Sites/coffeescript/test.coffee',
  '-abck' ]
```

I think we should remove the deprecation warning in the Usage text (“Previous versions of coffeescript required a `--` after the script to run, but this convention is now deprecated.”). There’s a [Breaking Changes](http://coffeescript.org/v2/#breaking-changes) section in the v2 docs, I think that’s a better place for that info.

Otherwise I think this is okay. @lydell or @connec, any notes?

cosmicexplorer:
I added a bit on the new argument parsing into the breaking changes section in the v2 docs, added test cases for the warnings described in that section, and committed the compiled files into `lib/`. Let me know if there's anything I missed or should fix.

test/importing/shebang-extra-args.coffee
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env coffee --
```
GeoffreyBooth: The convention in this repo is to use underscores in filenames.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
```diff
@@ -0,0 +1,69 @@
...
+
+``` bash
+> coffee -h
+
+Usage: coffee [options] [--] path/to/script.coffee [args]
+
+If called without options, `coffee` will run your script.
+
+
+```
```
GeoffreyBooth: Why the blank lines before and after the text?

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
```diff
@@ -0,0 +1,69 @@
...
+console.log x
+```
+
+``` bash
+> chmod +x ./executable-script.coffee
+> ./executable-script.coffee
+4
+```
+
+Due to a bug in the argument parsing of previous CoffeeScript versions, this used to fail when trying to pass arguments to the script. Some users on OSX worked around the problem by using `#!/usr/bin/env coffee --` at the top of the file instead. However, that won't work on Linux, which cannot parse shebang lines with more than a single argument. While these scripts will still run on OSX, CoffeeScript will now display a warning before compiling or evaluating files that begin with a too-long shebang line:
```
GeoffreyBooth: It’s “OS X”. Also the documentation uses curly quotes, e.g. “won’t”.

test/importing/shebang-extra-args.coffee
cosmicexplorer: I have changed all the files I have added to use underscores.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
cosmicexplorer: I actually screwed up the help text (see the most recent commit -- needed to change the help method to use `@rules.ruleList` instead of just `@rules`). I have also added tests in the most recent commit to ensure the help text will always display the banner and the flags.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
cosmicexplorer: Fixed both of those.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
```diff
@@ -11,6 +11,25 @@ Usage: coffee [options] [--] path/to/script.coffee [args]
...
+      --nodejs       pass options directly to the "node" binary
+      --no-header    suppress the "Generated by" header
+  -o, --output       set the output directory for compiled JavaScript
+  -p, --print        print out the compiled JavaScript
+  -r, --require      require the given module before eval or REPL
+  -s, --stdio        listen for and compile scripts over stdio
+  -l, --literate     treat stdio as literate style coffeescript
+  -t, --tokens       print out the tokens that the lexer/rewriter produce
+  -v, --version      display the version number
+  -w, --watch        watch scripts for changes and rerun commands
```
GeoffreyBooth: This is already documented in http://coffeescript.org/v2/#usage. The Breaking Changes section should only discuss breaking changes from 1.x.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
```diff
@@ -45,7 +64,7 @@ console.log x
 4
 ```
 
-Due to a bug in the argument parsing of previous CoffeeScript versions, this used to fail when trying to pass arguments to the script. Some users on OSX worked around the problem by using `#!/usr/bin/env coffee --` at the top of the file instead. However, that won't work on Linux, which cannot parse shebang lines with more than a single argument. While these scripts will still run on OSX, CoffeeScript will now display a warning before compiling or evaluating files that begin with a too-long shebang line:
+Due to a bug in the argument parsing of previous CoffeeScript versions, this used to fail when trying to pass arguments to the script. Some users on OS X worked around the problem by using `#!/usr/bin/env coffee --` at the top of the file instead. However, that won’t work on Linux, which cannot parse shebang lines with more than a single argument. While these scripts will still run on OSX, CoffeeScript will now display a warning before compiling or evaluating files that begin with a too-long shebang line:
```
GeoffreyBooth: Another “OS X”.

test/option_parser.coffee
```diff
@@ -65,3 +71,37 @@ test "throw if multiple flags try to use the same short or long name", ->
...
+    joined = [longFlagPat, descPat].join '.*'
+  new RegExp "^.*#{joined}.*$", 'm'
+
+test "help text shows banner and all switches", ->
+  bannerHelp = opt.help()
+  bannerPattern = new RegExp reQuote(banner), 'g'
+  ok bannerHelp.match bannerPattern
+  for flag in flags
+    linePat = flagPattern flag
+    ok bannerHelp.match linePat
```
GeoffreyBooth: This tests that the option parser correctly outputs a banner text string, but not necessarily that `coffee -h` actually outputs the correct help text. For that, you need to actually test `opt.help()` against the help text, _which should be written here in the test._ I know it doesn’t feel DRY to repeat a string that is elsewhere in the codebase, but if that string changes in the main code, we _want_ this test to fail—either to inform us that we unexpectedly changed the help text, or to remind us to update the test, too, so that we preserve this check for the future. The same applies to the list of switches.

cosmicexplorer: Ok, that makes sense. I'll make that change.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
cosmicexplorer: Fixed.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
cosmicexplorer: That makes sense. I have noted only the changed usage line now and left an `...` to denote the rest of the help text.

cosmicexplorer:
1. I have responded to your comments and merged in the `2` branch (the conflict was with the file `argument-parsing.md`, which I renamed with an underscore instead).
2. The tests for the help text now check the exact help text string. I have also added tests in `option_parser.coffee` which checks that the correct help text is output by the OptionParser class, not just the instance of it used in the `coffee` command.
3. I have slightly edited `breaking_changes_argument_parsing_and_shebang_lines.md` to make it more clear.

Let me know if there are any issues with these changes.

test/option_parser.coffee
```diff
@@ -72,36 +72,23 @@ test "throw if multiple flags try to use the same short or long name", ->
...
+  '''
+  ok opt.help() is expectedBanner
+
+  expected = [
+    ''
+    '  -r, --required     desc required'
+    '  -o, --optional     desc optional'
+    '  -l, --list         desc list'
+    ''
+  ].join('\n')
```
GeoffreyBooth: This could just be a `'''` string.

test/option_parser.coffee
```diff
@@ -72,36 +72,23 @@ test "throw if multiple flags try to use the same short or long name", ->
...
-test "help text shows switches, even without banner", ->
-  parser = new OptionParser flags
-  noBannerHelp = parser.help()
-  for flag in flags
-    linePat = flagPattern flag
-    ok noBannerHelp.match linePat
+test "outputs expected help text", ->
+  expectedBanner = '''
+
+banner text
```
GeoffreyBooth: Why is this here?

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
```diff
@@ -43,37 +31,44 @@ additional '--' before the path to your script.
...
 ``` bash
 > chmod +x ./executable-script.coffee
 > ./executable-script.coffee
 4
 ```
 
-Due to a bug in the argument parsing of previous CoffeeScript versions, this used to fail when trying to pass arguments to the script. Some users on OS X worked around the problem by using `#!/usr/bin/env coffee --` at the top of the file instead. However, that won’t work on Linux, which cannot parse shebang lines with more than a single argument. While these scripts will still run on OSX, CoffeeScript will now display a warning before compiling or evaluating files that begin with a too-long shebang line:
+##### Incorrect Example
+
+Due to a bug in the argument parsing of previous CoffeeScript versions, this used to fail when trying to pass arguments to the script. Some users on OSX worked around the problem by using `#!/usr/bin/env coffee --` at the top of the file instead. However, that won’t work on Linux, which cannot parse shebang lines with more than a single argument. While these scripts will still run on OSX, CoffeeScript will now display a warning before compiling or evaluating files that begin with a too-long shebang line:
```
GeoffreyBooth: “OS X”

cosmicexplorer:
I have added you as a collaborator to my branch -- let me know if you have push access.

test/option_parser.coffee
cosmicexplorer: This tests that the OptionParser class outputs the kind of help text that we expect, and not just the OptionParser instance that is used in `command.coffee`. It also tests the output without a banner.

test/option_parser.coffee
cosmicexplorer: Possibly, except that `'''` strings don't allow leading spaces unless one of the lines (e.g. the banner text) is not indented, so I don't think this would work as a `'''` string.

documentation/sections/breaking_changes_argument_parsing_and_shebang_lines.md
cosmicexplorer: I don't understand what the issue is -- should I add quotes around `OSX`? The space between `OS` and `X` was removed everywhere.

GeoffreyBooth:
Okay, I think we’re good now. I shortened the docs considerably, can you let me know if I took out anything important? We try to keep the docs as concise as possible. The generated output is here:
http://rawgit.com/cosmicexplorer/coffeescript/master/docs/v2/#breaking-changes-argument-parsing-and-shebang-lines

Anyone else have any final notes before this gets merged in?

src/command.coffee
```diff
@@ -25,7 +25,7 @@ hidden = (file) -> /^\.|~$/.test file
 
 # The help banner that is printed in conjunction with `-h`/`--help`.
 BANNER = '''
-  Usage: coffee [options] [--] path/to/script.coffee [args]
```
cosmicexplorer: I don't like removing `[--]`, the new version does not say anything about how `--` is parsed as an argument. `gawk --help` uses `[--]` in a very similar way.

test/argument_parsing.coffee
```diff
@@ -115,7 +115,7 @@ test "throw on invalid options", ->
 test "has expected help text", ->
   ok optionParser.help() is '''
 
-Usage: coffee [options] [--] path/to/script.coffee [args]
```
cosmicexplorer: As in `src/command.coffee`.

cosmicexplorer:
I really like the changes. My only concern is in the usage text, as noted inline.

GeoffreyBooth:
Why mention `--` in the help text if we don't want it to be used? Does it serve some purpose?

cosmicexplorer:
After thinking about it a bit, I think the main reason I wanted the `[--]` to stay is because awk's usage text uses `[--]` and it seems more precise. I completely agree that since we don't want it to be used (even though the command can handle it), it shouldn't be mentioned.

GeoffreyBooth:
I think it’s confusing to include the reference to it if we don’t define what it does (especially if the only thing it “does” is print a warning). If it did anything useful, then sure, let’s include it and describe its function like the other flags.

@lydell or anyone else, any final notes? This is an _old_ PR, thank you @cosmicexplorer for sticking with us for so long!

cosmicexplorer:
I completely 100% agree with that. Let me know if anything weird happens with these changes -- I feel comfortable after all the test cases, but I'm still wary. Glad this change finally saw light, turned out to be much, much harder than I expected.

PR #3966, opened 2015-04-30
Fix #3965, sourcemaps for bare programs.

### LINKED ISSUE #3965: Source maps are incorrect for bare programs
Coffeescript maps every generated newline and semicolon to the source position (0,0), which means that short bare programs (which are unindented) end up mapping the beginning of every generated line to (0,0).  So function calls written at the top-level in browser-based coffeescript do not source map correctly.

This causes problems for me on pencilcode.net, when students are trying to debug short programs all the time.

Here is the simplest possible repro:

test.coffee:
```
alert()
alert()
```

Now compile bare and get source maps with:
```
coffee -b -m -o . test.coffee
```

That produces this simple test.js:
```js
// Generated by CoffeeScript 1.9.2
alert();

alert();

//# sourceMappingURL=test.js.map
```

And the resulting test.js.map contains this encoded mapping string:
```
  "mappings": ";AAAA,KAAA,CAAA,CAAA,CAAA;;AAAA,KACA,CAAA,CADA,CAAA"
```

If you plug this into a source map decoder like http://murzwin.com/base64vlq.html, you see that maps our two-line program as follows:

```
([0,0](#0)=>[1,0]) | ([0,0](#0)=>[1,5]) | ([0,0](#0)=>[1,6]) | ([0,0](#0)=>[1,7]) | ([0,0](#0)=>[1,8])
([0,0](#0)=>[3,0]) | ([1,0](#0)=>[3,5]) | ([1,0](#0)=>[3,6]) | ([0,0](#0)=>[3,7]) | ([0,0](#0)=>[3,8])
```

The problem is on the second line (mappings for output line 3): whitespace and semicolons map from (0,0), which is bad, but what's worse is that the first column of output position (3, 0), which contains the "alert" call, is also mapped from (0, 0).  It actually comes from (1, 0) in the source.

In a bigger program, all generated code in column 0 in a bare program will get mapped to line (0,0) in the source program.  That's a side effect of the problem of mapping all whitespace to (0,0).

What we want is a source map that looks like this, with (1, 0) mapping to (3, 0):
```
([0,0](#0)=>[1,0]) | ([0,0](#0)=>[1,5]) | ([0,0](#0)=>[1,6])
([1,0](#0)=>[3,0]) | ([1,0](#0)=>[3,5]) | ([1,0](#0)=>[3,6])
```

In other words, for this program, we should have v3sourcemap like:
```
  "mappings": ";AAAA,KAAA,CAAA;;AACA,KAAA,CAAA"
```

We should probably not output sourcemapping for generated spaces and semicolons.

A simple fix coming soon.

davidbau:
Instead of mapping all generated spaces and semicolons and newlines
to the source position (0,0), we avoid generating sourcemap information
for generated space-or-semicolon-only fragments.

(In addition to shortening sourcemaps, this fixes a correctness issue
where an empty fragment at the beginning of each line maps from (0,0),
but in a bare program, that position at the begining of the line
should map from the actual source line.  When this conflict occurred,
(0,0) would win, resulting in an incorrect sourcemap, where each
top-level function call mapped to (0,0).)

### DIFF
--- a/src/coffee-script.coffee
+++ b/src/coffee-script.coffee
@@ -62,7 +62,8 @@ exports.compile = compile = withPrettyErrors (code, options) ->
   for fragment in fragments
     # Update the sourcemap with data from each fragment
     if options.sourceMap
-      if fragment.locationData
+      # Do not include empty, whitespace, or semicolon-only fragments.
+      if fragment.locationData and not /^[;\s]*$/.test fragment.code
         map.add(
           [fragment.locationData.first_line, fragment.locationData.first_column]
           [currentLine, currentColumn]


src/coffee-script.coffee
```diff
@@ -62,7 +62,8 @@ exports.compile = compile = withPrettyErrors (code, options) ->
   for fragment in fragments
     # Update the sourcemap with data from each fragment
     if options.sourceMap
-      if fragment.locationData
+      # Do not include empty, whitespace, or semicolon-only fragments.
+      if fragment.locationData and not /^;?\s*$/.test fragment.code
```
michaelficarra: `/^[;\s]*$/`

michaelficarra:
LGTM otherwise.

davidbau:
Adjusted the regexp.

lydell:
LGTM too. Few compilers I’ve seen map whitespace (and insignificant tokens).

carlsmith:
`+1` This will fix the only remaining bug in CoffeeShop's tracebacks.

src/coffee-script.coffee
```diff
@@ -62,7 +62,8 @@ exports.compile = compile = withPrettyErrors (code, options) ->
   for fragment in fragments
     # Update the sourcemap with data from each fragment
     if options.sourceMap
-      if fragment.locationData
+      # Do not include empty, whitespace, or semicolon-only fragments.
+      if fragment.locationData and not /^[;\s]*$/.test fragment.code
```
michaelficarra: Do we want to also ignore these insignificant characters? `(){}[]:,.`

davidbau: I have looked at a few constructs (not exhaustive by any means), and I have not found source mapping problems with those punctuation.  In other words, when they're mapped, they seem to be mapped between appropriate locations in the .coffee and .js (for example the ':' in a generated tenerary operator maps to an inline "if" in coffeescript).

Have you seen problems mapping fragments with those punctuation?  I'm inclined to just add cases where we've seen a problematic example.

michaelficarra: That's fine. I didn't know whether we were creating good mappings for any of them, so that's why I asked.

PR #3984, opened 2015-05-21
Use "Array.isArray" instead of "instanceof Array"

ide:
Array.isArray allows arrays from another JS context to be properly handled. The specific use case here is to support jest, which sets up JS contexts using Node/io.js's "vm" module.

The downside to this patch is that IE8 needs a polyfill for Array.isArray. If IE8 support for the in-browser transform (not the output of the coffeescript compiler) is important, it would be simple to include a polyfill with coffee-script.

### DIFF
--- a/lib/coffee-script/browser.js
+++ b/lib/coffee-script/browser.js
@@ -97,7 +97,7 @@
     execute = function() {
       var param;
       param = coffees[index];
-      if (param instanceof Array) {
+      if (Array.isArray(param)) {
         CoffeeScript.run.apply(CoffeeScript, param);
         index++;
         return execute();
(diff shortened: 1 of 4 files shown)


michaelficarra:
Sorry, we can't accept PRs that drop ES3 support.

michaelficarra:
Though we could do a similar check, `"[object Array]" is {}.toString.call element`

ide:
I was hoping some kind of shim would be acceptable for environments that don't support Array.isArray natively, like
```coffee
isArray = (value) ->
  if Array.isArray?
    Array.isArray value
  else
    '[object Array]" is Object::toString.call element
```

ide:
@michaelficarra: I updated the commit with your suggestion of checking with `toString`. Could you take a look please?

PR #3985, opened 2015-05-21
Replace "instanceof Array" in transformer with "[object Array]" comparison

ide:
Testing with `'[object Array]' is Object::toString.call element` allows arrays from another JS context to be properly handled. The specific use case here is to support jest, which sets up JS contexts using Node/io.js's "vm" module. This approach works in ES3 environments in contrast with ES5's `Array.isArray`.

### DIFF
--- a/lib/coffee-script/helpers.js
+++ b/lib/coffee-script/helpers.js
@@ -67,7 +67,7 @@
     flattened = [];
     for (i = 0, len1 = array.length; i < len1; i++) {
       element = array[i];
-      if (element instanceof Array) {
+      if ('[object Array]' === Object.prototype.toString.call(element)) {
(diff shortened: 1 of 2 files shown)


lydell:
In #3984 you also changed browser.coffee, but you haven't here. Is that intentional?

ide:
Yeah, since I didn't have a need for it and I didn't expect browser.coffee to run in a Node vm context (which was having trouble with instanceof Array in the first place).

michaelficarra:
LGTM.

DmitrySoshnikov:
Is it a good to go for merge? There is a dependency where we wait for this PR.

DmitrySoshnikov:
Thanks!

PR #4003, opened 2015-06-07
Ruby-esque symbol tokens

jaredp:
support `:foo == "foo"`

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -28,7 +28,7 @@
       code = this.clean(code);
       i = 0;
       while (this.chunk = code.slice(i)) {
-        consumed = this.identifierToken() || this.commentToken() || this.whitespaceToken() || this.lineToken() || this.stringToken() || this.numberToken() || this.regexToken() || this.jsToken() || this.literalToken();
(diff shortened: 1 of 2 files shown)


michaelficarra:
This doesn't appear to provide any value. See previous issues that have proposed this feature.

Ravenstine:
Symbols don't exist < ES6, but apparently they are a feature in ES6.  So I guess it depends on how much CoffeeScript wants to support ES6.  The problem is having to fall back to strings, but I guess that's the developer's job to provide a polyfill for it if they decide to use symbols.  Honestly, I wouldn't mind this as a feature because being able to use a shorthand for symbols is much simpler than writing `Symbol('foo')` everywhere.

christopherdumas:
:+1:

bjmiller:
ES6 symbols are not Ruby symbols.  They share a name, but they do very different things.

I believe that most ES6 features should not be supported by CS, because CS already does those things better.  A shorter syntax for ES6 Symbols may qualify for inclusion at some point, but it's not a priority if it can be used given existing syntax.

jashkenas:
This is a duplicate of a previous ticket.

And for the record, I tend to think that Symbols are one of the new "bad parts" of ES6. CoffeeScript shouldn't support them (although it may be inevitable in the future if they're required for certain APIs).

Argorate:
@jashkenas Why you said it's a "bad parts of ES6"?

Issue #3810, opened 2015-01-27
`TypeError` occurred whenever call `eval` on io.js 1.0.*


jacobbubu:
``` coffee
coffee = require 'coffee-script'
console.log coffee.eval '1+1'
```

```
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
  at Object.exports.eval (/Users/shenrong/tmp/coffeeeval/node_modules/coffee-script/lib/coffee-script/coffee-script.js:162:34)
  at Object.<anonymous> (/Users/shenrong/tmp/coffeeeval/index.coffee:2:13)
  at Object.<anonymous> (/Users/shenrong/tmp/coffeeeval/index.coffee:1:1)
...
```

jashkenas:
I'm not seeing this problem. `coffee.eval` is working alright on iojs 1.0.4 for me.

jacobbubu:
I'm on MacOS Yosemite 10.10.2, using npm version(1.8) of coffee. It appeals on 1.9 too with  subtle changes on error stack:

```
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
  at Object.exports.eval (/Users/shenrong/tmp/coffeeeval/node_modules/coffee-script/lib/coffee-script/coffee-script.js:176:34)
  at Object.<anonymous> (/Users/shenrong/tmp/coffeeeval/index.coffee:2:13)
  at Object.<anonymous> (/Users/shenrong/tmp/coffeeeval/index.coffee:1:1)
  at Module._compile (module.js:446:26)
...
```

jashkenas:
How about with the latest version of io.js, and the latest (1.9) version of CoffeeScript?

jacobbubu:
Has the same issue with the latest op.js (1.0.4) and CoffeeScript(1.9).

jacobbubu:
The exception should be thrown by https://github.com/jashkenas/coffeescript/blob/master/src/coffee-script.coffee#L158

GeoffreyBooth:
Saving the example as `test.coffee` and running `coffee test.coffee` in CS2 under Node 7.9.0 returns `2`.

PR #4019, opened 2015-06-22
Don't copy arguments or caller from require

rvagg:
Causes an error in io.js where strict-mode is set on internal modules.

Fixes: https://github.com/jashkenas/coffeescript/issues/3810

I'm trying to use CoffeeScript as part of the smoke tests for io.js releases and this is the blocker because it just doesn't run.

### DIFF
--- a/src/coffee-script.coffee
+++ b/src/coffee-script.coffee
@@ -157,7 +157,8 @@ exports.eval = (code, options = {}) ->
       sandbox.module  = _module  = new Module(options.modulename || 'eval')
       sandbox.require = _require = (path) ->  Module._load path, _module, true
       _module.filename = sandbox.__filename
-      _require[r] = require[r] for r in Object.getOwnPropertyNames require when r isnt 'paths'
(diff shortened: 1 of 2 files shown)


src/coffee-script.coffee
```diff
@@ -157,7 +157,8 @@ exports.eval = (code, options = {}) ->
       sandbox.module  = _module  = new Module(options.modulename || 'eval')
       sandbox.require = _require = (path) ->  Module._load path, _module, true
       _module.filename = sandbox.__filename
-      _require[r] = require[r] for r in Object.getOwnPropertyNames require when r isnt 'paths'
+      for r in Object.getOwnPropertyNames require when [ 'paths', 'arguments', 'caller' ].indexOf(r) is -1
```
lydell: `when r not in ['paths', 'arguments', 'caller']`

rvagg: nice, I thought there was a better way but couldn't find it in a quick scan of the docs, thanks

michaelficarra:
LGTM. Thanks.

Issue #1714, opened 2011-09-19
coffee -bpe "2 for [1..2]; a in b" fails


josephg:
```
$ coffee -bpe "2 for [1..2]; a in b"
Error: Parse error on line 1: Unexpected 'FORIN'
    at Object.parseError (/usr/local/lib/node_modules/coffee-script/lib/parser.js:472:11)
    at Object.parse (/usr/local/lib/node_modules/coffee-script/lib/parser.js:548:22)
    at Object.compile (/usr/local/lib/node_modules/coffee-script/lib/coffee-script.js:29:22)
    at /usr/local/lib/node_modules/coffee-script/lib/command.js:149:33
    at Object.run (/usr/local/lib/node_modules/coffee-script/lib/command.js:44:14)
    at Object.<anonymous> (/usr/local/lib/node_modules/coffee-script/bin/coffee:7:27)
    at Module._compile (module.js:411:26)
    at Object..js (module.js:417:10)
    at Module.load (module.js:343:31)
    at Function._load (module.js:302:12)
```

PR #4029, opened 2015-07-08
Reset @seenFor in lexer before tokenizing

yjerem:
The #1714 bug still occurs when the lexer is used multiple times, e.g. in the REPL:

```coffeescript
$ coffee
coffee> 0 for [1..2]
[ 0, 0 ]
coffee> 'a' in ['b']
[stdin]:1:5: error: unexpected in
'a' in ['b']
    ^^
coffee> 'a' in ['b']
false
```

This PR fixes it by setting `@seenFor = no` in the same place the rest of the lexer state is initialized.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -23,6 +23,7 @@
       this.indents = [];
       this.ends = [];
       this.tokens = [];
+      this.seenFor = false;
       this.chunkLine = opts.line || 0;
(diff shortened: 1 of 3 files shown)


michaelficarra:
Why did you also regenerate the parser?

yjerem:
Oops, I superstitiously did that at some point and left it in... amended the commit.

michaelficarra:
Thanks. Merging.

PR #4031, opened 2015-07-08
Add explicit instructions for compiling from source

JonathanPorta:
I was unable to find the instructions for compiling from source in any of the documentation. After poking at it, this is what I figured out. Apologies in advance if I missed the instructions somewhere or left something out.

I was inspired by this SO post: http://stackoverflow.com/questions/31185576/compiling-coffeescript-with-itself/31291745


michaelficarra:
Just use `bin/cake `.

Issue #4048, opened 2015-07-25
Npm releases do not contain extra directory


XVincentX:
Hello!

Given that Npm releases are without extras directory, I cannot require `/extras/coffee-script` from my code and have coffee-script support on the browser.

Would it be possible to include it back?

michaelficarra:
See my comment in #4056.

PR #4056, opened 2015-08-04
Include browser version in extras directory

XVincentX:
Will solve #4048

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,7 +5,6 @@
 Cakefile
 documentation/
 examples/
-extras/coffee-script.js
 raw/
 src/
 test/


michaelficarra:
Nope, this should not be distributed with the npm package. If you want it, extract it from the appropriate [github release](https://github.com/jashkenas/coffeescript/releases).

XVincentX:
Hey,thank you for your prompt reply.
According to your use case, I think I am missing something. How do I import the client side version of coffeescript from the Npm package?

michaelficarra:
You wouldn't. You can use bower to install CoffeeScript (including the browser build) or just fetch one of the release tarballs. npm distributes the files that are meant to be consumed by node when CoffeeScript is installed globally or as the dependency of a node project.

XVincentX:
I disagree.Today npm modules are not only used for node project, but for frontend dependencies as well. See angular, jquery, hammerjs, moment, minim and so on.
But I respect your position; that is why the fork button actually exists ;)

bjmiller:
If you're using CS for browser code, then use browserify with coffeeify.  Making the user take the performance hit of local transpilation is probably a bad idea.

XVincentX:
Hey,I am already using webpack with coffee-script loader, but I have a live editing part that should be transformed in the browser.
Thanks anyway for the tip.

bjmiller:
In that case...  You can point npm to the github address, and it will download master.  That should have what you need.

XVincentX:
I am afraid it won't do the trick. `.npmignore` file still gets respected.
Tested with
`npm install jashkenas/coffeescript#master`

and
`npm install jashkenas/coffeescript#2d1a6fa6ec76b4e721da7e5e4a9c6b90172309ae`

bjmiller:
I wonder if it makes sense to publish a separate entry in npm for just this one thing?  Just musing...

michaelficarra:
npm is just not the right tool for the job here. Bower or cURL/tar are what you want.

XVincentX:
Michael, are you sure of that? Personal experiences? Documentation?
http://blog.npmjs.org/post/101775448305/npm-and-front-end-packaging

michaelficarra:
Yep, that article basically makes the point for me. They point out a bunch of the pain points of using npm for managing front-end dependencies and show how bower solves those better. Package managers are designed with specific contexts, use cases, and distribution models in mind. npm wanting to be the single package manager for *all things JavaScript* is laughable.

XVincentX:
I see the article on the other side, but ok, your point of view is clear and I respect that.
Thanks!

XVincentX:
Ok, I see your point of view and I respect it.
Thanks!

PR #4059, opened 2015-08-16
Closes #4036: "Try catch" optimisation

### LINKED ISSUE #4036: "Try catch" optimisation
The following code compiles with additional self assignment in catch section when user set exception variable name as _error.

```coffeescript
try something()
catch _error
  console.log _error
```

compiles to:
```javascript
try {
  something();
} catch (_error) {
  // unnecessary operations
  _error = _error;

  console.log(_error);
}
```

Here is extra assignment operation

```coffeescript
try something()
catch error
  console.log error
```

compiles to:
```javascript

try {
  something();
} catch (_error) {
  // extra assignment
  error = _error;
  console.log(error);
}

```
Maybe coffeescript compiler may remove this unnecessary code?

BrunoBernardino:
Let me know if there's something I should be doing differently as this is my first contribution to coffeescript.

I fixed the reported issue where a generated variable could clash with a user-defined one in a try/catch block.

I added a test for a few scenarios with different variable names for a try/catch, to confirm the fix and avoid regressions.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1892,14 +1892,11 @@ exports.Try = class Try extends Base
   # is optional, the *catch* is not.
   compileNode: (o) ->
     o.indent  += TAB
+    scope     = o.scope
     tryPart   = @attempt.compileToFragments o, LEVEL_TOP
 
     catchPart = if @recovery
-      # Prevent user-defined error variable clashing with the generated one
-      if not @errorVariable or @errorVariable.value isnt '_error'
-        generatedErrorVariableName = '_error'
-      else
-        generatedErrorVariableName = '__error'
+      generatedErrorVariableName = scope.freeVariable '_error'
       placeholder = new Literal generatedErrorVariableName
       @recovery.unshift new Assign @errorVariable, placeholder if @errorVariable
       [].concat @makeCode(" catch ("), placeholder.compileToFragments(o), @makeCode(") {\n"),
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1896,7 +1896,7 @@ exports.Try = class Try extends Base
     tryPart   = @attempt.compileToFragments o, LEVEL_TOP
 
     catchPart = if @recovery
-      generatedErrorVariableName = scope.freeVariable '_error'
+      generatedErrorVariableName = scope.freeVariable 'error'
       placeholder = new Literal generatedErrorVariableName
       @recovery.unshift new Assign @errorVariable, placeholder if @errorVariable
       [].concat @makeCode(" catch ("), placeholder.compileToFragments(o), @makeCode(") {\n"),
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1895,12 +1895,17 @@ exports.Try = class Try extends Base
     tryPart   = @attempt.compileToFragments o, LEVEL_TOP
 
     catchPart = if @recovery
-      placeholder = new Literal '_error'
+      # Prevent user-defined error variable clashing with the generated one
+      if not @errorVariable or @errorVariable.value isnt '_error'
+        generatedErrorVariableName = '_error'
+      else
+        generatedErrorVariableName = '__error'
+      placeholder = new Literal generatedErrorVariableName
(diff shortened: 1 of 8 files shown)


vendethiel:
What happens if the user has a variable named `__error` then? :-)

BrunoBernardino:
It'll go back to `_error`. Tests prevent it. It's just one variable, both cases won't happen :)

vendethiel:
What I meant to say was that -- instead of hardcoding names, you should use `scope.freeVariable` (like the rest of the compiler does)

You can look at `scope.litcoffee`'s explanation of `freeVariable` to understand why it's more robust.

test/scope.coffee
```diff
@@ -33,6 +33,15 @@ test "catch statements should introduce their argument to scope", ->
...
+test "#4036 catch statements' error variable should not clash with user-defined variable", ->
+  compile = (code) -> CoffeeScript.compile(code, bare: yes).trim().replace(/\s+/g, " ")
+
+  eq "var _error; try { something(); } catch (__error) { _error = __error; console.log(_error); }", compile("try something()\ncatch _error\n  console.log _error")
+
+  eq "var __error; try { something(); } catch (_error) { __error = _error; console.log(__error); }", compile("try something()\ncatch __error\n  console.log __error")
+
+  eq "var error; try { something(); } catch (_error) { error = _error; console.log(error); }", compile("try something()\ncatch error\n  console.log error")
+
 test "loop variable should be accessible after for-of loop", ->
```
lydell: Don’t do fragile string comparisons of the output JavaScript. Test that the CoffeeScript works the way intended (just like other tests do). The tests regrading [variables named similar to generated variables](https://github.com/jashkenas/coffeescript/blob/2d1a6fa6ec76b4e721da7e5e4a9c6b90172309ae/test/assignment.coffee#L417-L466) might help.

BrunoBernardino:
Oh, thanks for that, I didn't know! I'll redo this PR with that then. :)

vendethiel:
Great! I'll close this one in the meantime :).

BrunoBernardino:
I thought I could just update this branch? No worries, I'll make another PR and mention this one, then.

lydell:
There’s nothing stopping you from updating this PR. It’ll get re-opened.

vendethiel:
Oh, both works! Sorry.

BrunoBernardino:
@lydell should I add the tests you mentioned, or do those _instead_ of the one I wrote?

vendethiel:
It'd be more robusts to use the new ones instead of the old ones (that's how it currently works)

BrunoBernardino:
This issue does depend on the fact the generated code (not functionality) is "wrong", though...

BrunoBernardino:
I'll test the variable's scope, because it should be the same, and with the previously generated code it wasn't.

vendethiel:
Indeed, please test that instead of the code.

BrunoBernardino:
Done. The test I added, with the previous code, would fail at the second `eq _error, 'foo'` with `AssertionError: Expected undefined to equal foo` :)

All good now.

vendethiel:
LGTM otherwise

src/nodes.coffee
```diff
@@ -1892,15 +1892,17 @@ exports.Try = class Try extends Base
   # is optional, the *catch* is not.
   compileNode: (o) ->
     o.indent  += TAB
+    scope     = o.scope
     tryPart   = @attempt.compileToFragments o, LEVEL_TOP
 
     catchPart = if @recovery
-      placeholder = new Literal '_error'
+      generatedErrorVariableName = scope.freeVariable '_error'
```
lydell: Why not simply `o.scope.freeVariable` and remove `scope = o.scope` (which btw could be written `{scope} = o`) above?

BrunoBernardino: It was just in case the variable was needed for something else. Do you prefer I write it like that?

BrunoBernardino:
Ok. I think it's got all suggestions in now :)

jashkenas:
Nice fix! Thanks for contributing.

Issue #4005, opened 2015-06-09
`([a = {}]..., b) ->` does not define `a`


xarvh:
With both 1.8.0 and 1.9.3, running
```
whatever = ([a = {}]..., b) ->
  console.log a

whatever()
```
will produce `ReferenceError: a is not defined`.

If the default assignment `= {}` is removed, `a` is defined correctly.

bjmiller:
I'm sort of amazed that this compiles.  What is it trying to do, exactly?

metalim:
Get the first array element of first argument and default it to empty object. Kinda cheesy...

xarvh:
@metalim Nice solution.
However, coffee should either reject the syntax, either define the variable.

pepkin88:
While ECMAScript 2015 does, CoffeeScript does not support default values for destructuring assignment.

xarvh:
The problem is not whether or not it should be supported.
The problem is that it produces non viable js code without any error or warning for the user.

lydell:
Fixed by #4069.

Issue #3288, opened 2013-12-17
ES6 Default values for destructuring


lydell:
Dug out of #1558. It felt buried there. See <https://github.com/jashkenas/coffee-script/issues/1558#issuecomment-22776449> and the following three comments for details.

nlundquist:
+1

DylanPiercey:
+1

lydell:
Fixed by #4069.

PR #4069, opened 2015-08-22
Implement ES2015-like destructuring defaults

### LINKED ISSUE #1558: Default values for destructuring parameters
What I'm proposing is the ability to supply default values for parameter lists (note the {}):
    

    # this works:
    constructor: (name, description = 'no description') ->

    # so does this:
    constructor: (@name, @description = 'no description') ->

    # this does not works:
    constructor: ({name, description = 'not described'}) ->

    # nor this:
    constructor: ({@name, @description = 'not described'}) ->

I was surprised when this didn't compile, considering you can set defaults for regular named arguments, and also destructuring arguments (i.e. all types of named arguments except parameter lists).

It'd also be nice to be able to describe parameter lists over multiple lines, but I couldn't think of a way to do it without making it rather ({ugly}), eg:

    constructor: ({
        @name
        @description 
    }) ->

edit: the above actually works.

Additionally, the destructuring parameter lists syntax isn't mentioned in the docs anywhere, and the syntax is really neat:

	class Man
		constructor: ({@name, @description }) ->

	dave = new Man
		name: 'Dave'
		description: 'a dude'
	console.log "#{dave.name} is #{dave.description}"

edit: rewritten to use the term 'destructuring parameter lists' instead of 'unknown auto-assigned option things huh?'. 
edit 2: clarified the ticket, as I didn't actually understand what I had written myself.

lydell:
This let's you do things like:

    fullName = ({first = 'John', last = 'Doe'}) -> "#{first} #{last}"

Note: CoffeeScrits treats `undefined` and `null` the same, and that's true in
the case of destructuring defaults as well, as opposed to ES2015 which only uses
the default value if the target is `undefined`. A similar ES2015 difference
already exists for function parameter defaults. It is important for CoffeeScript
to be consistent with itself.

    fullName2 = (first = 'John', last = 'Doe') -> "#{first} #{last}"
    assert fullName('Bob', null) is fullName2(first: 'Bob', last: null)

Fixes #1558, #3288 and #4005.


lydell:
Ah, I forgot single evaluation tests.

lydell:
While working on this I had a problem with a failing expansion test, which I solved by changing the semantics for that test. However, I now realized that isn’t necessary anymore, so I removed it. That should make this easier to merge. I’ll open a separate issue about the expansion stuff.

lydell:
Fixed merge conflicts.

jashkenas:
Pretty cool how (relatively) simple this is to implement.

For the record — historically, I haven't been in favor of adding the complexity of defaults to the destructuring syntax, which is already one of the most complex parts of the language.

But if JavaScript is doing it, okay.

aristidesfl:
The following results in "Cannot read property 'foo' of undefined"
```
example = (message, {foo = 1, bar}) -> return {foo, bar}
example(message) 
```
This works though:
```
example = (message, {foo = 1, bar} = {}) -> return {foo, bar}
example(message) 
```

Is this intended? Could we default to {} automatically?

lydell:
Yes, it is intended. We decided to make it work like ES6.

PR #4074, opened 2015-08-29
Document destructuring defaults

lydell:
A suggestion on how to document this new and shiny feature :) Feedback most welcome.


michaelficarra:
LGTM.

jashkenas:
I don't think that the null/undefined difference is worth mentioning again right here, but thanks!

PR #4075, opened 2015-08-30
Documentation improvements

lydell:
- Make the site easier to use on phones and other less wide screens.
- Get rid of the deprecated `<tt>` tag.

Before:

before

After:

navbar
code


vendethiel:
That looks good! :) (i'm not sure we need to load the compiler at all on mobie tho)

lydell:
True. Let’s do that some other time. :)

michaelficarra:
The examples should still all have a single line terminator at EOF.

lydell:
@michaelficarra They do, or am I mistaken?

DylanPiercey:
@vendethiel I think the compiler should stay. Even on mobile its handy to see the compiled js a tinker a bit. 

In my ideal world the repl would be mobile friendly as well. My thoughts are that (at mobile) you have a toggle that switches between coffeescript and JavaScript so that both don't have to fit on the screen at one time.

lydell:
@DylanPiercey I agree with you too. Ideally we’d on-demand load the repl and give it a nice interface. But not in this PR. I just tried to make minimal changes to improve the current state.

michaelficarra:
@lydell I'm not sure how to tell without checking it out. Some of the files in the diff have the no-newline octicon and some do not, so I'm sure *something* is off regarding trailing newlines.

lydell:
I'll double-check it tonight then. I'm using vim, though, and I don't know of any way to save a file in vim _without_ a newline character at EOF. (I use a small perl script to remove it afterwards in those rare occasions I cannot have one there.)

michaelficarra:
The only other thing I could imagine is Github is trying to indicate that the file used to not end with a newline, but that doesn't make sense since you are removing multiple trailing newlines.

jashkenas:
@lydell — You're a hero. This is wonderful. I'd like to cherry-pick this over to gh-pages and publish right away. Mind squashing into a single commit to make that easy?

lydell:
Sure, we could do that, though I was planning to make a v1.10.0 release PR right after my two open PRs were merged, which would get it out anyway.

jashkenas:
Lovely, that works too then.

lydell:
@michaelficarra I just ran the following program:

```coffee
fs = require 'fs'
path = require 'path'
dir = 'documentation/coffee'
for file in fs.readdirSync dir when path.extname(file) is '.coffee'
  contents = fs.readFileSync(path.join(dir, file)).toString()
  unless /\S\n$/.test contents
    console.log file
```

This is the output:

```
block_comment.coffee
cake_tasks.coffee
do.coffee
expressions.coffee
expressions_comprehension.coffee
fat_arrow.coffee
generators.coffee
scope.coffee
switch.coffee
```

So there _are_ files that lack a `\n` at EOF, but I didn’t touch any of those files in this PR.

lydell:
I’ll try to create the 1.10.0 PR tomorrow.

michaelficarra:
@lydell Open a PR? Should probably check other source/test files as well.

lydell:
@michaelficarra I’d like to make such a PR, but after the 1.10.0 PR, which I unfortunately didn’t get the time to do tonight because of unforeseen things to do :) I’ll try tomorrow night instead.

xixixao:
@lydell /clap

PR #4079, opened 2015-09-03
CoffeeScript 1.10.0

lydell:
I followed the steps I wrote after having made the 1.9.3 PR: https://github.com/jashkenas/coffeescript/wiki/%5BHowto%5D-Hacking-on-the-CoffeeScript-Compiler#11-preparing-a-new-release

Step 8:

> Make sure that you haven’t forgot to update the version number somewhere. If you update from 1.9.9, git grep '1\.9\.9' should only find matches in the change log.

However:

```
documentation/coffee/aliases.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/array_comprehensions.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/block_comment.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/cake_tasks.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/chaining.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/classes.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/comparisons.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/conditionals.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/constructor_destructuring.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/default_args.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/do.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/embedded.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/existence.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/expansion.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/expressions.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/expressions_assignment.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/expressions_comprehension.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/expressions_try.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/fat_arrow.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/functions.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/generators.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/heredocs.js:// Generated by CoffeeScript 1.9.3
documentation/coffee/heregexes.js:// Generated by CoffeeScript 1.9.3
[...]

PR #4080, opened 2015-09-03
Ensure that all source files end with a single `\n`

lydell:
This was done by first checking all files in the repository root by hand, and
then by doing the following:

    $ npm i get-stdin@4
    $ cat clean.coffee
    fs = require 'fs'
    stdin = require 'get-stdin'

    stdin (text) ->
      for file in text.trim().split('\n')
        contents = fs.readFileSync file
          .toString().replace /\s*$/, '\n'
        fs.writeFileSync file, contents
      return
    $ ls !(node_modules)/**/*.coffee | coffee clean.coffee

Ideally merged _after_ #4079.


michaelficarra:
Thanks.

Issue #3958, opened 2015-04-28
try without catch/finally is not documented


socketpair:
```
try
  42
```

is NOT (!) syntax error, and converted to

```
try {
  42;
} catch (_error) {}
```

effectively hiding any exceptions in ``try`` block. Either fix implementaion (require either ``finally`` or ``catch`` block), or documentation (say that ``catch``/``finally`` is optional and mean empty body in that case).

I prefer code fixing instead of documenting.

michaelficarra:
This feature is intentional. The `try` documentation needs some improvement.

SonOfLilit:
Currently it is documented as:

```
Try-expressions have the same semantics as try-statements in JavaScript, though in CoffeeScript, you may omit both the catch and finally parts.
```

I'd understand it as saying that this:

```
try
  f()
```

is the same as this:

```
try
  f()
finally
  null
```

which I would expect to do the same as:

```
f()
```

This is not a good description of the implemented behavior, which is to translate

```
try
  f()
```

to

```
try {
  f();
} catch (_error) {}
```

What's worse, this creates a kind of bug that is very hard to notice (hiding exceptions, thus hiding that there was a problem.) Beginners could spend days baffled by this.

Is there a reason not to change:

```
try
  f()
```

into a compiler error, or at least a warning? It would fail to compile some existing code, but I can't think of any situation where such code is written intentionally, and not because of wrong documentation or deleting some code without thinking of the surprising consequences it would have on remaining code.

I recently read a really good explanation of why these situations should be compile errors, by Eric Lippert from the C# design team in [this](http://www.informit.com/articles/article.aspx?p=2425867) article (under "#10: The empty statement does nothing for me"). It's not too late.

SonOfLilit:
Also, since I care, I'd love to submit a PR if the idea is accepted.

lydell:
I’ve seen _lots_ of CoffeeScript code where `try` without `catch` is used, as well as _lots_ of JavaScript code that uses empty `catch` blocks. People do that. This would be a _massive_ backwards compatible change.

I also don’t understand why beginners would have a hard time because of this. Otherwise they’d likely write `try fn() catch` and get the same problem.

SonOfLilit:
Because this:

```
try
  f()
```

doesn't indicate visually that it catches anything, and in fact looks like it should be identical to this:

```
try
  f()
finally
  # nothing
```

(and could be arrived at by erasing that empty finally block, not thinking you've changed anything semantically)

[...]

PR #4081, opened 2015-09-03
document try without catch nor finally

SonOfLilit:
try without catch/finally is not documented #3958

### DIFF
--- a/documentation/coffee/try-optional.coffee
+++ b/documentation/coffee/try-optional.coffee
@@ -1,0 +1,1 @@
+try data = JSON.parse(x)
--- a/documentation/index.html.js
+++ b/documentation/index.html.js
@@ -884,6 +884,10 @@ Expressions
       parts. The catch part may also omit the error parameter if it is not needed.
     </p>
     <%= codeFor('try') %>
+    <p>
+      Omitting both is a useful idiom for performing an operation and suppressing all exceptions:
+    </p>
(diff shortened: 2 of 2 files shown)


SonOfLilit:
Should I have compiled the .coffee example to .js, or is this done automatically when uploading to the website? I couldn't find a guide on doing it, I assume I could just look in the build system config though.

jashkenas:
I think it's already documented adequately. But thanks.

> Try-expressions have the same semantics as try-statements in JavaScript, though in CoffeeScript, you may omit both the catch and finally parts.

SonOfLilit:
I've asked a fellow programmer and he expected behavior to be not to suppress exceptions when `catch` is omitted.

Can I ask you to poll 3-4 non-CoffeeScript programmers on how they understood the existing documentation, and decide whether additional clarification is needed based on their responses? (Or if you trust me, I could do the experiment and report results.)

socketpair:
I prefer that `try` without a `catch` should be `syntax error`. Even in form `try xxx=do_some_fun()`.

Exceptions in javascript is very sensitive thing, so syntactically hiding situation where exception is swallowed is a very BAD thing.

carlsmith:
Maybe I missed something, but if a lone try-statement did not "suppress" exceptions, what should it do? You may not like the form, but is not ambiguous, in that there is only one sane expectation, which is the correct one.

PR #4085, opened 2015-09-06
Remove confusing * from the command help

xixixao:
The help does not describe how to require multiple modules (`-r first -r second`) but that star makes it look like there is some shortcut (which there isn't). Unless this is some UNIX convention I am not aware of.


michaelficarra:
We can probably get rid of all the brackets (`[]`) as well.

xixixao:
Eh, nope, that aint gonna work, these are used by the parser. Not great from user perspective though...

PR #4093, opened 2015-09-13
Remove unnecessary brackets and empty spaces around yield

alubbe:
This removes unnecessary brackets and empty spaces around yield when not needed. We still need brackets for if(a === (yield)).

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1837,9 +1837,11 @@ exports.Op = class Op extends Base
       else if @first.expression?
         parts.push @first.expression.compileToFragments o, LEVEL_OP
     else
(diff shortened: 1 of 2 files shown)


src/nodes.coffee
```diff
@@ -1837,9 +1837,11 @@ exports.Op = class Op extends Base
       else if @first.expression?
         parts.push @first.expression.compileToFragments o, LEVEL_OP
     else
-      parts.push [@makeCode "(#{op} "]
+      parts.push [@makeCode "("] if o.level >= LEVEL_PAREN
+      parts.push [@makeCode "#{op}"]
```
vendethiel: why is op interpolated?

Issue #3357, opened 2014-02-06
`for` comprehension doesn't works with `return` and `yield`


alexeypetrushin:
This code doesn't compiles

```
f = ->
  return for v in list
    v
```

GeoffreyBooth:
The initial example returns an “unexpected indentation” error in CS2, so I think that resolves this particular issue.

Added @xixixao’s invalid JS example to #3709, which appears to be related if not identical.

PR #4095, opened 2015-09-13
Add support for standalone yield

alubbe:
This breaks compatibility with
```coffee
->
  yield for i in [1..3]
    i * 2
```
and
```coffee
->
  yield
    i * 2
```
yield's behaviour now mirrors that of return in that it can be used stand alone as well as with expressions. Thus, it currently also inherits the above limitations.


vendethiel:
LGTM :+1: . Nice work on these PRs :).

alubbe:
It would be great to reenable the above, also for `return`. See https://github.com/jashkenas/coffeescript/issues/3357

alubbe:
Also, thanks so much to @lydell for being an awesome pair on these 3 PRs :+1:

vendethiel:
> yield's behaviour now mirrors that of return in that it can be used stand alone as well as with expressions. Thus, it currently also inherits the above limitations.

The only construct(s) that do not follow this rule, AFAIK, are function calls (and even those are very strict in what they allow) and functions (that's the implicit block rule)

alubbe:
Just to understand it a bit further, what is the consensus here, should the above cases be legal coffee-script?

vendethiel:
I think it should be allowed if possible (much the same for `return`)

lydell:
Agreed. Either way, it’s out of scope for this PR.

jashkenas:
Bravo.

PR #4111, opened 2015-09-26
Add inline sourcemap support

### LINKED ISSUE #3678: put source maps into a datauri

DylanPiercey:
This closes #3678 and is similar to #3697, however this PR also add's inline maps to "coffee-script/register" and reuses existing "browser.coffee" generation. This also allows for file names to be used in the source map.

It also allows CS to work with node-inspector by directly using register.

---

CLI Usage:
```console
coffee -c -M myFile.coffee
coffee -c --inline-map myFile.coffee
```
Require Hook Usage:
```javascript
require("coffee-script/register");
require("myFile.coffee");
```

---

Let me know if any modifications are needed for this PR as it is my first to coffeescript.


DylanPiercey:
Sorry for closing and opening - being dumb today :D

lydell:
I think not depending on btoa-lite would increase the chance of this getting merged.

DylanPiercey:
Alrighty, this no longer uses "btoa-light".

src/coffee-script.coffee
```diff
@@ -19,6 +19,11 @@ exports.FILE_EXTENSIONS = ['.coffee', '.litcoffee', '.coffee.md']
 # Expose helpers for testing.
 exports.helpers = helpers
 
+# Function that allows for btoa in both nodejs and the browser.
+toASCII = (src) ->
+  if typeof window isnt 'undefined' then btoa(src)
```
michaelficarra: This is the wrong kind of test. You should test for the existence of `btoa`, not assume it is there. Same with `Buffer`. If neither are there, give a nice error.

src/coffee-script.coffee
```diff
@@ -19,6 +19,11 @@ exports.FILE_EXTENSIONS = ['.coffee', '.litcoffee', '.coffee.md']
 # Expose helpers for testing.
 exports.helpers = helpers
 
+# Function that allows for btoa in both nodejs and the browser.
+toASCII = (src) ->
```
michaelficarra: `toASCII` is a poor name for this. I would call this `base64encode`.

DylanPiercey:
@michaelficarra Your recommended changes have been made. Let me know if any other changes are needed.

src/command.coffee
```diff
@@ -39,6 +39,7 @@ SWITCHES = [
   ['-i', '--interactive',     'run an interactive CoffeeScript REPL']
   ['-j', '--join [FILE]',     'concatenate the source CoffeeScript before compiling']
   ['-m', '--map',             'generate source map and save as .js.map files']
+  ['-I', '--inline',          'store sourcemap directly in compiled .js file']
```
michaelficarra: I would prefer

```
['-M', '--inline-map', 'generate source map and include it directly in output']
```

or something to that effect.

DylanPiercey: @michaelficarra thanks for your prompt review.

I've switched to your recommended commands and also ensured that `--map` is enabled by default with `--inline-map`.

Also I cleaned up URI generation a little. In `browser.coffee` it was encoding and decoding the uri for no reason (that I know of). I looked at how [babel is doing it](https://github.com/thlorenz/convert-source-map/blob/master/index.js#L65) and they don't do this.

Again, let me know if anything else is needed.

src/coffee-script.coffee
```diff
@@ -19,6 +19,12 @@ exports.FILE_EXTENSIONS = ['.coffee', '.litcoffee', '.coffee.md']
 # Expose helpers for testing.
 exports.helpers = helpers
 
+# Function that allows for btoa in both nodejs and the browser.
+base64encode = (src) ->
+  if typeof Buffer is 'function' then new Buffer(src).toString('base64')
+  else if typeof btoa is 'function' then btoa(src)
+  else throw new Error('Unable to base64 encode inline sourcemap.')
```
lydell: I find this a bit more readable:

```coffee
base64encode = (src) -> switch
  when typeof Buffer is 'function'
    new Buffer(src).toString('base64')
  when typeof btoa is 'function'
    btoa(src)
  else
    throw new Error('Unable to base64 encode inline sourcemap.')
```

DylanPiercey: Ah, I often do this. It's because originally there were only an if and an else. I'll get to refactoring this later in the day.

DylanPiercey: Done :+1:

src/coffee-script.coffee
```diff
@@ -84,10 +96,17 @@ exports.compile = compile = withPrettyErrors (code, options) ->
...
-    answer = {js}
-    answer.sourceMap = map
-    answer.v3SourceMap = map.generate(options, code)
-    answer
+    v3SourceMap = map.generate(options, code)
+
+    if options.inlineMap
+      sourceMapDataURI = "//# sourceMappingURL=data:application/json;base64,#{base64encode v3SourceMap}"
+      sourceURL        = "//# sourceURL=#{options.filename ? 'coffeescript'}"
+      "#{js}\n#{sourceMapDataURI}\n#{sourceURL}"
```
lydell: Mind explaining to me why sourceURL is needed? Also, are you sure things work (do you test several browsers?) having _both_ at the same time?

src/browser.coffee
```diff
@@ -26,10 +26,8 @@ return unless window?
 # Ported from https://developer.mozilla.org/en-US/docs/DOM/window.btoa
 if btoa? and JSON? and unescape? and encodeURIComponent?
```
lydell: I wonder if this check is still needed. Also, have you tested if the test and fix from commit b2d6997 still works after your changes?

src/coffee-script.coffee
DylanPiercey: This was taking from https://github.com/jashkenas/coffeescript/blob/master/src/browser.coffee#L32

Not sure if this is needed, I have only tested it in chrome where it appears to do nothing.

src/browser.coffee
DylanPiercey: I removed the `unescape` and `encodeURIComponent` checks as they are no longer used for the inline sourcemaps. As far as I know these functions simply undo each other so I am not sure why they were there in the first place.

I looked through other libraries that provide inline sourcemaps and none of them do this.

src/coffee-script.coffee
lydell: I see. I’ve never noticed that before, since the line was so long. Good job making it more readable! :+1:

lydell:
Would you mind squashing your two commits? Then I’m +1 on this :)

DylanPiercey:
I've merged the two commits, let me know if anything else is needed :).

DylanPiercey:
Is anything else needed to get this merged?

src/command.coffee
```diff
@@ -185,7 +186,7 @@ compileScript = (file, input, base = null) ->
     else
       compiled = CoffeeScript.compile t.input, t.options
       t.output = compiled
-      if o.map
+      if o.map and not o['inline-map']
```
michaelficarra: What is the reason for this change? I imagine someone may want to have both an inline and an external source map if they gave both options.

DylanPiercey: I'm not sure it's possible to have both an inline-map and an external map the browser would just go with whichever was last.

michaelficarra: I don't care much what the browser would do. One may want to generate inline source maps and external maps at the same time, and there's no reason for one of these options to cause another option to not take effect. If we really did want to make these options exclusive, we would error out immediately and tell the user to choose only one of them.

DylanPiercey: Since browsers wouldn't support this I would say that throwing when both options are used would be the best bet.

Is this okay?

michaelficarra: I'd prefer the alternative, but I'd be okay with that if that's what the other maintainers preferred.

DylanPiercey: Any other thoughts on this? I'll get to it on the weekend if this is the final verdict.

lydell: I've never seen a tool that supports both inlining _and_ a separate map file at the same time, and I've never seen anyone request it.

michaelficarra: :confused:

DylanPiercey: Would love to see a consensus on this as it would be handy to be merged :D.
My vote is for leaving as is or throwing an error.

DylanPiercey: @michaelficarra and @lydell I have made the decision to switch to babel which supports this currently; If any changes are needed in this PR i'd still be happy to make them but otherwise it can be closed.

XVincentX: @DylanPiercey What do you mean by switching to babel? Moving to ES6? Or - did you make it work with coffeescript? I'd love to see this merged as it would be really really beneficial for us! :+1:

DylanPiercey: @XVincentX I am no longer using coffee-script for any new projects, just maintaining old ones.

This PR does fix the issue, but the core contributors haven't been able to reach a decision to merge.

XVincentX: I am on your same boat (coffeescript for legacy purposes), but I'd love to be able to see this merged

svicalifornia: @DylanPiercey @michaelficarra Browsers will not fail simply because there are both inline and external source maps. The behavior may be indeterminate, but that is not itself a good reason to prevent this use case. Besides, browsers are not the only consumer of compiled CoffeeScript.

I say give the developer the power to use both options simultaneously. And let's get this merged in.

lydell: @svicalifornia :+1: I agree.

XVincentX: Indeed sirs, please let's do this.

michaelficarra: Waiting on @DylanPiercey to update the PR then.

DylanPiercey:
I think we are good to go?

src/command.coffee
DylanPiercey: @michaelficarra @lydell this was the only change needed right? If not just let me know.

src/command.coffee
```diff
@@ -39,6 +39,7 @@ SWITCHES = [
   ['-i', '--interactive',     'run an interactive CoffeeScript REPL']
   ['-j', '--join [FILE]',     'concatenate the source CoffeeScript before compiling']
   ['-m', '--map',             'generate source map and save as .js.map files']
+  ['-M', '--inline-map', 'generate source map and include it directly in output']
```
michaelficarra: The spacing here is inconsistent with the context.

jashkenas:
As @michaelficarra notes, this might still need a little tweak here or there later — but this is great stuff! Thanks!

DylanPiercey:
@jashkenas Thanks for merging this. Is it worth me fixing the spacing in a separate PR? I meant to get to it but got a little busy.

michaelficarra:
@jashkenas Actually, I think this is still very wrong. This will only generate an inline map if an external map is also being generated. There's no way to have just an inline map and no external map.

@DylanPiercey Will you submit a PR with that change, or should I revert this?

DylanPiercey:
@michaelficarra is this what you mean? https://github.com/DylanPiercey/coffeescript/commit/c5e87e72782989d060569c068cdb214f9a786b57

Sorry this PR is several months old so I am forgetting a bit what I did.

michaelficarra:
Yes, that looks like what I was thinking. I'd also feel a lot better if there were some tests, but I understand it's not exactly easy to automate a test for this.

michaelficarra:
For archeological purposes, I just wanted to note that this PR continued in https://github.com/jashkenas/coffeescript/pull/4193.

PR #4112, opened 2015-09-27
Fix compiler crash with renamed destrucured params with defaults

lydell:
`({a = 1}) ->` and `({a: b}) ->` worked, but not the combination of the two:
`({a: b = 1}) ->`. That destrucuring worked for normal assignments, though:
`{a: b = 1} = c`. This commit fixes the param case.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1532,6 +1532,9 @@ exports.Param = class Param extends Base
         obj = obj.variable
       # * assignments within destructured parameters `{foo:bar}`
       if obj instanceof Assign
+        # ... possibly with a default value
+        if obj.value instanceof Assign
+          obj = obj.value
         @eachName iterator, obj.value.unwrap()
       # * splats within destructured parameters `[xs...]`
       else if obj instanceof Splat
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -2169,6 +2169,9 @@
           obj = obj.variable;
         }
         if (obj instanceof Assign) {
+          if (obj.value instanceof Assign) {
+            obj = obj.value;
+          }
           this.eachName(iterator, obj.value.unwrap());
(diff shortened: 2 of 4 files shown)


src/nodes.coffee
```diff
@@ -1532,6 +1532,9 @@ exports.Param = class Param extends Base
         obj = obj.variable
       # * assignments within destructured parameters `{foo:bar}`
       if obj instanceof Assign
+        # ... possibly with a default value
+        if obj.value instanceof Assign
```
vendethiel: what happens with, say, `({a = b = 1}) ->`?

```
$ coffee -bpe '({a = b = 1}) ->'
(function(arg) {
  var a, b, ref;
  a = (ref = arg.a) != null ? ref : b = 1;
});
```

Ouch. Good catch. But it has nothing to do with this PR. I’ll fix it in another :)

lydell: It’s actually not a bug. It means `({a = (b = 1)}) ->`. That’s valid in JavaScript too.

vendethiel: No, I was just wondering if this nesting of checks were going to break this one. We should probably have tests for this behavior, though, if it's meant to exist...

vendethiel: You could argue it's **very** weird the first `=` has `?` semantics, when the second one doesn't (I guess it's why @satyr made `?` work in coco, huh.)

lydell: @vendethiel It’s no more weird than `(a = b = 1) ->`, which has been around for ages.

I’ll add a test, then.

lydell: Test added.

vendethiel: > which has been around for ages.

Yes, coco "changed" this one

lydell:
Could we try to merge this tiny bugfix? I have real-world code that breaks due to it.

PR #4120, opened 2015-10-02
Small repl fixes

zapu:
Removed a test in `repl.coffee` that fails in newer node version, because node repl behavior changed.

Removed unnecessary parenthesis unwrapping in repl.

### DIFF
--- a/lib/coffee-script/repl.js
+++ b/lib/coffee-script/repl.js
@@ -21,7 +21,6 @@
     "eval": function(input, context, filename, cb) {
       var Assign, Block, Literal, Value, ast, err, js, ref1, referencedVars, token, tokens;
       input = input.replace(/\uFF00/g, '\n');
-      input = input.replace(/^\(([\s\S]*)\n\)$/m, '$1');
       ref1 = require('./nodes'), Block = ref1.Block, Assign = ref1.Assign, Value = ref1.Value, Literal = ref1.Literal;
       try {
         tokens = CoffeeScript.tokens(input);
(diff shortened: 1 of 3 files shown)


michaelficarra:
This has to continue working in old versions of node, though.

zapu:
I was thinking about that too, because the same applies to the removed test - it's actually useful with node version that has the behavior described.

The changes are not very bleeding edge though, I can't easily deduct when those differences first occurred in node.js, but shouldn't people on very old node versions just use old coffee-script?

michaelficarra:
I don't see why this won't continue to work in new versions, though. If there's no wrapper, it won't strip anything.

zapu:
Oh, the unwrapping is completely fine, removing it is more of a code cleanup than fixing anything.

The broken test, however, affects new versions. I'll try to investigate what exactly has changed there, I have v0.10 installed as well and it prints undefined for empty statements, as well as for calls that return undefined.

```
» node --version
v0.10.35

» node
> console.log('test')
test
undefined
>
```

michaelficarra:
We should just bail out early and immediately prompt again if we get an empty (besides whitespace) input. Expressions that produce an undefined value should still show `undefined`.

carlsmith:
It's really helpful to not print anything when the value is undefined. If you work interactively, you want helper functions that work like coreutils, which equates to outputting something to stdout, then returning undefined.

If Python printed None values interactively, then `print "hello"` would print...

    hello
    None

In Coffee, we have the advantage of having two null values, `null` for explicit nothing, and `undefined` for a valueless value. It makes a lot of sense to print null values, but not undefined ones interactively. From doing a lot with IPython, I personally wish Python had two kinds of Null type value, specifically for shell stuff.

zapu:
What about skipping certain tests (or having different versions of a test) based on node version? Would this work?

zapu:
This looks fixed in node v6.2.1. Something must have changed in the meantime, sorry for keeping this pull request open.

PR #4126, opened 2015-10-14
Add incrementing index to make sourceURL unique

deram:
Firefox debugging with source maps and multiple in browser transpiled coffeescript segments fails miserably.
Adding index helps debugger to be able to use breakpoints etc.


GeoffreyBooth:
Fixed in the `2` branch.

PR #4135, opened 2015-10-24
Allow `throw null`

alxarch:
Throwing `null` results in a `TypeError` due to a property check in `helpers.updateSyntaxError` and an `err.stack` access in `command.compileScript()`

### DIFF
--- a/test/exception_handling.coffee
+++ b/test/exception_handling.coffee
@@ -140,3 +140,10 @@ test "parameter-less catch clause", ->
   try throw new Error 'failed' catch finally ok true
 
   ok try throw new Error 'failed' catch then true
+
+test "Allow throwing `null`", ->
+	try
+		throw null
+	catch err
+		error = err
+	ok error is null
--- a/test/helpers.coffee
+++ b/test/helpers.coffee
@@ -2,7 +2,7 @@
 # -------
 
 # pull the helpers from `CoffeeScript.helpers` into local variables
-{starts, ends, repeat, compact, count, merge, extend, flatten, del, baseFileName} = CoffeeScript.helpers
+{starts, ends, repeat, compact, count, merge, extend, flatten, del, baseFileName, updateSyntaxError} = CoffeeScript.helpers
 
 
 # `starts`
@@ -123,3 +123,6 @@ test "the `baseFileName` helper returns the file name to write to", ->
     name = baseFileName sourceFileName, yes
     filename = name + ext
     eq filename, expectedFileName
+
+test "the `updateSyntaxError` helper handles `null` argument", ->
+    eq null, updateSyntaxError null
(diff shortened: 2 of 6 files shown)


test/helpers.coffee
```diff
@@ -123,3 +123,11 @@ test "the `baseFileName` helper returns the file name to write to", ->
     name = baseFileName sourceFileName, yes
     filename = name + ext
     eq filename, expectedFileName
+
+test "the `updateSyntaxError` helper allows `throw null`", ->
+    try
+        error = updateSyntaxError null
+    catch err
+        error = err
+        console.error err.stack
```
michaelficarra: What is this?

michaelficarra:
Please squash your commits.

test/helpers.coffee
alxarch: Just added a test to ensure the updateSyntaxHelper does not block null throwing in the future

alxarch:
@michaelficarra I squashed the commits as you requested

alxarch:
@michaelficarra I was thinking of a more generic test executing `throw null` and testing the coffeescript interpreter response. Is there a place for such tests somewhere?

test/helpers.coffee
```diff
@@ -123,3 +123,10 @@ test "the `baseFileName` helper returns the file name to write to", ->
     name = baseFileName sourceFileName, yes
     filename = name + ext
     eq filename, expectedFileName
+
+test "the `updateSyntaxError` helper allows `throw null`", ->
```
lydell: Couldn’t this whole test be replaced by `eq null, updateSyntaxError null`?

alxarch: Without the try/catch the previous code would throw a `TypeError` that would not be caught in a tidy fashion from the test. If this is not considered an issue I can change it. I think a better place for testing this fix is `test/exception_handling.coffee` with a simple `throws (->throw null), null`. Should I proceed thus?

vendethiel:
LGTM. @lydell @michaelficarra ?

test/exception_handling.coffee
```diff
@@ -140,3 +140,10 @@ test "parameter-less catch clause", ->
   try throw new Error 'failed' catch finally ok true
 
   ok try throw new Error 'failed' catch then true
+
+test "Allow throwing `null`", ->
+	try
```
michaelficarra: This indentation is off.

test/exception_handling.coffee
```diff
@@ -140,3 +140,10 @@ test "parameter-less catch clause", ->
   try throw new Error 'failed' catch finally ok true
 
   ok try throw new Error 'failed' catch then true
+
+test "Allow throwing `null`", ->
```
michaelficarra: Would this test have failed without your changes? Did you try it?

Issue #3949, opened 2015-04-17
Proposal: Allow splats as last expressions of loops


connec:
This would solve some of the past issues to do with nested array comprehensions, and in general makes flat-map type transformations easier to write.

``` coffee
array... for array in array_of_arrays
# // Unchanged
# for (i = 0, len = array_of_arrays.length; i < len; i++) {
#   array = array_of_arrays[i];
#   array;
# }

result = array... for array in array_of_arrays
# Unchanged ("error: unexpected ...", or more specific error, maybe)

result = (array... for array in array_of_arrays)
# result = (function() {
#   var i, len, results;
#   results = [];
#   for (i = 0, len = array_of_arrays.length; i < len; i++) {
#     array = array_of_arrays[i];
#     results.push.apply(results, array);
#   }
#   return results;
# })();

result = for i in [0..2] then ( [i, j] for j in [0..2] )...
# result = (function() {
#   var k, results;
#   results = [];
#   for (i = k = 0; k <= 2; i = ++k) {
#     results.push.apply(results, ((function() {
#       var l, results1;
#       results1 = [];
#       for (j = l = 0; l <= 2; j = ++l) {
#         results1.push([i, j]);
#       }
#       return results1;
#     })()));
#   }
#   return results;
# })();

result = for array in array_of_arrays
  if array.length > 5
    array...
  else
    array
# result = (function() {
#   var i, len, results;
#   results = [];
#   for (i = 0, len = array_of_arrays.length; i < len; i++) {
#     array = array_of_arrays[i];
#     if (array.length > 5) {
#       results.push.apply(results, array);
#     } else {
#       results.push(array);
#     }
#   }
#   return results;
# })();
```

lydell:
Cool idea!

DylanPiercey:
+1

GeoffreyBooth:
Closing due to lack of interest.

snarlynarwhal:
The spread operator is quite useful when using the [mapGetters helper of Vuex](https://vuex.vuejs.org/en/getters.html).

PR #4136, opened 2015-10-25
Support splats at the end of loop bodies

connec:
I decided to take a stab at implementing #3949 and allow splats to occur at the end of loop bodies, allowing you to 'flat map' within a comprehension:

``` coffee
for a in [1..3]
  (a * b for b in [1..3])...
# [ 1, 2, 3, 2, 4, 6, 3, 6, 9 ]
```

**Compiling**

As splats will now parse almost anywhere, a `safe` flag has been added to the `Splat` node.  If this is not true, then calling `splat.compileToFragments` will throw a parse error: "Splats are not allowed here".  This is used when building a splat param.

For the same reason, `Splat` no longer implements `unwrap` (we want to compile the splat itself, no longer just the name).

In other cases, consumers of splats need to call `compileSplat` instead of `compileToFragments` (currently only `Splat` itself, as all other access happens through `Splat.compileSplattedArray`).

**Parsing**

My first instinct was to add `Splat` as a branch of `Expression`, however this leads to ambiguity.  After some trial and error, adding `Splat` as another branch of `Line` produces an unambiguous grammar.  I'm not terribly happy with how it parses, some issues include:
- You can splat a comprehension without parentheses:
  
  ```
  for a in [1..3]
    a * b for b in [1..3]...
  # actual: [ 1, 2, 3, 2, 4, 6, 3, 6, 9 ]
  # desired: Splats are not allowed here
  # [1..3]...
  # ^^^^^^^^^
  ```
- You **must** use parentheses when using postfix:
  
  ```
  arrayOfArrays = [ [ 1, 2 ], [ 3, 4 ] ]
  array... for array in arrayOfArrays
  # actual: unexpected for
  # array... for array in arrayOfArrays
  #          ^^^
  # desired: [ 1, 2, 3, 4 ]
  ```

Any hints at how to tackle this?


vendethiel:
Mh, this needs to go through @jashkenas 's approval first. I'm not too happy with it being in Line, but it seems [Splat is actually enough of a special case that it's not an expression, it's explicitly in `arg`](https://github.com/connec/coffeescript/blob/splats-in-loops/src/grammar.coffee#L418)

connec:
Sure, I appreciate this is a new language feature, and so needs review.  I just wanted to help the discussion with a sample implementation.

I missed the `Arg` case, at least it's consistent then :)

vendethiel:
Could you add a test to make sure

```
for a in b
  (a * b for b in [1..3])...
  1
```

fails?

connec:
Good shout. I'll rebase and add that in soon.

Still looking for feedback about the implementation, and/or usefulness 😛

vendethiel:
I definitely like the idea.

edemaine:
[Unearthing this via #4384]  What about putting Splat into Statement instead of Line?  As far as I can see, that would fix the second issue you list.

Ultimately, though, I think it might make sense to move this into Expression, so that we can use precedence to deal with your first issue.  We'd need to remove other scenarios where Splat is special-cased (like Arg) to remove the ambiguity, but it should be possible.

jashkenas:
Sorry, bud, but — this looks medium-to-highly problematic to me!

I agree that splatting out a comprehension is a neat use case.

But comprehensions are just expressions (as you note) — and splatting out an arbitrary expression doesn't make sense for most values, and isn't something we can distinguish at compile time. Because of that, there isn't going to be a good way to parse this syntax. To do what you're looking for, you'd have to add it to the `FOR` syntax somewhere.

And I don't think it's a good enough use-case to merit special syntax. It's rare, and learning when to use it is more onerous than just writing it out. This already works:


list = []
list.push(a * b) for a in [1..3] for b in [1..3]

connec:
That sounds fair enough. I think it *is* something we could work out at compile time, but it would need some sort of mechanism like `makeReturn` that forces every node to be aware that it might one day be splatted, which would definitely be too many shenanigans.

#4384 already captures the alternative of adding it to the `FOR` syntax.

jashkenas:
> it would need some sort of mechanism like makeReturn that forces every node to be aware that it might one day be splatted, which would definitely be too many shenanigans.

Right, but even then — almost every node can be returned. Very few nodes can be unsplatted.

edemaine:
Following up on @jashkenas's suggestion to modify `for`, what about the following changes to the grammar?

```coffeescript
  For: [
    o 'Statement  ForBody',                     -> new For $1, $2
    o 'Expression ForBody',                     -> new For $1, $2
    o 'Splat      ForBody',                     -> new For $1, $2  ## NEW
    o 'ForBody    SplatBlock',                  -> new For $2, $1  ## CHANGED
  ]
  ...
  SplatBlock: [  ## NEW
    o 'INDENT OUTDENT',                         -> new Block
    o 'INDENT SplatBody OUTDENT',               -> $2
  ]
  SplatBody: [  ## NEW
    o 'Body',                                   -> $1
    o 'Splat',                                  -> Block.wrap [$1]
    o 'Body TERMINATOR Splat',                  -> $1.push $3
  ]
```

Admittedly, I haven't for ambiguity, but I think it should be resolvable...  (This doesn't support `return splat...`, but does support it at the end of a `for` block.)

jashkenas:
> Following up on @jashkenas's suggestion to modify for, what about the following changes to the grammar?

That's not actually a suggestion. For a use case that can be easily, and more clearly written as:

```coffeescript
list = []
list.push(a * b) for a in [1..3] for b in [1..3]
```

It ain't worth adding a new form of **for** for everyone to learn. Come up with a more compelling example that occurs with regularity in real-world code, and it would be more reasonable.

edemaine:
I'm not sure they're going to get much more complicated, but here's some real (geometry) code I wrote last week which got me interested in this issue:

```coffeescript
## Find all intersections between list of polygons and given halfspace.
intersectPolygons = (polygons, halfspace) ->
  all = []
  for polygon in polygons
    some = intersectPolygon polygon, halfspace
    all.push some...
  all
```

In my opinion, this would be a lot cleaner as

```coffeescript
## Find all intersections between list of polygons and given halfspace.
intersectPolygons = (polygons, halfspace) ->
  for polygon in polygons
    intersectPolygon(polygon, halfspace)...
```

Another example from the same codebase (which is what spurred #4384):

```coffeescript
  endpoints = []
  for path in paths
    continue unless path.length
    endpoints.push path
    endpoints.push path[..].reverse()
```

which could rewrite to

```coffeescript
  endpoints = for path in paths
    continue unless path.length
    [
      path
      path[..].reverse()
    ]...
```

or

```coffeescript
  endpoints = [path, path[..].reverse()]... for path in paths when path.length
```

(This example gets more interesting if we get fancy with optimization: in the special case of array literal splat, we can directly give the arguments to `push` instead of forming a list and then unsplatting.)

But this notation is never really going to save much more space than a few lines, just like the existing `for` return value accumulation.  I like it because it's an intuitive extension of that accumulation -- it feels natural to be able to automatically `result.push returnValue...` just as you can currently `result.push returnValue`.  But I understand if you feel it's too complicated to learn.

jashkenas:
> But I understand if you feel it's too complicated to learn.

I do. And I also think that — for your second example in particular, that the first version is much easier to understand than the second one.

A nice cleanup — depending on how perf-critical these paths are, is a use of a `flatten` function for these cases:

```coffeescript
intersectPolygons = (polygons, halfspace) ->
  flatten(intersectPolygon(poly, halfspace) for poly in polygons)
```

```coffeescript
endpoints = flatten([path, path[..].reverse()] for path in paths when path.length)
```

PR #4140, opened 2015-11-02
Fix #4137: Caught errors named `undefined`

### LINKED ISSUE #4137: The caught error in `catch`-less `try`s are named `undefined`
```
$ bin/coffee -bpe 'try'
try {

} catch (undefined) {}
```

It should be called `error` just like when including `catch`:

```
$ bin/coffee -bpe 'try catch'
try {

} catch (error) {

}
```

lydell:
Previously, `catch`-less `try`s named the caught error `undefined`, instead of
`error` like usual.

I don’t know how/if to test this, which is why I included no test.

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -2740,7 +2740,9 @@
       tryPart = this.attempt.compileToFragments(o, LEVEL_TOP);
       catchPart = this.recovery ? (generatedErrorVariableName = o.scope.freeVariable('error', {
         reserve: false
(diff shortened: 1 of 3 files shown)


vendethiel:
An issue-specific test case would be to try+throw and in the catch, `assert !undefined?` I guess...

michaelficarra:
@vendethiel `undefined` is a keyword. No tests needed.

vendethiel:
We don't compile with "use strict" thoigh, do we?

michaelficarra:
You can still create a binding named undefined in strict mode. And we compile `undefined` to `void 0` anyway.

vendethiel:
If you create a binding named undefined in strict mode, it wouldn't retain its value, right? So the `!undefined?` test would be okay... Barring what you just reminded me about it being compiled to `void 0` – thanks.
I guess it's really not worth to test it with backticks, so I'll drop it. Thanks @lydell !

michaelficarra:
I'm not sure why you keep bringing up strict mode, though. The only restricted binding names in strict mode are `arguments` and `eval`. Strict mode shouldn't have an effect on this problem, even if we were compiling to the identifier `undefined`.

vendethiel:
I'm just mixing up forbidden **assignment** to undefined in strict mode (which is an exception), and binding (here in a catch clause) (it would've broken all the javascript libraries doing stuff like `!function (undefined){}()`, I guess? Not sure why that wasn't disallowed as well).

Issue #3600, opened 2014-08-21
Feature Request: Elixir's pipe operator


alisnic:
Suppose you have several method chains:

``` coffee
normalize: (data)->
   filter(strip(split(data))
```

Using the piping operator it might look like this:

``` coffee
normalize: (data)->
   split(data) |> strip |> filter
```

It's much more understandable conceptually and can be used to create long chains. Seriously this is one of the most awesome things ever. It allows thinking about code in terms of transformation, which I think is very powerful.

Issue #1339, opened 2011-05-08
Pipeline support


klumsy:
I know that this has been written up before , and the current consensus is that 1) it has a limitation that it doesn't allow functions with multiple arguments, and that it is a niche functionality only useful for DSLs.. I think that 1) it doesn't have to have that limitation, and you can look at F#'s implementation to show this, and 2 it can be very generic and useful, as seen in both F# and more so in powershell.

here is a link to how it is done in F# http://lorgonblog.wordpress.com/2008/03/30/pipelining-in-f/

Here is some coffee of mine doing a scenario. first how i have to call it now, then how i'd like to be able to call it, and third how i can do it with a helper function, then following up with how it can be done in kaffiene currently. The example is rather contrived but it should be sufficient to communicate 

``` ruby
where = (list,func) -> item for item in list when (func.call _ :item)  
select = (list, prefix,postfix) -> "#{prefix}#{item}#{postfix}" for item in list
getnums = (x) -> [1..x]
```

 # is there any easier way i can write the following below

``` ruby
results = select( (where (getnums 10) ,-> @_ > 5) , "pre","post" )
```

alert res for res in results
### 

what it really would like to do is be able, whether using | or say |> for a pipe do

``` ruby
getnum | where -> @_ > 5 | select "pre","post"
```

basically the pipe takes the output of the previous command and puts it as teh first argument, and then takes any other arguments after that (so 1st argument becomes 2nd, and second becomes third..

with my own pipe i can do something like
pipe getnum .to where, -> @_ > 5 .to select , "pre","post"

kaffiene though not complete lets me do this with

``` javascript
getnums = { [1,2,3,4,5,6,7,8,9,10] }select = (list,prefix,postfix) {
  res = []
  for x of list {
   res.push("#{prefix}#{x}#{postfix}")
    }
  return res
  }
```

and call it like this in kaffiene

``` javascript
getnums | select "pre","post"
```

generates the following javascript

``` javascript
var getnums, select; getnums = function() { return [1,2,3,4,5,6,7,8,9,10] }select = function(list,prefix,postfix) {
  var _a, x, res; res = []
  for(_a = 0; _a < list.length; _a++) { x = list[_a];
   res.push("" + (prefix) + (x) + "" + (postfix) + "")
    }
  return res
  }
__.select.call(this, getnums, "pre","post")
```

and

``` javascript
getnums | select "pre","post" | select "yo","dude"
```

``` javascript
[...]

PR #4144, opened 2015-11-12
Pipe Operator

hzamani:
Pipe operator `|>` adds its right hand expression as first argument to the left hand function invocation.

Selecting first operand was a smart choice in Elixir as they don't have function currying. I think it's also a good choice for CoffeeScript, because of JS libraries common structure on accepting "subject" (i.e. `Object|Array|String|...`) as first arg.

An answer to #1339 and #3600

```coffeescript
_ = require 'lodash'

users = [
  { user: 'barney',  active: false, age: 37}
  { user: 'fred',    active: true,  age: 21}
  { user: 'pebbles', active: true,  age: 19}
]

youngest = users
  |> _.select('active')
  |> _.sortBy('age')
  |> _.first
  |> _.get('user')
  |> _.capitalize

console.log youngest
```

### DIFF
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -582,6 +582,9 @@ grammar =
     # [The existential operator](http://jashkenas.github.com/coffee-script/#existence).
     o 'Expression ?',                           -> new Existence $1
 
+    o 'Expression PIPE Value',                  -> new Call $3, [$1]
+    o 'Expression PIPE Invocation',             -> $3.pipe $1
+
     o 'Expression +  Expression',               -> new Op '+' , $1, $3
     o 'Expression -  Expression',               -> new Op '-' , $1, $3
 
@@ -631,6 +634,7 @@ operators = [
   ['left',      'RELATION']
   ['left',      'COMPARE']
   ['left',      'LOGIC']
+  ['left',      'PIPE']
   ['nonassoc',  'INDENT', 'OUTDENT']
   ['right',     'YIELD']
   ['right',     '=', ':', 'COMPOUND_ASSIGN', 'RETURN', 'THROW', 'EXTENDS']
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -377,7 +377,6 @@ exports.Lexer = class Lexer
   # as being "spaced", because there are some cases where it makes a difference.
   whitespaceToken: ->
     return 0 unless (match = WHITESPACE.exec @chunk) or
-                    (match = /^\s+(?=\|>)/.exec @chunk) or
                     (nline = @chunk.charAt(0) is '\n')
     [..., prev] = @tokens
     prev[if match then 'spaced' else 'newLine'] = true if prev
@@ -685,7 +684,7 @@ exports.Lexer = class Lexer
   unfinished: ->
     LINE_CONTINUER.test(@chunk) or
     @tag() in ['\\', '.', '?.', '?::', 'UNARY', 'MATH', 'UNARY_MATH', '+', '-',
-               '**', 'SHIFT', 'RELATION', 'COMPARE', 'LOGIC', 'THROW', 'EXTENDS', 'PIPE']
+               '**', 'SHIFT', 'RELATION', 'COMPARE', 'LOGIC', 'THROW', 'EXTENDS']
 
   formatString: (str) ->
     str.replace STRING_OMIT, '$1'
@@ -871,7 +870,7 @@ POSSIBLY_DIVISION   = /// ^ /=?\s ///
 # Other regexes.
 HERECOMMENT_ILLEGAL = /\*\//
 
-LINE_CONTINUER      = /// ^ \s* (?: , | \??\.(?![.\d]) | :: ) ///
+LINE_CONTINUER      = /// ^ \s* (?: , | \??\.(?![.\d]) | :: | \|> ) ///
 
 INVALID_ESCAPE      = ///
   ( (?:^|[^\\]) (?:\\\\)* )        # make sure the escape isn’t escaped
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -377,6 +377,7 @@ exports.Lexer = class Lexer
   # as being "spaced", because there are some cases where it makes a difference.
   whitespaceToken: ->
     return 0 unless (match = WHITESPACE.exec @chunk) or
+                    (match = /^\s+(?=\|>)/.exec @chunk) or
                     (nline = @chunk.charAt(0) is '\n')
     [..., prev] = @tokens
     prev[if match then 'spaced' else 'newLine'] = true if prev
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -420,6 +420,7 @@ exports.Lexer = class Lexer
       tag = 'TERMINATOR'
     else if value in MATH            then tag = 'MATH'
     else if value in COMPARE         then tag = 'COMPARE'
+    else if value in PIPE            then tag = 'PIPE'
     else if value in COMPOUND_ASSIGN then tag = 'COMPOUND_ASSIGN'
     else if value in UNARY           then tag = 'UNARY'
     else if value in UNARY_MATH      then tag = 'UNARY_MATH'
@@ -683,7 +684,7 @@ exports.Lexer = class Lexer
   unfinished: ->
     LINE_CONTINUER.test(@chunk) or
     @tag() in ['\\', '.', '?.', '?::', 'UNARY', 'MATH', 'UNARY_MATH', '+', '-',
-               '**', 'SHIFT', 'RELATION', 'COMPARE', 'LOGIC', 'THROW', 'EXTENDS']
+               '**', 'SHIFT', 'RELATION', 'COMPARE', 'LOGIC', 'THROW', 'EXTENDS', 'PIPE']
 
   formatString: (str) ->
     str.replace STRING_OMIT, '$1'
@@ -806,6 +807,7 @@ NUMBER     = ///
 
 OPERATOR   = /// ^ (
   ?: [-=]>             # function
+   | \|>               # pipe
    | [-+*/%<>&|^!?=]=  # compound assign / compare
    | >>>=?             # zero-fill right shift
    | ([-+:])\1         # doubles
@@ -900,6 +902,8 @@ LOGIC = ['&&', '||', '&', '|', '^']
 
 # Bit-shifting tokens.
 SHIFT = ['<<', '>>', '>>>']
+
+PIPE = ['|>']
 
 # Comparison tokens.
 COMPARE = ['==', '!=', '<', '>', '<=', '>=']
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -609,6 +609,10 @@ exports.Call = class Call extends Base
 
   children: ['variable', 'args']
 
+  pipe: (arg) ->
+    @args.unshift arg
+    this
+
   # Tag this invocation as creating a new instance.
   newInstance: ->
     base = @variable?.base or @variable
--- a/test/pipe.coffee
+++ b/test/pipe.coffee
@@ -27,3 +27,13 @@ test "Pipe chain", ->
   eq 45, 4 |> mul(10) |> add 5
   eq 70, 4 |> add(10) |> mul 5
 
+test "Pipe on the next line", ->
+
+  fn = (x,y) -> "#{x},#{y}"
+
+  eq '1,2', 1
+    |> fn(2)
+
+  eq '1,2', 1
+    |> fn 2
+
--- a/test/pipe.coffee
+++ b/test/pipe.coffee
@@ -1,0 +1,29 @@
+
+test "Pipe to function", ->
+
+  fn = (x) -> x * 100
+
+  eq 100, 1 |> fn
+  eq 300, 3 |> fn
+
+test "Pipe to partialy filled function", ->
+
+  fn = (x,y) -> "#{x},#{y}"
+
+  eq '1,2', 1 |> fn(2)
+  eq '1,2', 1 |> fn 2
+
+test "Pipe chain", ->
+
+  f = (x) -> x * 10
+  g = (x) -> x + 5
+
+  eq 15, 1 |> f |> g
+  eq 60, 1 |> g |> f
+
+  add = (x,y) -> x + y
+  mul = (x,y) -> x * y
+
+  eq 45, 4 |> mul(10) |> add 5
+  eq 70, 4 |> add(10) |> mul 5
+


lydell:
Wow, what I simple and clean patch!

aurium:
This looks nice to me. :+1:

vendethiel:
Pretty nice patch! Does it work with the pipe on the first line, and the call on the second line as well? (probably should "just work")

hzamani:
@vendethiel yes it works
```coffeescript
f = (x,y) -> x + y
g = (x) -> x * 10
h = (x,y,z) -> [x,y,z]

10 |> f(1) |>
g
|> h(1,2)
```
will be `[ 110, 1, 2 ]`

seanstrom:
@hzamani Will I be able to pipe a value to an inlined, partially applied function?
```coffeescript
add = x -> y -> x + y
10 |> add(10)
```

hzamani:
@seanstrom No, pipe is just a syntactic sugar for adding parameters to a function call, `10 |> add(20)` is same as `add(10, 20)` .

jamiter:
+1

zweifisch:
:+1: one more reason to stick with coffeescript

jamiter:
So what needs to be done to get this into CoffeeScript? :smile:

vendethiel:
@jashkenas 's approval

aurium:
I second @jamiter's question.

That waits so long to be reviewed by some commiter, that becomes conflicting.

davinov:
Lodash already adresses this concern by `_.chain` or its short `_` function. However, for many other libs that doesn't provide partial functions, that's a great idea! :+1:

taylorlapeyre:
Really looking forward to getting this really useful tool!

zimt28:
Really nice, what does @jashkenas think?

blitmap:
**Yes, please!**

dbackeus:
With functional programming styles on the rise this would be a very welcome addition to coffeescript.

Some debaters in previous discussions have insisted OO style method chaining provides equally nice syntax. However there are certainly situations where it's not possible or wanted to have those methods on the objects worked on.

A general trend can currently be seen where OO style is giving way for functional style so definitely time to rethink previous decisions. Adding pipes would enable nicer ways of dealing with more functional code designs.

hflw:
Having used this in Elixir, the pipe operator really encourages you to split things up into small functions and chain them together, it feels extremely satisfying and leads to some great code. Also leads to much simpler code compared to the OO-style chaining approach, since you're just returning the value instead of a scope that you need to hold onto in a closure. Would absolutely love to see this in Coffeescript.

cullophid:
This would be a great addition. I don't agree with the idea of adding the statement as the first argument though. It would make the feature unusable for functional programming.

cullophid:
I think `<|` would be a great addition too

blitmap:
Just wanted to add another idea:

What if the pipe operator were a forward slash (`\`) rather than `|>` ?

I realize `|>` is used in other languages, but a forward slash looks more "pipe-like" in my opinion.

```
a \ b \ c \ d
```

Might be worth considering. :>

dbackeus:
@blitmap forward slash means division so don't think that's happening.

Actually using pipe `|` would obviously be the most "pipe-like" thing but I think due to potential confusion with the bitwise operator mimicking Elixir's syntax is a good idea.

blitmap:
@dbackeus Ahh, I confused myself - but if you look in the examples I used the backslash :smile:

vendethiel:
Backslash is already used by CoffeeScript (line continuations).

blitmap:
@vendethiel Huh - TIL.  I thought `then` was used for line continuations.  So yes, my idea doesn't make sense anymore :smile:

nojvek:
Been six months this is still open. I see there is a es7 proposal for this too. Would be awesome to get this in.

lzrski:
It's a beautiful idea :)

ch1c0t:
@lydell By any chance, could it get into the next release?

nojvek:
What's the status of this in ES spec? Has it made it out of strawman?

taylorlapeyre:
okay 👋 coffeescript see ya later

nojvek:
Although I agree, it would be nice to see this in coffeescript driving the
spec forward.

Would make a ton of JavaScript cleaner.

igl:
@nojvek tc39 https://esdiscuss.org/topic/the-pipeline-operator-making-multiple-function-calls-look-great
babel-pr: https://github.com/babel/babel/pull/3159

Generally people do not understand. It's not a java-esque oo-feature, so tc39 does not like it.

While it seems like coffee stalled since people jumped to es6... this and the es6-module PR give me hope and i think the time is right for coffee to spearhead into the future with it's own principles.

lzrski:
> It's not a java-esque oo-feature, so tc39 does not like it.

Sad but true.

I also really hope that CS will move forward and survive the ES[0-9]{1,4} hype. The new JS is getting some nice features indeed, but at the same time it gets bloated and inconsistent. Even if somewhat stalled, Coffee is so much better! It would be an awful shame if it fades away.

On the main subject, I'd like to share a (crazy?) idea. Playing with Elm I developed much affection to currying and having `|>` apply the value to the last argument. Now I curry a lot in CS / JS as well. That said hzamani's position that:

> Selecting first operand [...] is a good choice for CoffeeScript, because of JS libraries common structure on accepting "subject" (i.e. `Object|Array|String|...`) as first arg.

makes a lot of sense from a pragmatic point of view. Also it's not really true that:

> adding the statement as the first argument [...] would make the feature unusable for functional programming.

It would only make it harder to work with some functional libs. Even then you can [`flip`](https://lodash.com/docs/4.16.1#flip) or [`rearg`](https://lodash.com/docs/4.16.1#rearg) on the fly and be good. It doesn't seem that data last is a necessity for FP. OP's already mentioned Elixir, which is an FP language and uses data first style - I guess the reason is compatibility with OTP, which also makes it a good analogy for us.

So how about having another operator to pass value as a last argument, maybe `!>` or `:>`? Bang looks better IMO, but it's visually very similar to pipe, so it may be confusing.

Anyway 💯 for this PR.

igl:
If TS is leaning towards OOP, coffee could lean more towards FP.
But you could argue that there is ELM and PureScript for FP and that leaves coffee in the middle again, where assigning to the left would maybe more appreciated.

With Object/List-Spread/Splat and RXJS i can live without _.fp/mori/ramda or just fall back to the regular lodash.

> The golden rule of CoffeeScript is: "It's just JavaScript"

GeoffreyBooth:
I’m not clear on why passing the first parameter makes sense. In Node it's almost universal that the first parameter of a callback is `err`, with the parameter you want usually the second one (whether there are only two or there are several).

lzrski:
@GeoffreyBooth in callbacks, yes. But they are async. I don't imagine it's feasible to pipe something to an async callback.

lzrski:
@igl:

All the languages you mentioned are awesome, but let's not underestimate the fact that CS is already a very popular and well established language.  There is a lot of existing codebase and esp.  in enterprise (where you have managers 🕴🕴🕴 and colleagues 🤔😳😱 who do not always want to live on the cutting edge) it's way easier to upgrade the CS compiler and then add more functional style to the code then migrate to a different language. Also the CS being an OO-FP hybrid (like JS) and having this awesome syntax and stuff has a unique value proposition.

In case of Elm I suspect that for quite some time we will often need to use ports and reach to JS, simply because there is so many libraries there. I would love to write the code on the other side of the port using CS.

GeoffreyBooth:
You didn’t show what your example is supposed to compile to, so taking the one from the [esdiscuss.org thread](https://esdiscuss.org/topic/the-pipeline-operator-making-multiple-function-calls-look-great):

```coffee
doubleSay = (str) -> "#{str}, #{str}"
capitalize = (str) -> str[0].toUpperCase() + str.substring(1)
exclaim = (str) -> "#{str}!"

# Proposed syntax:
result = 'hello'
  |> doubleSay
  |> capitalize
  |> exclaim

# Or:
result = 'hello' |> doubleSay |> capitalize |> exclaim

# Compiles to: var result = exclaim(capitalize(doubleSay("hello")));

# Result:
"Hello, hello!"
```

You could achieve the same in various ways without needing a new operator:
```coffee
# 1
result = exclaim capitalize doubleSay 'hello'

# 2
result = \
  exclaim \
  capitalize \
  doubleSay \
  'hello'

# 3
_ = require 'underscore' # or lodash
result = _.compose(exclaim, capitalize, doubleSay, -> 'hello')()
```
Your actual example happens to be [Lodash’s example for `_.chain`](http://devdocs.io/lodash~4/index#chain). In CoffeeScript, it could be written as:
```coffee
youngest = _.chain users
  .sortBy 'age'
  .first()
  .get 'user'
  .capitalize()
  .value()
```
Several of these examples are so similar to the `|>` syntax that I don’t see the benefit that the new operator provides.

lzrski:
Hey, @GeoffreyBooth! I'm very happy that you took interest in this PR. I also appreciate you taking critical position - it's important to avoid feature bloat. That said I wouldn't agree that the alternative examples you have provided are similar.

The first three require developer to reverse the order of functions. This is not intuitive, esp. for beginner programmers. Also if the composed functions take more then one argument, the syntax becomes much more complex, wouldn't it? The solution proposed in this example solves both problems.

The third and fourth example obviously require a third party library. Again this raises the barrier of entry - now a beginner not only needs to learn the language, understand the FP technics but also chose a library (there are more than two alternatives here) and learn it's API. Also the last one (involving `chain`) requires parens when there are no additional arguments, which is somewhat confusing and IMO ugly.

As a person who cares about teaching I would very appreciate a standard and elegant way of doing functional composition. Please reconsider :)

igl:
@GeoffreyBooth _.chain and _.compose exist only because function composition is not in the language.

@lzrski CS **used to be** a very popular and well established language ;)

GeoffreyBooth:
> The first three require developer to reverse the order of functions. This is not intuitive, esp. for beginner programmers.

But JavaScript and CoffeeScript are right-to-left when it comes to functions. We write `alert 'hello'`, not `'hello' alert`. That’s just how it is. I would argue that it’s more counterintuitive to make an exception.

> Also if the composed functions take more then one argument, the syntax becomes much more complex, wouldn't it? The solution proposed in this example solves both problems.

No. If a function took more than one argument, the return value is passed into the first argument and the remaining arguments are undefined. That’s what’s happening in the `chain`/`compose` examples; [`_.get` takes up to three arguments](http://devdocs.io/lodash~4/index#get), but we’re only passing in one value to the first argument. The proposed operator doesn’t change any of this.

> The third and fourth example obviously require a third party library. Again this raises the barrier of entry - now a beginner not only needs to learn the language, understand the FP technics but also chose a library (there are more than two alternatives here) and learn it's API.

What you’re really proposing is essentially adopting `_.compose` into CoffeeScript. My point was that that functionality is already easily accessible from Underscore/Lodash, or someone could import just that one function into their project if they want to. You could make an argument that any of Lodash’s functions are so incredibly useful that they should become part of CoffeeScript or JavaScript; that’s a slippery slope.

> Also the last one (involving chain) requires parens when there are no additional arguments, which is somewhat confusing and IMO ugly.

Again, that’s just how CoffeeScript and JavaScript are. Simply typing `alert` without parentheses is referring _to_ the function, not calling it. That’s just basic syntax.

I don’t think we should be adding language features just to make educational examples simpler, especially when the way to achieve this using existing syntax (my example 1) is already simpler than the new proposal.

nojvek:
I believe this feature is inspired by UNIX pipes.  The output of one
process is redirected into the next. Reading a |> b |> c is mentally less
taxing than c(b(a))

The other patter in js is .chaining e.g a.filter(x).sort(). This only works
if all the functions return objects with those methods.

The pipe pattern allows chaining functions without adding them to a giant
class or using a library function like _.chain

Powershell makes use of the chaining pattern extensively, so does elixr.
It's like connecting a circuit. It's not a new concept. It does make code
look cleaner.

I think coffee pushing for this will drive the es6 feature as well.

My 2 cents.

lzrski:
> But JavaScript and CoffeeScript are right-to-left when it comes to functions.

No, they are not. It's rather inside out. Please consider this:

```coffeescript
c (b (a z))
```

now let's add second argument to each function call. This is what we get:

```coffeescript
c (b (a z 2), y), x
```

It's not easy to see, but the last call is at the beginning and the end of the expression, while the fist one is in the middle and second - all over the place! Parsing it gives me instant headache and I don't think it's because I'm particularly stupid or lack experience. In contrast the new syntax is really readable:

```
a z |> b y |> c x
```

Also as @nojvek rightly pointed chaining is left to right and it is a widely used pattern exactly because it's easier to understand the data flow that way. Having ability to chain any function, regardless of it's implementation is a big plus - the reasons were given multiple times above.

> You could make an argument that any of Lodash’s functions are so incredibly useful that they should become part of CoffeeScript or JavaScript; that’s a slippery slope.

Maybe, but so far nobody is making such an argument and this PR is only about pipe operator, so please let's keep our focus on that particular feature\*

> Simply typing alert without parentheses is referring to the function, not calling it.

How about:

```coffeescript
alert "Hello"
```

I don't see any parens here, and yet I'm pretty sure it will **call** the function \#sarcasm

Lastly:

> I don’t think we should be adding language features just to make educational examples simpler

Me neither, but it's only one of multiple good reasons expressed above by several participants of this thread.

lzrski:
> You could achieve the same in various ways without needing a new operator

For every operation you can do in CS it is true that *you could achieve the same in JS without needing a different language.* In fact this is exactly the argument that ES20?? fanboys are making against CS. While the statement is true, it's not a valid argument against having a better syntax.

GeoffreyBooth:
I’m open to being convinced. Also my opinion is not the only one that matters. Why don’t you start by showing what JavaScript your initial example is supposed to compile to? How do you handle the issues brought up in the esdiscuss.org thread, such as how to handle generators or async/await?

Lots of operators have been proposed over the years; just search the issues and pull requests for “operator.” I think it’s important to keep CoffeeScript simple, not least because we need to maintain it. For a new operator to be added, I would think a compelling case needs to be made that it adds something significant to justify the cost of maintenance and of the rest of the community needing to learn it.

I’m also not sure we should add this if it might someday appear in ECMAScript. We don’t want another `class` or `...`, where they implement a feature differently than we do.

nojvek:
> I’m also not sure we should add this if it might someday appear in
ECMAScript. We don’t want another class or ..., where they implement a
feature differently than we do.

I think this is a chicken and egg problem. A huge motivation for Es6
features came from coffee script. For things to get into es6, a solid
implantation in a language like coffee-script can really help.

A programming language is an translation of an idea in the brain to
something than can run inside a machine. It's an interface.

We are arguing that a pipe operator allows a cleaner abstraction. It's
scope is limited to functions, there are already implantations in other
languages.

Coffee was an inspiration for ES6, I don't think we should be scared of the
future when we have a chance to influence it.

Coffee is not a strict superset of JavaScript unlike Typescript. I believe
that is one of its appeal, it was always one step ahead of JavaScript.

Why can't we ship this as an experimental feature?

"Hey we introduced this great pipe operator, tell us how you're using it.
Don't build huge production systems that rely on it just yet, it may have
minor changes in the near future"

kirly-af:
I just wanted to mention there's actually a [proposal](https://github.com/mindeavor/es-pipeline-operator) for that in JS, a (closed) [babel PR](https://github.com/babel/babel/pull/3159) and also a similar [babel plugin](https://github.com/miraks/babel-plugin-pipe-operator).

Although I feel this feature appealing, I kinda share @GeoffreyBooth opinion about not implementing a feature that might work differently in JS. On the one hand this feature might never come to JS (it's not even a stage 0 proposal), and on the other hand CS suffered a lot from implementation conflicts when ES6 came. Both positions make sense and it's quite hard to decide which direction CS should take: keep adding cool nice features or adopt a _wait and see_ policy...

Inve1951:
That kind of general refusal against new sugar let's me see CS's death closing in. Which is kinda sad because CS is nice.
The way it looks to me implementing all that ESNext stuff could certainly (re)gain some devs who switched since. But that decimates CS to 'less typing required'-JS.

I however want to write CS **instead** of JS so I don't care at all if they got different syntax or not. Especially if it's for a feature which isn't even available in plain JS (yet).

To me this discussion seems needless until [discuss/newSugars](https://github.com/coffeescript6/discuss/issues/44) is resolved.

vendethiel:
> That kind of general refusal against new sugar let's me see CS's death closing in. Which is kinda sad because CS is nice.

It'd be really nice if people would stop doomsaying every few weeks. Back then it was because Coffee didn't have source maps, then yield, then import, then...

Inve1951:
@vendethiel i see your point on that.
Maybe care to comment on the rest of my post.

shreeve:
@lydell - any shot you could include this? @lzrski makes some great points above and this feature is clean and elegant

nojvek:
Folks with pull request accepting permissions, Could you please summarize
what is blocking this and what requirements need to be met for merging this?

lydell:
I really like the `|>` operator (at least in Elm)!

If I have this `test.coffee` …

```coffee
2 |> Math.pow 3 |> console.log
2 |> Math.pow(3) |> console.log
2 |> (Math.pow 3) |> console.log
```

… and I run `coffee test.coffee`, what will be printed to the console?

If enough people give the same answer as this PR, I guess we might be ready to merge, otherwise we’re probably not.

kirly-af:
@lydell: I'd say:

```coffee
8
8
"TypeError: [...] is not a function"
```

@vendethiel: well I would say @Inve1951 comment was legitimate. The thing is that it's not clear anymore what the CoffeeScript policy will be about new features adoption regarding JS fast evolution. 

IMHO, that's more global question than something specifically related to this feature (or `await`, or whatever else). Because we love this language, doesn't it make sense to care about its future ?

lzrski:
As for the output I agree with @kirly-af. First two both compile to:

```js
console.log(Math.pow((2), 3));
```

Since 2^3 is 8, that's what is going to be printed. I've put parens around first arg to `Math.pow` to emphasise that it's coming from the pipe.

The third one will get compiled to:

```js
console.log((Math.pow(3))(2));
```

since `Math.pow(3)` returns `NaN` which is not a function, it would give a TypeError at runtime.

@GeoffreyBooth I appreciate your last comment regarding the concerns with generators and future compatibility. Unfortunately I haven't enough time to formulate a reasonable comment on that yet. I'll try to address it as soon as I have some time on my hand. As for the following (also reasonable) request:

> Why don’t you start by showing what JavaScript your initial example is supposed to compile to?

```coffeescript
youngest = users
  |> _.select('active')
  |> _.sortBy('age')
  |> _.first
  |> _.get('user')
  |> _.capitalize
```

it should compile to:

```js
var youngest = _.capitalize( _.get( _.first( _.sortBy( _.select( users, 'active' ), 'age' ) ), 'user' ) )
```

I think it's a pretty compelling example of turning inside-out style into left-to-right.

My abstract example:

```coffeescript
a z |> b y |> c x
```

would become:

```js
c( b( a( z ), y ), x );
```

I hope you find these satisfactory. Also please accept my apology if I got too emotional in my last comment :)

lydell:
Here is what I thought the program would output:

```
8
8
8
```

Here is the actual result:

```
3
8
TypeError: Math.pow(...) is not a function
```

Here is the compiled JS:

```js
Math.pow(2, console.log(3));
console.log(Math.pow(2, 3));
console.log((Math.pow(3))(2));
```

Let's break it down:

1. All three of us who took the challenge failed on the first line. This shows that this PR most likely has to adjust the precedence of the `|>` operator, so that it is lower than implicit function invocation.

2. The second line wasn't that hard. Great!

3. I was surprised that two people successfully understood that the third line would result in an error. Why didn't I expect that? Probably because I learned about the "`|>` concept" in a different language (Elm). I guess it was also due to my being used to that you can always change `fn 1, 2` into `(fn 1, 2)` if needed. Does this show that many people might be confused by the `|>` operator (as currently implemented)?

Actually, point 3. above got me thinking. In my opinion, it is very confusing to allow `a |> b`. I think the feature will be easier to understand if the right side of the pipe _has to be a function invocation._ In other words, that you'd need to type `a |> b()` or `a |> do b` instead. (Note: `a |> do b` is currently a syntax error.) Why? Because in `a |> (b(1))`, `b(1)` is actually evaluated as an expression, and then that result is called as a function. In `a |> b(1)`, `b(1)` is also an expression, but it is _not_ evaluated; instead, it is kind of like modified into `b(a, 1)`. I find that special treatment of certain expressions confusing.

Other than the above, there are a few implementation details that need improvements:

- The error messages when putting something invalid on the right side of the pipe need to be improved.
- Execution order needs to be guaranteed. Consider `a() |> b()(1)` – `a()` should be executed before `b()`, but it is currently the other way around.
- Information is lost in the parser. If you write `a |> b(1)`, CoffeeScript actually thinks that you wrote `b(a, 1)`. While the code is really clever and short, this information loss in the parser has always bitten us before. Instead, `a |> b(1)` should be parsed into a `Pipe` node or something.

---

In summary:

- The precedence of `|>` most likely needs to be lowered.
- The allowed grammar needs to be discussed.
- Implementation details need to be improved.

GeoffreyBooth:
Thank you @lydell. I didn't want to stoke an argument before, but I wasn't trying to be snarky when I pointed out that it matters that parentheses are required when calling functions without arguments. `f callback` and `f callback()` are both very different and very common use cases.

lzrski:
Thanks @lydell - very valuable research. I'm happy that we can discuss the merits of this PR.

Fully agree on the precedence and error message.

Regarding the parens I'm on a fence. I've also learned about the concept of pipe in Elm where auto-currying works great together with this operator. Unfortunate reality is that we don't have it in JS and as the PR's author pointend - it doesn't fit the ecosystem. That said, I don't think that this is a valid argument here:

> parentheses are required when calling functions without arguments

because in this case we are calling it **with argument**. The argument is stated on the left side of the pipe operator instead of the right side of a function name.

Since in CS we can generally call functions with or without parens (including calling it without arguments using `do` keyword or a hackish ``` fn `` ``` syntax), I would say we should stick to this style and have

```coffeescript
val |> fn
```

be equivalent to

```coffeescript
val |> fn()
```

both compiling to `fn(val)`.

In case of factory function (i.e. when we want a returned function to be a sink for the pipe), the syntax would be:

```coffeescript
val |> (factory args...)
```

or without args:

```coffeescript
val |> (factory())
```

IMO `do` should have precedent over pipe, esp. that most of the times I use it to create a closure with IIFE. That's how it's introduced on coffeescript.org:

> When using a JavaScript loop to generate functions, it's common to insert a closure wrapper in order to ensure that loop variables are closed over, and all the generated functions don't just share the final values. CoffeeScript provides the `do` keyword, which immediately invokes a passed function, forwarding any arguments.

So `do` with pipe would work like this:

```coffeescript
a = 1
b = 2
c = 4
10 |> do (a, b, c) -> (d) -> a / b / c / d
# 1 / 2 / 4 / 10 = 0.0125
```

otherwise it would be unclear what `a` stands for in the factory: 1 or 10?

Calling with `do` without arguments, e.g. `val |> do fn` would compile to following JS: `fn()(val)`.

jashkenas:
Just to leave my stinky two cents lying around in here — I don't think this feature should be moved forward with and merged.

* It's glyph-eriffic, and uncoffeescripty to that extent.
* As @lydell demonstrated above, it's confusing even to its experts and champions.
* If we do call chaining, it should probably be in more of a Dart-esque cascades way, and not a pipe way.
* It semi-confusingly provides more than one way to do a thing.

If someone can demonstrate a clean and clear tutorial of how you'd explain this feature to beginners and why you'd use it instead of regular function calls, then we can continue to putter. But I don't think we should merge.

thepeoplesbourgeois:
> - It's glyph-eriffic, and uncoffeescripty to that extent.

I was under the impression that CoffeeScript was kind of fond of glyphs? If I'm mistaken, for my curiosity, what exactly do we call the tokens, `->`, `=>`, `@`, `#{}`, `<num>..<num>`, `<num>...<num>`, `<list>...`, `**`, `//`, and `?`?

> As @lydell demonstrated above, it's confusing even to its experts and champions.

@lydell's demonstration felt like it was more to highlight the specific implementation of the pipeline operator, as defined in this pull request.  My research of functional languages indicates that pipeline operators are rather uncontroversial, used in a way very similar to what's being suggested here, and implemented within languages where the inverse-nesting examples provided by @GeoffreyBooth and @lzrski would also work.

For my money, 
```
[1,2,3,true,"dog"] 
  |> the 
  |> transforms(options)
  |> iWould 
  |> like(:totally) 
  |> toapply
```

actually **is** far more readable (and writable) than

```
toApply(like(iWould(transforms(the([1,2,3,true,"dog"]), options)), :totally))
```

and far more maintainable than 

```
ref = [1,2,3,true,"dog"]
ref = the(ref)
ref = transforms(options)
ref = iWould(ref)
ref = like(ref, :totally)
toApply(ref)
```

> If we do call chaining, it should probably be in more of a Dart-esque cascades way, and not a pipe way.

I'd be all for that! … What is it? 😶 

> It semi-confusingly provides more than one way to do a thing.

screen shot 2016-10-12 at 3 20 49 pm

thepeoplesbourgeois:
As an aside, I find it interesting that the maintainer of CoffeeScript feels this strongly about not introducing glyphs to the syntax, as it was originally CoffeeScript's own splat `...` that made me think of how I would implement a `|>`

laurentpayot:
If you want coffeescript with pipe operator you can use [livescript](http://livescript.net/#operators-piping), a  Haskell-flavored coffeescript fork.

After using livescript for a while I realized  pipe operators and other livescript syntaxic niceties were actually making my code less readable because there are too many symbols/rules to remember.

Keep coffeescript simple please.

thepeoplesbourgeois:
> After using livescript for a while I realized pipe operators and other livescript syntaxic niceties were actually making my code less readable because there are too many symbols/rules to remember.

@laurentpayot A fair number of people have written blogs over the years to levy the same complaint against CoffeeScript. I wonder what makes the argument more powerful in this particular instance, compared to previous times we've all ignored it.

laurentpayot:
@thepeoplesbourgeois coffeescript added the much needed noise removal (semicolon, brackets) with  pythonic indentation, arrow shortcuts, list comprehension etc. A relatively small set of changes with great effects.
_IMHO_ pipes are less bang for the buck than the changes listed above but I agree it is a subjective opinion.

GeoffreyBooth:
I’m not finding this thread very constructive. How about someone takes @jashkenas up on his suggestion and writes a “clean and clear tutorial of how you'd explain this feature to beginners and why you'd use it instead of regular function calls”?

Personally I still don’t see what great savings this feature adds, and I _do_ feel that it adds a nontrivial cost, both in terms of one more thing we need to maintain (and that could eventually get broken by ECMA) and one more thing that everyone using CoffeeScript needs to learn, if they encounter this operator in others’ CoffeeScript code. A tutorial explaining how _and why_ to use this operator would go a long way in persuading me why it’s worth the cost. Simply producing cleaner syntax isn’t enough in my opinion.

Doesn’t this also encourage a pattern that is out of fashion? Isn’t the general best practice nowadays to move toward object-oriented programming, where you create an object or class with methods, i.e. `joe = new Employee(); joe.setSalary(1000000)` etc., where the methods/functions act on the object itself rather than being static and acting on their input? I know there’s another school of thought that pushes immutable functions, but even looking at [immutable.js](https://facebook.github.io/immutable-js/) it very much uses an object-oriented style. Or is this operator compatible with an object-oriented pattern, and/or non-static functions? If so, how?

Along those lines, can this be used as an alternative to chaining, for libraries that already support chains? Is there a way to write `$('button').css('background-color', 'red').delay(300).fadeOut()` using the pipe operator?

thepeoplesbourgeois:
@GeoffreyBooth, 

> Along those lines, can this be used as an alternative to chaining, for libraries that already support chains?

tl;dr, no, this is something very different. Read on for more.

> How about someone takes @jashkenas up on his suggestion and writes a “clean and clear tutorial of how you'd explain this feature to beginners and why you'd use it instead of regular function calls”?

```coffeescript
Me.try(this) # => ...
```

# The Pipe Operator

Think of any time you've used `Array.prototype.<yourMethodHere>.call(theThingYouWantToMutate, <yourMethodArgs>)`. You've written this function call, ostensibly, because `theThingYouWantToMutate` has not implemented the specific `yourMethodHere` you want to use, and you either can't / won't implement `theThingYouWantToMutate.yourMethodHere`. It might look like this in your code right now:

```coffeescript
mutatedObj = originalObj
  .withSome()
  .chainedMethods()

Array.prototype.yourMethodHere.call(mutatedObj, yourMethodArgs)
  .andNow()
  .moreCoolObjectChains()
```

In order to make the call to your Array function, you need to stop in the middle of your dot chain, pass your `mutatedObj` to it. If you need to invoke other libraries, you may need to introduce even more references, or consider wrapping your function calls like this:

```coffeescript
SomeOtherLib.toCoolObject( 
  Array.prototype.yourMethodHere.call(
    originalObj
      .withSome()
      .chainedMethods(), yourMethodArgs
  )
    .andNow()
    .moreCoolObjectChains(), transformArgs
)
  .awesomeFinalMutation()
```

Already, we need to look down three lines to see that `originalObj` is the first variable we reference, but if we don't look down two more lines, we will fail to see that the argument being passed in to `Array.prototype.yourMethodHere.call` is actually whatever we return from `orginalObj.withSome().chainedMethods()`.  Besides that, because of the nesting, it looks like `yourMethodArgs` and `transformArgs` are somehow related to `.chainedMethods()` and `.moreCoolObjectChains()`, instead of to `Array.prototype.yourMethodHere.call` and `SomeOtherLib.toCoolObject`

With the pipe operator, it becomes trivial for any developer not only to see that...

```coffeescript
originalObj
  .withSome()
  .chainedMethods()
  |> Array.prototype.yourMethodHere.call(yourMethodArgs)
  .andNow()
  .moreCoolObjectChains()
  |> SomeOtherLib.toCoolObject(transformArgs)
  .awesomeFinalMutation()
```

...transforms us from `originalObj` to `SomeOtherLib.CoolObject.awesomeFinalMutation()`, but they can also freely apply other transformations from their object, or any accessible library, with little confusion.

```coffeescript
originalObj
  .withSome()
  .chainedMethods()
  |> Array.prototype.yourMethodHere.call(yourMethodArgs)
  .andNow()
  .moreCoolObjectChains()
  |> _.merge(["I","love","function","chaining"])
  |> SomeOtherLib.toCoolObject(transformArgs)
  .awesomeFinalMutation()
```

thepeoplesbourgeois:
I also just came up with a hypothetical tl;dr for `|>` and a corresponding `|<` or `<|`...

### tl;dr

Think of `42 |> function(args)` like it's saying `function(args.unshift(42))` and `42 |< function(args)` like it's saying `function(args.push(42))`, or, colloquially, `pipe into the left` and `pipe into the right`.

lydell:
How would somebody explain the function invocation expression special-case?

GeoffreyBooth:
@thepeoplesbourgeois Is your last example the same as this?
```coffee
mutatedObj = originalObj
  .withSome()
  .chainedMethods()
mutatedObj = Array.prototype.yourMethodHere.call(mutatedObj, yourMethodArgs)
  .andNow()
  .forSomethingCompletelyDifferent()
  .moreCoolObjectChains()
mutatedObj = _.merge(mutatedObj, ["I","love","function","chaining"])
mutatedObj = SomeOtherLib.toCoolObject(mutatedObj, transformArgs)
  .awesomeFinalMutation()
```

lydell:
For reference, here is what this PR produces of that example:

```js
SomeOtherLib.toCoolObject(transformArgs).awesomeFinalMutation(_.merge(Array.prototype.yourMethodHere.call(yourMethodArgs).andNow().forSomethingCompletelyDifferent().moreCoolObjectChains(originalObj.withSome().chainedMethods()), ["I", "love", "function", "chaining"]));
```

nojvek:
I really don't understand the against point. If someone doesn't like the
operator then they shouldn't use it. I'm personally not a big fan of @ ==
this so I don't use it.

The pipe is additive. It doesn't break existing code. Its another pattern
of expressing an idea to a computer in an elegant manner.

Regarding ECMA, coffee is quite diverged already.

I believe against arguments are coming from the fear of uncertain future.
Coffee won't be first. Many stable languages already implement this.

Should we create a pro/con summary and put it to a vote?

thepeoplesbourgeois:
@GeoffreyBooth 

> Is your last example the same as this?

If we want to split hairs, no, because I never needed to introduce a new variable reference

GeoffreyBooth:
@thepeoplesbourgeois, I meant are they functionally identical? If both were compiled and evaluated, would they produce identical results?

Per @lydell’s point, assuming my version is identical, then there’s a bug in this PR. The first argument of `_.merge` and of `.call` below should be the intermediate object you’re passing along, and each function should be getting two arguments instead of just one:
```js
SomeOtherLib.toCoolObject(transformArgs)
  .awesomeFinalMutation(
    _.merge(
      Array.prototype.yourMethodHere.call(
        yourMethodArgs
      )
      .andNow()
      .forSomethingCompletelyDifferent()
      .moreCoolObjectChains(
        originalObj
        .withSome()
        .chainedMethods()
      ), ["I", "love", "function", "chaining"]));
```

Aside from this issue, is the only benefit to this operator that in an example like this one, people can avoid intermediate variables? Going back to @jashkenas’ request, is that the only reason “why you'd use it instead of regular function calls”?

thepeoplesbourgeois:
> Per @lydell’s point, assuming my version is identical, then there’s a bug in this PR. The first argument of _.merge and of .call below should be the intermediate object you’re passing along, and each function should be getting two arguments instead of just one:

There's actually one more nesting error to it. Notice that in my hybrid-chain and your reference chain, the result of `_.merge(list, otherList)` is passed as the first argument to `SomeOtherLib.toCoolObject(arg, transformArgs)`.  The full referenceless function wrap would actually be

```coffeescript
SomeOtherLib.toCoolObject((
  _.merge(
    Array.prototype.yourMethodHere.call(
      originalObj
        .withSome()
        .chainedMethods(), yourMethodArgs
    )
    .andNow()
    .forSomethingCompletelyDifferent()
    .moreCoolObjectChains(), ["I","love","function","chaining"]
  ), transformArgs)
  .awesomeFinalMutation()
```

> Aside from this issue, is the only benefit to this operator that in an example like this one, people can avoid intermediate variables? Going back to @jashkenas’ request, is that the only reason “why you'd use it instead of regular function calls”?

Any piece of syntactic sugar is meant to let developers express the work that needs to be done in straightforward, clear terms. When you don't have to think as hard about the work you're trying to accomplish, you tend to work more efficiently. CS already implements `=>` expressly to avoid binding `this` to an intermediate variable, and the immediate benefit to developers was a clearer understanding of how their `=>` functions behaved.

```coffeescript
# this
self = this
(something) ->
  self.isBoundProperly()

# is definitively kludgier than
(something) =>
  @isBoundProperly()

# and we love coffeescript so much more for it!
```

Let me answer your question with a question: Is avoiding unnecessary punctuation _the only reason_ why you'd use CoffeeScript over vanilla JS?

GeoffreyBooth:
Where would you use this operator in the CoffeeScript codebase? Can you point to a few examples where that code’s legibility would be improved by the use of this operator?

I can do without the entreaties to why syntactic sugars in general are awesome. The question is why _this_ sugar is worth its cost in implementation, maintainability and the education of the community.

thepeoplesbourgeois:
> The question is why this sugar is worth its cost in implementation, maintainability and the education of the community.

What part of this discussion is insufficient in answering that question?

thepeoplesbourgeois:
> Where would you use this operator in the CoffeeScript codebase? 

Considering how no other CoffeeScript operator is used in the CoffeeScript codebase, besides `->`, how is this relevant?.

thepeoplesbourgeois:
> I can do without the entreaties to why syntactic sugars in general are awesome.

I was specifically pointing out why it's useful when syntactic sugars obviate the need for intermediate variable declaration.

thepeoplesbourgeois:
@GeoffreyBooth I've found some code I believe could be improved by this operator, but before I spend my weekend completely rewriting `rewriter.coffee`, I should get clarification around a few things:

1. What are the specific criteria by which you'll judge the success or failure of my refactor?
1. What's the threshold for maximum changes I can/should make to demonstrate the utility of this operator? ie., how much code can I change before you tell me to dial it back?
1. So that I have a reference of how much code you want to see beneficially affected by this change, what was the last feature CoffeeScript implemented that underwent this kind of stress test, and whereabouts in the source code can I find a few samples of its usage?
1. Are there any potential details of a refactor that would immediately bar my changes from being considered for the code base, and do any recent feature suggestions exemplify such a restriction?

Thank you for your answers.

GeoffreyBooth:
@thepeoplesbourgeois to be honest, @jashkenas coming out against something is usually the death knell for a proposal. So I wouldn’t spend much if any more time on this.

Personally I was trying to keep an open mind, but I feel like a lot of the arguments are just rehashes of “look how beautiful it is.” And sure, I get that beautiful syntax can lead to more understandable code, which leads to fewer bugs. So I’m open to being convinced.

But I feel like this becomes _less_ understandable for all but the simplest examples. Sure, this:
```coffee
result = 'hello'
  |> doubleSay
  |> capitalize
  |> exclaim
```
is better than the various ways it could be written in CoffeeScript today. But this:
```coffee
originalObj
  .withSome()
  .chainedMethods()
  |> Array.prototype.yourMethodHere.call(yourMethodArgs)
```
troubles me, because at a glance it looks like `.call` takes only one argument, when in fact `yourMethodArgs` is the _second_ argument, after `originalObj.withSome().chainedMethods()`. This strikes me as very irregular, and something that I would think that a lot of people would find confusing and/or objectionable.

Obviously another issue is that this currently isn’t compiling correctly, which also means it doesn’t have enough tests. So there’s a lot of work to do on that front, aside from designing the syntax itself such that people aren’t confused by it and that smart people can reliably predict what the output will be. As we saw above with @lydell’s example, it currently also fails that benchmark.

The reason I asked about real-world examples where you’d want to use this, is I wanted to see how often this would likely actually be used. I think the CoffeeScript codebase is a pretty good corpus, because there’s a great variety of code in there and many of the most arcane CoffeeScript features are used in once place or another. You’re welcome to choose something else if there’s another open-source project you know better. And I truly am open to considering examples where the pipe operator would significantly improve the legibility of real-world code as it stands today. But . . . I have a hunch that there wouldn’t be many clear-cut examples you could find where the pipe operator would be a dramatic improvement. How often do people really string together several functions, other than with chaining (which I consider superior to the pipe operator)? And even then, when it does happen I feel like most people do the intermediate-variable approach, like my `mutatedObj =` example. One could argue that the intermediate-variable pattern is easier to understand than the pipe operator equivalent, even if it’s slightly more verbose.

But this goes back to my earlier question that also went unanswered: isn’t this promoting an anti-pattern? I feel like the current recommended way of dealing with stringing together functions is to take the object-oriented approach, chaining methods: `$('button').fadeIn().delay(100).fadeOut()`. When is chaining functions, like would be encouraged by a pipe operator, a best practice?

So that’s where I am at the moment. This feels _potentially_ helpful, but also very potentially confusing; it requires a lot more work, which I’m not sure it deserves; and I’m not sure how often it would ever get used, or if we should even be encouraging its use.

danschumann:
A business needs to be resistant to change because one customer said so.
This is maybe more of a small crowd, however.

I think it raises a question about order of operations.  Everyone knows
multiplication comes before addition for non parenthesized operators.
2*2+2*2 = 8

But we have more operators than that, like 'or'

foo or 2 |> double

Do we get
foo || double(2)
double(foo||2)

How priority is this bad boy, relative to other smooth operators?

lydell:
I think it was kind of decided upon that this PR should be changed so that `|>` gets the lowest priority (precedence) of all (even lower than implicit function invocation!).

lydell:
I think it was kind of decided upon that this PR should be changed so that `|>` gets the lowest priority (precedence) of all (even lower than implicit function invocation!).

vendethiel:
@lydell No, I don't think you want lowest of them all. Think what happens for `val = x |> foo` in this case.

lydell:
Ah, good catch. So kinda lowest or lowest except assignment operators or something then.

jashkenas:
New syntax should have to pass a very high bar to be included, and this is falling quite short.

I think that there's room for discussion about a pipeline-y feature, but it would:

* Need to be intuitive.
* Need to be able to support Dart cascade-style repeated calls for side effects (OO style).
* Need to be able to support repeated operations against the same value. (Functional style).

If it can't do those things, then it probably isn't worth including — and this particular PR+talk doesn't seem like a great base to start from. Closing.

lzrski:
@jashkenas

I'm happy that you remain open to the pipe idea. Also you are obviously right that design for a new syntax needs to be really sound. Thanks for taking time on this issue.

> - Need to be able to support Dart cascade-style repeated calls for side effects (OO style).
> - Need to be able to support repeated operations against the same value. (Functional style).

I like the idea cascading. Do you mean that the considered new operator should cover both cascading and piping? The functionality seems pretty different to me - maybe I'm narrow in my thinking.

jashkenas:
I see them as being very related. One is the object-oriented way of **applying repeated operations to a value (either for side effects, or to mutate the value itself)**

```coffeescript
div.show()
  .animate(duration.slow)
  .moveTo(coordinates)
  .blink()
```

And the other is the functional way to **apply repeated operations to a value (either for side effects, or to mutate the value itself)**

```coffeescript
div |> show
  |> animate(duration.slow)
  |> moveTo(coordinates)
  |> blink
```

Note that there's strangeness with the piping operator there because it's the first argument to `animate` and `moveTo`, but it doesn't appear to be.

If we're going to introduce new syntax, and it *can't* address both of these at once, the object-oriented style would be preferable to accomplish over the functional one. This is JavaScript, after all.

lzrski:
Yes, as soon as more than one argument come into play it gets somewhat confusing.

Do you think we can consider any kind of partial application syntax? It seems to me that together they would make more sense.

lzrski:
Re. relation between cascading and piping.

If we think of `this` as an additional, implicit argument to the function, then piping and cascading does seem very related. Maybe it should be a hint in eventual future syntax design.

yetone:
Just like livescript piping?

lzrski:
@yetone LiveScript has some awesome qualities and personally I would love to use it, but it's way more difficult than CoffeeScript to introduce in a workplace. It's less known and seems like learning curve for developers accustomed to OO and imperative style is too steep. [Same arguments](https://github.com/jashkenas/coffeescript/pull/4144#issuecomment-249206361) apply here as for the PureScript or Elm mentioned earlier by @igl .

To all those disappointed by the rejection of this PR, here is a somewhat hackish way to avoid inside-out function application without big dependencies. It works by using single-element array as an applicative functor.

Inspiration comes from [Professor Frisbee's Mostly Adequate Guide to Functional Programming](https://drboolean.gitbooks.io/mostly-adequate-guide/content/ch8.html)

```coffeescript
# Coerce number to a range and make sure it's a multiple of a different number:

[ result ] = [ number ]
  .map coerce minimum, maximum
  .map round multipleOf
  .map adjust minimum, maximum, multipleOf
  .map log "Coerced number"

# Provided that:

coerce = curry (minimum, maximum, number) ->
  switch
    when number < minimum
      minimum
    when number > maximum
      maximum
    else
      number

round = curry (step, number) ->
  switch
    when not step?
      number
    when number % step is 0
      number
    when (number % step) < (step / 2)
      step * (number // step)
    else
      step * (number // step) + step

adjust = curry (minimum, maximum, step, number) ->
  switch
    when number < minimum
      number + step
    when number > maximum
      number - step
    else
      number

log = curry (label, x) ->
  console.log label, x
  return x
```

By no means it's perfect as it requires instantiation of an array and only works with curried functions. I hope it's worth sharing anyway. Where there's a will, there's a way :)

bbugh:
Every day I don't have a pipe operator `|>` (#4144) in any language is a sad day. It's so much more natural than nested calls especially in languages that already chain. I would love if Coffeescript added this one, it fits pretty well I think with how Coffeescript works already. Hopefully Coffeescript 2 will consider this!

thepeoplesbourgeois:
I will say that it's odd but somewhat comforting that the CoffeeScript committee is showing restraint compared to TC39 around [questions like this one]((https://github.com/tc39/proposal-bind-operator/issues/44))

aurium:
https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Pipeline_operator

jessaustin:
@aurium the example at that link is amusing because the expression _f∘ g∘ f_ is always commutative. It seems deliberately confusing about the fact that `f(g(h(x)))` is equivalent to `x |> h |> g |> f`.

danschumann:
The pipe is the language equivalent of "go into"
x goes into h
x |> h
h(x)

x goes into h, which go into g
(x |> h) |> g
(or apparently by the order of operations is not needed to have parenthesis)
g(h(x))

I would probably not recommend using too many of these.. but I can see how it would be nice to have.

PR #4145, opened 2015-11-17
Fix spelling mistake.

randruc:
Hi,

While reading the annotated source code, I noticed what seems to be a spelling mistake in a annotation comment.
As English isn't my mother tongue I'm not quite sure about that.

Regards,

Rémi Andruccioli

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -18,7 +18,7 @@ locationDataToString,  throwSyntaxError} = require './helpers'
 # The Lexer Class
 # ---------------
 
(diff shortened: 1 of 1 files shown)


randruc:
It's ok, I've just been told about the strange verb :-)

carlsmith:
It sounds better as *divides it into*, not *divides it up into*. Otherwise, dropping *divvies it up* makes the docs clearer I think.

randruc:
Well... In my opinion *divides* is a verb that is much more common for non English-speaking people.

PR #4155, opened 2015-12-01
[wip][rfc] Introduce annotation-comment (breaking syntax change)

ryoqun:
# Summary

Currently, CoffeeScript provides no reliable way to retain comments through
compilation.

Block comments are just fine most of time, however we're faced to a
kind-of-hard dilemma described below.

Simply put, we want some comments used for annotation are passed down to the
compiled JavaScript as is and at the exact position of arbitrary token boundary
as long as they are allowed. Currently supported block comments aren't suffice.
# Problem background

We are developing an Angular app with CoffeeScript using ng-annotate.
Then, we stumbled upon this issue (ref).

(As for justification of annotation itself, I actually don't want to resort to
such things. Though, Angular practically requires such practice otherwise we're
forced to write many repetition for the Dependancy-Inject thingy..)
# Problem

``` CoffeeScript
# ??? means some annotation syntax

???@ngInject???
MainCtrl = (Page) ->
  console.log("say hello")
```

We want the above CoffeeScript code is compiled likes this:

``` JavaScript
/* @ngInject */
MainCtrl = function(Page) {
  return console.log("say hello");
};
```

Then, JavaScript code is fed into ng-annotate, producing this as a final result:

``` JavaScript
/* @ngInject */
MainCtrl = function(Page) {
  return console.log("say hello");
};
MainCtrl.$inject = ["Page"];
```
# Current working solution

the "use strict" style:

``` CoffeeScript
MainCtrl = (Page) ->
  "ngInject"
  console.log("say hello")
```

=>

``` CoffeeScript
bash-4.3$ cat examples/annotate.coffee | ./bin/coffee --compile --stdio --bare --no-header | ng-annotate --add -
var MainCtrl;

MainCtrl = ["Page", function(Page) {
  "ngInject";
  return console.log("say hello");
}];
```

However this isn't applicable to more complex annotation usage.
# Almost working solution: the block comment style

``` CoffeeScript
### @ngInject ###
MainCtrl = (Page) ->
  console.log("say hello")

### @ngInject ###
MainCtrlNotAtTop = (Page) ->
  console.log("say hello")

```

=>

``` JavaScript
/* @ngInject */
var MainCtrl, MainCtrlNotAtTop;

MainCtrl = function(Page) {
  return console.log("say hello");
};


/* @ngInject */

MainCtrlNotAtTop = function(Page) {
  return console.log("say hello");
};
MainCtrlNotAtTop.$inject = ["Page"];
```

The first block comment is moved at the first of file. I believe this is intentional,
after all, there is ambiguity between a given block comment at the first of file is
a file header comment or commenting about the following code..

However this isn't applicable to more complex annotation usage.
# Almost working solution: javascript comment inside javascript label

``` CoffeeScript
`/* @ngInject */`
MainCtrl = (Page) ->
  console.log("say hello")

`/* @ngInject */`
MainCtrlNotAtTop = (Page) ->
  console.log("say hello")
```

=>

``` JavaScript
/* @ngInject */;
var MainCtrl, MainCtrlNotAtTop;

MainCtrl = function(Page) {
  return console.log("say hello");
};

/* @ngInject */;

MainCtrlNotAtTop = function(Page) {
  return console.log("say hello");
};
```

Maybe the `;` is preventing ng-annotate from working correctly...

There is top of file or scope does exist here as well..

Also, this isn't applicable to more complex annotation usage.
## Proposed solution: introducing _annotation-comment_ syntax

The newly-introduced syntax: 'Annotation comment'

So, we resorted to make changes to the CoffeeScript to implement a
general-purpose transpilation-resistent annotation comment syntax.

If this change would be accepted, I'd commit my available time to realize it.

Currently, I've just hacked up a very proof-of-concept pull request.
## New syntax considerations

```
`/* ngInject */`
```
- good: maximum compatibility is retained
- bad: more chars

```
#`ngInject`
```
- good: less chars
- bad: moderate compatibility is retained
- bad: hard to parse because we're trained to ignore the whole line (subjective?)

ref: https://github.com/olov/ng-annotate/issues/152


ryoqun:
ping @jashkenas Could you take a look a bit if you have some time?

src/coffee-script.coffee
```diff
@@ -62,6 +62,7 @@ exports.compile = compile = withPrettyErrors (code, options) ->
   js = ""
   for fragment in fragments
     # Update the sourcemap with data from each fragment
+    fragment.annotateCode()
```
GeoffreyBooth: This is definitely not the way to do this. Your annotation comment would be another type of token, like block comments. You need to define them in grammar.coffee.

src/command.coffee
```diff
@@ -393,7 +393,11 @@ printTokens = (tokens) ->
   strings = for token in tokens
     tag = token[0]
     value = token[1].toString().replace(/\n/, '\\n')
-    "[#{tag} #{value}]"
+    if annotation = token[2]?.annotation
```
GeoffreyBooth: This file should not be changed.

src/grammar.coffee
```diff
@@ -52,7 +52,9 @@ o = (patternString, action, options) ->
   action = action.replace /LOC\(([0-9]*)\)/g, addLocationDataFn('$1')
   action = action.replace /LOC\(([0-9]*),\s*([0-9]*)\)/g, addLocationDataFn('$1', '$2')
 
-  [patternString, "$$ = #{addLocationDataFn(1, patternCount)}(#{action});", options]
+  a = [patternString, "$$ = #{addLocationDataFn(1, patternCount)}(#{action});", options]
+  #console.log(a)
```
GeoffreyBooth: Please take care to not commit debugging code. Also, it would appear that this is more your issue: the insertion of a semicolon after a block comment. Without that semicolon, would your problem be solved?

src/helpers.coffee
```diff
@@ -84,13 +84,15 @@ exports.invertLiterate = (code) ->
...
 buildLocationData = (first, last) ->
+  #console.log("build:", first, last)
   if not last
     first
   else
     first_line: first.first_line
     first_column: first.first_column
     last_line: last.last_line
     last_column: last.last_column
+    annotation: first.annotation || last.annotation # Bad?
```
GeoffreyBooth: Annotation comments are not part of the location data.

GeoffreyBooth:
Implemented via #4572.

Issue #2518, opened 2012-08-28
`do for item in list` ?


jashkenas:
An interesting proposal raised by @lehni, here:

https://github.com/jashkenas/coffee-script/issues/2382#issuecomment-8032361

... is to somehow combine the `do` and `for` keywords to give you a quick and easy way to write a loop that closes over its variables ... _without_ hiding the fact that you're working with an implied inner function, and all of the scope changes that involves. I think it might work out nicely.

```
do for num in [1, 2, 3]
```

... becomes the equivalent of:

```
for num in [1, 2, 3]
  do (num) ->
```

Any ideas?

GeoffreyBooth:
A pull request for this would be welcome.

connec:
I think for CS2 the right solution to the actual problem here would be a `for let` syntax.

```coffee
for let i in [1..5]
  setTimeout -> console.log i

for let key, value of object
  ...
```

Another option would be to take the proposal here and combine it with #4514 to produce the same output.

connec:
Just gonna drop a link to https://github.com/jashkenas/coffeescript/issues/959#issuecomment-634034 - the original issue adding `do` also makes several mentions of possible shortcuts for IIFEs in loops.

treeform:
I was going to propose `for bar in foo ->` syntax. I feel like its better syntax then `for let bar in foo ->`, `for bar in foo do ->`, `for scope bar in foo ->`, `do for scope bar in foo ->` and feels just like a shorter version of `for bar in foo then do (bar) ->` which we use a ton.

Is this issue closed or patches welcome?

jashkenas:
Let's leave this closed — it's been long enough that we don't need to be adding a new iteration construct at this point...

PR #4167, opened 2015-12-26
Fix scoping error

robkinyon:
I understand these are generated files, but there's an error in the generated javascript. The `ext` variable isn't scoped, so the `ext` in the error string is always the last value in the array, regardless of what the actual error is.

I expect you will decline this PR, but I hope this will point you in the right direction.

### DIFF
--- a/lib/coffee-script/coffee-script.js
+++ b/lib/coffee-script/coffee-script.js
@@ -209,9 +209,11 @@
     for (i = 0, len = ref.length; i < len; i++) {
       ext = ref[i];
       if ((base = require.extensions)[ext] == null) {
+        (function (ext) {
         base[ext] = function() {
           throw new Error("Use CoffeeScript.register() or require the coffee-script/register module to require " + ext + " files.");
         };
+        })(ext);
       }
     }
   }


vendethiel:
Oh, good catch. [indeed](https://github.com/jashkenas/coffeescript/blob/master/src/coffee-script.coffee#L181)

vendethiel:
@jashkenas #2518 ?

lydell:
```coffee
diff --git a/src/coffee-script.coffee b/src/coffee-script.coffee
index 69957fd..b6efad6 100644
--- a/src/coffee-script.coffee
+++ b/src/coffee-script.coffee
@@ -175,7 +175,7 @@ exports.register = -> require './register'
 
 # Throw error with deprecation warning when depending upon implicit `require.extensions` registration
 if require.extensions
-  for ext in @FILE_EXTENSIONS
+  for ext in @FILE_EXTENSIONS then do (ext) ->
     require.extensions[ext] ?= ->
       throw new Error """
       Use CoffeeScript.register() or require the coffee-script/register module to require #{ext} files.
```

robkinyon:
I would argue that, as in Python and Ruby, `then do (ext) ->` should be the default behavior for all for-loops. Otherwise, the scoping is wrong and you end up with the wrong behavior.

vendethiel:
Well; loops used to capture every locals "a long time ago", but it's been changed for performance reasons

robkinyon:
@vendethiel Unless the compiler can detect when the locals are going to be bound to a function and auto-capture in that case, then you need to capture otherwise you end up with a bug.

I mean, the coffeescript devs were caught by this. What does that bode for us mortals?

PR #4169, opened 2015-12-31
Update copyright date

zachpanz88:
Update copyright year

Happy new year!


michaelficarra:
There's no content in this repo created in 2016, so I don't see a need for this.

zachpanz88:
I assumed that it would be good to change in the event that files are changed in the future and just to keep the copyright date updated to the current year

bjmiller:
That's not how copyright works.  Your copyright year is the last year that you created content.  You can't claim copyright protection "just in case" like that.  Goodness, could you imagine the havoc if that were the case?

PR #4198, opened 2016-02-02
Refactor `Literal` into several subtypes

### LINKED ISSUE #2009: Bug with backtick
Consider this code:

    class jQueryCalendar
      saveDate: (inputField) ->
        alert 'Saving Date' + $(inputField).val()
    
      initCalendarObjects: ->
        params =
          onSelect: (dateText, inst) =>
            @saveDate `this`
        $('.timelineCalendar').datepicker params

After running coffee -c -b on this file, the following javascript is generated:

    var jQueryCalendar;
    
    jQueryCalendar = (function() {
    
      function jQueryCalendar() {}
    
      jQueryCalendar.prototype.saveDate = function(inputField) {
        return alert('Saving Date' + $(inputField).val());
      };
    
      jQueryCalendar.prototype.initCalendarObjects = function() {
        var params,
          _this = this;
        params = {
          onSelect: function(dateText, inst) {
            return _this.saveDate(_this);
          }
        };
        return $('.timelineCalendar').datepicker(params);
      };
    
      return jQueryCalendar;
    
    })();

Notice that what was in backticks incorrectly replaced. "this" was replaced with "_this"

A workaround is to put a space after the backtick: \` this\` instead of \`this\`

But it seems like this is a bug that needs to be fixed.

### LINKED ISSUE #4192: Expand string interpolation expressions into dedicated AST
Hi, as seen in #3831, string interpolation like `"A #{a} B #{b}"` is lexed as if it was `'A '+a+' B '+b`.

as mentioned in juliankrispel/decaf#2, there are use cases for having the semantic information survive lexing.

Therefore it would be awesome if we could introduce a new `StringInterpolation` AST node which is later converted to a concatenation expression.

lydell:
Previously, the parser created `Literal` nodes for many things. This resulted in
information loss. Instead of being able to check the node type, we had to use
regexes to tell the different types of `Literal`s apart. That was a bit like
parsing literals twice: Once in the lexer, and once (or more) in the compiler.
It also caused problems, such as `` `this` `` and `this` being indistinguishable
(fixes #2009).

Instead returning `new Literal` in the grammar, subtypes of it are now returned
instead, such as `NumberLiteral`, `StringLiteral` and `IdentifierLiteral`. `new
Literal` by itself is only used to represent code chunks that fit no category.

`StringWithInterpolations` has been added as a subtype of `Parens`, and
`RegexWithInterpolations` as a subtype of `Call`. This makes it easier for other
programs to make use of CoffeeScript's "AST" (nodes). For example, it is now
possible to distinguish between `"a #{b} c"` and `"a " + b + " c"`. Fixes #4192.

`SuperCall` has been added as a subtype of `Call`.

Note, though, that some information is still lost, especially in the lexer. For
example, there is no way to distinguish a heredoc from a regular string, or a
heregex without interpolations from a regular regex.

After the new subtypes were added, they were taken advantage of, removing most
regexes in nodes.coffee. `SIMPLENUM` (which matches non-hex integers) had to be
kept, though, because such numbers need special handling in JavaScript (for
example in `1..toString()`).

An especially nice hack to get rid of was using `new String()` for the token
value for reserved identifiers (to be able to set a property on them which could
survive through the parser). Now it's a good old regular string.

In range literals, slices, splices and for loop steps when number literals
are involved, CoffeeScript can do some optimizations, such as precomputing the
value of, say, `5 - 3` (outputting `2` instead of `5 - 3` literally). As a side
bonus, this now also works with hexadecimal number literals, such as `0x02`.

Finally, this also improves the output of `coffee --nodes`:

    # Before:
    $ bin/coffee -ne 'while true
      "#{a}"
      break'
    Block
      While
        Value
          Bool
        Block
          Value
            Parens
              Block
                Op +
                  Value """"
                  Value
                    Parens
                      Block
                        Value "a" "break"

    # After:
    $ bin/coffee -ne 'while true
      "#{a}"
      break'
    Block
      While
        Value BooleanLiteral: true
        Block
          Value
            StringWithInterpolations
              Block
                Op +
                  Value StringLiteral: ""
                  Value
                    Parens
                      Block
                        Value IdentifierLiteral: a
          StatementLiteral: break


vendethiel:
Looks pretty good on the surface.

Can you please add a test with the new (fixed) behavior of ``` `this` ``` (not getting rewritten to `_this` in a bound function)?

lydell:
I actually thought about adding a test, but then I chose not to because I wasn’t sure if we’d then lock us down on implementation details. I was thinking something like this:

```coffee
nonce = {}
obj: fn: => `this`
eq nonce, obj.fn.call nonce
```

But let’s say we wanted to compile `=>` to ES2015 `=>` – then the test wouldn’t work anymore. But I can add a test if that’s wanted (but not tonight).

bjmiller:
This may be a little out of scope, but I don't think that we ever want to compile CS `=>` to ES6 arrow functions.  The semantics are very different, and they would be breaking changes for each other.  There may need to be, unfortunately, another symbol. :-/

src/nodes.coffee
```diff
@@ -380,59 +381,63 @@ exports.Block = class Block extends Base
...
-    ' "' + @value + '"'
+exports.ThisLiteral = class ThisLiteral extends Literal
+  constructor: ->
+    super 'this'
 
-class exports.Undefined extends Base
-  isAssignable: NO
-  isComplex: NO
+  compileNode: (o) ->
+      code = if o.scope.method?.bound then o.scope.method.context else @value
```
connec: This function is indented too much, I think.

lydell:
Fixed indentation, added test … and also removed `parseNum()` (turns out `Number()` does the job).

(Regarding the test: What if we decide to compile to ES5 some day? `(foo) =>` → `function(foo) {}.bind(this)`? :) But whatever, we'll think about that then.)

lydell:
@michaelficarra Do you have any opinion on this PR? (Asking since you :+1:ed #4192.)

src/nodes.coffee
```diff
@@ -380,59 +381,63 @@ exports.Block = class Block extends Base
...
+  constructor: -> super 'undefined'
   compileNode: (o) ->
     [@makeCode if o.level >= LEVEL_ACCESS then '(void 0)' else 'void 0']
 
-class exports.Null extends Base
-  isAssignable: NO
-  isComplex: NO
-  compileNode: -> [@makeCode "null"]
+class exports.Null extends Literal
+  constructor: -> super 'null'
```
michaelficarra: The ThisLiteral constructor is on two lines and this is on one line. Be consistent.

src/nodes.coffee
```diff
@@ -380,59 +381,63 @@ exports.Block = class Block extends Base
...
 
-class exports.Bool extends Base
-  isAssignable: NO
-  isComplex: NO
-  compileNode: -> [@makeCode @val]
-  constructor: (@val) ->
+class exports.Bool extends Literal
```
michaelficarra: Why do some of these constructor names not end in "Literal"?

src/nodes.coffee
```diff
@@ -498,15 +503,15 @@ exports.Value = class Value extends Base
...
+  isString       : -> @bareLiteral(StringLiteral)
+  isRegex        : -> @bareLiteral(RegexLiteral)
   isAtomic       : ->
     for node in @properties.concat @base
       return no if node.soak or node instanceof Call
     yes
 
-  isNotCallable  : -> @isSimpleNumber() or @isString() or @isRegex() or
+  isNotCallable  : -> @isNumber() or @isString() or @isRegex() or
                       @isArray() or @isRange() or @isSplice() or @isObject()
```
michaelficarra: Shouldn't Boolean, null, undefined be in this list?

src/nodes.coffee
```diff
@@ -573,7 +578,7 @@ exports.Value = class Value extends Base
         fst = new Value @base, @properties[...i]
         snd = new Value @base, @properties[i..]
         if fst.isComplex()
-          ref = new Literal o.scope.freeVariable 'ref'
+          ref = new IdentifierLiteral o.scope.freeVariable 'ref'
```
michaelficarra: Can freeVariable return an IdentifierLiteral for us?

src/nodes.coffee
```diff
@@ -380,59 +381,63 @@ exports.Block = class Block extends Base
...
+
+exports.NumberLiteral = class NumberLiteral extends Literal
+
+exports.StringLiteral = class StringLiteral extends Literal
+
+exports.RegexLiteral = class RegexLiteral extends Literal
+
+exports.PassthroughLiteral = class PassthroughLiteral extends Literal
+
+exports.IdentifierLiteral = class IdentifierLiteral extends Literal
```
michaelficarra: I'd like to see if we can not use IdentifierLiteral for anything other than identifiers. Using it for properties and then doing this reserved test is gross.

src/nodes.coffee
```diff
@@ -2065,20 +2096,20 @@ exports.For = class For extends While
...
-      if (name or @own) and not IDENTIFIER.test svar
+      if (name or @own) and @source.unwrap() not instanceof IdentifierLiteral
         defPart    += "#{@tab}#{ref = scope.freeVariable 'ref'} = #{svar};\n"
         svar       = ref
       if name and not @pattern
         namePart   = "#{name} = #{svar}[#{kvar}]"
       if not @object
         defPart += "#{@tab}#{step};\n" if step isnt stepVar
-        lvar = scope.freeVariable 'len' unless @step and stepNum and down = (parseNum(stepNum[0]) < 0)
+        lvar = scope.freeVariable 'len' unless @step and stepNum? and down = (stepNum < 0)
```
michaelficarra: Doing an assignment in this condition is really disgusting. I know you didn't design it, but while we're here can you restructure it?

test/functions.coffee
```diff
@@ -86,6 +86,17 @@ test "self-referencing functions", ->
   changeMe()
   eq changeMe, 2
 
+test "#2009: don't touch `` `this` ``", ->
+  nonceA = {}
+  nonceB = {}
+  fn = null
+  (->
+    fn = => [this, `this`]
```
michaelficarra: Can't you just do the equality comparison here and assert the return value?

michaelficarra:
Agreed with @vendethiel about super having its own node. But I think it should be done as special case subclasses of Call and Access. That's the only way super is used after all.

lydell:
Special case of Call _and Access?_ Mind elaborating on that?

michaelficarra:
Sorry, I had ECMAScript super on the mind. You're right, just Call.

src/nodes.coffee
lydell: There are 27 occurrences of `freeVariable` in nodes.coffee. I touched 11 of those in this commit: Those which had `new Literal` in front of them before. The rest use the return variable name as a string in different string concatenations.

It would also cause nodes.coffee and scope.litcoffee to become circular dependencies. While that would most likely still work, I think it's nice to avoid.

So, in my opinion, freeVariable would _ideally_ return an IdentifierLiteral, but right now it feels like too much work. Let's revisit that some other time.

michaelficarra: Sounds good.

src/nodes.coffee
lydell: No tests break if you change the next line to `isAssignable: YES`. Not sure what to do.

lydell:
I've fixed most things now. Would be happy for some feedback regarding IdentifierLiteral and SuperCall.

test/functions.coffee
```diff
@@ -86,6 +86,15 @@ test "self-referencing functions", ->
   changeMe()
   eq changeMe, 2
 
+test "#2009: don't touch `` `this` ``", ->
+  nonceA = {}
+  nonceB = {}
+  fn = null
+  (->
+    fn = => [this is nonceA, `this` is nonceB]
```
michaelficarra:
```coffee
fn = => this is nonceA and `this` is nonceB
```

And then you can just `ok` it below.

src/nodes.coffee
```diff
@@ -745,6 +756,23 @@ exports.Call = class Call extends Base
       answer = answer.concat fun
     answer = answer.concat @makeCode(".apply(#{ref}, "), splatArgs, @makeCode(")")
 
+#### Super
+
+exports.SuperCall = class SuperCall extends Call
+  constructor: (args) ->
+    super null, args ? [new Splat new IdentifierLiteral 'arguments']
+    @isSuper = true
```
michaelficarra: Why is this necessary? Shouldn't it be observable via `instanceof`?

src/nodes.coffee
```diff
@@ -600,12 +612,11 @@ exports.Comment = class Comment extends Base
 # Node for a function invocation. Takes care of converting `super()` calls into
 # calls against the prototype's function of the same name.
 exports.Call = class Call extends Base
```
michaelficarra: All the super-related methods in Call should be moved to SuperCall.

src/grammar.coffee
```diff
@@ -135,36 +135,35 @@ grammar =
...
   Identifier: [
-    o 'IDENTIFIER',                             -> new Literal $1
+    o 'IDENTIFIER',                             -> new IdentifierLiteral $1
   ]
 
   # Alphanumerics are separated from the other **Literal** matchers because
   # they can also serve as keys in object literals.
   AlphaNumeric: [
-    o 'NUMBER',                                 -> new Literal $1
+    o 'NUMBER',                                 -> new NumberLiteral $1
```
michaelficarra: We might want to separate InfinityLiteral from NumberLiteral. Whenever `parseFloat` returns an `Infinity` value, we can render it simply as `2e308`.

lydell:
```
Block
  Value IdentifierLiteral: Infinity
```

Hehe.

CoffeeScript actually allows `Infinity = 0`. (That's not valid in strict mode.) Not sure if it's intentional.

Don't really know what to do now.

vendethiel: Well, we don't really do "strict"... #2337 has been open for a while.

lydell: But we have test/strict.coffee

michaelficarra: @lydell I'm not talking about a reference to the `Infinity` global, I'm talking about infinity values (such as the initial value of the `Infinity` global or the value created by the literals `2e308` or 500 `9`s in a row).

lydell: Is this what you mean?

```coffee
exports.NumberLiteral = class NumberLiteral extends Literal
  constructor: (@value) ->
    return new InfinityLiteral if parseFloat(@value) is Infinity

exports.InfinityLiteral = class InfinityLiteral extends Literal
  constructor: ->
    super '2e308'
```

src/nodes.coffee
vendethiel: more tests?

src/grammar.coffee
michaelficarra: The latter please. Sorry for the delay here.

lydell: Done. (See the next line.)

test/numbers.coffee
```diff
@@ -74,3 +76,10 @@ test "#2224: hex literals with 0b or B or E", ->
   eq 176, 0x0b0
   eq 177, 0x0B1
   eq 225, 0xE1
+
+test "Infinity", ->
+  eq Infinity, CoffeeScript.eval "0b#{Array(1024 + 1).join('1')}"
+  eq Infinity, CoffeeScript.eval "0o#{Array(342 + 1).join('7')}"
```
lydell: The above two lines actually failed before adding `InfinityLiteral`. Invalid JS was generated: `0xInfinity`.

lydell:
InfinityLiteral is done. I’ll now continue with improving IdentifierLiteral.

test/numbers.coffee
michaelficarra: Awesome :smile:.

michaelficarra:
@lydell Maybe submit it as a separate PR? This should be good to merge right now, right?

lydell:
Ok, I’ll do it in a separate PR. Yes, I consider this PR good to merge.

vendethiel:
:+1:

michaelficarra:
Thanks. Great work, @lydell.

lydell:
The follow-up PR became three: #4219, #4220 and #4221.

PR #4204, opened 2016-02-10
Fix getSourceMap with extention ".coffee.md"

wdavidw:
The call `path?.extname(filename) in exports.FILE_EXTENSIONS` fails for extension ".coffee.md" because extname return only ".md". This produce  incorrect line numbers The current approad fixes the issue by testing if the end of the filename match at least one extension.


wdavidw:
This feels good, thanks for your hard work, long life to coffee.

PR #4219, opened 2016-03-05
Treat Infinity and NaN as reserved words

### LINKED ISSUE #4218: treat Infinity and NaN as reserved words
In the same way we compile `undefined` to `void 0`, I think we should compile the identifiers `Infinity` and `NaN` to `2e308` and `0/0` respectively. If someone wants to access the values of these global variables,  they can do so explicitly via `global.undefined`, `global.Infinity`, and `global.NaN` instead of doing it through scope resolution.

lydell:
Fixes #4218.


src/grammar.coffee
```diff
@@ -142,7 +142,7 @@ grammar =
   # they can also serve as keys in object literals.
   AlphaNumeric: [
     o 'NUMBER',                                 -> new NumberLiteral $1
-    o 'INFINITY',                               -> new InfinityLiteral $1
+    o 'INFINITE_NUMBER',                        -> new InfiniteNumberLiteral $1
```
michaelficarra: I liked the old names. Let's not distinguish between `Infinity` and `2e308`. They should each be an `InfinityLiteral`.

src/nodes.coffee
```diff
@@ -446,6 +446,20 @@ exports.NullLiteral = class NullLiteral extends Literal
 
 exports.BooleanLiteral = class BooleanLiteral extends Literal
 
+exports.InfinityLiteral = class InfinityLiteral extends Literal
```
michaelficarra: Same as above. This doesn't need to be a separate thing.

src/nodes.coffee
```diff
@@ -446,6 +446,20 @@ exports.NullLiteral = class NullLiteral extends Literal
...
+
+  compileNode: ->
+    [@makeCode '2e308']
+
+exports.NaNLiteral = class NaNLiteral extends Literal
+  constructor: ->
+    super 'NaN'
+
+  compileNode: ->
+    [@makeCode '0/0']
```
michaelficarra: Since this compilation includes an infix operator, we should either surround it with parens or make sure `NaNLiteral`s have the same precedence as `/`.

src/nodes.coffee
```diff
@@ -404,6 +404,13 @@ exports.InfinityLiteral = class InfinityLiteral extends NumberLiteral
   compileNode: ->
     [@makeCode '2e308']
 
+exports.NaNLiteral = class NaNLiteral extends NumberLiteral
+  constructor: ->
+    super 'NaN'
+
+  compileNode: ->
+    [@makeCode '(0/0)']
```
lydell: I went with the simple approach of _always_ wrapping in parentheses here. Do we want to only do it if strictly needed? (If so, how?)

test/numbers.coffee
```diff
@@ -83,3 +83,6 @@ test "Infinity", ->
   eq Infinity, CoffeeScript.eval "0x#{Array(256 + 1).join('f')}"
   eq Infinity, CoffeeScript.eval Array(500 + 1).join('9')
   eq Infinity, 2e308
+
+test "NaN", ->
+  ok isNaN 1/NaN
```
lydell: This test would have failed in the first version of this PR, where `1/NaN` became `Infinity`.

src/nodes.coffee
michaelficarra: If you wanted to do it the better way, you could either create an `Op` node that represents the division and call compile on that, or you could just add the parens yourself by checking `o.level`. But this works fine for a first implementation if you just feel like leaving it. It's up to you.

src/nodes.coffee
```diff
@@ -404,6 +404,14 @@ exports.InfinityLiteral = class InfinityLiteral extends NumberLiteral
   compileNode: ->
     [@makeCode '2e308']
 
+exports.NaNLiteral = class NaNLiteral extends NumberLiteral
+  constructor: ->
+    super 'NaN'
+
+  compileNode: (o) ->
+    code = [@makeCode '0/0']
+    if o.level >= LEVEL_OP then @wrapInBraces code else code
```
lydell: I settled on using `@makeCode` instead of `new Op` since that allows `NaN` to be rendered as `0/0` instead of `0 / 0`.

lydell:
@michaelficarra Any more comments? :)

michaelficarra:
Yeah, I had just wanted to clone it and try some things out myself. LGTM. Merging.

PR #4220, opened 2016-03-05
Split out properties from identifiers

lydell:
- Split out a PROPERTY token from the IDENTIFIER token.
- Split out Property from the Identifier in the grammar.
- Split out PropertyLiteral from IdentifierLiteral.


src/nodes.coffee
```diff
@@ -413,6 +413,9 @@ exports.PassthroughLiteral = class PassthroughLiteral extends Literal
 exports.IdentifierLiteral = class IdentifierLiteral extends Literal
   isAssignable: -> @value not in RESERVED
 
+exports.PropertyLiteral = class PropertyLiteral extends Literal
```
michaelficarra: `PropertyLiteral` is a pretty strange name. I think `PropertyName` is better.

src/lexer.coffee
```diff
@@ -114,12 +114,15 @@ exports.Lexer = class Lexer
...
+    tag =
+      if colon or prev? and
+         (prev[0] in ['.', '?.', '::', '?::'] or
+         not prev.spaced and prev[0] is '@')
+        'PROPERTY'
+      else
+        'IDENTIFIER'
 
-    if not forcedIdentifier and (id in JS_KEYWORDS or id in COFFEE_KEYWORDS)
+    if tag == 'IDENTIFIER' and (id in JS_KEYWORDS or id in COFFEE_KEYWORDS)
```
michaelficarra: Use `is`, not `==`.

lydell: Thanks. Seems like I’ve gotten too used to the style guide used in another project :) I went ahead and fixed all other `==`s and `!=`s in the diff as well.

michaelficarra:
LGTM. Please rebase and merge.

PR #4221, opened 2016-03-05
Unify, simplify and fixup assignment errors

lydell:
- Show the same type of error message for compound assignment as for `=`
  assignment when the LHS is invalid.
- Show the same type of error message when trying to assign to a CoffeeScript
  keyword as when trying to assign to a JavaScript keyword.
- Now longer treat `&& =` as `&&=`. The same goes for `and=`, `||=` and `or=`.
- Unify the error message to: `<optional type> '<value>' can't be assigned`.


src/lexer.coffee
```diff
@@ -411,14 +411,20 @@ exports.Lexer = class Lexer
...
+      skipToken = false
+      if value is '=' and prev[1] in ['||', '&&'] and not prev.spaced
         prev[0] = 'COMPOUND_ASSIGN'
         prev[1] += '='
-        return value.length
+        prev = @tokens[@tokens.length - 2]
+        skipToken = true
+      if prev and prev.variable
+        origin = prev.origin ? prev
+        if message = isUnassignable prev[1], origin[1]
```
michaelficarra: I find this pattern of doing an assignment in a condition really unreadable and error-prone.

src/lexer.coffee
```diff
@@ -782,15 +803,12 @@ RESERVED = [
   'protected', 'public', 'static'
 ]
 
-STRICT_PROSCRIBED = ['arguments', 'eval', 'yield*']
+STRICT_PROSCRIBED = ['arguments', 'eval']
```
michaelficarra: This should include `yield`, shouldn't it?

lydell: But `yield` is a CoffeeScript keyword?

michaelficarra: Oh sure, if it's in that list then it shouldn't need to be in this list.

lydell: (It’s in `JS_KEYWORDS` (“keywords that CoffeeScript shares in common with JavaScript”).)

src/lexer.coffee
lydell: Fixed all occurrences in the diff.

Issue #4156, opened 2015-12-02
Getter/setter syntax after IE8


woubuc:
I think everyone here knows how @jashkenas feels about getters and setters (for those who don't: he hates them). The official reason for their absence in Cofeescript has always been to maintain browser support, as per #3698 and a bunch of others, but that is obviously nothing more than an excuse.

But here's the thing: whether or not it is an excuse or an actual compatibility consideration, on january 12th 2016 the support for IE8 will [officially end](https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=Windows%20Internet%20Explorer%208&Filter=FilterNO), meaning the only remaining browser to not support getters and setters will be officially dead. After that, [all browsers and Javascript platforms will support it](http://kangax.github.io/compat-table/es5/).

So what I want to know is: now that there is literally no more reason not to support this feature (other than an apparent personal grudge of some kind that you hold against the words get and set), **will a proper getter/setter syntax finally be implemented?** If not for you, then for all those devs who have been asking for this for a long time.

vendethiel:
> The official reason for their absence in Cofeescript has always been to maintain browser support, as per #3698 and a bunch of others, but that is obviously nothing more than an excuse.

Not really. You can find several instances of @jashkenas and @michaelficarra saying they're a bad part of JavaScript.

[here in #3698](https://github.com/jashkenas/coffeescript/issues/3698#issuecomment-60769197), [here in #2902](https://github.com/jashkenas/coffeescript/pull/2902#issuecomment-15881233), #64, #322, #451, #949

michaelficarra:
To summarise the sentiment, we don't think it's a good thing that getters/setters enable you to add effects to an expression that does not *look* like it should have effects. If you want an expression to have some effect, use a function to make it obvious to consumers.

woubuc:
@michaelficarra For an existing in-house application I'm currently working on, there is a property in an object that we need to adjust the value of slightly, to be compatible with a new add-on we're implementing. It is already used extensively throughout the code so implementing a getter and setter here in the Coffeescript code instead of having to implement a set of hacks to achieve the same result would be the best and easiest option.

For 95% of the cases, I agree that get/set shouldn't be used, but it would just be very useful to have them available in edge cases like this.

lydell:
@woubuc Considering what you said in the above comment, I’d say use something like the following:

```coffee
Object.defineProperty(obj, 'key', {
  get: -> ...
  enumerable: true
})
```

[...]

Issue #3698, opened 2014-10-28
Document official policy on future ES features


DavidSouther:
I've seen snippets of a conversation in various threads (#3596, #1363, 3240). 2011 was very ES3, 2012 had a lot of ES5, 2013 started adding ES6. End of 2014, Google and Microsoft both seem to be pushing forward with TypeScript, and people in Google are pushing for an even further extension with AtScript. IE8 support is only another year and a half. It seems safe to say that the JS we have today is not the JS enterprise and browser vendors want.

Is there an official policy documented on when ES features will be willing to add to CS? (Beyond the threads I've pointed out - which come down to "@jashkenas says no, until enough devs complain, then he grudgingly says yes" ;) )

umhan35:
> That's because this is the jashkenas/coffeescript fork. There's also the satyr/coco fork, the gkz/LiveScript fork, and the michaelficarra/CoffeeScriptRedux fork, just to name a few.

This doesn't make sense at all.

> There are no "official policies" in CoffeeScript ;) This is an open source project, not a government.

This doesn't make sense at all too. Most issues are closed by one person. This is a despotic land.

The JS getter and setter have been here so long. People writing JS or CS not only work with browsers but also Node.js.

J4G-:
+1 for getter and setter support

zimt28:
I'd also like support for getters/setters and named functions.
Generally speaking I don't see the point of not adding new features - if they were just included in CoffeeScript people could decide on their own if they wanted to to use those or not, just like they can with plain JavaScript.

PR #4234, opened 2016-03-24
Accessor descriptor operator (~>)

dadleyy:
```coffeescript
class UserSearch

  consturctor: () ->
    @sort = {rel: "name", order: 1}

  orderby: ~> "#{if @sort.order then "" else "-"}#{@sort.rel}"

  search: (query) ->
    {promise, resolve, reject} = Promse.defer()

    request =
      where: name: like: query
      orderby: @orderby

    failed = (err) ->
      reject err

    finished = (user) ->
      resolve user

    User.query request
      .then finished
      .catch failed

    promise
```


michaelficarra:
@dadleyy Did you read any of the issues about accessors before doing this work? We don't want syntax for defining them to be part of the language. If you want an accessor, you can use `Object.defineProperty`. It shouldn't be convenient because you shouldn't be using them frequently.

dadleyy:
I did read the issues - #4156 and #3698 in particular. It seems like there are multiple people in the community asking for this feature and I figured I'd put together a proof-of-concept solution for it. 

I agree that developers shouldn't be using them frequently, but that doesn't mean it should not be convenient for them to use them, especially when using them makes the code more readable like my example above.

In the coffeescript lexer there is the [tag()](https://github.com/jashkenas/coffeescript/blob/master/src/lexer.coffee#L682-L684) method for getting the last tag to be pushed into the token array.  Having read through the source code myself, I think it would have been much more readable to see:

```coffeescript
if id is 'own' and @top[TAG] is 'FOR'
  @token 'OWN', id
  return id.length
```

where `@top` was a property overloaded (in js) like:


```
defineProperty(Lexer.prototype, "top", {
});
```

dadleyy:
My argument for getters is that functions not having any parameters would be better made as a getter. (similar to the [mixin-or-placeholder](http://www.sitepoint.com/sass-mixin-placeholder/) argument for sass).

Whats the point of having:

```coffeescript
class Lexer

  tag: ->
```

which means you need to `@tag()`, when you could just `@tag`. You have created a function that is purely for accessing things only because you were unable to define a getter.

Issue #4252, opened 2016-04-14
REPL throws "Error: EBADF: bad file descriptor, write" on exit


benbuckman:
node v4.4.1
coffee v1.10.0

Running a simple coffee REPL with `coffee`, and exiting with `.exit`, it outputs this error:

```
fs.js:60
      throw err;  // Forgot a callback but don't know where? Use NODE_DEBUG=fs
      ^

Error: EBADF: bad file descriptor, write
  at Error (native)
```

Running with `NODE_DEBUG=fs` like it suggests, it outputs:

```
fs.js:53
        throw backtrace;
        ^

Error: EBADF: bad file descriptor, write
  at rethrow (fs.js:48:21)
  at maybeCallback (fs.js:66:42)
  at Object.fs.write (fs.js:646:14)
  at REPLServer.<anonymous> (./node_modules/coffee-script/lib/coffee-script/repl.js:142:12)
  at emitOne (events.js:82:20)
  at REPLServer.emit (events.js:169:7)
  at REPLServer.Interface._onLine (readline.js:211:10)
  at REPLServer.Interface._line (readline.js:550:8)
  at REPLServer.Interface._ttyWrite (readline.js:827:14)
  at ReadStream.onkeypress (readline.js:106:10)
  at emitTwo (events.js:92:20)
  at ReadStream.emit (events.js:172:7)
  at emitKeys (readline.js:1251:14)
  at [object Generator].next (native)
  at ReadStream.onData (readline.js:919:36)
  at emitOne (events.js:77:13)
  at ReadStream.emit (events.js:169:7)
  at readableAddChunk (_stream_readable.js:153:18)
  at ReadStream.Readable.push (_stream_readable.js:111:10)
  at TTY.onread (net.js:529:20)
```

benbuckman:
It appears to be because it's trying to `fd.write` after `fd.close()` has been called.
This seems like a simple thing, I'm surprised no one's reported this yet (looking at repo-issue and google searches)?

lydell:
```
$ node -v
v5.10.0
$ coffee -v
CoffeeScript version 1.10.0
$ coffee
coffee> .exit
$ 
```

michaelficarra:
Fixed by #4253.

PR #4253, opened 2016-04-14
Fix 'Error: EBADF: bad file descriptor, write' on `.exit` in REPL

benbuckman:
Fixes issue #4252.

> Running a simple coffee REPL with `coffee`, and exiting with `.exit`, it outputs this error:

> ```
fs.js:60
      throw err;  // Forgot a callback but don't know where? Use NODE_DEBUG=fs
      ^
Error: EBADF: bad file descriptor, write
  at Error (native)
```

> It appears to be because it's trying to fd.write after fd.close() has been called.
> This seems like a simple thing, I'm surprised no one's reported this yet (looking at repo-issue and google searches). Am I missing something?

The issue only manifests with `.exit`, not `ctrl+d`, so this fixes it simply by not trying to write on `.exit`.

Thank you.

### DIFF
--- a/lib/coffee-script/repl.js
+++ b/lib/coffee-script/repl.js
@@ -138,7 +138,7 @@
     } catch (error) {}
     fd = fs.openSync(filename, 'a');
     repl.rli.addListener('line', function(code) {
-      if (code && code.length && code !== '.history' && lastLine !== code) {
+      if (code && code.length && code !== '.history' && code !== '.exit' && lastLine !== code) {
         fs.write(fd, code + "\n");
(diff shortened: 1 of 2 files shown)


michaelficarra:
This works, but it'd probably be better to just check that the file descriptor is still open before writing. Do we want `.exit` to be included in the history or not?

vendethiel:
-1 from me.

vendethiel:
(I don't want the `exit` in the history)

michaelficarra:
Okay, then merging.

PR #4300, opened 2016-08-31
Support import and export of ES2015 modules

GeoffreyBooth:
This pull request adds support for ES2015 modules, by recognizing `import` and `export` statements. The following syntaxes are supported, based on the MDN [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) pages:

```js
import "module-name"
import defaultMember from "module-name"
import * as name from "module-name"
import { } from "module-name"
import { member } from "module-name"
import { member as alias } from "module-name"
import { member1, member2 as alias2, … } from "module-name"
import defaultMember, * as name from "module-name"
import defaultMember, { … } from "module-name"

export default expression
export class name
export { }
export { name }
export { name as exportedName }
export { name as default }
export { name1, name2 as exportedName2, name3 as default, … }

export * from "module-name"
export { … } from "module-name"
```

As a subsitute for ECMAScript’s `export var name = …` and `export function name {}`, CoffeeScript also supports:
```js
export name = …
```

CoffeeScript also supports optional commas within `{ … }`.

This PR converts the supported `import` and `export` statements into ES2015 `import` and `export` statements; it **does not resolve the modules**. So any CoffeeScript with `import` or `export` statements will be output as ES2015, and will need to be transpiled by another tool such as Babel before it can be used in a browser. If this PR is accepted, we will need to add a warning to the documentation explaining this.

This should be fully backward-compatible, as `import` and `export` were previously reserved keywords. No flags are used.

There are extensive tests included, though because no current JavaScript runtime supports `import` or `export`, the tests compare strings of what the compiled CoffeeScript output is against what the expected ES2015 should be. I also conducted two more elaborate tests:

* I forked the [ember-piqu](https://github.com/pauc/piqu-ember) project, which was an Ember CLI app that used ember-cli-coffeescript and [ember-cli-coffees6](https://github.com/alexspeller/ember-cli-coffees6) (which adds “support” for `import`/`export` by wrapping such statements in backticks before passing the result to the CoffeeScript compiler). I removed `ember-cli-coffees6` and replaced the CoffeeScript compiler used in the build chain with this code, and the app built without errors. [Demo here.](https://github.com/GeoffreyBooth/coffeescript-modules-test-piqu)
* I also forked the [CoffeeScript version of Meteor’s Todos example app](https://github.com/meteor/todos/tree/coffeescript), and replaced all of its `require` statements with the `import` and `export` statements from the original ES2015 version of the app on its `master` branch. I then updated the `coffeescript` Meteor package in the app to use this new code, and again the app builds without errors. [Demo here.](https://github.com/GeoffreyBooth/coffeescript-modules-test-meteor-todos)

The discussion history for this work started [here](https://github.com/jashkenas/coffeescript/pull/4160) and continued [here](https://github.com/GeoffreyBooth/coffeescript/pull/2). @lydell provided guidance, and @JimPanic and @rattrayalex contributed essential code.


test/error_messages.coffee
```diff
@@ -989,3 +991,39 @@ test "`&&=` and `||=` with a space in-between", ->
...
+    export ->
+           ^^
+  '''
+
+test "anonymous classes cannot be exported", ->
+  assertErrorFormat '''
+    export class
+      @constructor: ->
+        console.log 'hello, world!'
+  ''', 'SyntaxError: anonymous classes cannot be exported'
```
lydell: I'm not sure if this is "your fault" or if it belongs to this PR, but it's sad that this error is so ugly (no `file:line:col` and no code snippet with `^^^` marker) compared to the other ones here ...

test/modules.coffee
```diff
@@ -0,0 +1,492 @@
...
+# export default ->
+# export { name1 as default, … }
+
+# export * from …
+# export { name1, name2, …, nameN } from …
+# export { import1 as name1, import2 as name2, …, nameN } from …
+
+# Syntaxes from the MDN documentation that are *not* supported, because of ambiguous grammar and similarity to unsupported `var foo, bar = 'baz'`:
+# export name1, name2, …, nameN
+# export name1 = …, name2 = …, …, nameN
```
lydell: These three lines can be removed. Those syntaxes don't exist. (You can also remove that from your PR description. It is confusing.)

test/modules.coffee
```diff
@@ -0,0 +1,492 @@
...
+  CoffeeScript.compile str, bare: yes
+  .replace /^\s+|\s+$/g, '' # Trim leading/trailing whitespace
+
+
+# Import statements
+
+test "backticked import statement", ->
+  input = "`import { foo, bar as baz } from 'lib'`"
+  output = "import { foo, bar as baz } from 'lib';"
+  eq toJS(input), output
```
lydell: Why does this test exist?

test/modules.coffee
```diff
@@ -0,0 +1,492 @@
...
+    bar.barMethod()"""
+  output = """
+    import {
+      foo as bar
+    } from 'lib';
+
+    bar.barMethod();"""
+  eq toJS(input), output
+
+test "import a multiple members of a module, adding the members to the current scope", ->
```
lydell: `a multiple` → `multiple`

test/modules.coffee
```diff
@@ -0,0 +1,492 @@
...
+      foo,
+      bar
+    } from 'lib';
+
+    foo.fooMethod();
+
+    bar.barMethod();"""
+  eq toJS(input), output
+
+test "import a multiple members of a module where some are aliased, adding the members or aliases to the current scope", ->
```
lydell: `a multiple` → `multiple`

test/modules.coffee
```diff
@@ -0,0 +1,492 @@
...
+    var foo;
+
+    foo = function(bar) {
+      return console.log(bar);
+    };
+
+    export default foo;"""
+  eq toJS(input), output
+
+# Uncomment this test once ES2015+ `class` support is added
```
lydell: Please forgive me if we might have discussed this before. Why can't we support `export default class foo`?

lydell:
While quickly glancing through all tests, I couldn't find any that tests that optional commas work inside `import {...}` and `export {...}`

test/modules.coffee
GeoffreyBooth: We do support it (see line 353). This is a test that checks that `export default class foo` _outputs_ `export default class foo`, e.g. it outputs the ES2015 `class` keyword. Currently `class` gets output as `function`, which is what we’re testing for below; but eventually we want `class` (or some other keyword like `esclass`) to output as `class`, and when that day comes we should uncomment this test.

test/modules.coffee
GeoffreyBooth: Legacy code that uses ES2015 modules probably has backticked `import` and `export` statements. I wanted to make sure we aren’t breaking such code. Perhaps it’s trivial to test specifically for this since we test backticks elsewhere. Do you think we should remove this test?

test/modules.coffee
GeoffreyBooth: I removed it from the description, and I don’t suggest we include it in the documentation, but I feel like this should still be mentioned somewhere. A comment seems like a pretty minor place to note it.

src/nodes.coffee
```diff
@@ -1223,18 +1223,128 @@ exports.Class = class Class extends Base
...
+
+  children: ['clause', 'moduleName']
+
+  isStatement: YES
+  jumps:       THIS
+  makeReturn:  THIS
+
+  compileNode: (o) ->
+    if o.indent.length isnt 0
+      @error "#{@type} statements must be at top-level scope"
```
GeoffreyBooth: Meteor wants code blocks like this:

```js
if (Meteor.isServer) {
  import { foo } from 'lib';
}
```

which of course violates the spec, but their build tool somehow parses this. I guess we should support such a syntax, and let the user violate the “`import` or `export` statements must be at top level” requirement if they choose? If we allow this, how can I rewrite the other places where I look at `o.indent` to use some other way of determining if we’re in an `export` statement, rather than an expression within an `export` statement?

src/grammar.coffee
```diff
@@ -349,6 +351,48 @@ grammar =
...
   ]
 
+  Import: [
+    o 'IMPORT String',                          -> new Import null, $2
+    o 'IMPORT ImportClause FROM String',        -> new Import $2, $4
+  ]
+
+  ImportClause: [
+    o '{ }',                                    -> new ModuleList [], yes
+    o 'ModuleList OptComma',                    -> new ModuleList $1, no
```
lydell: This makes `import a, from 'b'` valid (which is invalid JS). (The output is still correct, though.) Do we really want to allow that?

lydell:
Here's a bug:

```
$ ./bin/coffee -pe 'export class a' 
var a;

export a = (function() {
  function a() {}

  return a;

})();
```

The output is invalid JavaScript. I expect this output:

```js
export var a = (function() {
  function a() {}

  return a;

})();
```

src/nodes.coffee
lydell: I'm strongly against this. Meteor is doing it wrong.

> how can I rewrite the other places where I look at `o.indent` to use some other way of determining if we’re in an `export` statement, rather than an expression within an `export` statement?

Sorry, I don't follow along here …

lydell:
A test with comments might be nice as well. Something like this:

```
import {
    a # foo
    b # bar
  } from "c"
```

src/nodes.coffee
```diff
@@ -1223,18 +1223,128 @@ exports.Class = class Class extends Base
...
 
+#### Import and Export
+
+exports.Module = class Module extends Base
+  constructor: (@type, @clause, @moduleName, @default = no) ->
+    if @type isnt 'import' and @type isnt 'export'
+      @error 'module type must be import or export'
+
+    if @moduleName? and @moduleName instanceof StringWithInterpolations
+      @error 'the name of the module to be imported from must be an uninterpolated string'
```
lydell: I noticed that this error becomes much nicer if you use `@moduleName.error` instead of `@error`:

```
$ ./bin/coffee -pe 'import a from "a#{}"'
[stdin]:1:15: error: the name of the module to be imported from must be an uninterpolated string
import a from "a#{}"
              ^^^^^^
```

src/nodes.coffee
```diff
@@ -1223,18 +1223,128 @@ exports.Class = class Class extends Base
...
+
+    if @type is 'export' and @default is no and @clause instanceof Class and not @clause.variable?
+      @error 'anonymous classes cannot be exported'
```
lydell: Same thing here: `@clause.error` gives a better message.

src/nodes.coffee
```diff
@@ -1223,18 +1223,128 @@ exports.Class = class Class extends Base
     klass = new Assign @variable, klass if @variable
     klass.compileToFragments o
 
+#### Import and Export
+
+exports.Module = class Module extends Base
+  constructor: (@type, @clause, @moduleName, @default = no) ->
+    if @type isnt 'import' and @type isnt 'export'
```
lydell: `if @type not in ['import', 'export']`

test/error_messages.coffee
```diff
@@ -989,3 +991,39 @@ test "`&&=` and `||=` with a space in-between", ->
...
+  ''', '''
+    [stdin]:1:8: error: unexpected ->
+    export ->
+           ^^
+  '''
+
+test "anonymous classes cannot be exported", ->
+  assertErrorFormat '''
+    export class
+      @constructor: ->
```
lydell: Do you really mean to include that `@` here? It sure is valid, but it's some very odd code :)

test/modules.coffee
lydell: That syntax still doesn't exist in JavaScript. Try it out in the babel repl! It's invalid JavaScript.

src/lexer.coffee
```diff
@@ -201,6 +215,12 @@ exports.Lexer = class Lexer
   stringToken: ->
     [quote] = STRING_START.exec(@chunk) || []
     return 0 unless quote
+
+    # If the preceding token is `from` and this is an import or export statement,
+    # properly tag the `from`.
+    if @tokens.length and @tokens[@tokens.length - 1][1] is 'from' and (@seenImport or @seenExport)
```
lydell: There is `@peek()` and `@value()` to look at the last token.

src/lexer.coffee
```diff
@@ -113,7 +115,15 @@ exports.Lexer = class Lexer
     if id is 'from' and @tag() is 'YIELD'
       @token 'FROM', id
       return id.length
+    if id is 'as' and (@seenImport or @seenExport) and @tokens[@tokens.length - 1][0] in ['IDENTIFIER', 'IMPORT_ALL']
```
lydell: `@tag()`

test/modules.coffee
GeoffreyBooth: My comment is a CoffeeScript interpretation of the JavaScript syntax. The actual JavaScript syntax per MDN is:
```js
export let name1, name2, …, nameN; // also var
export let name1 = …, name2 = …, …, nameN; // also var, const
```
which _is_ valid JavaScript per the Babel REPL. In my version I remove the `let`/`var`, because CoffeeScript. This could be clearer though.

src/nodes.coffee
GeoffreyBooth: Good point re backticks, maybe that’s the best solution. My question was just about line 1246 above, using `o.indent.length` as a check to determine if we’re in an `export` statement itself as opposed to a block (expression) within the `export` statement, e.g. the body of the `foo` function in `export default foo ->`. Using `o.indent` works, assuming we never cave to Meteor’s desired syntax and `export` statements must always begin at the start of a line, but it feels brittle to me and a bit indirect. I feel like there should be some way of explicitly getting the “parent” of a block and seeing if that’s an `export`, rather than counting indentation characters.

src/grammar.coffee
GeoffreyBooth: I think we do. We allow `a = { b, }`, as ugly as that looks.

GeoffreyBooth:
`export class a` is fixed in f32f78fa0e0c07bcfeabde72229d27669d68d486.

test/modules.coffee
lydell: Perhaps we should convert the test to checking that

```coffee
if Metor.stuff
  `import { foo, bar as baz } from 'lib'`
```

works.

src/grammar.coffee
lydell: `a = { b, }` is a completely different case. I don't think `import a, from 'b'` should be allowed. CoffeeScript allows you to _remove_ commas here and there, not _add_ them.

test/modules.coffee
```diff
@@ -183,6 +183,28 @@ test "multiline complex import", ->
...
+      foo,
+      bar
+    } from 'lib';"""
+  eq toJS(input), output
+
+test "multiline import with optional commas", ->
+  input = """
+    import {
+      foo,
+      bar,
```
lydell: Nice! Extra trailing comma is now accounted for. We also need to test _removed_ commas:

```coffee
import {
  foo
  bar
} from 'lib'
```

src/grammar.coffee
GeoffreyBooth: Revised in 52640fe6a95eb6c247843b7a721e83f1495c22e1

test/modules.coffee
GeoffreyBooth: Added in 2d60c310aa0a15db69a09b3d99adf4308fbb58cd

GeoffreyBooth:
@lydell, all of your notes (as of now) have been addressed as of 2d60c310aa0a15db69a09b3d99adf4308fbb58cd.

GeoffreyBooth:
Okay: 21a5a05c8f6012f7c7544bc538b8127d6f660c65

lydell:
Nice! :+1:

Here are a few more things I found:

Needs to be disallowed:

```
import a, b from 'c'
import a, b, c from 'c'

export *
```

Needs to be allowed:

```
export { }
```

test/modules.coffee
```diff
@@ -0,0 +1,586 @@
...
+# export { variable1 as name1, variable2 as name2, …, nameN }
+# export name1 = …
+# export class …
+# export default expression
+# export default ->
+# export { name1 as default, … }
+
+# export * from …
+# export { name1, name2, …, nameN } from …
+# export { import1 as name1, import2 as name2, …, nameN } from …
```
lydell: Here is how I would phrase that section:

```
# This is the CoffeeScript import and export syntax, closely modeled after the ES2015 syntax
# https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import
# https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

# import "module-name"
# import defaultMember from "module-name"
# import * as name from "module-name"
# import { } from "module-name"
# import { member } from "module-name"
# import { member as alias } from "module-name"
# import { member1 , member2 as alias2 , … } from "module-name"
# import defaultMember, * as name from "module-name"
# import defaultMember, { … } from "module-name"

# export default expression
# export { }
# export { name }
# export { name as exportedName }
# export { name as default }
# export { name1 , name2 as exportedName2 , name3 as default , … }
# export class name
#
# export * from "module-name"
# export { … } from "module-name"
#
# As a subsitute for `export var name = …` and `export function name {}`,
# CoffeeScript also supports:
# export name = …

# CoffeeScript also supports optional commas within `{ … }`.
```

(More coverage, less repetition, more consistency, more explicit JavaScript/CoffeeScript differences.)

GeoffreyBooth:
@lydell the scenarios from your [last comment](https://github.com/jashkenas/coffeescript/pull/4300#issuecomment-244099562) are now all handled. You couldn’t have thought of all these before I submitted the PR? 😅

lydell:
@GeoffreyBooth Nice!

I’ll see if I can come up with a grammar that only accepts valid import/export syntax so you don’t have to do all that post-validation.

> You couldn’t have thought of all these before I submitted the PR?

It’s different to review a PR. Then you _have_ to really look at everything and look for every corner case etc. rather than just casually looking at code if it looks on the right track or not. It’s also easier to work with a PR: github tells me if there are code changes since I last viewed the PR and let’s me easily comment on any change.

GeoffreyBooth:
I thought about trying to use the grammar to exclude bad forms, but wouldn’t such an approach preclude more informative error messages?

lydell:
> I thought about trying to use the grammar to exclude bad forms, but wouldn’t such an approach preclude more informative error messages?

Babel is used by lots of people. Babel seems to generally say “unexpected token” when doing the same errors. I don’t hear lots of people complaining about this in Babel. Now, I’m not saying that we can’t be better than Babel. I just don’t think that these errors are the ones where we need to invest extra thought to produce special-case errors. Especially when it makes grammar.coffee more confusing to upcoming contributors: They’ll wonder why the grammar described there does not correspond to what is valid CoffeeScript. I know that other parts of the grammar already violate this, but let’s not make it worse, for the sake of maintainability? :)

What I’d to is following the grammar described in the spec:

- Import: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-imports
- Export: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-exports

As you can see, the syntax used there is very similar to the Jison DSL used in grammar.coffee.

Then, I’d make the AST for import/export (as defined by classes in nodes.coffee) as close to Babylon’s AST as possible. That’ll be very, very useful when working on outputting ES2015+ code in the future. Let’s make it “right” from the beginning, so that we don’t break coffeelint unnecessarily if we change the AST later. I don’t think that’s very difficult at all: Basically renaming a few of the node classes, and perhaps adding a few more of those “dummy” classes we already have, like this one:

```coffee
exports.ExportDefault = class ExportDefault extends Module
  constructor: (@clause) ->
    super 'export', @clause, null, yes
```

Use https://astexplorer.net/ to explore Babel’s AST – that’s a fantastic tool that makes it super simple!

This is what I had in mind doing this weekend, but I don’t think I’ll get to it. Maybe next week. But @GeoffreyBooth you seem to be a tad bit more motivated than me and have done a terrific work so far – perhaps you can make it? I’m confident you’ll manage in little time!

The above should be the last touches needed for this PR. The hard parts are already done.

GeoffreyBooth:
@lydell done, though that was a fairly significant refactor. I had to diverge `Import` and `Export` in the grammar, introducing some unfortunate duplication, because things like `export { member as default }` are allowed while `import { member as default }` are not. So now the grammar is more verbose, but with no overlap between `Import` and `Export` it should be easier to follow.

The nodes are as DRY as possible, with some `extends` to map the names from the grammar. I split up the former `Module::compileNode` method into import and export versions because that function was really becoming unwieldy trying to do its job for both import and export; there are enough differences between the two that it’s worth just having two separate functions.

I also renamed many of the classes to follow Babel’s AST. I searched through its source but couldn’t find a direct equivalent to our `grammar.coffee`, so that’s inferred based on astexplorer.net (https://astexplorer.net/#/x5DSU1n2hp/1).

lydell:
@GeoffreyBooth Wow you are amazing! :+1: Awesome work!

I had a quick glance at it just now, and it all seems really, really good. I'll have a closer look tonight.

Unfortunately, I forgot to mention a few little things yesterday:

Compare this:

```coffee
o 'EXPORT Identifier = Expression'
```

```coffee
Assign: [
  o 'Assignable = Expression',                -> new Assign $1, $3
  o 'Assignable = TERMINATOR Expression',     -> new Assign $1, $4
  o 'Assignable = INDENT Expression OUTDENT', -> new Assign $1, $4
]
```

The export one needs to allow the `TERMINATOR Expression` and `INDENT Expression OUTDENT` versions as well, in order to keep symmetry:

```coffee
foo1 =
42

export foo2 =
42

foo3 =
  42

export foo4 =
  42
```

Then, the spec mentions that it is an early syntax error to have duplicates in `import …`. We already enforce that there are no function parameter duplicates, and Babel won't compile `import`s with duplicates, so it is something that needs doing (should be easy though).

```coffee
import {a, a} from 'a' # Error
import {a, a as b} from 'a' # NOT an error
import {a, b as a} from 'a' # Error

import a, {a} from 'a' # Error
import a, {a as b} from 'a' # NOT an error
import a, {b as a} from 'a' # Error
```

Note that for instance `export {foo, foo}` is still valid.

(Finally, here's a link to [Babel's/Babylon's AST spec](https://github.com/babel/babylon/blob/master/ast/spec.md). You seem to be doing fine with astexplorer, though – as do I :) )

GeoffreyBooth:
@lydell okay, multiline export assignments are handled, and importing the same member multiple times now throws an error.

GeoffreyBooth:
When this PR is merged, afterward can we please consolidate some of the knowledge shared about how `lexer.coffee`/`grammar.coffee`/`nodes.coffee` work? I learned a lot from these threads about how the parser and compiler work, and I feel like this info should go into [a document](https://github.com/joliss/coffee-script/wiki/%5BHowto%5D-Hacking-on-the-CoffeeScript-Compiler) for others to find more easily. Especially if we go the route of updating the original compiler to output ESNext, and several more people start editing the original compiler, this will be good info to consolidate.

jashkenas:
Guys, this is looking awesome. @lydell should be the one to do the final approval/merge — but it has my blessing, for whatever that's worth.

test/error_messages.coffee
```diff
@@ -1074,3 +1073,40 @@ test "imports and exports must be top-level", ->
...
+    import foo, { foo } from 'lib'
+                  ^^^
+  '''
+  assertErrorFormat '''
+    import foo, { bar as foo } from 'lib'
+  ''', '''
+    [stdin]:1:15: error: already imported
+    import foo, { bar as foo } from 'lib'
+                  ^^^^^^^^^^
+  '''
```
lydell: Splendid! I just thought about one/two more case that could need a test:

```coffee
import a from 'a'
import a from 'b'
```

```coffee
import * as a from 'a'
import {a} from 'b'
```

GeoffreyBooth:
@lydell done, we now check scope to see if a symbol is already used before allowing it to be used for an import. This catches your two new test cases.

What about this though:
```coffee
import a from 'lib'
a = 2
```
This currently compiles just like it looks, though Babel throws an error:
```
repl: "a" is read-only
  1 | import a from 'lib'
> 2 | a = 2;
    | ^
```
Is this something we want to catch? It would introduce a significant complexity, tracking which variables are read-only. I’m not sure it’s worth the effort in this pass, since Babel will catch the error anyway. Perhaps we could open an issue to add this check, and implement it the same time we add support for `const` 😉

lydell:
Quick and dirty hack code:

```diff
diff --git a/src/nodes.coffee b/src/nodes.coffee
index 64be9f5..bbb9e98 100644
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1401,6 +1401,9 @@ exports.Assign = class Assign extends Base
         else if @param
           o.scope.add varBase.value, 'var'
         else
+          if Object::hasOwnProperty.call(o.scope.positions, varBase.value) and
+             o.scope.variables[o.scope.positions[varBase.value]].type is 'import'
+            varBase.error 'foobar'
           o.scope.find varBase.value
 
     val = @value.compileToFragments o, LEVEL_LIST
```

I think it might work. Just need to clean up and use it for `import a from ''; export a = 2` as well.

About the AST, we need a few more things like this one:

```coffee
exports.BooleanLiteral = class BooleanLiteral extends Literal
```

Kinda like this:

```coffee
exports.ImportDefaultSpecifier = class ImportDefaultSpecifier extends ImportSpecifier
```

Because now some of your nice grammar are lost in the parser:

```
$ ./bin/coffee -ne 'import a from ""'
Block
  ImportDeclaration
    ImportSpecifier IdentifierLiteral: a StringLiteral: ""
```


Perhaps we can also use stuff like `instance of ImportDefaultSpecifier` instead of `@default` then. I don't know – I haven't been able to look into it enough, so please disregard this paragraph if it doesn't make sense :)

It might also make sense to use the same names as Babel for some of the AST Node _properties._ For example, `.moduleName` vs. `.source`. I mean, the closer we are to the Babel AST from the beginning, the easier it will be to convert it to Babel AST later, right? Please disregard this paragraph too if it doesn't make sense. Again, I apologize for not having read the code carefully enough before making such comments.

We are crazy close to merging now, I'd say! :)

GeoffreyBooth:
I haven’t looked into the code yet, but what do you mean by `BooleanLiteral`? A literal that’s read-only?

For whatever reason, the module name (`''` in your example here) [is just a `Literal` in astexplorer](http://astexplorer.net/#/iWnW6MVZT5). Currently we have it as a `StringLiteral`, which I do a special check on to ensure that it’s not interpolated. I could do a simple additional check that it’s not an empty string, [though Babel doesn’t throw an error for such code](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=import%20a%20from%20''). I’m not sure it’s worth creating a new type of literal here, when even Babel doesn’t have one. Or did you mean `BooleanLiteral` to refer to `a`?

Agreed about parameter names. I started doing that but then it complicated DRYing up `ImportSpecifier` and `ExportSpecifier`, since `ImportSpecifier` uses `imported` and `local` while `ExportSpecifier` uses `local` and `exported`. But I think I can square the circle, by using those names in the child class constructors while `ModuleSpecifier` uses the current `original` and `alias`.

lydell:
The line with `BooleanLiteral` in it is an actual line from nodes.coffee. What I tried to say is that you have made all these nice rules in grammar.coffee, such as 'ImportDefaultSpecifier', but there's no node in nodes.coffee called `ImportDefaultSpecifier`, so we don't get an AST like Babel. I tried to show that it can be as easy as adding a few one-liners. (`BooleanLiteral` is the node used for `true` and `false`.)

I did not mean that we should make a new node type for the module name. It should be a `StringLiteral`. And the empty string should be allowed. And it has nothing to do with `BooleanLiteral`. I meant that on the AST node, you save the module name on the `.moduleName` property, while Babel seems to save it on `.source`.

GeoffreyBooth:
Okay @lydell, your quick and dirty code actually works quite well. I’ve added a test for it, and I’ve renamed as much as I could find that diverged from Babel’s naming. How are we looking now?

test/error_messages.coffee
```diff
@@ -1126,3 +1126,13 @@ test "cannot import the same member more than once", ->
...
+
+test "imported members cannot be reassigned", ->
+  assertErrorFormat '''
+    import { foo } from 'lib'
+    foo = 'bar'
+  ''', '''
+    [stdin]:2:1: error: 'foo' is read-only
+    foo = 'bar'
+    ^^^
+  '''
```
lydell: This is good, but you missed this part of my comment (can't blame though you considering how buried this line was among all the other stuff :) ): 

> I think it might work. Just need to clean up and use it for `import a from ''; export a = 2` as well.

src/nodes.coffee
```diff
@@ -1220,21 +1220,151 @@ exports.Class = class Class extends Base
...
+    code
+
+exports.ImportSpecifier = class ImportSpecifier extends ModuleSpecifier
+  constructor: (imported, local) ->
+    super imported, local, 'import'
+
+  compileNode: (o) ->
+    # Per the spec, symbols can’t be imported multiple times
+    # (e.g. `import { foo, foo } from 'lib'` is invalid)
+    if o.importedSymbols.indexOf(@identifier) isnt -1
```
lydell: `if @identifier in o.importedSymbols`

src/nodes.coffee
```diff
@@ -1220,21 +1220,151 @@ exports.Class = class Class extends Base
...
+    super @clause, @source
+
+  compileNode: (o) ->
+    @checkScope o, 'export'
+
+    code = []
+    code.push @makeCode "#{@tab}export "
+    code.push @makeCode 'default ' if @ instanceof ExportDefaultDeclaration
+
+    if @ instanceof ExportDefaultDeclaration is no and (@clause instanceof Assign or @clause instanceof Class)
```
```coffee
if @ not instanceof ExportDefaultDeclaration and
   (@clause instanceof Assign or @clause instanceof Class)
```

src/nodes.coffee
```diff
@@ -1220,21 +1220,151 @@ exports.Class = class Class extends Base
...
+    if @source?.value?
+      code.push @makeCode ' from ' unless @clause is null
+      code.push @makeCode @source.value
+
+    code.push @makeCode ';'
+    code
+
+exports.ExportDeclaration = class ExportDeclaration extends ModuleDeclaration
+  constructor: (@clause, @source) ->
+    super @clause, @source
```
lydell: This constructor can be dropped, can't it?

GeoffreyBooth:
@lydell okay, it catches read-only assignment in `export` now too. Apparently this check shouldn’t apply to function parameters, [if we’re following Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=es2015%2Cstage-3&code=import%20%7B%20a%20%7D%20from%20'lib'%0A%0Afunction%20foo%20(a)%20%7B%0A%20%20a%20%3D%201%3B%0A%7D%0A%0Aa%20%3D%202%3B).

src/nodes.coffee
```diff
@@ -1263,17 +1263,15 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
...
-
   compileNode: (o) ->
     @checkScope o, 'export'
 
     code = []
     code.push @makeCode "#{@tab}export "
     code.push @makeCode 'default ' if @ instanceof ExportDefaultDeclaration
 
-    if @ instanceof ExportDefaultDeclaration is no and (@clause instanceof Assign or @clause instanceof Class)
+    if @ instanceof ExportDefaultDeclaration is no and
```
lydell: I think that `if @ instanceof ExportDefaultDeclaration is no` reads really badly and would like it to be `if @ not instanceof ExportDefaultDeclaration` as I suggested in my previous comment.

src/nodes.coffee
```diff
@@ -1263,17 +1263,15 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
...
+    (@clause instanceof Assign or @clause instanceof Class)
```
lydell: I think this looks really confusing when the line isn’t indented. I’d suggest adding three spaces at the beginning of the line (like I showed in my last comment).

src/nodes.coffee
```diff
@@ -1366,6 +1366,11 @@ exports.Assign = class Assign extends Base
   isStatement: (o) ->
     o?.level is LEVEL_TOP and @context? and (@moduleDeclaration or "?" in @context)
 
+  checkAssignability: (o, varBase) ->
+    if Object::hasOwnProperty.call(o.scope.positions, varBase.value) and
+    o.scope.variables[o.scope.positions[varBase.value]].type is 'import'
```
lydell: I think this looks really confusing when the line isn’t indented. I’d suggest adding three spaces at the beginning of the line.

lydell:
Nice! You are correct regarding function parameters.

src/nodes.coffee
```diff
@@ -1220,21 +1220,154 @@ exports.Class = class Class extends Base
...
+exports.ExportDeclaration = class ExportDeclaration extends ModuleDeclaration
+  compileNode: (o) ->
+    @checkScope o, 'export'
+
+    code = []
+    code.push @makeCode "#{@tab}export "
+    code.push @makeCode 'default ' if @ instanceof ExportDefaultDeclaration
+
+    if @ instanceof ExportDefaultDeclaration is no and
+    (@clause instanceof Assign or @clause instanceof Class)
```
lydell: I think this looks really confusing when the line isn’t indented. I’d suggest adding three spaces at the beginning of the line (like I showed in my last comment).

src/nodes.coffee
```diff
@@ -1220,21 +1220,154 @@ exports.Class = class Class extends Base
...
+
+exports.ExportDeclaration = class ExportDeclaration extends ModuleDeclaration
+  compileNode: (o) ->
+    @checkScope o, 'export'
+
+    code = []
+    code.push @makeCode "#{@tab}export "
+    code.push @makeCode 'default ' if @ instanceof ExportDefaultDeclaration
+
+    if @ instanceof ExportDefaultDeclaration is no and
```
lydell: I think `if @ instanceof ExportDefaultDeclaration is no` reads really badly. I’d like it to be `if @ not instanceof ExportDefaultDeclaration` (like I showed in my last comment).

src/nodes.coffee
GeoffreyBooth: Okay. Anything else?

src/coffee-script.coffee
```diff
@@ -63,6 +63,13 @@ exports.compile = compile = withPrettyErrors (code, options) ->
     token[1] for token in tokens when token[0] is 'IDENTIFIER'
   )
 
+  # Check for import or export; if found, force bare mode
+  unless options.bare? and options.bare is yes
+    for token in tokens
+      if token[0] in ['IMPORT', 'EXPORT']
+        options.bare = yes
```
vendethiel: is it really better to silently force bare rather than error out because of misconfiguration?

lydell: Yes, it is, in my opinion. This allows compiling legacy files with `--bare` _and_ newer files with import/export in them using one single `coffee --watch`.

GeoffreyBooth: The wrapper function’s purpose is to [prevent accidental leaks into the global scope](http://coffeescript.org/#lexical-scope) when running in a browser. By definition a project with module imports is a multiple-file project, and something that will undergo further processing before a browser gets it. I don’t think we need the protection that the wrapper provides for such a project. Conversely, an `import` inside such a wrapper will fail to compile; and not all build chains provide easy ways to pass flags to the CoffeeScript compiler. I don’t want to make people scratch their heads trying to figure out how to pass flags to the compiler in Webpack/Gulp/Broccoli/Browserify/`ember-cli-coffeescript`/whatever, when the benefits of the safety wrapper are negligible or perhaps unnecessary in this case.

test/error_messages.coffee
```diff
@@ -989,3 +995,160 @@ test "`&&=` and `||=` with a space in-between", ->
...
+    [stdin]:1:20: error: already imported
+    import { foo, bar, foo } from 'lib'
+                       ^^^
+  '''
+  assertErrorFormat '''
+    import { foo, bar as foo } from 'lib'
+  ''', '''
+    [stdin]:1:15: error: already imported
+    import { foo, bar as foo } from 'lib'
+                  ^^^^^^^^^^
```
lydell: In this case, we are actually importing two different things – `foo` and `bar` – but erroneously try assign `foo` twice. So the problem isn't really that something is already imported. Babel's error message is `Duplicate declaration "foo"`. Not sure what the best phrasing is.

lydell: Thinking about it some more, I now realized that there's a "duplicate declaration" error as well – I think we should collapse both into "duplicate declaration". What do you think?

src/coffee-script.coffee
vendethiel: ? I'm not arguing about not allowing both bare and modules. I'm arguing about doing it silently.

test/error_messages.coffee
GeoffreyBooth: You mean just rename "already imported" to "duplicate declaration"? I think that would work, technically "duplicate declaration" could refer to both importing the same symbol twice or trying to declare and assign the same symbol twice.

src/coffee-script.coffee
GeoffreyBooth: We haven’t written the documentation for this feature yet. When we do, we will note that using modules forces `bare` mode automatically for the files in which modules are imported. As @lydell points out, users might not want `bare` mode for all files in a project, and the flag doesn’t provide a way to enable it selectively. Simply telling people to set `bare` to true isn’t an option for that reason.

We will also be writing big warnings about how CoffeeScript won’t be resolving modules and therefore Babel or similar will be necessary. This will all get covered. Thanks for pointing it out, though, it’s important that we don’t forget to note this.

vendethiel: That sounds good, thanks

src/nodes.coffee
```diff
@@ -1220,21 +1220,154 @@ exports.Class = class Class extends Base
...
+  constructor: (imported, local) ->
+    super imported, local, 'import'
+
+  compileNode: (o) ->
+    # Per the spec, symbols can’t be imported multiple times
+    # (e.g. `import { foo, foo } from 'lib'` is invalid)
+    if @identifier in o.importedSymbols
+      @error 'already imported'
+    else if o.scope.check(@identifier)
+      @error 'duplicate declaration'
```
lydell: > You mean just rename "already imported" to "duplicate declaration"?

@GeoffreyBooth I'm answering your question here :) I'm thinking of collapsing the above lines to:

```coffee
if @identifier in o.importedSymbols or o.scope.check(@identifier)
  @error 'duplicate declaration'
```

lydell:
Here's some code that fails to compile, but should actually do compile:

```
$ ./bin/coffee -bpe 'import {} from ""'
[stdin]:1:9: error: unexpected }
import {} from ""
        ^
```

src/nodes.coffee
JimPanic: How about also adding the identifier that is duplicated?

`Duplicate error: @identifier was declared more than once.`

src/grammar.coffee
```diff
@@ -349,6 +351,65 @@ grammar =
...
+  ImportDefaultSpecifier: [
+    o 'Identifier',                             -> new ImportDefaultSpecifier $1
+  ]
+
+  ImportNamespaceSpecifier: [
+    o 'IMPORT_ALL AS Identifier',               -> new ImportNamespaceSpecifier new Literal($1), $3
+  ]
+
+  Export: [
+    o 'EXPORT { }',                                          -> new ExportNamedDeclaration new Literal '{}'
```
lydell: `new Literal '{}'` should be replaced with an empty `ExportSpecifierList` here.

src/grammar.coffee
```diff
@@ -349,6 +351,65 @@ grammar =
     o 'CLASS SimpleAssignable EXTENDS Expression Block', -> new Class $2, $4, $5
   ]
 
+  Import: [
+    o 'IMPORT String',                                                                -> new ImportDeclaration null, $2
+    o 'IMPORT ImportDefaultSpecifier FROM String',                                    -> new ImportDeclaration [$2], $4
+    o 'IMPORT ImportNamespaceSpecifier FROM String',                                  -> new ImportDeclaration [$2], $4
+    o 'IMPORT { ImportSpecifierList OptComma } FROM String',                          -> new ImportDeclaration [new ImportSpecifierList $3], $7
+    o 'IMPORT ImportDefaultSpecifier , ImportNamespaceSpecifier FROM String',         -> new ImportDeclaration [$2, new Literal(', '), $4], $6
+    o 'IMPORT ImportDefaultSpecifier , { ImportSpecifierList OptComma } FROM String', -> new ImportDeclaration [$2, new Literal(', '), new ImportSpecifierList($5)], $9
```
lydell: Those `new Literal(', ')` commas should not be in the AST.

```
$ ./bin/coffee -ne 'import a, {b} from ""'
Block
  ImportDeclaration
    ImportDefaultSpecifier IdentifierLiteral: a Literal: , 
    ImportSpecifierList
      ImportSpecifier IdentifierLiteral: b StringLiteral: ""
```

GeoffreyBooth:
I updated `error: duplicate declaration` to `error: 'foo' has already been declared` per @JimPanic’s suggestion. I like including the variable name, but I wonder if talking about declaration is really the proper phrasing for us. CoffeeScript doesn’t really have variable declarations; the compiler auto-generates the variable declarations for us. Perhaps the error should be something like `error: 'foo' is the name of an import and cannot be used as a variable name` or somesuch? Though then we would need to revive the separate error for `already imported`. I dunno.

src/nodes.coffee
```diff
@@ -1336,7 +1337,7 @@ exports.ImportSpecifier = class ImportSpecifier extends ModuleSpecifier
     # Per the spec, symbols can’t be imported multiple times
     # (e.g. `import { foo, foo } from 'lib'` is invalid)
     if @identifier in o.importedSymbols or o.scope.check(@identifier)
-      @error 'duplicate declaration'
+      @error "'#{@identifier}' has already been declared"
```
lydell: Nice! 👍

test/modules.coffee
```diff
@@ -79,6 +79,16 @@ test "import an entire module's contents as an alias, adding the alias to the cu
     foo.fooMethod();"""
   eq toJS(input), output
 
+test "import empty object", ->
```
lydell: The `{}` is not an object here.

src/nodes.coffee
```diff
@@ -1252,7 +1252,8 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
     code.push @makeCode "#{@tab}import "
 
     if @clause? and @clause.length isnt 0
-      for subclause in @clause
+      for subclause, index in @clause
+        code.push @makeCode ', ' if index isnt 0
         code = code.concat subclause.compileNode o
```
lydell: Isn't it somewhat strange to use an array for `@clause` when it can only contain one or two values? This code kinda implies that `@clause` could be of any length. I guess it doesn't matter in the end though…

GeoffreyBooth: An earlier version had `firstClause` and `secondClause` but that looked sloppy. I suppose this could technically say `if index is 1` because the index will never hit 2, but that seems unnecessarily opaque.

GeoffreyBooth:
So, as we continue to bicker over code style, does anyone want to run any more real-world tests with this PR? I already tested this code on two more-than-example-size apps that I linked to in the top comment, but if anyone has any real CoffeeScript projects of their own that do module imports (whether via CommonJS or backticked ES2015), perhaps those could be refactored to use this syntax and compiled using this PR? And let us know if anything fails to work as expected. Bonus points if your app has tests and you can run the tests before and after refactoring your module imports/exports.

src/nodes.coffee
lydell: Ok. Let’s keep it the way it is.

lydell:
What about unifying the `'foo' is read-only` and `'foo' has already been declared` to `the name 'foo' cannot be used here because it clashes with an imported name`? Not totally happy with that, but something along those lines.

Btw, the error messages are the only thing left. Then everything looks good to me.

My idea is to use github’s squash and merge feature and use this commit message:

```
<PR title> (#4300)

<PR description>
```

GeoffreyBooth:
What, I don’t get to add a hundred commits onto `master`? 😄 

I don’t think combining those error messages works. `the name 'foo' cannot be used here` implies that the user is intending to declare a variable; they just as easily might be intending to reassign it. In CoffeeScript there’s no way to know for sure. `read-only` is the language Babel uses.

lydell:
@GeoffreyBooth What you say about the error messages makes sense. Do you have any other ideas, or are we good with the ones we have? @JimPanic do you have any input?

@vendethiel Any more comments?

@michaelficarra Would you like to take a look before merging? (I think nobody in the CoffeeScript community knows the ECMAScript spec like you do!)

JimPanic:
I'd keep the error message as simple as possible unless we have more information at hand in this particular case. But I think we don't?

vendethiel:
Nothing that wouldn't be minor nitpicking. Amazing job!

jashkenas:
A few thoughts:

* Beautiful work. I know how much of a bear it can be to wrestle with large additions to the grammar, and it looks like this turned out just about as well as it could, considering ...

* ... what a shame it is that JS modules — which add more or less zero new functionality to JavaScript — require such large and semi-alien additions to JS grammar, and now CoffeeScript grammar. The `{ ... }` for member access feels particularly unCoffeeScript-y, not that I have an alternative to offer at this time.

* ... that modules after all these years still aren't done: image

* ... and that their implementation seems to maybe even be moving backwards: https://github.com/nodejs/node-eps/pull/39

I'd be interested in hearing from some of the folks planning on using this patch. Are you all going to be pumping compiled CoffeeScript through another transpiler? Do you not think it would be wiser to wait until modules can be used directly in runtimes before emitting them?

That said, if y'all think this is essential, feel free to merge it in.

rattrayalex:
Really impressed @GeoffreyBooth @simonlydell et al. Amazing work here. 

> Are you all going to be pumping compiled CoffeeScript through another transpiler? 

tl;dr, yes, and happily so. On the server this may look like `babel-node compiled.js` and on the frontend this may look like a webpack config with coffee and then babel loaders in sequence, which is relatively trivial to setup. 

Thanks for your support, @jashkenas !

lydell:
To me, the fact that no browser supports import/export natively doesn't matter to me that much. The important thing is that the import/export syntax is _standardized._ Finally there's a way to write modules that any tool can consume. No more hacks to get it working with commonjs, AMD and browser globals at the same time (which is called UMD). Just write standard code. Let the consumers of your module use the tools they need to get it going.

Another great thing with import/export is tree shaking, which means that tools can statically analyze your code and remove the parts that don't end up being used, which results in smaller download sizes.

Linters can also do better analysis on import/export than `require()`.

It's not just about the runtime support, it is also about the tooling support.

(Finally, the `{ ... }` for member access feels particularly CoffeeScript-y, considering how similar it is to `data = {foo, bar, baz}`. ;) )

zeekay:
Agree with @lydell. Native support isn't that important because most browser projects will ship bundled code to reduce size + load times. CommonJS-based bundlers like browserify and webpack are [quite inferior](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/) to ES6 module-based tools like [rollup](https://github.com/rollup/rollup) (and even Closure Compiler in a lot of cases). It can even improve performance on Node.js, as it allows libraries to flatten sub-modules into a single CommonJS compatible bundle which improves install times and require overhead.

Looking forward to taking advantage of ES6 modules so I can optimize browser applications and interoperate with existing ES6 module tooling!

GeoffreyBooth:
Agree with everything stated above. I think the time has come that we need to start looking beyond browser support to _ecosystem_ support, meaning what does the JavaScript world treat as a standard? Clearly modules are a feature that Meteor and Ember, to name two frameworks I’m familiar with, take not just as a given but as a required feature that you need to use to use their frameworks. Using React without classes is technically possible, but good luck finding examples for that syntax. These frameworks don’t care that browser support is lacking; they assume that people will transpile. CoffeeScript basically invented transpiling, so it’s hardly something unfamiliar to our community.

CoffeeScript must be able to interoperate with the rest of the JavaScript community, or people won’t be able to use CoffeeScript. I’ve written a lot more about this in the [coffeescript6 readme](https://github.com/coffeescript6/discuss).

jashkenas:
Very persuasive, folks. Sounds great.

chrisdickinson:
:wave: Hey folks! I'm Chris, a CTC member over on the Node project, currently focussed on the ES modules effort.

I only just now saw this PR, so forgive me for throwing up a warning flag so late in the process: the loading semantics that babel is _currently_ using are (IIUC) incompatible with those proposed by the spec & by TC39, and _may_ be changing in the near future. I have this info from @bmeck: I'm looping in @loganfsmyth from Babel who may be able to speak more to that. I wanted to make sure that y'all had this information before pulling something with the potential to change drastically into the language. Thanks for your time!

GeoffreyBooth:
@chrisdickinson Is the _syntax_ changing? In other words, what I see on the MDN [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) pages for `import` and `export` statements?

All we’re doing here is outputting `import` and `export` statements for Babel or another tool to go and actually resolve the modules. If the syntax is finalized and not changing, we’re okay.

bmeck:
@GeoffreyBooth the semantics of what goes on is definitely changing:

https://docs.google.com/presentation/d/1aq_QjBUQTovj9aQZQrVzS7l1aiOs3ZNlk7wgNTUEMy0/edit#slide=id.g16ab11d101_0_5

See notes on named imports and timing in particular.

GeoffreyBooth:
@bmeck but is the syntax of `import` or `export` statements changing?

bmeck:
@GeoffreyBooth what they do according to ES spec is different from current behavior, but syntax is same. You will see errors from using existing babel impl likely if:

1. you use any named imports to try to get something from a NCJS file
2. are not prepared to have your compiled module be wrapped in a Promise / only eval once stack unwinds.

GeoffreyBooth:
@bmeck @chrisdickinson Thanks for letting us know. Fortunately I don’t think the changes you’re describing should affect CoffeeScript. All that this PR is doing (that CoffeeScript will do) regarding modules is to take a CoffeeScript file like this:

```coffee
# main.coffee
import { _ } from 'underscore'

console.log _.now()
```

and transpile it into a JavaScript file like this:
```js
// main.js
import { _ } from 'underscore';

console.log(_.now());
```

Actually doing something with that `import` statement is the responsibility of some downstream tool like Babel, which the developer is responsible for setting up. So as long as the syntax of `import` and `export` isn’t changing, we don’t need to do anything. The CoffeeScript compiler is essentially just a converter from CoffeeScript syntax to JavaScript syntax, from one string to another string; we don’t actually evaluate the code.

chrisdickinson:
@GeoffreyBooth Thanks — sorry, I misunderstood! On first read, I thought this implementing ESM-on-top-of-CommonJS in the style of the Babel transpiler, and wanted to give advance warning that those semantics may be changing. If it's a passthrough that should be totally fine!

GeoffreyBooth:
@lydell and @michaelficarra, is there anything else we need to do here?

lydell:
Yes. I started playing around with trying to get babel-generator to create a string of JavaScript from CoffeeScript's AST – but only import/export. Then I discovered that the ASTs are a bit different.

```
$ node -p 'require("util").inspect(require("babylon").parse("import a, {b, c as d} from \"lib\"", {sourceType: "module"}).program.body[0], {depth: null, colors: true})'
Node {
  type: 'ImportDeclaration',
  start: 0,
  end: 32,
  loc: 
   SourceLocation {
     start: Position { line: 1, column: 0 },
     end: Position { line: 1, column: 32 } },
  specifiers: 
   [ Node {
       type: 'ImportDefaultSpecifier',
       start: 7,
       end: 8,
       loc: 
        SourceLocation {
          start: Position { line: 1, column: 7 },
          end: Position { line: 1, column: 8 } },
       local: 
        Node {
          type: 'Identifier',
          start: 7,
          end: 8,
          loc: 
           SourceLocation {
             start: Position { line: 1, column: 7 },
             end: Position { line: 1, column: 8 },
             identifierName: 'a' },
          name: 'a' } },
     Node {
       type: 'ImportSpecifier',
       start: 11,
       end: 12,
       loc: 
        SourceLocation {
          start: Position { line: 1, column: 11 },
          end: Position { line: 1, column: 12 } },
       imported: 
        Node {
          type: 'Identifier',
          start: 11,
          end: 12,
          loc: 
           SourceLocation {
             start: Position { line: 1, column: 11 },
             end: Position { line: 1, column: 12 },
             identifierName: 'b' },
          name: 'b' },
       local: 
        Node {
          type: 'Identifier',
          start: 11,
          end: 12,
          loc: 
           SourceLocation {
             start: Position { line: 1, column: 11 },
             end: Position { line: 1, column: 12 },
             identifierName: 'b' },
          name: 'b',
          __clone: [Function: __clone] } },
     Node {
       type: 'ImportSpecifier',
       start: 14,
       end: 20,
       loc: 
        SourceLocation {
          start: Position { line: 1, column: 14 },
          end: Position { line: 1, column: 20 } },
       imported: 
        Node {
          type: 'Identifier',
          start: 14,
          end: 15,
          loc: 
           SourceLocation {
             start: Position { line: 1, column: 14 },
             end: Position { line: 1, column: 15 },
             identifierName: 'c' },
          name: 'c' },
       local: 
        Node {
          type: 'Identifier',
          start: 19,
          end: 20,
          loc: 
           SourceLocation {
             start: Position { line: 1, column: 19 },
             end: Position { line: 1, column: 20 },
             identifierName: 'd' },
          name: 'd' } } ],
  source: 
   Node {
     type: 'StringLiteral',
     start: 27,
     end: 32,
     loc: 
      SourceLocation {
        start: Position { line: 1, column: 27 },
        end: Position { line: 1, column: 32 } },
     extra: { rawValue: 'lib', raw: '"lib"' },
     value: 'lib' } }
```

```
$ node -p 'require("util").inspect(require("./").nodes("import a, {b, c as d} from \"lib\"").expressions[0], {depth: null, colors: true})'
ImportDeclaration {
  clause: 
   [ ImportDefaultSpecifier {
       original: 
        IdentifierLiteral {
          value: 'a',
          locationData: { first_line: 0, first_column: 7, last_line: 0, last_column: 7 } },
       alias: undefined,
       moduleDeclarationType: 'import',
       identifier: 'a',
       locationData: { first_line: 0, first_column: 7, last_line: 0, last_column: 7 } },
     ImportSpecifierList {
       specifiers: 
        [ ImportSpecifier {
            original: 
             IdentifierLiteral {
               value: 'b',
               locationData: { first_line: 0, first_column: 11, last_line: 0, last_column: 11 } },
            alias: undefined,
            moduleDeclarationType: 'import',
            identifier: 'b',
            locationData: { first_line: 0, first_column: 11, last_line: 0, last_column: 11 } },
          ImportSpecifier {
            original: 
             IdentifierLiteral {
               value: 'c',
               locationData: { first_line: 0, first_column: 14, last_line: 0, last_column: 14 } },
            alias: 
             IdentifierLiteral {
               value: 'd',
               locationData: { first_line: 0, first_column: 19, last_line: 0, last_column: 19 } },
            moduleDeclarationType: 'import',
            identifier: 'd',
            locationData: { first_line: 0, first_column: 14, last_line: 0, last_column: 19 } } ],
       locationData: { first_line: 0, first_column: 0, last_line: 0, last_column: 31 } } ],
  source: 
   StringLiteral {
     value: '"lib"',
     locationData: { first_line: 0, first_column: 27, last_line: 0, last_column: 31 } },
  locationData: { first_line: 0, first_column: 0, last_line: 0, last_column: 31 } }
```

GeoffreyBooth:
So to replicate that, I would need to rename the `children` properties of `class ImportDeclaration` . . . but then how would I avoid the level of `ImportSpecifierList`? Somehow within `grammar.coffee` create an array of `ImportSpecifier` objects?

Why are we trying to match Babel’s AST? Surely many of CoffeeScript’s other features diverge from it?

lydell:
> Why are we trying to match Babel’s AST?

Because I thought it was more or less decided over at the coffeescript6/discuss repo that a future goal is to make the compiler part of CoffeeScript simply turn its own CoffeeScript AST into Babel AST and then hand that off to Babel's stringifier (babel-generator). (Or you could pass the Babel AST directly to Babel for transforming into ES5). So my thinking has been that the closer we are to the Babel AST from the beginning, the easier we make things for us down the line.

rattrayalex:
Hmm, I definitely leave that up to @GeoffreyBooth , but my impression is that while it might be nice to compile to a Babel AST at some point, it's not an immediate goal and not something we want to hold us back at this point.

vendethiel:
Congrats on your contributor stripes, @GeoffreyBooth :).

lydell:
@GeoffreyBooth Thinking about the AST some more, Babel (and the other ESTree(-inspired) parsers) don't actually have the best AST here. Mixing the default binding and named imports doesn't really make sense. babel-generator has some [pretty ugly code](https://github.com/babel/babel/blob/de975b9660014a4533c90190b297352495636530/packages/babel-generator/src/generators/modules.js#L81-L107) for turning that into a string again. It might make sense to go with what we have.

GeoffreyBooth:
@vendethiel Thanks! I think the `class` keyword would be so much cooler renamed to `klass` . . .

@lydell agreed. I also don’t think we should make assumptions just yet about how we might theoretically implement a new compiler. I agree that the AST-matching approach holds a lot of promise, but until we know what exactly we’re trying to produce as nodes output I don’t think we should make our current code unnecessarily less comprehensible. I also feel like it’s likely that we take the approach of Decaf, and have the new compiler only handle certain types of nodes and fall back on the old compiler for others. In the case of modules, if the old compiler is already outputting ESNext then there’s not really a need for redoing it (until the day comes that the new compiler is ready to compile _everything,_ and we’re eliminating the last few old compiler-generated nodes).

vendethiel:
> @vendethiel Thanks! I think the class keyword would be so much cooler renamed to klass . . .

Rubyists around the world agree :P.

lydell:
@GeoffreyBooth What do you think about this:

```diff
diff --git a/src/grammar.coffee b/src/grammar.coffee
index 93a7c0b..bf236a4 100644
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -353,12 +353,12 @@ grammar =
 
   Import: [
     o 'IMPORT String',                                                                -> new ImportDeclaration null, $2
-    o 'IMPORT ImportDefaultSpecifier FROM String',                                    -> new ImportDeclaration [$2], $4
-    o 'IMPORT ImportNamespaceSpecifier FROM String',                                  -> new ImportDeclaration [$2], $4
-    o 'IMPORT { } FROM String',                                                       -> new ImportDeclaration [new ImportSpecifierList []], $5
-    o 'IMPORT { ImportSpecifierList OptComma } FROM String',                          -> new ImportDeclaration [new ImportSpecifierList $3], $7
-    o 'IMPORT ImportDefaultSpecifier , ImportNamespaceSpecifier FROM String',         -> new ImportDeclaration [$2, $4], $6
-    o 'IMPORT ImportDefaultSpecifier , { ImportSpecifierList OptComma } FROM String', -> new ImportDeclaration [$2, new ImportSpecifierList($5)], $9
+    o 'IMPORT ImportDefaultSpecifier FROM String',                                    -> new ImportDeclaration new ImportClause($2, null), $4
+    o 'IMPORT ImportNamespaceSpecifier FROM String',                                  -> new ImportDeclaration new ImportClause(null, $2), $4
+    o 'IMPORT { } FROM String',                                                       -> new ImportDeclaration new ImportClause(null, new ImportSpecifierList []), $5
+    o 'IMPORT { ImportSpecifierList OptComma } FROM String',                          -> new ImportDeclaration new ImportClause(null, new ImportSpecifierList $3), $7
+    o 'IMPORT ImportDefaultSpecifier , ImportNamespaceSpecifier FROM String',         -> new ImportDeclaration new ImportClause($2, $4), $6
+    o 'IMPORT ImportDefaultSpecifier , { ImportSpecifierList OptComma } FROM String', -> new ImportDeclaration new ImportClause($2, new ImportSpecifierList $5), $9
   ]
 
   ImportSpecifierList: [
diff --git a/src/nodes.coffee b/src/nodes.coffee
index c0ad7a3..ed928f6 100644
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1250,11 +1250,7 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
 
     code = []
     code.push @makeCode "#{@tab}import "
-
-    if @clause? and @clause.length isnt 0
-      for subclause, index in @clause
-        code.push @makeCode ', ' if index isnt 0
-        code = code.concat subclause.compileNode o
+    code.push @clause.compileNode(o)... if @clause?
 
     if @source?.value?
       code.push @makeCode ' from ' unless @clause is null
@@ -1263,6 +1259,24 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
     code.push @makeCode ';'
     code
 
+exports.ImportClause = class ImportClause extends Base
+  constructor: (@defaultBinding, @namedImports) ->
+
+  children: ['defaultBinding', 'namedImports']
+
+  compileNode: (o) ->
+    code = []
+
+    if @defaultBinding
+      code.push @defaultBinding.compileNode(o)...
+      code.push @makeCode ', ' if @namedImports
+
+    if @namedImports
+      code.push @namedImports.compileNode(o)...
+
+    code
+
+
 exports.ExportDeclaration = class ExportDeclaration extends ModuleDeclaration
   compileNode: (o) ->
     @checkScope o, 'export'
```

GeoffreyBooth:
@lydell doesn’t that push us even farther from Babel’s AST? Wouldn’t the nodes generated by that be ImportDeclaration » ImportClause » ImportSpecifierList » ImportSpecifier?

Would we create an ExportClause too? This PR used to have ImportClause and ExportClause but I removed them because we were trying to get closer to Babel.

lydell:
Yes, that would move us farther away from Babel. But we kinda decided not to make the AST exactly like Babel anyway, and then I think it's nicer to have `@clause` be some kind of Node, rather than an array. An array doesn't really make sense, but a ImportClause with one field for "either side of the import comma" does. For export, `@clause` already seems to always be a Node, and doesn't really benefit from having a special ExportClause, does it?

Again, in short, if we're _not_ going "full Babel AST" (which I have nothing against, now that I've realized that Babel's AST is worse on this particular point) that means we're going to stick with what we have instead. If we stick with what we have, we could just as well make it a tiny bit better. IMO, the code became easier to understand.

GeoffreyBooth:
Alrighty. I added some existential operators rather than checking just for truthiness, but otherwise it’s exactly your diff. `code.push @clause.compileNode(o)... if @clause?` is pretty slick.

With regard to the AST, let’s leave that for when we actually start implementing that approach. I have a feeling we’re going to have some kind of intermediate layer where we rewrite our AST tree to conform to the target tree. (Doesn’t Decaffeinate have something like that?) A rewrite layer might be preferable to awful contortions in our grammar. Anyway, let’s deal with it later when we’ll have a better sense of which is the less-bad option.

Anything else?

lydell:
> I have a feeling we’re going to have some kind of intermediate layer where we rewrite our AST tree to conform to the target tree. (Doesn’t Decaffeinate have something like that?) A rewrite layer might be preferable to awful contortions in our grammar.

Correct! Since CoffeeScript isn't JavaScript, it is going to need to have its own AST nodes for many things. The parsing phase builds the AST from a string. The compiling phase takes the AST as input, and outputs JavaScript AST. Another tool can then turn that JS AST into a string.

So, you're right. There's no need to match Babel exactly here – we won't be able to in many places anyway. I think we've landed on a very nice spot right now – our AST is not unnecessarily different from Babel's, and it's better where it makes sense for us.

> With regard to the AST, let’s leave that for when we actually start implementing that approach. … let’s deal with it later when we’ll have a better sense of which is the less-bad option.

Definitely! 👍 

> Anything else?

I don't think so. We'll probably merge tonight.

lydell:
There we go, merged! Now, I think we should:

1. Document import/export.
2. Make a new release. See #4079 for inspiration.

JimPanic:
Amazing! 🙌

GeoffreyBooth:
Thank you @lydell! Should I start a new branch for the documentation? On my repo or the main one?

If I handle the docs do you want to handle the release? 😄

lydell:
> Should I start a new branch for the documentation? On my repo or the main one?

I usually do stuff in branches on my own fork and then submit a pull request.

> If I handle the docs do you want to handle the release? 😄

Ok.

jashkenas:
> There we go, merged! Now, I think we should:
> 
> Document import/export.
> Make a new release. See #4079 for inspiration.

Fantastic. @lydell, you're now an owner of the npm module as well, so feel free to release when you're ready.

vendethiel:
👍

test/error_messages.coffee
```diff
@@ -989,3 +995,160 @@ test "`&&=` and `||=` with a space in-between", ->
...
+           ^^
+  '''
+
+test "anonymous classes cannot be exported", ->
+  assertErrorFormat '''
+    export class
+      constructor: ->
+        console.log 'hello, world!'
+  ''', '''
+    SyntaxError: Unexpected token export
```
GeoffreyBooth: Dammit, you’re completely correct. The generated JavaScript for that test is:
```js
export var (function() {
  function _Class() {
    console.log('hello, world!');
  }

  return _Class;

})();
```
which is completely invalid. I’ll add a check to catch this within the CoffeeScript compiler. Thanks for pointing this out.

zapu: Thanks! I'm glad I was able to help :)

Issue #2141, opened 2012-02-26
Parser trips on hard tabs with misleading error message: Unexpected '{'


dkamins:
Presence of hard tabs in certain places in .coffee source causes confusing and irrelevant parser error: **Unexpected '{'**

Tested with v1.2.0.

Specifically see example when object property line with tab is followed by regularly space-indented property line ("Test 2" below).

Test case with proper tabs preserved: http://pastebin.ca/2121616

Or test case inline below, but github converted the tab character to spaces, so you must convert it back yourself:

``` coffeescript
# Tabs vs. Spaces in CoffeeScript v1.2.0
#
# This file contains both TABS and SPACES - that's the whole point!
# Copy/Paste carefully.

# Test 1:
ok_in_this_order:
  spaces_on_this_line: 2
    tabs_on_this_line: 1

# Test 2:
confuses_parser_in_this_order:
    tabs_on_this_line: 1
  spaces_on_this_line: 2

# Test 2 yields the confusing message on compilation:
#  Parse error on line 7: Unexpected '{'

# Error: In test.coffee, Parse error on line 7: Unexpected '{'
#     at Object.parseError (/opt/local/lib/node_modules/coffee-script/lib/coffee-script/parser.js:470:11)
#     at Object.parse (/opt/local/lib/node_modules/coffee-script/lib/coffee-script/parser.js:546:22)
#     at Object.compile (/opt/local/lib/node_modules/coffee-script/lib/coffee-script/coffee-script.js:40:22)
#     at /opt/local/lib/node_modules/coffee-script/lib/coffee-script/command.js:140:33
#     at /opt/local/lib/node_modules/coffee-script/lib/coffee-script/command.js:110:18
#     at [object Object].<anonymous> (fs.js:115:5)
#     at [object Object].emit (events.js:64:17)
#     at afterRead (fs.js:1112:12)
#     at Object.wrapper [as oncomplete] (fs.js:254:17)

```

PR #4303, opened 2016-09-03
Allow mixing of tabs and spaces if an explicit tabstop size is set

eelco:
[“CoffeeScript just counts tabs and counts spaces -- you can't possibly expect it to guess how wide your text editor is configured to visually show tab characters.”](https://github.com/jashkenas/coffeescript/issues/2141#issuecomment-4180847)

But, what if you could tell CoffeeScript how wide a tab stop is? Following the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) I think it would be great if you could instruct CoffeeScript to treat indentation just as it is visually. This of course requires cooperation of the editor / environment or explicit configuration by the user, making this an opt-in feature ([we](https://framerjs.com) can do this).

I’ve read the discussions about tabs/spaces in a couple of issues and have seen that the sentiment generally is to not allow mixing, although it’s currently not enforced. However, while CoffeeScript is a really nice language to learn to program, the requirement to carefully manage your whitespace is quite unfriendly. The current behavior, where a tab is equivalent to a space in width is imho a poor choice: no editor is configured like that. (Python 2’s behavior of having a tabstop width of 8 makes a bit more sense).

The implementation in this pull request is mostly a ‘proof of concept’, to show what I’m talking about. Having this option would make parsing dependent on a specific option, meaning that some files won’t parse unless the exact right size is passed. A better way to do it (albeit still breaking backwards compatibility) might be to have it set via modeline at the start of the file.

I’m interested to hear from the maintainers and the community if anybody thinks pursuing a feature like this has merit.

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -44,6 +44,8 @@ exports.Lexer = class Lexer
     @tokens     = []             # Stream of parsed tokens in the form `['TYPE', value, location data]`.
     @seenFor    = no             # Used to recognize FORIN and FOROF tokens.
 
+    @tabSize    = opts.tabSize or 1 # Width of tab stop
+
     @chunkLine =
       opts.line or 0         # The start line for the current @chunk.
     @chunkColumn =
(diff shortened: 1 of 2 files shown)


jashkenas:
Mixing tabs and spaces in a single file is a recipe for madness … and then sadness. Let's not add code to support something that is unsupportable, given different people with differently configured editors.

eelco:
Hi @jashkenas, thanks for replying! :) 

How would you feel about changing the tab width to something more sensible than 1 (I’d prefer 4), *plus* making mixed spaces and tabs for indentation an error in the parser?

Currently, something like this is valid code:
```
layer = new Layer
\t\tx: 1
  y: 0
```

which is definitely a recipe for sadness when seeing it in an editor (without invisibles on) and then trying to align it. Changing the tab width would not fix this (only make it a bit more likely that working code also aligns).

jashkenas:
We don't want to make it slightly more sensible, because slightly more sensible is still deeply, fundamentally, broken.

I would be open to making mixed tabs and spaces a parse error — but I can also imagine that breaking various folks who are doing complicated or sophisticated tricky things. Really, I think that the best behavior is the behavior we currently have:

* Don't mix tabs and spaces.
* For purposes of indentation, a tab is one indent in a file of tabs, and a space is one indent in a file of spaces.

eelco:
I see your point, but that’s not the behavior it currently has, that’s an undocumented expectation of the people that use Coffeescript.

I’m often seeing code, especially from beginners (so they might have copy-pasted some code from examples) that mixes tabs and spaces. For them, code not working even though it’s visually aligned is baffling, especially because the errors they’ll get might not make sense. (Try it with the example above, align the 3rd line with spaces and you’ll get “Error on line 3: unexpected :”)

While my proposals might have been a bit too radical (or, well, just wrong), I don’t believe this can’t be improved. It would be great to do this in the context of the official version of Coffeescript, to improve the situation for everybody.

Issue #3832, opened 2015-02-05
No way to do ES6 'for .. of' loops for generators?


jagill:
Given a generator

``` coffeescript
gen = ->
  j = 0
  while j < 9
    yield j++
  return
```

how do we loop through it via the javascript `for .. of` loops?  In javascript, we'd do

``` javascript
for (i of gen()) {
  console.log(i);
}
```

The coffeescript `for i in gen()` gives the C-style `for (_i = 0; _i < gen().length; i++)` loops, and `for i of gen()` gives `for (i in gen())`.  Is there some way to get `for (i of gen())` that I didn't see in the documentation?

ssboisen:
I came to ask the same question. Having to manually run next and check for done is quite counter to the syntax-goodness that CS normally prides itself of.

jashkenas:
Shoot. That's a major bummer.

@alubbe et. al — any ideas here? Do all of the runtimes that support generators also support ES2015's `for of`?

alubbe:
@jashkenas to my knowledge only chrome, firefox and opera support generators and all three support ```for .. of``` - so the answer would seem to be yes.

Syntax wise, this will be a bit of a challenge. ```for all .. of/in```? ```each ... of/in```?

michaelficarra:
We might have to add a flag for this to compile CoffeeScript for-in to ES6 for-of. I don't see any better way.

jagill:
@michaelficarra That sound problematic -- a given script could either loop through generators, or arrays.

Node 0.11 and iojs also support `for of`.

I'd much rather have a special syntax, like `for gen x of myGen()`.

michaelficarra:
@jagill: ES6 for-of works with arrays because they are iterables.

epidemian:
I think @michaelficarra's suggestion is the most sensible approach.

I personally would prefer releasing a backward incompatible version of CS (2.0, 3.0, what have you) that aligned `for in` and `for of` semantics with JS and have a conversion tool that can be run reliably and transform CS code from one version to the other. After all, it's just a deterministic syntax transformation.

I'm not in favour of adding yet another looping construct to CS.

vendethiel:
if they're supposed to be used by generators, can't we use something like the currently-invalid `from x from y()`?

lydell:
@vendethiel did you mean `for x from y()`? I like that.

A current workaround would be something like:

```coffee
forOf = (gen, fn) ->
  `for (value of gen()) fn(value)`

[...]

PR #4306, opened 2016-09-12
Add a `for .. from ..` loop for generators, see #3832

atg:
This implements a `for x from g` loop as argued for in [this comment](https://github.com/jashkenas/coffeescript/issues/3832#issuecomment-203739975).

Currently CoffeeScript supports generators using `yield` and `yield from`, but it doesn't support a way to _use_ these generators like `for (x of g)` in ES6.

CoffeeScript already has `for in` (for Arrays) and `for of` (for Objects), however users expect both to generate es5-compatible code. It's impossible to modify either of these syntaxes to have all three of es5 compatibility, efficiency and support for generators/iterables.

That means the choice is either a compile-time flag or a new loop syntax. A compile-time flag is likely to create subtle bugs: if the code is expected to be compiled in es6 mode but is actually compiled in es5 mode, then a `for in` loop over an iterator will be a normal for loop `for (var i = 0; i < iterator.length; i++)`. Iterators are objects, and `.length` on an iterator is usually 0. The loop will therefore silently fail to run, with no error message.

That leaves a new loop syntax.

The best argument for introducing a new for loop syntax for generators is that they are fundamentally different from either arrays or objects. Iterating over a generator _mutates_ that generator, the next time you iterate over it, it will be empty. Generators cannot be iterated over in reverse order (because they are not indexed). Generators can be infinite.

So aside from being the only option, a new syntax has the benefit of making it clear from the code which loops are meant for Arrays, and which loops are meant for generators (or any iterable).

As for the specific syntax, `for x from y` is a nice complement to `yield from y`.


lydell:
I like this. 👍

src/nodes.coffee
```diff
@@ -2178,6 +2180,9 @@ exports.For = class For extends While
     if @object
       forPartFragments   = [@makeCode("#{kvar} in #{svar}")]
       guardPart = "\n#{idt1}if (!#{utility 'hasProp', o}.call(#{svar}, #{kvar})) continue;" if @own
+    if @from
+      forPartFragments   = [@makeCode("#{kvar} of #{svar}")]
+      guardPart = "\n"
```
vendethiel: why this?

src/nodes.coffee
```diff
@@ -2095,13 +2095,15 @@ exports.For = class For extends While
     @body    = Block.wrap [body]
     @own     = !!source.own
     @object  = !!source.object
-    [@name, @index] = [@index, @name] if @object
+    @from    = !!source.from
+    [@name, @index] = [@index, @name] if @object or @from
     @index.error 'index cannot be a pattern matching expression' if @index instanceof Value
     @range   = @source instanceof Value and @source.base instanceof Range and not @source.properties.length
     @pattern = @name instanceof Value
     @index.error 'indexes do not apply to range loops' if @range and @index
```
vendethiel: this can be an issue:

```
./bin/coffee -bce '1 for a from [1..5]'
[stdin]:1:7: error: indexes do not apply to range loops
1 for a from [1..5]
```

src/lexer.coffee
```diff
@@ -121,7 +121,11 @@ exports.Lexer = class Lexer
         'PROPERTY'
       else
         'IDENTIFIER'
-
+    
+    if tag is 'IDENTIFIER' and @seenFor and id is 'from'
```
vendethiel: add `FROM` to `RELATIONS` instead (~L939)

atg: `from` is not really a relation like `in`, `of` or `instanceof`. It can't be used as an infix operator. You can do `x in y` and `x of y` but not `x from y`.

src/nodes.coffee
atg: You mean the guardPart? Yeah, I'll delete that.

src/nodes.coffee
atg: OK, I've made `@range` conditional on `@from` not being enabled and added some tests for range support.

src/nodes.coffee
vendethiel: :+1:

src/lexer.coffee
vendethiel: Fair enough. It's still a bit weird to have this part disjointed from the others.

atg: Well `in` and `of` are reserved keywords, whereas `from` is a context-sensitive identifier. So the lexer can't handle them all in the same way.

```
$ coffee -te 'from'
[IDENTIFIER from] [TERMINATOR \n]

$ coffee -te 'in'
[RELATION in] [TERMINATOR \n]
```

I've merged the `if tag is 'IDENTIFIER'` conditionals into one, maybe that helps it look cleaner.

atg:
I'm actually starting to understand how `For#compileNode` works now. I've refactored my first try so that `@index` and `@name` are not flipped, and added support for patterns.

jashkenas:
I like this too. @lydell should we merge it?

lydell:
I get this error when running the tests:

```
$ npm run test-harmony

> coffee-script@1.11.0 test-harmony /home/lydell/forks/coffeescript
> node --harmony ./bin/cake test

failed 1 and passed 755 tests in 10.75 seconds 

  for-from comprehensions over generators 
  AssertionError: Expected  to deep equal 70,20
  ...
 
  function () {
    var array1, array2, array3, array4, gen, iterator, x;
    array1 = [50, 30, 70, 20];
    gen = function*() {
      return (yield* array1);
    };
    array2 = [];
    array3 = [];
    array4 = [];
    iterator = gen();
    for (x of iterator) {
      array2.push(x);
      if (x === 30) {
        break;
      }
    }
    for (x of iterator) {
      array3.push(x);
    }
    for (x of iterator) {
      array4.push(x);
    }
    arrayEq(array2, [50, 30]);
    arrayEq(array3, [70, 20]);
    return arrayEq(array4, []);
  }
```

lydell:
@atg I'd like to merge this too. Could you please look into the failing test whenever you get the time? :)

Inve1951:
To me this `iterator = gen();` looks wrong.
I guess it should be `iterator = gen;`.

Jamesernator:
> I'd like to merge this too. Could you please look into the failing test whenever you get the time? :)

I took a look at this, it seems like there's a bug in V8 with resuming iterators in a for-of loop (it did nothing with either Chrome or Node), it works fine on SpiderMonkey though.

atg:
Oops, didn't see the updates. I'll see if I can get this up to date.

atg:
@lydell OK the branch is conflict-free now.

The test is ok insofar as CoffeeScript is generating the right code. But the code doesn't run as expected.

As @Jamesernator said, the test works in Firefox but not in v8. So I tried in Safari, which agrees with **v8** that the test is wrong.

I dug into the spec, and it's difficult to understand, but from what I can determine my test & spidermonkey are wrong and v8/nitro are doing the right thing. [This is the relevant section](https://tc39.github.io/ecma262/#sec-runtime-semantics-forin-div-ofbodyevaluation-lhs-stmt-iterator-lhskind-labelset).

You can see that all paths lead to `IteratorClose()`. If you loop over a generator object once and break in the middle, then it will be closed forever. Personally I think this is a silly decision because it shuts the door on many cool uses of generators. But arguing with the spec is like arguing with a wall. ¯\_(ツ)_/¯

GeoffreyBooth:
@atg so does the test pass (in Node 6.9.1 or 7.0.0, with `--harmony`)? We can’t merge in a PR with broken tests.

atg:
I'll change the test to mark accept either the firefox or v8/nitro behaviour as a pass.

atg:
```
$ node --harmony ./bin/cake test
passed 771 tests in 9.13 seconds
```

test/compilation.coffee
```diff
@@ -123,3 +123,10 @@ test "#3001: `own` shouldn't be allowed in a `for`-`in` loop", ->
 
 test "#2994: single-line `if` requires `then`", ->
   cantCompile "if b else x"
+
+test "indexes are not supported in for-from loops", ->
```
GeoffreyBooth: Please put these tests in `error_messages.coffee`.

test/generators.coffee
```diff
@@ -250,3 +250,44 @@ test "yield handles 'this' correctly", ->
...
+    array3.push(x)
+  
+  for x from iterator
+    array4.push(x)
+  
+  arrayEq array2, [50, 30]
+  # Different JS engines have different opinions on the value of array3:
+  # https://github.com/jashkenas/coffeescript/pull/4306#issuecomment-257066877
+  # As a temporary measure, either result is accepted.
+  ok array3.length == 0 or array3.join(',') == '70,20'
```
GeoffreyBooth: Please always use `is` instead of `==`.

src/lexer.coffee
```diff
@@ -141,34 +141,38 @@ exports.Lexer = class Lexer
...
-          tag = 'RELATION'
-          if @value() is '!'
-            poppedToken = @tokens.pop()
-            id = '!' + id
-
-    if tag is 'IDENTIFIER' and id in RESERVED
-      @error "reserved word '#{id}'", length: id.length
-
+    if tag is 'IDENTIFIER'
+      if @seenFor and id is 'from'
```
GeoffreyBooth: Can this be refactored so that we don't branch this `if` block again? In other words, don't indent the following 15ish lines.

atg: Not sure. You can copy the `tag is 'IDENTIFIER'` into each of the ifs, but then the `else if` condition becomes comically long (see below).

I do agree that probably the if at the bottom of the big if (`if id in RESERVED`) should broken out and returned to the main body of the method.

```coffee
if tag is 'IDENTIFIER' and @seenFor and id is 'from'
  tag = 'FORFROM'
  @seenFor = no
else if tag is 'IDENTIFIER' and (id in JS_KEYWORDS or id in COFFEE_KEYWORDS) and
   not (@exportSpecifierList and id in COFFEE_KEYWORDS)
  tag = id.toUpperCase()
  if tag is 'WHEN' and @tag() in LINE_BREAK
    tag = 'LEADING_WHEN'
  else if tag is 'FOR'
    @seenFor = yes
  else if tag is 'UNLESS'
    tag = 'IF'
  else if tag is 'IMPORT'
    @seenImport = yes
  else if tag is 'EXPORT'
    @seenExport = yes
  else if tag in UNARY
    tag = 'UNARY'
  else if tag in RELATION
    if tag isnt 'INSTANCEOF' and @seenFor
      tag = 'FOR' + tag
      @seenFor = no
    else
      tag = 'RELATION'
      if @value() is '!'
        poppedToken = @tokens.pop()
        id = '!' + id

if tag is 'IDENTIFIER' and id in RESERVED
  @error "reserved word '#{id}'", length: id.length
```

GeoffreyBooth: https://github.com/atg/coffeescript/pull/1

atg:
Alright, do it or don't. I got places to be.

test/compilation.coffee
```diff
@@ -123,3 +123,6 @@ test "#3001: `own` shouldn't be allowed in a `for`-`in` loop", ->
 
 test "#2994: single-line `if` requires `then`", ->
   cantCompile "if b else x"
+
+test "own is not supported in for-from loops", ->
```
GeoffreyBooth: This should also go in `error_messages.coffee`.

GeoffreyBooth:
Closed per #4355

PR #4308, opened 2016-09-15
Be much more careful about parsing `*` in import and export statements

GeoffreyBooth:
While writing the documentation for modules, I stumbled across a bug. Yes, I broke it already.

I was writing this as an example:
```coffee
export square = (x) -> x * x
```

Which choked on the `*`, because our check for whether `*` should be treated as `EXPORT_ALL` was whether we’ve seen an `EXPORT` token and that the indent was 0. [I knew that indent check was going to be trouble](https://github.com/GeoffreyBooth/coffeescript/pull/2#issuecomment-241963249); but I gave it a lot more thought this time, and noticed two things:

- In `import` statements, an `*` always precedes an `as`.
- In `export` statements, the only time `*` can appear is in a statement like `export * from 'lib'`

So I made the check for converting `*` to `IMPORT_ALL` or `EXPORT_ALL` much narrower, based on these two rules. This made my example compile properly, and broke no other tests. I added two new tests like this example, where `*` is used on the same line (and indentation level) as an `export` keyword. I also re-ran the [Meteor todos test](https://github.com/GeoffreyBooth/coffeescript-modules-test-meteor-todos).

@lydell you were wondering why we can’t just put `*` in the grammar? It’s because the lexer converts `*` to tag `MATH`, and I don’t want to write grammar like `EXPORT MATH FROM String`. That doesn’t make much sense, and then in `ExportAllDeclaration` I would need to filter whether `MATH` was an `*` and throw an error otherwise. Better to intercept `*` in the lexer and convert it to `IMPORT_ALL` or `EXPORT_ALL` there rather than `MATH`.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -68,7 +68,7 @@
     };
 
     Lexer.prototype.identifierToken = function() {
-      var alias, colon, colonOffset, id, idLength, input, match, poppedToken, prev, ref2, ref3, ref4, ref5, ref6, tag, tagToken;
+      var alias, colon, colonOffset, id, idLength, input, match, poppedToken, prev, ref2, ref3, ref4, ref5, tag, tagToken;
       if (!(match = IDENTIFIER.exec(this.chunk))) {
         return 0;
       }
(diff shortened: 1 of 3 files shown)


lydell:
Nice catch!

Issue #3981, opened 2015-05-16
highlight.js github repo is missing on build


rwdim:
bower highlight.js#~8.0.0      ECMDERR Failed to execute "git ls-remote --tags --heads git://github.com/danielditgens/highlight.js.git", exit code of #128 fatal: remote error:    Repository not found.

michaelficarra:
What command are you running? It looks like you're installing dependencies through bower?

olleicua:
I got the same error:

```
$ git clone git@github.com:jashkenas/coffeescript.git

$ cd coffeescript

$ bower install
bower uglify-js#~2.2        not-cached git://github.com/mishoo/UglifyJS.git#~2.2
bower uglify-js#~2.2           resolve git://github.com/mishoo/UglifyJS.git#~2.2
bower highlight.js#~8.0.0   not-cached git://github.com/danielditgens/highlight.js.git#~8.0.0
bower highlight.js#~8.0.0      resolve git://github.com/danielditgens/highlight.js.git#~8.0.0
bower underscore#~1.5.2         cached git://github.com/jashkenas/underscore.git#1.5.2
bower underscore#~1.5.2       validate 1.5.2 against git://github.com/jashkenas/underscore.git#~1.5.2
bower highlight.js#~8.0.0      ECMDERR Failed to execute "git ls-remote --tags --heads git://github.com/danielditgens/highlight.js.git", exit code of #128 fatal: remote error:    Repository not found.

Additional error details:
fatal: remote error: 
  Repository not found.
```

Are there instructions for contributers somewhere that I'm missing?

GeoffreyBooth:
Is there a reason we have this `bower.json`? It doesn’t seem to be used.

GeoffreyBooth:
We’ve since removed the dependencies from `bower.json` and now `bower install` doesn’t throw any errors (though it also doesn’t do anything). The file is still there for _other_ tools to use the CoffeeScript module via Bower.

PR #4309, opened 2016-09-15
Modules documentation

GeoffreyBooth:
First draft. Relies on #4308.


GeoffreyBooth:
For whatever reason, the JavaScript syntax highlighting doesn’t treat `import` and `export` as keywords:

image

lydell:
Looks nice!

Did we discuss somewhere else something about documenting `--bare`+import/export stuff?

documentation/index.html.js
```diff
@@ -937,6 +938,24 @@ Expressions
...
+      <span id="modules" class="bookmark"></span>
+      <b class="header">Modules</b>
+      ES2015 modules are supported in CoffeeScript, with a very similar <code>import</code>
+      and <code>export</code> syntax:
+    </p>
+    <%= codeFor('modules') %>
+    <p>
+      Note that the CoffeeScript compiler <strong>does not resolve modules</strong>; writing an
+      <code>import</code> or <code>export</code> statement in CoffeeScript will produce an
+      <code>import</code> or <code>export</code> statement in the resulting output JavaScript.
```
vendethiel: resulting output JavaScript? I'd propose "resulting JavaScript" or "resulting output" or "JavaScript output" or...

src/lexer.coffee
```diff
@@ -115,7 +115,11 @@ exports.Lexer = class Lexer
     if id is 'from' and @tag() is 'YIELD'
       @token 'FROM', id
       return id.length
-    if id is 'as' and (@seenImport or @seenExport) and @tag() in ['IDENTIFIER', 'IMPORT_ALL', 'EXPORT_ALL']
+    if id is 'as' and @seenImport and (@tag() is 'IDENTIFIER' or @value() is '*')
```
vendethiel: not sure why the commit from #4308 is here...

vendethiel:
@lydell we just said it'd be nice to document it, IIRC.

src/lexer.coffee
GeoffreyBooth: It was necessary to get my example code to compile.

GeoffreyBooth:
@lydell, good catch. Updated.

I wonder if we really need to say anything else? I like the rather minimal documentation that lets the example speak for itself. Or should I link to something explaining what modules are, or to the MDN import or export pages, or anything like that?

Cakefile
```diff
@@ -54,6 +54,10 @@ codeFor = ->
     js = js.replace /^\/\/ generated.*?\n/i, ''
 
     cshtml = "<pre><code>#{hljs.highlight('coffeescript', cs).value}</code></pre>"
+    # Temporary fix until highlight.js adds support for newer CoffeeScript reserved words
```
GeoffreyBooth: @lydell how do you feel about this? How important is clear highlighting? I feel like we don’t want to wait until highlight.js updates its definition for CoffeeScript, and then we update the version of highlight.js we’re using.

bower.json
```diff
@@ -14,7 +14,7 @@
   "devDependencies": {
     "uglify-js": "~2.2",
     "jison": ">=0.2.0",
-    "highlight.js": "~8.0.0",
+    "highlight.js": "~9.6.0",
```
lydell: I'm confused. <https://github.com/jashkenas/coffeescript/issues/3981#issuecomment-247511095>:

> Is there a reason we have this bower.json? It doesn’t seem to be used.

documentation/index.html.js
```diff
@@ -937,6 +938,29 @@ Expressions
...
+      <a href="http://babeljs.io/">Babel</a>&nbsp;or
+      <a href="https://github.com/rollup/rollup">Rollup</a>, to convert this ES2015 syntax into
+      code that will work in your target runtimes.
+    </p>
+    <p>
+      Also note that any file with an <code>import</code> or <code>export</code> statement will
+      be output without a <a href="#lexical-scope">top-level function safety wrapper</a>;
+      in other words, importing or exporting modules will automatically trigger
+      <a href="#usage">bare</a> mode for that file. This is because per the ES2015 spec,
+      <code>import</code> or <code>export</code> statements must occur at the topmost scope.
```
lydell: I like that you link to other parts of the documentation. Here's a quote from the `#lexical-scope` section:

> If you'd like to create top-level variables for other scripts to use, attach them as properties on window, or on the exports object in CommonJS. The existential operator (covered below), gives you a reliable way to figure out where to add them; if you're targeting both CommonJS and the browser: `exports ? this`

That mentions even CommonJS – should it perhaps mention import/export as well?

Also, I'm having a hard time deciding where this `--bare` stuff fits best – in the `#modules` section, the `#lexical-scope` section, the `#usage section` (does the `coffee --help` output need to be updated as well?), all of them … ? What's your opinion?

Cakefile
lydell: I think it's fine special-case here. However, I think we should use the power of regex instead of a `for` loop ;)

```coffee
cshtml = cshtml.replace /import|export|from|as|default/g, '<span class="reserved">$&</span> '
```

index.html
```diff
@@ -159,10 +160,10 @@
 opposite = <span class="literal">true</span>;
 
 <span class="keyword">if</span> (opposite) {
-  number = -<span class="number">42</span>;
+  number = <span class="number">-42</span>;
 }
 
-square = <span class="function"><span class="keyword">function</span><span class="params">(x)</span> {</span>
+square = <span class="function"><span class="keyword">function</span>(<span class="params">x</span>) </span>{
```
lydell: Are all these little changing due to you updating highlight.js?

I think this PR will be easier to review if you leave out the compiled files. They will be committed when the 1.11.0 release is done.

lydell:
> I wonder if we really need to say anything else? I like the rather minimal documentation that lets the example speak for itself. Or should I link to something explaining what modules are, or to the MDN import or export pages, or anything like that?

I think it's good the way it is. You've matched the tone and style of the rest of the documentation nicely, I think.

src/lexer.coffee
connec: I suppose it can be rebased out now.

bower.json
vendethiel: The `bower.json` is (was) for people using Bower as part of their compilation steps. Not sure it's needed anymore.

documentation/index.html.js
vendethiel: `--help` seems fine to me.

Cakefile
GeoffreyBooth: They’re actually not the same. We need to search for the space after the keyword, or else `as` matches within `class="`. But if we replace including the space, we get `<span class="reserved">as </span>`. Maybe this could be done with an even cleverer regex, but I’m not sure it’s worth the effort.

documentation/index.html.js
GeoffreyBooth: I don’t understand `exports ? this`. Is that even CoffeeScript? Shouldn’t it be something like `exports ?= this`?

Cakefile
lydell: Ah, sorry, I didn’t see that space! This should do the trick:

```coffee
cshtml = cshtml.replace /(import|export|from|as|default) /g, '<span class="reserved">$1</span> '
```

documentation/index.html.js
lydell: I think this is what is hinted as:

> attach them as properties on window, or on the exports object

```coffee
# Browser globals only:
window.myExport = myExport

# CommonJS only:
exports.myExport = myExport

# Support either:
global = exports ? window
global.myExport = myExport

# `this` at the top level is the global object. Theoretically support more stuff:
global = exports ? this
global.myExport = myExport
```

Not sure if we should go down this sub-quest of improving that paragraph of documentation “while at it,” or just leave it be and be happy with the addition of mentioning `export` statements.

GeoffreyBooth: There’s no convenient way to support all three. You either have to know you’re using modules _or_ you’re targeting the browser/CommonJS. You can’t even do some kind of check as far as I’m aware, like “if `export` is supported...”. I guess just leave it as I’ve amended it, with the mention of `export`?

GeoffreyBooth:
Alright @lydell, anything else? I wish I could somehow mark your review as done. I think I might prefer the review style we were using in #4300, since when I would commit changes that addressed line notes, those comments would disappear from the stream. In this mode the comments stick around commit after commit, with no way to mark them as “closed” or addressed, as far as I can tell.

GeoffreyBooth:
Sweet! Is there anything else to do before cutting a release?

I was just noticing that the introduction at the top of the docs says that CoffeeScript “will work in every JavaScript runtime” . . .

lydell:
> Sweet! Is there anything else to do before cutting a release?

I don’t think so. Last Friday I started plowing through the commit log since the last release (that’s one year ago!) to make a change log, but I didn’t get very far.

> I was just noticing that the introduction at the top of the docs says that CoffeeScript “will work in every JavaScript runtime” . . .

Well, that’s not been true since `yield` was added …

jashkenas:
> Well, that’s not been true since yield was added …

Maybe it should be, though? I kind of like the notion of CoffeeScript 1.X being super-compatible just-works-anywhere — even in bizarre environments like Adobe Illustrator ... and 2.X being the "modern" flavor. Probably too late for that though.

GeoffreyBooth:
A bit late since we added generators. Modules I don’t think anyone would expect to work without further build-chain steps. Maybe we should just add an asterisk after “every JavaScript runtime,” linking to a footnote or tooltip like “(except for certain ES2015+ features, modules and generators, that are opt-in by using the feature)”.

PR #4311, opened 2016-09-20
[CS2] Output ES2015 arrow functions, default parameters, rest parameters

GeoffreyBooth:
**WIP**. Just thought I’d get the ball rolling and try to establish a precedent for merging into the `2` branch. I’ve only spent a few minutes on this, and I was surprised how easy it was to get `=>` to output as `=>`; although I get the sense that there are a lot of considerations to `=>` that I’m not considering. Please enlighten me.

Only two tests fail with this patch, but I’m not sure that they should pass:
```coffee
test "`new` works against bare function", ->
  eq Date, new ->
    eq this, new => this
    Date

test "#2009: don't touch `` `this` ``", ->
  nonceA = {}
  nonceB = {}
  fn = null
  (->
    fn = => this is nonceA and `this` is nonceB
  ).call nonceA
  ok fn.call nonceB
```

@lydell actually [commented](https://github.com/jashkenas/coffeescript/pull/4198#issuecomment-178812597) back when he wrote the second test that it would fail when we output `=>`. So maybe it’s time to remove that test?

As for the first one, I get the sense that the test was written expecting the arrow function to be wrapped by an IIFE. I presume that the point of outputting `=>` for `=>` is to get our output closer to its input, and therefore we shouldn’t be dropping unexpected wrapper functions into the output:

```bash
./bin/coffee -bpe 'foo = (bar) => console.log bar, @'
```
```js
var foo;

foo = (bar) => {
  return console.log(bar, this);
};
```

Is this correct? Or should this test pass somehow?

There must be lots of additional code scattered across the repo that can be removed now that we needn’t do so much work to parse `=>`. What cleanup needs to be done?


lydell:
```js
new (function () {}) // Valid JavaScript
new (() => {}) // Runtime error.
```

Both tests should be removed, in my opinion.

Are there no tests that include both fat arrows and parameter splats? I'd expect that to fail, since splats are implemented with `arguments`, which is not available in fat arrows.

```coffee
a = (args...) => args
```

connec:
> Are there no tests that include both fat arrows and parameter splats? I'd expect that to fail, since splats are implemented with arguments, which is not available in fat arrows.

There's probably a whole debate to be had about whether we really _want_ to output ES2015 arrow functions. I guess we do, but perhaps we should add special handling for `arguments`? Perhaps not since it's very 'magical'. But then users need to remember that there's a concrete difference between `->` and `=>` beyond the value of `this`.

connec:
Just another thought (forgive me if this is already handled in the PR): would we want to special-case single-JS-statement fat arrow functions and generate shorthand methods in that case?

```coffee
a = => console.log 'hello'
# var a = () => console.log('hello')

a = => console.log 'hello' ; console.log 'world'
# var a = () => {
#   console.log('hello')
#   console.log('world')
# }
```

lydell:
@connec `arguments` is not a problem. Here’s a lazy solution:

```js
a = (..._arguments) => {
  ...
}
```

connec:
OK, I guess I was just wondering aloud whether or not we should continue to deal with `arguments` in fat arrow functions or whether we should adopt ES6 behaviour (`arguments is undefined`).

E.g. what should this CS produce:

```coffee
() => arguments
```

Currently it compiles to:

```js
function() {
  return arguments;
}
```

With this change:

```js
() => {
  return arguments;
}
```

Ultimately, would we want to catch calls to `arguments` in bound functions and set them up to instead output:

```js
(..._arguments) => {
  return _arguments;
}
```

lydell:
I think it makes sense to simply disallow `arguments` in fat arrow functions.

bjmiller:
I vote no for outputting es6 fat arrows.  Original CS fat arrows are objectively better,  and there is no advantage to having es6 arrows in the compiled code.

Can everyone please stop trying to add es6 things to the compiled output just because they're new?

lydell:
@bjmiller I would be very interesting to hear what makes CS fat arrows better! :)

GeoffreyBooth:
Re the one-line syntax for arrow functions, I think we shouldn't bother. We don't output `if foo then bar()` or `a = { b: 1 }` as one-liners.

Re `arguments`, one of the goals of the 2.x effort is to align more closely with ECMAScript. Part of that involves no more sugar to “correct” ECMAScript’s oversights as we may perceive them, primarily because ES is evolving so fast and they might fix said oversight themselves soon enough and we don't want our polyfill to behave differently than their new syntax. In the case of `arguments` specifically, we should just follow the spec: if they left it out, presumably they have their reasons, and it's no more difficult in CoffeeScript to need to know that arrow functions lack `arguments` than it is to know the same in JavaScript.

atg:
My intuition of CoffeeScript functions is that `->` has a free `this` but `=>` has a bound `this`. That's it. It's kind of weird for `arguments` to depend on the whether this was bound or not. If you are going to ban `arguments` then ban it from both or neither.

lydell:
I wouldn't mind "banning" `arguments` altogether – in fact, I don't think I've _ever_ used in in CoffeeScript, since we have splat parameters.

atg:
I _personally_ would not mourn the loss of `arguments`, however removing it has a high potential to break existing code. ¯\\_(ツ)_/¯

My view is that it's desirable for `->` and `=>` to behave as identically as possible, and not to introduce too many surprises or gotchas. The current (ES3) implementation performs perfectly in this regard. It binds `this` and otherwise works exactly the same as `->`.

In some ways, ES6 arrows are worse. You can't `yield` inside an ES6 arrow, so CoffeeScript would have to fall back to the current implementation to continue to support that combination of features.

So I agree with @bjmiller there's really nothing wrong with the way things currently work, and it has the added benefit of being compatible with ES3 and ES5 as well.

GeoffreyBooth:
When would you ever put a `yield` inside an arrow function? [You can’t create a generator arrow function](http://stackoverflow.com/questions/27661306/can-i-use-es6s-arrow-function-syntax-with-generators-arrow-notation), either in CoffeeScript or in ECMAScript.

What’s “wrong” with current arrow functions is that they unnecessarily deviate from the ES2015 spec. Now that JavaScript has arrow functions, we should output arrow functions. Anything else is counterintuitive.

So assuming I’m correct about `yield`, the only breaking change by outputting a true `=>` is that people would no longer be able to reference `arguments` inside an arrow function. This isn’t even a loss; it just means people might have to refactor away from the magic `arguments` object into `(args...) =>`.

GeoffreyBooth:
If you want to create a bound generator function after this change, you would do so the same way in CoffeeScript as you would in ECMAScript: just declare `self = this` in the enclosing function, like the old days. This doesn’t strike me as terrible. Likewise, having to write `(args...) =>` instead of referencing the magic `arguments` object also doesn’t strike me as terrible.

What’s “wrong” with current arrow functions is that they unnecessarily deviate from the ES2015 spec. Now that JavaScript has arrow functions, we should output arrow functions. Anything else is counterintuitive.

lydell:
I was confused in my previous comments. `arguments` is not _disallowed_ in ES arrow functions. They refer to the outside `arguments` (just like `this`).

 ```
$ ./bin/coffee -e 'console.log(((a...) -> a) 1, 2, 3)'
[ 1, 2, 3 ]

$ ./bin/coffee -e 'console.log(((a...) => a) 1, 2, 3)'
[]
```

The second example returns an empty array, because of the `arguments` from CoffeeScript's IIFE wrapper.

I'm surprised that there were no tests for this! (Both examples above should of course log the same thing: `[1, 2, 3]`).

**Comment by GeoffreyBooth** [comment]
To complete @lydell’s example, here’s the ES2015 equivalent:
```js
console.log(((...a) => a)(1, 2, 3))
// [1, 2, 3]
```
So there _is_ a way to access a limitless number of arguments in an arrow function, so long as we make sure to compile `...` into ES2015 `...` in any parameters (e.g. `args...` to `...args`).

That just leaves the generator issue. Is it really so common to make a bound function generator? Is `self = this` in the outer scope not good enough for what I presume to be an uncommon use case?

GeoffreyBooth:
Okay @lydell, I did some work on splats/rest parameters. I could do a lot more to eliminate now-unnecessary code, but I just want to get this working first. Anyway, as of my latest commit your example:
```coffee
a = (args...) => args
```
becomes
```js
var a;

a = (...args) => {
  return args;
};
```

Since we’re not converting `args...` into ES5 code involving `arguments` anymore, we needn’t worry about `arguments` within arrow functions. This also fixes your other example:
```sh
./bin/coffee -e 'console.log(((a...) => a) 1, 2, 3)'
[ 1, 2, 3 ]
```

An interesting disadvantage of the ES2015 rest parameter syntax is that the rest parameter must be the _last_ parameter. So now we have two failing tests:
```
  /Users/Geoffrey/Dropbox/Sites/coffeescript/test/function_invocation.coffee:173
    mult = function(x, ...mids, y) {
                              ^
SyntaxError: Rest parameter must be last formal parameter

  /Users/Geoffrey/Dropbox/Sites/coffeescript/test/functions.coffee:124
    arrayEq([0, 1], (function(...splat, _, _1) {
                                      ^
SyntaxError: Rest parameter must be last formal parameter
```
That `SyntaxError` is thrown by Node. I guess this should just become part of our syntax as well that you can’t use `...` except in the last parameter?

lydell:
Yes, splats have to be the _last_ thing in ES. My first idea for handling that (rather than dropping support) was to do something like:

```coffee
f = (a, b, args..., c, d) ->
```
```js
f = function (a, b, c, d, ...args) {
  [c, d] = args.slice(-2);
  args = args.slice(0, -2);
};
```

Also, could you merge `master` into `2` and then rebase this branch on `2`? It’s impossible to review this PR now with so much crap in the diffs.

GeoffreyBooth:
The diff should be sensible now.

Your example doesn’t work for me:
```js
f = function (a, b, c, d, ...args) {
  [c, d] = args.slice(-2);
  args = args.slice(0, -2);
  console.log(a, b, args[0], c, d);
};

f(1, 2, 3, 4, 5)
// 1 2 undefined 5 undefined
```
`args`, at the start of the function, is just `[5]`. So `args.slice(-2)` returns `[5]`, thereby setting `d` to undefined. We can’t use `args` as a substitute for `arguments`.

We can, however, recreate `arguments` (this works for arrow functions too):
```js
f = function (a, b, c, d, ...args) {
  let arguments = [a, b, c, d].concat(args);
  [c, d] = arguments.slice(-2);
  args = arguments.slice(2, -2);
  console.log(a, b, ...args, c, d);
};

f(1, 2, 3, 4, 5)
// 1 2 3 4 5
f(1, 2, 3, 4, 5, 6, 7)
// 1 2 3 4 5 6 7
```

I guess the question is whether we should. This strikes me as a lot of extra JavaScript to be outputting. I suppose we should do it to prevent breaking backward compatibility, but it feels messy. It’s also so much more difficult to figure out: the `slice` lines require knowing how many arguments precede and follow the splat, so that you know the indices to slice by. The `[c, d]` part requires knowing which arguments follow the splat. It’s just a lot of processing, prone to edge cases I can’t imagine, for a feature that I can’t imagine people use all that often.

GeoffreyBooth:
Better version without intermediate variable:
```coffee
f = (a, b, args..., c, d) =>
  console.log a, b, args..., c, d
```
becomes:
```js
f = (...args) => {
  [a, b] = args.slice(0, 2);
  [c, d] = args.slice(-2);
  args = args.slice(2, -2);
  console.log(a, b, ...args, c, d);
}
```
And `f(1, 2, 3, 4, 5, 6, 7)` still prints `1 2 3 4 5 6 7`.

This is quite complicated to produce, so the question still stands: is this a feature we want to support?

Likewise with expanding parameters:
```coffee
f = (a, ..., b) =>
```

This isn’t even mentioned in the docs, though it’s in the grammar. Are we keeping this too?

lydell:
Sorry about that broken compilation example. It sure was a "something like." Here's a version that I've actually tested:

```coffee
var f;

f = function (a, b, c, d, ...args) {
  [c, d, ...args] = args.slice(-2).concat([c, d], args.slice(0, -2))
}
```

Regarding that it is more complicated to compile: Yes, it is. But at the same time, that's what we've been doing here for like half a decade now.

Regarding whether the feature is useful or not: I can't really think of a case (from the top of my mind) when I've used a splat parameter somewhere else than at the end. However, I _have_ used it in `[first, middle..., last] = someArray`.

Regarding `a, ..., b`: Yes, it is missing from the documentation. That's a shame. The feature is not very old (it was added in 1.7.0), and is useful for getting the last item of an array: `[..., last] = someArray`. That's used quite a bit in CoffeeScript's source code. When it comes to compiling, it's no different than `a, args..., b`, except that `args` won't be available for the user to use.

Regarding your suggested compilation: `(...args) =>` means that we loose `.length` of the function. That's true for todays' compiled output too, which is a shame. Redux preserves `.length`. I think we should, too, when we have the chance.

Summary:

- I've never missed non-last-splat-parameter when writing ES2015, but I _have_ missed it in array destructuring.
- I believe that dropping support for splats-anywhere-in-the-list will be a major backwards incompatibility hurdle for people.
- We already have lots of code for dealing with it.
- My suggested output is just one extra line of code.

lydell:
Here's an example with parameter defaults:

```js
// f = (a = 1, b = 2, args..., c = 3, d = 4) ->
var f;

f = function (a = 1, b = 2, c, d, ...args) {
  [c = 3, d = 4, ...args] = args.slice(-2).concat([c, d], args.slice(0, -2))
  console.log(a, b, args, c, d)
}
```

And here's an example with destructuring:

```js
// f = ({a = 1} = {}, b, args..., c, d) ->
var f;

f = function ({a = 1} = {}, b, c, d, ...args) {
  [c, {d = 4} = {}, ...args] = args.slice(-2).concat([c, d], args.slice(0, -2))
  console.log(a, b, args, c, d)
}
```

GeoffreyBooth:
I was only referring to function parameters. Limiting splats to the last argument in a function parameter list, the way ES does, doesn’t mean we would need to stop using splats in any position in destructuring.

Assuming splats elsewhere are untouched, are you still sure we want to support splats at any function parameter position? And expansions at any function parameter position?

lydell:
I've never been sure :)

Personally, I wouldn't mind dropping splats at any position in parameter lists. And dropping expansion in parameter lists.

The only "bad" thing is that array destructuring and "parameter list destructuring" would no longer be equivalent. The following would no longer be true:

```coffee
f = (<params>) ->
# equivalent to:
f = (args...) ->
  [<params>] = args
```

Couldn't we use the same code in both cases?

---

Another route we could take, is to:

- Remove expansion.
- Make splats work like in ES everywhere.
- Flip the `...` to the ES side (`...args` instead of `args...` – this regularly trips me up when switching between the languages ;) )

GeoffreyBooth:
One of our other goals is to make splats conform to the ES spec. Obviously we don’t need to and could just continue outputting what we do now, but if we want to output `...args` in the JS then we need to play by ES’ rules. They call `...` in destructuring “spread syntax” and as the last parameter of a function the “rest parameter.” They follow different rules.

And even now in CS it’s not the same code. Splats in parameters produce code involving the `arguments` object, which is obviously irrelevant to splats in destructuring. That’s why my current commit only breaks not-last-parameter argument splats, while all the destructuring tests still pass.

As for the flipping, I’ve been considering that too. It feels like a pointless breaking change, though. What if we simply allow it on either side?

lydell:
> They follow different rules.

They do?

> It feels like a pointless breaking change, though.

I thought since we were breaking compatibility anyway ...

> What if we simply allow it on either side?

Just need to be careful there ...

```coffee
f = (...args...) ->
```

GeoffreyBooth:
They follow different rules in that spread syntax can be at any position, like my `console.log(a, b, ...args, c, d)` example above.

Either side, not both sides. Allowing either side would be like we currently allow either `is` or `==`.

lydell:
I thought that you meant that there is a difference between `[...splat] = foo` and `(...splat) => {}` in ES.

vendethiel:
> Just need to be careful there ...

The parsing rules should not allow `Splat Splat` :)

GeoffreyBooth:
Why don’t we start a document where we track the breaking changes in 2? Perhaps a wiki page. And I think for now let’s just say that per the ES spec, splats in the parameters of function definitions must always be in the last parameter. And we can also make a note that links to this thread that says that if we want to add sugar to enable a splat in any parameter, we can do so.

Again, this only applies to the _definition_ of the function. Calling a function, like `console.log(a, b, ...args, c, d)`, can have a splat as any argument.

Anyway I think we should separate out that potential sugar from the rest of this feature. If we want to implement that, it should get its own PR. Ditto with implementing left-side splat (`...args`) as an additional syntax.

What do you think? If you agree, then the only work left in this PR is removing all the code that is now unnecessary because of the features implemented here; and rewriting or removing the tests that break because of function definitions that include splats in a non-final parameter.

lydell:
I don't really know what to think anymore. I guess we need input from more people.

Note to any commenter: Let's not talk about using splats for spreading out an array, such as in `[a, array..., b]` and `f a, array..., b`. Only about soaking up items into an array, such as in `[a, array...] = foo` and `(a, array...) ->`. Specifically:

- Parameter lists vs. array destructuring.
- Splats at any position vs. only last.
- Expansion (`[first, ..., last] = array` and `(first, ..., last) ->`).
- _Your_ use cases vs. current CoffeeScript vs. ES2015 vs. backwards compatibility.

jashkenas:
Reading through this, I think — although it's tempting to just break compatibility and go with the ESNext semantics — it would be worth outputting an extra bit of code to maintain compatibility, and more importantly, parallelism, here.

But I don't know the details about how hard it would be to maintain compatibility & destructuring/splatting parallelism while compiling into the new ESNext destructuring, so — the two options, the way I see them, are:

* Keep splats postfix `args...` CoffeeScript-style, keep compatibility and parallelism, and compile into as much ESNext as is convenient.

* Or move the ellipsis to ES-style `...args`, compile straight into the ES equivalent, and remove the CoffeeScript features that ES doesn't support.

Perhaps @lydell can choose between the two? If we want to break compatibility with respect to this feature, now is the time to do it.

GeoffreyBooth:
@lydell, the ES restriction _does not_ apply to destructuring. Look at the tests that fail right now: it’s code like `mult = (x, mids..., y) ->`, function definitions, not destructuring. ES doesn’t like splats in a non-last argument in a function definition. That’s all we’re concerned with right now. All the destructuring splat tests still pass.

lydell:
@GeoffreyBooth 

```
$ node -p 'let [...args, last] = [1, 2, 3, 4, 5]'
[eval]:1
let [...args, last] = [1, 2, 3, 4, 5]
     ^^^^^^^
SyntaxError: Rest element must be last element in array
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:556:32)
    at bootstrap_node.js:357:29
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

GeoffreyBooth:
What I meant was, we don’t need to restrict splats in destructuring just because we restrict them in function parameters. But I see your point, we should probably either fix it in both places or neither. It’s just that non-last splats are common in destructuring, but almost unheard-of in function parameters.

jashkenas:
> It’s just that non-last splats are common in destructuring, but almost unheard-of in function parameters.

Not necessarily:

```coffeescript
transformer = (args..., callback) ->
  doSomethingWith args, callback
```

vendethiel:
I don't see why the last element is any kind of thing. ES doesn't even have infinite lists to make this a reasonable restriction (like i.e. Perl 6 does).

I'm against adding such a specialcase into the language. Getting first/last element that way is useful. Slicing them off when you want to recurse useful as well

GeoffreyBooth:
Okay, okay, so we’ll polyfill this 😄

lydell:
```coffee
string.replace regex, (match, subMatches..., offset) ->
  return foobar
```

bjmiller:
I've been thinking on this one.  While I despise the idea of outputting arrow functions, I kind of like the idea of making splats work either on the left or on the right, in all cases.

ES6 was, in some part, written the way it was to make things harder for c2js languages, especially CS.  That's why they put the ellipsis on the left instead of the right - to make it deliberately harder to use CS.  The best way I can think to thwart that effort is to make splats work on either side.  It's backwards-compatible, and it removes the cognitive load of having to remember which way it goes, since it would always just work.

connec:
@bjmiller I think it's important to try and maintain a "one way" of things. Having multiple syntaxes for explicitly the same thing (e.i. it's not an indirect method such as `class` vs. `function` + `prototype`) feels to me like a confusing thing for the language.

I like the idea of emitting arrow functions in order to avoid CS having to deal with `this` binding (no more `_this`!), but I don't see the need to change any other semantics.

I also thing that part of the joy of CS comes from the parallelism of its features: if you can do something in a `->`'s arguments you can do it in a `=>`'s arguments and you can do it in destructuring. I like that there are few, if any, artificial limitations such as splats having to go at the end, and it would be a shame if that changed.

GeoffreyBooth:
So implementing this “splats for any parameter” requirement means implementing _all_ the ESNext parameter features, including parameter default values and destructuring. So it’s a fair bit of work. The key bit of code seems to be in `nodes.coffee`, in the `Code` class `compileNode` method. There’s a nice comment at the top of [that method](http://coffeescript.org/documentation/docs/nodes.html#section-123) that explains how it handles `=>` (now irrelevant) but then no other commentary. Does anyone who understands this stretch of code care to explain how it works? What does `param.isComplex()` mean and why does it matter? Should the code for compiling parameter default values (i.e. compiling `(a = 1) ->` into `function (a = 1) {}`) and destructured parameters happen here, or in `Param.compileNode` or somewhere else? Wherever it goes, is there a reason to compile it into `new Assign` rather than string JavaScript?

Also do we want to do all this in one pull request, or maybe treat this one as simply arrow functions (albeit with two broken tests) and have a separate PR implement the parameter-related features?

jashkenas:
> So implementing this “splats for any parameter” requirement means implementing all the ESNext parameter features, including parameter default values and destructuring. So it’s a fair bit of work.

Maybe it's not worth it then...



> param.isComplex()

IIRC, that was a flag to distinguish a simple param, like `a`, from a complex param that needs to be destructured, like `{name: a}`

GeoffreyBooth:
> > So implementing this “splats for any parameter” requirement means implementing all the ESNext parameter features, including parameter default values and destructuring. So it’s a fair bit of work.

> Maybe it's not worth it then...

We were planning on implementing those anyway; compiling parameter default values into parameter default values is definitely a win for legibility, I think. But it might be a big enough effort to do in its own PR; but that would mean either making this PR dependent on that one, or accepting this one with broken tests. So I dunno, maybe we rename this PR to be “ESNext functions” and it covers all the new function-related features.

GeoffreyBooth:
@lydell It appears you don’t need to repeat the default assignment inside the function body for it to be preserved:
```coffee
f = (a = 1, b = 2, args..., c = 3, d = 4) -> console.log a, b, args..., c, d
```
```js
var f,
  slice = [].slice;

f = function(a = 1, b = 2, c = 3, d = 4, ...args) {
  [c, d, ...args] = args.slice(-2).concat([c, d], args.slice(0, -2));

  return console.log.apply(console, [a, b].concat(slice.call(args), [c], [d]));
};
```
In Chrome console:
```js
f(1)
// 1 2 3 4
f(1, 2, 3, 4, 5, 6, 7)
// 1 2 3 4 5 6 7
```

lydell:
Nice! Didn't think about that. What about destructured parameters?

GeoffreyBooth:
But unfortunately your one-liner doesn’t work when the number of arguments after the splat is greater than the number of arguments contained by the splat:
```js
// f = (a, b, args..., c, d, e)
f = function(a, b, c, d, e, ...args) {
  [c, d, e, ...args] = args.slice(-3).concat([c, d, e], args.slice(0, -3));
  console.log(a, b, ...args, c, d, e);
};

f(1, 2, 3, 4, 5, 6, 7)
// 1 2 4 5 6 7 3
```
Here, `args.slice(-3)` is just `[6, 7]`—only two values, not three.

GeoffreyBooth:
@lydell I haven’t looked into destructured parameters yet. There’s still a long way to go to just get these splats working. Feel free to take a crack at it, or look at my last commit and tell me what I’m doing wrong so far.

src/nodes.coffee
```diff
@@ -1695,14 +1697,18 @@ exports.Code = class Code extends Base
...
+                                                                     [new Access new PropertyName 'slice']),
+                                                                     [
+                                                                        new NumberLiteral 0
+                                                                        (new Op '-',
+                                                                          new Value(
+                                                                            new IdentifierLiteral(splatParamName),
+                                                                            [new Access new PropertyName 'length']
+                                                                          ),
+                                                                          new NumberLiteral paramsAfterSplat)
+                                                                     ]
```
lydell: W00t crazy indentation, eh? ;)

GeoffreyBooth: My personal style is to just let lines run as long as they need to, and rely on word wrap in my editor. I'm not sure what to do in this codebase's style for a ridiculous line like this. Improvements welcome.

src/nodes.coffee
```diff
@@ -1610,70 +1611,129 @@ exports.Code = class Code extends Base
...
+    # function definition; and dealing with splats or expansions, including
+    # adding expressions to the function body to declare all parameter
+    # variables that would have been after the splat/expansion parameter.
+    for param, i in @params
+      # Was `...` used with this parameter? (Only one such parameter is allowed per function.)
+      # Splat/expansion parameters cannot have default values, so we need not worry about that.
+      if param.splat or param instanceof Expansion
+        haveSplatParam = yes
+        if param.name.value?
+          splatParamName = param.name.value
```
GeoffreyBooth: How do I safely get the name of the splat parameter in all cases? Including `@splat...`. I tried `param.asReference` but that triggers a `var` in the function body.

And I have no idea how to handle `[]...`.

src/nodes.coffee
lydell: Personally, I always try to avoid horizontal alignment, because no matter how pedantic I try to be, it always goes out-of-sync after a while :) It also creates larger diffs than necessary, which increases the risk of merge conflicts (and who likes those?). But that's just my preference. I guess this is one of those things that can rapidly turn into a flame war :) That being said, I haven't really seen this style in the CoffeeScript code base before.

GeoffreyBooth: I was following the style of grammar.coffee. I'm happy to pull it back left though, that would be my preference; I just thought this was the intended style.

GeoffreyBooth:
@lydell two questions:

1. Are we supporting splats like `[]...` and `{}...`? What should these become? Their direct ES equivalents (`...[]`, `...{}`) are syntax errors.
2. What code can we use to process the reassignment of variables when an expansion or non-final splat is involved? See my [comment](https://github.com/jashkenas/coffeescript/pull/4311#issuecomment-250074739) above on why our initial plan doesn’t work.

I’ve been messing with the second question for quite a while. I want to at least output in the function definition all the parameters that are to the left of the splat/expansion, and then in the body of the function only reassign the parameters to the right and the splat itself. But it’s tricky finding logic that works for all these cases:
```coffee
  withSplats = (a = 2, b..., c = 3, d = 5) -> a * (b.length + 1) * c * d
  eq 30, withSplats()
  eq 15, withSplats(1)
  eq  5, withSplats(1,1)
  eq  1, withSplats(1,1,1)
  eq  2, withSplats(1,1,1,1)
  eq  3, withSplats(1,1,1,1,1)
```
I guess one option for handling an expansion or a non-final splat is that the function parameters should just always become `(...args)` and we use the old logic for declaring all the parameter variables in the body. That gets around the issue of the lack of arrow-function-scope `arguments`; basically `...args` becomes that for us.

It would be nice, though, if the example above became `(a = 2, ...b)` rather than `(...args)`. Maybe a modified version of the current construction?
```js
a = arguments[0], b = 4 <= arguments.length ?
  slice.call(arguments, 1, i = arguments.length - 2) :
  (i = 1, []), c = arguments[i++], d = arguments[i++];
```

Kovensky:
Wait, why would `...[]`  be a syntax error? That's just a spread of empty iterator. `...{}` is only stage-3, though.

Kovensky:
> Only two tests fail with this patch, but I’m not sure that they should pass:

```coffee
test "`new` works against bare function", ->
  eq Date, new ->
    eq this, new => this
    Date

test "#2009: don't touch `` `this` ``", ->
  nonceA = {}
  nonceB = {}
  fn = null
  (->
    fn = => this is nonceA and `this` is nonceB
  ).call nonceA
  ok fn.call nonceB
```

The first test definitely should _not_ pass when outputting ES2015 arrow functions. Arrow functions cannot be called as constructors.

As for the second test, arrow functions (and bound functions) ignore any `this` override as given by further calls to `.bind`, or to `.call` / `.apply`; so both `this` and \`this\` will refer to the same value.

lydell:
```
$ node -p '((...[]) => 1337)()'
1337
```

GeoffreyBooth:
Okay, so following the approach I suggested in my last comment, this:
```coffee
withSplats = (a = 2, b..., c = 3, d = 5) -> a * (b.length + 1) * c * d
```
now compiles to this:
```js
var withSplats,
  slice = [].slice;

withSplats = function(a = 2, ...b) {
  var c, d, i, ref;
  ref = b, b = 3 <= ref.length ? slice.call(ref, 0, i = ref.length - 2) : (i = 0, []), c = ref[i++], d = ref[i++];
  if (c == null) {
    c = 3;
  }
  if (d == null) {
    d = 5;
  }
  return a * (b.length + 1) * c * d;
};
```
and the test it came from:
```coffee
test "default values with splatted arguments", ->
  withSplats = (a = 2, b..., c = 3, d = 5) -> a * (b.length + 1) * c * d
  eq 30, withSplats()
  eq 15, withSplats(1)
  eq  5, withSplats(1,1)
  eq  1, withSplats(1,1,1)
  eq  2, withSplats(1,1,1,1)
```
now passes.

There’s no need for `slice = [].slice`, since `b` here is always an array and will always have `slice`; but that’s an optimization for another day. Aside from that I think this is the best we can do for non-final splats.

GeoffreyBooth:
@lydell I’m working through the tests that are still failing. There are two categories of tests that I think _should_ fail, which we should probably remove.

The first category is tests that fail because they assume it should be possible to name a parameter `case`, which ES doesn’t allow. I’m not sure why CS should allow parameters to be reserved words.

#### 1:

```coffee
test "reserved keyword at-splat", ->
  f = (@case...) -> @case
```
```error
    f = function(...case) {
                    ^^^^
  SyntaxError: Unexpected token case
```

#### 2:

```coffee
test "duplicate formal parameters are prohibited", ->
  strictOk '(@case...,_case)->'
```
```error
  duplicate formal parameters are prohibited
  AssertionError: Got unwanted exception. (@case...,_case)->
  at Function.<anonymous> (coffeescript/test/strict.coffee:116:3)
```

The other category is tests that fail because when you pass in `null` as a value for a parameter that has a default value, ES honors the `null` rather than the default value. In other words, the ES code `((a = 1) => a)(null)` returns `null`, whereas its CS equivalent returns `1`.

I think ES’ behavior makes more sense here. If you want an ES equivalent to the old CS behavior, you can use `undefined`: ES `((a = 1) => a)(undefined)` returns `1`.

#### 3:

```coffee
test "default values", ->
  nonceA = {}
  nonceB = {}
  a = (_,_1,arg=nonceA) -> arg
  eq nonceA, a(0,0,null)
```
```error
  default values
  AssertionError: Expected [object Object] to equal null
  at Function.<anonymous> (coffeescript/test/functions.coffee:183:3)
```

#### 4:

```coffee
test "default values", ->
  nonceA = {}
  nonceB = {}
  b = (_,arg=nonceA,_1,_2) -> arg
  eq nonceA, b(0,null)
```
```error
  default values
  AssertionError: Expected [object Object] to equal null
  at Function.<anonymous> (coffeescript/test/functions.coffee:191:3)
```

#### 5:

```coffee
test "default values", ->
  nonceA = {}
  nonceB = {}
  c = (arg=nonceA,_,_1) -> arg
  eq nonceA, c(null)
```
```error
  default values
  AssertionError: Expected [object Object] to equal null
  at Function.<anonymous> (coffeescript/test/functions.coffee:199:3)
```

Any objection to removing these tests?

lydell:
**1 and 2:**

Imagine this chain of code editing:

```coffee
class Foo
  constructor: (_case) ->
    this.case = _case
```
↓
```coffee
class Foo
  constructor: (_case) ->
    @case = _case
```
↓
```coffee
class Foo
  constructor: (@case) ->
```
↓
“Why you no compile?!”

**3, 4 and 5:**

I had completely forgotten to think about that ES only uses the default for `undefined` in earlier discussions. Having learned parameter defaults in CoffeeScript, that regularly trips me up when writing ES2015 these days. I have this habit of using `null` everywhere and almost never `undefined`, and then I find myself with `TypError: foo is null` errors because parameters defaults somewhere didn’t do what I’m used to.

Part of me wants to say: “It what nice having this behavior in CoffeeScript, but it is time to move on to how the ES standard does things. That might also reduce confusion for people.”

But I also think about this way. Many people dislike that there’s both `undefined` and `null` in JavaScript. I big thing in CoffeeScript is that CoffeeScript always treats both of those the same. That’s quite nice. Imagine CoffeeScript being invented today. Saying that “CoffeeScript pretty much does away with the `undefined` _and_ `null` concept, letting you think of them as one” would still be a pretty good selling point, wouldn’t it?

So I’m torn. I don’t really know what I think.

Perhaps “treating `undefined` and `null` the same” makes the most sense for the `?` (and `?.` etc.) operator, and aren’t as important in parameter and destructuring defaults.

lydell:
Also, do not think that it is worth preserving `.length` of functions?

GeoffreyBooth:
Assuming we remove the above tests, there are only remaining failing tests per the current code:

#### 6:

```coffee
test "destructuring in function definition", ->
  context = {}
  (([{a: [b, c = 2], @d, e = 4}]...) ->
    eq 1, b
    eq 2, c
    eq @d, 3
    eq context.d, 3
    eq e, 4
  ).call context, {a: [1], d: 3}
```
```error
  test/functions.coffee:144:28: error: unexpected =
  (([{a: [b, c = 2], @d, e = 4}]...) ->
                           ^
```

#### 7:

```coffee
test "#4005: `([a = {}]..., b) ->` weirdness", ->
  fn = ([a = {}]..., b) -> [a, b]
  deepEqual fn(5), [{}, 5]
```
```error
  #4005: `([a = {}]..., b) ->` weirdness
  AssertionError: [ [], undefined ] deepEqual [ {}, 5 ]
  at Function.<anonymous> (coffeescript/test/functions.coffee:173:3)
```

Any ideas how to deal with these?

GeoffreyBooth:
Sorry, just saw your comment after I posted mine. Regarding `_case` to `case`, I don’t see why this would surprise: `case` is a reserved word and `_case` is not. The error is pretty direct: `SyntaxError: Unexpected token case`. Rather than letting the runtime catch it, we could catch it ourselves and explain in more detail. I guess the larger point is, is it really a “feature” of CoffeeScript that we let people use reserved keywords as function parameter names? Is this something worth preserving?

As for `null` vs. `undefined`, I had always thought of `foo?` becoming `foo != null` as a clever hack—taking advantage of `!=` instead of `!==` to let `null` stand in for all sorts of falsy values, including `undefined`. I never thought that CoffeeScript was trying to erase the concept of `undefined`. I occasionally write `if foo isnt undefined` because I specifically want to know if a variable is undefined rather than `null`, and I know that `?` will lump `undefined` and `null` together. I use `null` to represent when I want to say that a value exists but just happens to be null—like a form was submitted, and a field was left blank rather than being missing from the submission. Aside from `?` and function parameters, are there other places where CoffeeScript conflates the `undefined` and `null`?

If we want to preserve CoffeeScript’s behavior of conflating `null` and `undefined` with regard to function parameter defaults, then we can never output ES like `foo = function (a = 1) {}`. We would _always_ need to add a statement inside the function like we have now, just to set `a` to `1` if `a` is `null`.

As for function length, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) the property only counts parameters without defaults _until_ a rest parameter. That should mean that the current implementation actually returns the correct values for this. `(a, b, c..., d) ->` becomes `function (a, b, ...c) {}` and has a length of 2. Because I’m taking pains to preserve the parameters to the left of the splat/expansion, the function length should be as expected; which is better than the 1.x implementation, where `(a, b, c..., d) ->` becomes `function ()` with all the parameters defined in the body.

lydell:
case/_case: In `(@case) ->` there’s no variable called `case`. Only `@case` is available in scope. CoffeeScript compiles it to `(hiddenVariableNotAccessibleToTheUser) => this.case = hiddenVariableNotAccessibleToTheUser`.

Function length: “ccording to MDN the property only counts parameters without defaults until a rest parameter.” Of course. ES doesn’t allow parameters after the rest parameter.

GeoffreyBooth:
@lydell okay, I don’t see much harm in allowing reserved words for `@` parameters, and it’s not that big of a change. [Fixed](https://github.com/jashkenas/coffeescript/pull/4311/commits/a509f539ed5e67599db28d31fc5ad4bdbb2560f5).

I also [updated](https://github.com/jashkenas/coffeescript/pull/4311/commits/4f4e9512562b55369d8f009336589f930bdff450) the code that adds statements to the function body to set default parameters, now in _all_ cases. So `f = (a = 1) => a` now becomes:

```js
f = (a = 1) => {
  if (a == null) {
    a = 1;
  }
  return a;
};
```

The `if (a == null)` block is there _only_ for the edge case of someone calling `f(null)` and expecting to get `1` back (CoffeeScript 1.x behavior) and not `null` (ES2015 behavior).

I’m still not sure how I feel about this. On the one hand, why break backward compatibility if we don’t have to? And this keeps function definitions consistent with the existential operator. But on the other, this triggers a sizeable chunk of extra code _every_ time a parameter default value is specified, which is a pretty commonly-used feature. One of our goals with CS 2 is to reduce the amount of extra code added to the output and conform more closely to ES. Like you were saying, I wonder if it’s more confusing for people if `null` behaves differently in terms of how parameter default values behave in CoffeeScript versus ECMAScript. This seems to go against the “It’s just JavaScript” principle (though so does compiling `==` into `===`, I suppose) and would be one more thing people would need to learn to use CoffeeScript. Anyone else have an opinion? @jashkenas?

Whichever way we decide that question, this PR is getting close. I’m down to only two failing tests now, numbers 6 and 7 in [my comment above](https://github.com/jashkenas/coffeescript/pull/4311#issuecomment-252471603). Any ideas on those?

lydell:
- Regarding function length, I think it's fine to let `.length === 2` for `(a, b, c..., d) ->`.

- Regarding parameter defaults and undefined/null, I've started to lean against compiling to ES and dropping "null support".

Regarding "tests 6 and 7" – what in particular are you having problems with there?

GeoffreyBooth:
8e84792 fixes test 6. This was a tricky one. Basically, destructured parameters can be output in ES as-is; even complicated ones, like this:
```js
(function(...[
    {
      a: [b],
      c: c
    }
  ]) {
  eq(1, b);
  return eq(2, c);
})({
  a: [1],
  c: 2
});
```
Because of this rather good support for destructured parameters in ES, a lot of tests have been passing that I otherwise would’ve had to worry about 😄 The one that didn’t pass, though, had a destructured splat parameter that _wasn’t_ last.

So for non-destructured function definitions where the splat parameter wasn’t last, e.g. `f = (a, b..., c) =>`, basically the approach has been to output `(a, ...b)` and define `c` within the function body based on an element within `b`—basically treating `b` here the same way CS 1 used the `arguments` object, which we can’t use inside ES2015 arrow functions. Calling `f(1, 2, 3)` creates a `b` of `[2, 3]`, so an assignment like `b = b[0], c = b[1]` is possible. (It’s more complicated than that, to account for calling `f` with less than or greater than three arguments, but this is the basic idea.)

But I don’t get a parameter like `b` if I convert `f = ([a = {}]..., b) =>` to `f = (...[a = {}]) =>`. How would I define `b` from the `...[a = {}]` parameter? So for these cases, we need to use a placeholder variable instead: `f = (...arg) =>`, and then both parameters are defined within the function, exactly like the CS 1 approach except we use this `...arg` instead of `arguments`.

GeoffreyBooth:
And 91396cb fixes test 6. This kinda sucks. I basically just reverted to the CS 1 behavior for processing destructured splat parameters, letting `...arg` be the function parameter and leaving the destructuring to expressions in the function body. That sidesteps messiness with `this` or default values in the destructured parameters. A better solution would investigate the destructured object to see if it has anything we can’t simply pass through to ES, and only in those rare cases declare it in the function body. That could be a future challenge. As of now all the tests pass (aside from unrelated #4336).

So . . . I guess let the review begin?

nickdima:
Will compiling to ES6 arrow functions be optional? Just trying to figure out how it would work for older JS environments.

GeoffreyBooth:
This is intended for a 2.x version of CoffeeScript that targets modern ECMAScript. The 1.x line will remain more or less as is for legacy support.

nickdima:
Is there a written explanation somewhere about why we're going down this path?
One of the compelling "features" of CS was that it would compile to a very wide used version of JS.
Throughout the years I used CS not only in the browser but also in more "exotic" environments like QuartzComposer or MaxMSP which usually use old JS versions.
I'm not complaining here, just trying to figure out what the future will look like. :)

carlmathisen:
@nickdima Have a look at https://github.com/coffeescript6/discuss for an in-depth explanation of what is going on. The discussions over there are very much alive and you will probably find some topics of interest.

jashkenas:
> Is there a written explanation somewhere about why we're going down this path?
> One of the compelling "features" of CS was that it would compile to a very wide used version of JS.
> Throughout the years I used CS not only in the browser but also in more "exotic" environments like QuartzComposer or MaxMSP which usually use old JS versions.
> I'm not complaining here, just trying to figure out what the future will look like. :)

@nickdima — In short, that's what CoffeeScript 1.X is for. You can continue using it anywhere old-fashioned JavaScript is still served.

test/classes.coffee
```diff
@@ -471,13 +471,6 @@ test "`new` shouldn't add extra parens", ->
   ok new Date().constructor is Date
 
 
-test "`new` works against bare function", ->
-
-  eq Date, new ->
-    eq this, new => this
```
connec: We could choose to remove this line only, and continue to test `new -> ...`.

Optionally we could also add a test that `new => ...` fails.

GeoffreyBooth: So the 1.x output for this is:
```js
eq(Date, new function() {
  eq(this, new ((function(_this) {
    return function() {
      return _this;
    };
  })(this)));
  return Date;
});
```
The output per this PR is:
```js
eq(Date, new function() {
  eq(this, new (() => {
    return this;
  }));
  return Date;
});
```
The latter block doesn’t throw any compilation errors per the Babel REPL, but throws `TypeError: (intermediate value) is not a constructor` in the Node runtime. See [MDN: TypeError: "x" is not a constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Not_a_constructor). Per [MDN: Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), arrow functions “can not be used as constructors.” (Not to educate you @connec, since you obviously know this, but just for others who might find this thread.)

So if by definition you can’t use `new` with an arrow function, I don’t think we need to write a test that proves that our runtime enforces JavaScript rules. One could argue that we should throw an error in the compiler if someone tries `new =>`, but we don’t currently check or throw errors for all the prohibited constructors, like `new Math`.

There’s nothing wrong with keeping just the first line though, and the last `Date` line. That at least preserves the test that `new` works against regular functions. I’ll restore that.

src/nodes.coffee
```diff
@@ -1611,70 +1611,133 @@ exports.Code = class Code extends Base
...
+        else if param instanceof Expansion and @params.length is 1
+          param.error 'an expansion parameter cannot be the only parameter in a function definition'
+
+        haveSplatParam = yes
+        if param.splat
+          params.push ref = param.asReference o
+          splatParamName = fragmentsToText ref.compileNode o
+          if param.isComplex() # Parameter is destructured or attached to `this`
+            exprs.push new Assign new Value(param.name), ref, '=', param: yes
+            # TODO: output destrucutred parameters as is, *unless* they contain
```
connec: I was messing about yesterday with compiling destructuring to ES2015 in all cases, and discovered that `this` parameters can be compiled quite neatly when destructuring:

```coffee
{ @a } = b
[ @a ] = b
```

This can become:

```js
{ a: this.a } = b
[ this.a ] = b
```

lydell: Wow, I had no idea that was valid JS! Thanks for enlightening us!

GeoffreyBooth: Unfortunately it doesn’t appear to be allowed in function parameters, [per Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&code=f%20%3D%20(this.a)%20%3D%3E%20a%3B):
```
repl: Invalid left-hand side in arrow function parameters (1:5)
> 1 | f = (this.a) => a;
    |      ^
```
or
```
repl: Binding member expression (1:5)
> 1 | f = (this.a = a) => a;
    |      ^
```

connec: Yes, it seems to be restricted in destructuring too (node v6.8.1):

```js
f = ({ a: this.a }) => {}
          ^^^^
SyntaxError: Unexpected token this
```

GeoffreyBooth:
@connec did you have any other notes? Should we merge this into `2`?

lydell:
I would still like to see tests for https://github.com/jashkenas/coffeescript/pull/4311#issuecomment-248517199 (did you get my email?). :)

GeoffreyBooth:
@lydell ah, sorry, I forgot about that 😄 Will do soon.

lydell:
```
$ ./bin/coffee -bpe '(a=fn()) ->'
(function(a = fn()) {
  if (a == null) {
    a = fn();
  }
});
```

I'm a bit unsure if it makes sense to use ES parameter defaults if we're doing a manual check afterwards anyway. In this example, if `a` is `undefined` and `fn()` returns `undefined` or `null` the first time it is called, `fn()` will end up being called twice. I'm not sure, but perhaps it makes more sense to _either_ use ES param defaults _only_ (dropping "`null` support") or not using ES param defaults at all and keeping the "manual defaults" code.

GeoffreyBooth:
You’re right, it should do one or the other, but not both. I guess for the best backward compatibility it should always do the `if (a == null)` block . . . but then we’re not really outputting ES at all for a feature that should be ES. I guess it’s a question of whether one of CoffeeScript’s improvements on JavaScript is conflating `undefined` and `null`, or if we’re going to walk away from that (excepting the existential operator). I guess the question could be, does anyone care how clean the output is?

GeoffreyBooth:
Okay, we now follow the ES spec for applying default parameters only when a parameter is missing or `undefined`. I updated the tests to reflect this.

Should we add any additional tests to make sure things are working properly, especially when a splat or expansion is not the last parameter?

lydell:
Well, now things are confusing if you use both parameter defaults and destructuring defaults at the same time.

```
$ ./bin/coffee -bpe '(a = 1, [b = 2]) ->'
(function(a = 1, arg) {
  var b, ref;
  b = (ref = arg[0]) != null ? ref : 2;
});
```

Perhaps it doesn't make sense to land this PR without re-doing destructuring at the same time?

> Should we add any additional tests to make sure things are working properly, especially when a splat or expansion is not the last parameter?

Sounds like a good idea.

GeoffreyBooth:
Destructuring will need to stay in the function body unless we want to search for `this` inside the destructured parameter, and only if `this` is found move it inside. But it will definitely need to go inside if `this` appears.

There’s no need to do everything in one PR. The handling of destructured parameters in this PR works. When destructuring in general is refactored, yes, we can improve the handling of destructured parameters. But that shouldn’t hold up this PR.

Regarding tests, I was hoping others might have ideas for some. :)

lydell:
Regarding destructuring, I wasn't talking about what the generated JS looks like. I was talking about the difference in `null` handling.

Regarding tests, it sounded like you already had a few test cases in mind but hesitated adding them :) Guess we're fine then.

connec:
> I guess it’s a question of whether one of CoffeeScript’s improvements on JavaScript is conflating undefined and null

So we've decided that's not an improvement? I would rather lose ES2015's defaults and keep existence checks consistent. I really don't like the idea of defaults behaving differently to `?`, `?=`, etc. (though `?()` gives precedent for that...).

To illustrate a problem with this, currently you can use argument defaults to make values in functions 'safe' for existence checks:

```coffee
action = (options = {}) ->
  # We can safely call options.whatever here, since we know it's not undefined or null
  options.whatever
```

Admittedly this is also an issue with ES2015, but it's one that CoffeeScript *solves* by treating `null` and `undefined` as non-existent values, for existence checks and defaults.

lydell:
This might possibly be a very bad idea, but what if we made `(a = 1) ->` use ES’ semantics, and `(a ?= 1) ->` use CS’ semantics?

vendethiel:
Sounds like a recipe for disaster and unteachability.

GeoffreyBooth:
Personally I always thought parameter default values only applied when an argument was missing (which I guess translates to `undefined`). I'm not bothered by the ES behavior. There’s no `?` in there, I wouldn't have expected parity with the existential operator.

As for destructuring, currently this PR doesn't allow any destructuring in the parameters. If it encounters a destructured parameter, it defines that parameter in the function body, just like in 1.x. We could simply keep this behavior, and then we’ll never need worry about destructuring behaving differently in function parameters.

As for the `options = {}` example, it’s only an issue if you ever call that function with `null` for that parameter. I don’t see why that would commonly happen. The CoffeeScript codebase it littered with `options = {}`, but I didn’t need to refactor anything as a result of this PR.

GeoffreyBooth:
@connec @jashkenas @lydell Not to short-circuit the discussion in https://github.com/coffeescript6/discuss/issues/51, but can we merge in this PR as is? And if we ultimately decide to go the other way regarding `null`, I can submit another PR. I don’t feel strongly either way on that question, but I want to get this change merged in so that the classes PR can start using it.

lydell:
If merging this helps you who actually work on the 2 branch, then go for it, I'd say. The 2 branch doesn't need be be as stable as master.

GeoffreyBooth:
So I just tried running the tests in this branch under `--harmony` mode, and I got a surprise. This test fails only in harmony (`generators.coffee:55`):
```coffee
test "bound generator", ->
  obj =
    bound: ->
      do =>
        yield this
    unbound: ->
      do ->
        yield this
    nested: ->
      do =>
        yield do =>
          yield do =>
            yield this
```
It only fails in harmony mode because we’re not running `generators.coffee` except in harmony mode, because back when generators were added to CoffeeScript that’s the only way they could be run. Now with Node 6 the `--harmony` is unnecessary, and will be removed as soon as #3757 lands.

Anyway the reason this test fails is because of the `yield this` inside a bound function. For a simpler example, this:
```coffee
f = => yield this
```
compiles (via this PR) to this:
```js
f = () => {
  return (yield this);
};
```
which Node doesn’t like:
```js
  return (yield this);
                ^^^^
SyntaxError: Unexpected token this
```
because [arrow functions cannot be generators](http://stackoverflow.com/questions/27661306/can-i-use-es6s-arrow-function-syntax-with-generators-arrow-notation). So this is another breaking change as a result of this PR. And I need to find a way to check that `yield` is inside a bound function as opposed to a regular function, and throw an error in the former case.

The workaround, ironically enough, is to just use a regular function and the old `self = this` hack:
```coffee
self = this
f = -> yield self
```

GeoffreyBooth:
Fixed. Everyone okay with this change? I suppose the alternative is to fall back to the CS 1.x output for bound functions in the event of a bound generator function, but I feel like that will cause more debugging trouble than it’s worth, by potentially introducing inconsistencies between bound functions with and without `yield`. Hopefully bound generator functions aren’t so common, and we’re only aligning ourselves with ES here by forcing people to do the `self = this` rigamarole for bound generator functions. I’ve added this to the [breaking changes wiki document](https://github.com/jashkenas/coffeescript/wiki/%5BWIP%5D-Breaking-changes-in-CoffeeScript-2).

@lydell, others, is there anything I’m missing here? If not, please let me know and I’ll merge this PR into `2`.

test/error_messages.coffee
```diff
@@ -1142,3 +1142,10 @@ test "imported members cannot be reassigned", ->
     export foo = 'bar'
            ^^^
   '''
+
+test "bound functions cannot be generators", ->
+  assertErrorFormat 'f = => yield this', '''
+    [stdin]:1:5: error: bound (fat arrow) functions cannot contain 'yield'
+    f = => yield this
+        ^^^^^^^^^^^^^
```
lydell: What will this look like if the function is multi-line? Is it better to underline `yield` only?

GeoffreyBooth: It just underlines whatever the first line of the function is, starting at the `=>`:
```
./bin/coffee -bpe 'f = =>
quote>   1
quote>   2
quote>   yield this'
[stdin]:1:5: error: bound (fat arrow) functions cannot contain 'yield'
f = =>
    ^^
```

GeoffreyBooth:
Updated. Now it only underlines the `yield` line (assuming I understand what `o.scope.method` refers to):
```
./bin/coffee -bpe 'f = =>
  1
  2
  yield this'
[stdin]:4:3: error: yield cannot occur inside bound (fat arrow) functions
  yield this
  ^^^^^^^^^^
```

GeoffreyBooth:
@lydell any other notes?

lydell:
No :=

PR #4312, opened 2016-09-20
Ensure indentation is consistent

eelco:
This is a follow up of #4303, but takes more or less the opposite approach:

This is a proposal to make the parser more strict, and ensure that indentation always matches the literal indentation that came before it.

It makes this code not parse anymore:
```
new Layer
 x: 0
	y: 1
```
(2nd line uses a space, 3rd line a tab, this is currently legal code.)

This prevents using a different style of indentation (e.g. mixing spaces and tabs) in the same ‘block’ of code (where indentation > 0).

Mixing is still allowed if each line uses the same mix and if the indentation level returns to 0. (Each block can use it own style.)

(Breaks the literate coffeescript test that mixes spaces and tabs.)

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -40,6 +40,7 @@ exports.Lexer = class Lexer
     @indebt     = 0              # The over-indentation at the current level.
     @outdebt    = 0              # The under-outdentation at the current level.
     @indents    = []             # The stack of all current indentation levels.
+    @indentLiteral = ''          # The indentation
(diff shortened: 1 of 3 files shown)


eelco:
Also see the follow-up #4313, which is a even more strict (and IMHO better) proposal to prevent (accidental) whitespace mixing.

jashkenas:
Cool. If we're going to consider this, we should consider the strict version of it. Closing in favor of #4313.

PR #4318, opened 2016-09-26
Pass through octal and binary literals as-is

JimPanic:
See https://github.com/coffeescript6/discuss/issues/45

Note: I'm deliberately creating this PR towards branch 2 because it is not opt-in and I'm not sure how many current JS environments implement the binary and octal literals as ES2015 does.

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -211,9 +211,8 @@ exports.Lexer = class Lexer
       when 'o' then 8
       when 'x' then 16
       else null
+
     numberValue = if base? then parseInt(number[2..], base) else parseFloat(number)
-    if number.charAt(1) in ['b', 'o']
-      number = "0x#{numberValue.toString 16}"
 
     tag = if numberValue is Infinity then 'INFINITY' else 'NUMBER'
     @token tag, number, 0, lexedLength


src/lexer.coffee
```diff
@@ -202,14 +204,12 @@ exports.Lexer = class Lexer
...
-      numberValue = parseInt(octalLiteral[1], 8)
-      number = "0x#{numberValue.toString 16}"
-    else if binaryLiteral = /^0b([01]+)/.exec number
-      numberValue = parseInt(binaryLiteral[1], 2)
-      number = "0x#{numberValue.toString 16}"
-    else
+
+    numberValue = number
+
+    unless /^(0o[0-7]+)/.test(number) and /^(0b[01]+)/.test(number)
```
lydell: Isn’t this a constant condition (`unless false`)?

test/numbers.coffee
```diff
@@ -13,9 +13,26 @@
 # Binary Integer Literals
 # Binary notation is understood as would be decimal notation.
 
+toJS = (str) ->
+  CoffeeScript.compile str, bare: yes
+  .replace /^\s+|\s+$/g, '' # Trim leading/trailing whitespace
+
```
lydell: Following the style of all other tests (except import/export), all these added tests should be removed.

JimPanic: Oh, I didn't know that. So no compilation tests at all?

src/lexer.coffee
JimPanic: Good catch, thanks.

test/numbers.coffee
lydell: Yeah, that’s been the rule. Only check that the CoffeeScript does what is expected.

JimPanic: Ok, cool. 👍 I like removing things.

JimPanic:
Removed the tests and fixed the logic.

lydell:
@JimPanic While reviewing this PR, I found a bug in master. I fixed that bug, and have also updated the 2 branch with the fix. Could you please rebase this PR?

JimPanic:
Just to be clear: is there a reason to parse hex, oct and binary in `2`?

JimPanic:
Oh, of course they need to be parsed - for the Infinity check.

lydell:
Yep. Btw, you seem to have completely messed up the commits now :) git is so much fun

JimPanic:
Yes, totally. And your fix confused me. :) Basically you refactored it so well, I just have to remove two lines again. So this will be a new force push to my branch; on top of your last commit on 2.

lydell:
Nice! LGTM. @GeoffreyBooth what about you?

GeoffreyBooth:
Is this really just +1 −2 changes? It doesn’t get much simpler than that. Looks good to me assuming we don’t need any tests (?)

lydell:
I've missed compiled output tests from time to time. For example, when reviewing this PR, I found an output bug that still gave the same runtime result. It would be nice to write tests for such things.

JimPanic:
Hence why I added the tests in the first place, but I didn't know there was this rule.

We could have compilation tests in a seperate folder/repository maybe?

lydell:
Compilation tests shouldn't be a blocker for this PR, though? That's a separate issue?

JimPanic:
Yeah, that's definitely something for a different issue.

PR #4319, opened 2016-09-26
Fix REPL tests by trimming the output

JimPanic:
I had problems running the REPL tests on my machine because `output.lastWrite()` returns the last line with a `\r` in it, causing `eq` to fail with error messages like `AssertionError: Expected 2 to equal 2`.

### DIFF
--- a/test/repl.coffee
+++ b/test/repl.coffee
@@ -42,31 +42,31 @@ ctrlV = { ctrl: true, name: 'v'}
 
 testRepl 'reads history file', (input, output, repl) ->
   input.emitLine repl.rli.history[0]
-  eq '3', output.lastWrite()
+  eq '3', output.lastWrite().trim()
 
 testRepl "starts with coffee prompt", (input, output) ->
   eq 'coffee> ', output.lastWrite(0)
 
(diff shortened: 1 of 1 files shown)


lydell:
That does not happen for me when I run the tests.

OS: Ubuntu 16.04
Node.js: 6.5.0

JimPanic:
Here it's:

OS X 10.10
Node 6.4.0 (and I think 6.6.0 as well not sure though; didn't test 6.5.0)

Gonna test tomorrow, when I'm at the work machine again.

JimPanic:
Okay, I couldn't reproduce it with Node 6.6.0, and if 6.5.0 works for you as well, this fix is not necessary. Thanks!

shreeve:
I just upgraded to node 6.7.0 and this PR is not required.

Perhaps it was a transient thing that only impacted node versions on Mac between 6.4.x and 6.6.x?

Having said that, this small fix might help  people on those versions and shouldn't hurt anything otherwise. It just just does this:

```diff
-    @written[@written.length - 1 + fromEnd].replace /\n$/, ''
+    @written[@written.length - 1 + fromEnd].replace /\r?\n$/, ''
```

PR #4329, opened 2016-10-02
strip \r (if present) before final \n

shreeve:
Fixes 'cake test', which was failing due to "\r\n" terminators

### DIFF
--- a/test/repl.coffee
+++ b/test/repl.coffee
@@ -25,7 +25,7 @@ class MockOutputStream extends Stream
     @written.push data
 
   lastWrite: (fromEnd = -1) ->
-    @written[@written.length - 1 + fromEnd].replace /\n$/, ''
+    @written[@written.length - 1 + fromEnd].replace /\r?\n$/, ''
 
 # Create a dummy history file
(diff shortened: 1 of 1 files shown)


lydell:
This sounds like #4319. Could you please look through the comments over there, and then say if this PR is still relevant or not?

shreeve:
Thanks for the reference to #4319, and yes this is the same issue. It appears to affect Mac OS only.

This PR has a better fix than the one from #4319 and should safely close both issues.

lydell:
So the Node.js version doesn’t matter for you?

shreeve:
I just upgraded to node 6.7.0 and this PR is not required.

Perhaps it was a transient thing that only impacted node versions on Mac between 6.4.x and 6.6.x?

Having said that, this small fix might help  people on those versions and shouldn't hurt anything otherwise. It just just does this:

```diff
-    @written[@written.length - 1 + fromEnd].replace /\n$/, ''
+    @written[@written.length - 1 + fromEnd].replace /\r?\n$/, ''
```

Issue #4233, opened 2016-03-23
Extending an ES6 class with a CS class fails


bcherny:
CoffeeScript calls classes as functions (rather than newing them as constructors) when instantiating a class that inherits from an ES6 class.

The fix is to interoperate with ES6 classes - no need to go all the way and compile Coffee to ES6 classes. We should check to see if a class is an ES6 class, and treat it differently (eg. [see what angular does here](https://github.com/angular/angular.js/commit/8955cfb6462f79a32caa641ffc002f1522f08220#diff-ced481bc839279a343e9f8bb8dcf6e98R824)).

Test case:

js:

``` js
class A {}
```

coffee:

``` coffee
class B extends A {}
new B # TypeError: Class constructor A cannot be invoked without 'new'
```

lydell:
(Actually, you’re doing `A({})` (that is, invoking `A` without `new`) in the _first_ line. But you’re issue still stands if that accidental `{}` is removed.)

bcherny:
This is still a major blocker for us - any way you guys can push a patch to interoperate with classes??

michaelficarra:
@bcherny Someone would have to suggest a fix first. I don't believe that there's going to be a way to do this. It could be done with `Function.prototype.bind`, but we can't use that in CoffeeScript's target language.

bcherny:
I see what you're saying, what a bummer. What about an opt-in compiler flag, something like `--allow-bind` so this kind of interop is supported?

michaelficarra:
The CoffeeScript maintainers have rejected compilation target options many times in the past, but it may be accepted if that's the only way to do something.

vendethiel:
I think an ES6 compilation target was almost-agreed upon?

bcherny:
What's the compile target today? It seems that CS already makes exceptions for generators (ES6). This one special case for interoperable classes would be a huge win for my team!

carlsmith:
I'm not certain on the details, but I think generators are different, because there, you opt in to ES6 dependance by using generators. Generators compile to ES6 code. Classes are different, as people use CoffeeScript classes in code that is expected to work in old browsers, so you'd need a flag or something.

albertywu:
:+1:

connec:
You could use the same trick angular uses to determine whether or not to use `bind`:

```js
...
function B () {
  if (/^class /.test(A.__super__.constructor)) {
    // A extends an ES6 class so can't be called directly
[...]

PR #4330, opened 2016-10-03
[wip] Compile classes to ES2015

connec:
I started looking at adding some tests for executable class bodies to check for regressions when we begin to compile ES2015 classes, and I ended up writing an implementation. Sorry if this clashes with anyone else's work for https://github.com/coffeescript6/discuss/issues/22.

This is some unfinished, work in progress changes to compile CoffeeScript classes into ES2015 classes. I've only committed the source files, so if y ou want to run it you'll need to `cake build && cake build:parser`.

The compiler successfully self hosts, however **THERE ARE DEFINITELY A LOT OF BUGS AND THE CODE IS NOT PRETTY**. It might be useful as prior art and/or to resolve some discussions on how CS should proceed with ES2015 classes.

I tried not to remove any CS functionality, including executable class bodies, soaked super invocations, etc. Quite a bit of code could be removed/cleaned up if they were removed.

I'll add some comments in-line, and later update this with a TL;DR of the significant issues.

See https://github.com/coffeescript6/discuss/issues/22 for more context.

Top issues:
- Dynamic method names referencing variables that change in the class body is broken.
- Dynamic method names need to be cached.
- Super referencing is convoluted.
- Class method hoisting is hacky.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -225,6 +225,7 @@ exports.Base = class Base
 # `if`, `switch`, or `try`, and so on...
 exports.Block = class Block extends Base
   constructor: (nodes) ->
+    super()
     @expressions = compact flatten nodes or []
 
   children: ['expressions']
@@ -497,6 +498,8 @@ exports.YieldReturn = class YieldReturn extends Return
 # or vanilla.
 exports.Value = class Value extends Base
   constructor: (base, props, tag) ->
+    super()
+
     return base if not props and base instanceof Value
     @base       = base
     @properties = props or []
@@ -547,10 +550,6 @@ exports.Value = class Value extends Base
   isSplice: ->
     [..., lastProp] = @properties
     lastProp instanceof Slice
-
-  looksStatic: (className) ->
-    @base.value is className and @properties.length is 1 and
-      @properties[0].name?.value isnt 'prototype'
 
   # The value can be unwrapped as its inner node, if there are no attached
   # properties.
@@ -687,12 +686,20 @@ exports.Call = class Call extends Base
 
     fragments = []
     if this instanceof SuperCall
-      preface = @superReference(o) + ".call(#{@superThis(o)}"
-      if compiledArgs.length then preface += ", "
+      preface = @superReference o
+      if preface isnt 'super'
+        preface += ".call(#{@superThis o}"
+        preface += ', ' if compiledArgs.length
+      else
+        preface += '('
       fragments.push @makeCode preface
     else
-      if @isNew then fragments.push @makeCode 'new '
+      complex = @variable.isComplex()
+      if @isNew
+        fragments.push @makeCode 'new '
+        fragments.push @makeCode '(' if complex
       fragments.push @variable.compileToFragments(o, LEVEL_ACCESS)...
+      fragments.push @makeCode ')' if @isNew and complex
       fragments.push @makeCode "("
     fragments.push compiledArgs...
     fragments.push @makeCode ")"
@@ -706,19 +713,21 @@ exports.Call = class Call extends Base
   # splatArgs is an array of CodeFragments to put into the 'apply'.
   compileSplat: (o, splatArgs) ->
     if this instanceof SuperCall
-      return [].concat @makeCode("#{ @superReference o }.apply(#{@superThis(o)}, "),
-        splatArgs, @makeCode(")")
+      reference = @superReference o
+      if reference is 'super' or reference.indexOf('super.') is 0
+        return [].concat @makeCode("#{reference}(..."), splatArgs, @makeCode ')'
+      else
+        return [].concat @makeCode("#{reference}.apply(#{@superThis(o)}, "),
+          splatArgs, @makeCode(")")
 
     if @isNew
-      idt = @tab + TAB
-      return [].concat @makeCode("""
-        (function(func, args, ctor) {
-        #{idt}ctor.prototype = func.prototype;
-        #{idt}var child = new ctor, result = func.apply(child, args);
-        #{idt}return Object(result) === result ? result : child;
-        #{@tab}})("""),
-        (@variable.compileToFragments o, LEVEL_LIST),
-        @makeCode(", "), splatArgs, @makeCode(", function(){})")
+      complex = @variable.isComplex()
+      answer  = []
+      answer.push @makeCode 'new '
+      answer.push @makeCode '(' if complex
+      answer.push @variable.compileToFragments(o, LEVEL_LIST)...
+      answer.push @makeCode ')' if complex
+      return answer.concat @makeCode('(...'), splatArgs, @makeCode(')')
 
     answer = []
     base = new Value @variable
@@ -737,7 +746,7 @@ exports.Call = class Call extends Base
       else
         ref = 'null'
       answer = answer.concat fun
-    answer = answer.concat @makeCode(".apply(#{ref}, "), splatArgs, @makeCode(")")
+    answer = answer.concat @makeCode("(..."), splatArgs, @makeCode(")")
 
 #### Super
 
@@ -753,7 +762,11 @@ exports.SuperCall = class SuperCall extends Call
   # method.
   superReference: (o) ->
     method = o.scope.namedMethod()
-    if method?.klass
+    if method?.isMethod and method?.ctor
+      'super'
+    else if method?.isMethod
+      "super#{method.name.compile o}"
+    else if method?.klass
       {klass, name, variable} = method
       if klass.isComplex()
         bref = new IdentifierLiteral o.scope.parent.freeVariable 'base'
@@ -772,7 +785,8 @@ exports.SuperCall = class SuperCall extends Call
     else if method?.ctor
       "#{method.name}.__super__.constructor"
     else
-      @error 'cannot call super outside of an instance method.'
+      'unknown'
+      # @error 'cannot call super outside of an instance method.'
 
   # The appropriate `this` value for a `super` call.
   superThis : (o) ->
@@ -945,7 +959,6 @@ exports.Slice = class Slice extends Base
   children: ['range']
 
   constructor: (@range) ->
-    super()
 
   # We have to be careful when trying to slice through the end of the array,
   # `9e9` is used because not all implementations respect `undefined` or `1/0`.
@@ -1039,6 +1052,8 @@ exports.Obj = class Obj extends Base
 # An array literal.
 exports.Arr = class Arr extends Base
   constructor: (objs) ->
+    super()
+
     @objects = objs or []
 
   children: ['objects']
@@ -1068,6 +1083,106 @@ exports.Arr = class Arr extends Base
     no
 
 #### Class
+
+# A node representing a Javascript class.
+# Initialize a **JsClass** with its name, an optional superclass, and a
+# list of methods.
+exports.JsClass = class JsClass extends Base
+  constructor: (@name, @parent, @methods = []) ->
+    @boundMethods = []
+
+  isStatement: YES
+
+  setConstructor: (source, constructor) ->
+    source.error 'cannot define more than one constructor in a class' if @ctor
+
+    if constructor instanceof Code
+      @ctor = constructor
+    else
+      @generatedCtor = true
+
+      apply = new Value constructor, [new Access new PropertyName 'apply']
+      @ctor = new Code
+      @ctor.body.push new Literal 'super(...arguments)' if @parent
+      @ctor.body.push new Call apply, [new ThisLiteral, new IdentifierLiteral 'arguments']
+      @ctor.body.makeReturn()
+
+    @ctor.ctor = true
+    @ctor.noReturn = true
+    @ctor.isDerivedCtor = @parent?
+
+    name = new PropertyName 'constructor'
+    @addMethod name, @ctor
+
+  addMethod: (name, method) ->
+    method.name = name
+    method.isMethod = true
+
+    if method.bound and not method.static
+      method.bound = false
+      @boundMethods.push name
+
+    # Override `method`'s `compileToFragments` to append to the class's method fragments
+    # This allows `method.compileToFragments` to be called with the correct scope whilst still
+    # being generated in the class body.
+    compileToFragments = method.compileToFragments
+    method.compileToFragments = (o) =>
+      o.indent += TAB
+      @methodFragments.push compileToFragments.apply(method, arguments)
+      o.indent = o.indent[0..-1 - TAB.length]
+      []
+
+    @methods.push method
+    method
+
+  addStaticMethod: (name, method) ->
+    method.static = true
+    @addMethod name, method
+
+  compileNode: (o) ->
+    { indent } = o
+    o.indent += TAB
+
+    @bindMethods o
+
+    code = []
+    code.push @makeCode "#{indent}class "
+    code.push @name.compileNode(o)...
+    code.push @makeCode ' '
+    if @parent
+      code.push @makeCode 'extends '
+      code.push @parent.compileNode(o)...
+      code.push @makeCode ' '
+    code.push @makeCode '{'
+    if @methods.length
+      @bodyFragment    = @makeCode ''
+      @methodFragments = []
+      code.push @makeCode("\n")
+      code.push @ctor.compileToFragments(o)... if @generatedCtor
+      code.push @bodyFragment, @makeCode("\n#{indent}")
+    code.push @makeCode '}'
+
+    code
+
+  bindMethods: (o) ->
+    # If there are no bound methods, there's nothing to do
+    return if @boundMethods.length is 0
+
+    # If there's no ctor defined we need to generate one
+    if not @ctor
+      @generatedCtor = true
+      ctor = new Code
+      @setConstructor null, ctor
+
+    for name in @boundMethods
+      lhs = new Value(new ThisLiteral, [name]).compile o
+      @ctor.body.unshift new Literal "#{lhs} = #{utility 'bind', o}(#{lhs}, this)"
+
+    ctor?.body.unshift new Literal 'super(...arguments)' if @parent
+
+  finalize: ->
+    return if @methods.length is 0
+    @bodyFragment.code = fragmentsToText @joinFragmentArrays @methodFragments, "\n\n"
 
 # The CoffeeScript class definition.
 # Initialize a **Class** with its name, an optional superclass, and a
@@ -1117,50 +1232,63 @@ exports.Class = class Class extends Base
 
   # Merge the properties from a top-level object as prototypal properties
   # on the class.
-  addProperties: (node, name, o) ->
+  addProperties: (klass, name, o, node) ->
     props = node.base.properties[..]
     exprs = while assign = props.shift()
       if assign instanceof Assign
         base = assign.variable.base
-        delete assign.context
         func = assign.value
+
         if base.value is 'constructor'
-          if @ctor
-            assign.error 'cannot define more than one constructor in a class'
-          if func.bound
-            assign.error 'cannot define a constructor as a bound function'
           if func instanceof Code
-            assign = @ctor = func
+            klass.setConstructor new Access(base), func
           else
-            @externalCtor = o.classScope.freeVariable 'ctor'
-            assign = new Assign new IdentifierLiteral(@externalCtor), func
+            ctor = new IdentifierLiteral o.classScope.freeVariable 'ctor'
+            klass.setConstructor new Access(base), ctor
+            new Assign ctor, func
+        else if func instanceof Code and node in @body.expressions
+          @detectMethod klass, assign, required: true
         else
-          if assign.variable.this
-            func.static = yes
+          delete assign.context
+          accesses = if base.value == name
+            assign.variable.properties
           else
-            acc = if base.isComplex() then new Index base else new Access base
-            assign.variable = new Value(new IdentifierLiteral(name), [(new Access new PropertyName 'prototype'), acc])
-            if func instanceof Code and func.bound
-              @boundFuncs.push base
-              func.bound = no
-      assign
+            [
+              new Access new PropertyName 'prototype'
+              if base.isComplex() then new Index base else new Access base
+            ]
+          assign.variable = new Value new IdentifierLiteral(name), accesses
+          assign
+      else
+        assign
     compact exprs
 
   # Walk the body of the class, looking for prototype properties to be converted
   # and tagging static assignments.
-  walkBody: (name, o) ->
+  walkBody: (klass, name, o) ->
     @traverseChildren false, (child) =>
       cont = true
       return false if child instanceof Class
       if child instanceof Block
-        for node, i in exps = child.expressions
-          if node instanceof Assign and node.variable.looksStatic name
-            node.value.static = yes
+        for node, i in child.expressions
+          if node instanceof Assign and node.variable.base.value == name and node.value instanceof Code
+            child.expressions[i] = [m] if m = @detectMethod klass, node
           else if node instanceof Value and node.isObject(true)
             cont = false
-            exps[i] = @addProperties node, name, o
-        child.expressions = exps = flatten exps
-      cont and child not instanceof Class
+            child.expressions[i] = @addProperties klass, name, o, node
+        child.expressions = flatten child.expressions
+      cont
+
+  detectMethod: (klass, node, { required = false } = {}) ->
+    { base, properties } = node.variable
+    if properties.length is 0 and base.constructor
+      klass.addMethod new Access(base), node.value
+    else if properties.length is 1 and properties[0].name?.value isnt 'prototype'
+      klass.addStaticMethod properties[0], node.value
+    else if properties.length is 2 and properties[0].name.value is 'prototype'
+      klass.addMethod properties[1], node.value
+    else if required
+      node.error "can't determine method for #{node.constructor.name}"
 
   # `use strict` (and other directives) must be the first expression statement(s)
   # of a function body. This method ensures the prologue is correctly positioned
@@ -1172,24 +1300,6 @@ exports.Class = class Class extends Base
       node instanceof Value and node.isString()
     @directives = expressions.splice 0, index
 
-  # Make sure that a constructor is defined for the class, and properly
-  # configured.
-  ensureConstructor: (name) ->
-    if not @ctor
-      @ctor = new Code
-      if @externalCtor
-        @ctor.body.push new Literal "#{@externalCtor}.apply(this, arguments)"
-      else if @parent
-        @ctor.body.push new Literal "#{name}.__super__.constructor.apply(this, arguments)"
-      @ctor.body.makeReturn()
-      @body.expressions.unshift @ctor
-    @ctor.ctor = @ctor.name = name
-    @ctor.klass = null
-    @ctor.noReturn = yes
-
-  # Instead of generating the JavaScript string directly, we build up the
-  # equivalent syntax tree and compile that, in pieces. You can see the
-  # constructor, property assignments, and inheritance getting built out below.
   compileNode: (o) ->
     if jumpNode = @body.jumps()
       jumpNode.error 'Class bodies cannot contain pure statements'
@@ -1202,25 +1312,30 @@ exports.Class = class Class extends Base
     args  = []
     o.classScope = func.makeScope o.scope
 
-    @hoistDirectivePrologue()
     @setContext name
-    @walkBody name, o
-    @ensureConstructor name
-    @addBoundFunctions o
-    @body.spaced = yes
-    @body.expressions.push lname
+    @hoistDirectivePrologue()
 
     if @parent
       superClass = new IdentifierLiteral o.classScope.freeVariable 'superClass', reserve: no
-      @body.expressions.unshift new Extends lname, superClass
+      @body.expressions.unshift new Assign \
+        new Value(lname, [new Access new PropertyName '__super__']),
+        new Value(superClass, [new Access new PropertyName 'prototype'])
       func.params.push new Param superClass
       args.push @parent
 
-    @body.expressions.unshift @directives...
+    klass = new JsClass lname, superClass
+    @walkBody klass, name, o
 
-    klass = new Parens new Call func, args
-    klass = new Assign @variable, klass, null, { @moduleDeclaration } if @variable
-    klass.compileToFragments o
+    @body.spaced = true
+    @body.expressions.unshift @directives..., klass
+    @body.expressions.push lname
+
+    result = new Parens new Call func, args
+    result = new Assign @variable, result, null, { @moduleDeclaration } if @variable
+    fragments = result.compileToFragments o
+
+    klass.finalize()
+    fragments
 
 #### Import and Export
 
@@ -1596,15 +1711,18 @@ exports.Assign = class Assign extends Base
 # has no *children* -- they're within the inner scope.
 exports.Code = class Code extends Base
   constructor: (params, body, tag) ->
+    super()
+
     @params      = params or []
     @body        = body or new Block
     @bound       = tag is 'boundfunc'
     @isGenerator = !!@body.contains (node) ->
       (node instanceof Op and node.isYield()) or node instanceof YieldReturn
+    @isMethod    = false
 
   children: ['params', 'body']
 
-  isStatement: -> !!@ctor
+  isStatement: -> !!@method
 
   jumps: NO
 
@@ -1666,20 +1784,30 @@ exports.Code = class Code extends Base
       node.error "multiple parameters named #{name}" if name in uniqs
       uniqs.push name
     @body.makeReturn() unless wasEmpty or @noReturn
-    code = 'function'
-    code += '*' if @isGenerator
-    code += ' ' + @name if @ctor
-    code += '('
-    answer = [@makeCode(code)]
-    for p, i in params
-      if i then answer.push @makeCode ", "
-      answer.push p...
-    answer.push @makeCode ') {'
-    answer = answer.concat(@makeCode("\n"), @body.compileWithDeclarations(o), @makeCode("\n#{@tab}")) unless @body.isEmpty()
-    answer.push @makeCode '}'
+    code = []
+    code.push @makeCode 'static ' if @static
+    code.push @makeCode 'function' unless @isMethod
+    code.push @makeCode '*' if @isGenerator
+    if @isMethod
+      name = @name.compileToFragments o
+      name.shift() if name[0].code is '.'
+      code.push name...
 
-    return [@makeCode(@tab), answer...] if @ctor
-    if @front or (o.level >= LEVEL_ACCESS) then @wrapInBraces answer else answer
+      if @isDerivedCtor then for param in @params when param.name.this
+        superCall = @body.contains (child) -> child instanceof SuperCall
+        superCall?.error 'explicit super not allowed with @ params'
+        @body.expressions.unshift new SuperCall
+        break
+    code.push @makeCode '('
+    for p, i in params
+      if i then code.push @makeCode ", "
+      code.push p...
+    code.push @makeCode ') {'
+    code = code.concat(@makeCode("\n"), @body.compileWithDeclarations(o), @makeCode("\n#{@tab}")) unless @body.isEmpty()
+    code.push @makeCode '}'
+
+    return [@makeCode(@tab), code...] if @isMethod
+    if @front or (o.level >= LEVEL_ACCESS) then @wrapInBraces code else code
 
   eachParamName: (iterator) ->
     param.eachName iterator for param in @params
@@ -1774,6 +1902,7 @@ exports.Splat = class Splat extends Base
   isAssignable: YES
 
   constructor: (name) ->
+    super()
     @name = if name.compile then name else new Literal name
 
   assigns: (name) ->
@@ -1834,6 +1963,8 @@ exports.Expansion = class Expansion extends Base
 # flexibility or more speed than a comprehension can provide.
 exports.While = class While extends Base
   constructor: (condition, options) ->
+    super()
+
     @condition = if options?.invert then condition.invert() else condition
     @guard     = options?.guard
 
@@ -1890,6 +2021,8 @@ exports.While = class While extends Base
 exports.Op = class Op extends Base
   constructor: (op, first, second, flip ) ->
     return new In first, second if op is 'in'
+
+    super()
     if op is 'do'
       return @generateDo first
     if op is 'new'
@@ -2241,6 +2374,8 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Parens
 # you can map and filter in a single pass.
 exports.For = class For extends While
   constructor: (body, source) ->
+    super()
+
     {@source, @guard, @step, @name, @index} = source
     @body    = Block.wrap [body]
     @own     = !!source.own
--- a/test/assignment.coffee
+++ b/test/assignment.coffee
@@ -449,7 +449,7 @@ test "#1591, #1101: splatted expressions in destructuring assignment must be ass
 
 test "#1643: splatted accesses in destructuring assignments should not be declared as variables", ->
   nonce = {}
-  accesses = ['o.a', 'o["a"]', '(o.a)', '(o.a).a', '@o.a', 'C::a', 'C::', 'f().a', 'o?.a', 'o?.a.b', 'f?().a']
+  accesses = ['o.a', 'o["a"]', '(o.a)', '(o.a).a', '@o.a', 'C::a', 'f().a', 'o?.a', 'o?.a.b', 'f?().a']
   for access in accesses
     for i,j in [1,2,3] #position can matter
       code =
--- a/test/classes.coffee
+++ b/test/classes.coffee
@@ -26,7 +26,9 @@ test "classes with a four-level inheritance chain", ->
     @array = [1, 2, 3]
 
   class ThirdChild extends SecondChild
-    constructor: -> thirdCtor.call this
+    constructor: ->
+      super
+      thirdCtor.call this
 
     # Gratuitous comment for testing.
     func: (string) ->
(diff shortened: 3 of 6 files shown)


src/nodes.coffee
```diff
@@ -225,6 +225,7 @@ exports.Base = class Base
 # `if`, `switch`, or `try`, and so on...
 exports.Block = class Block extends Base
   constructor: (nodes) ->
+    super()
```
connec: ES2015 derived constructors need explicit `super` calls in order to access `this`.

src/nodes.coffee
```diff
@@ -687,12 +686,20 @@ exports.Call = class Call extends Base
...
+        preface += ', ' if compiledArgs.length
+      else
+        preface += '('
       fragments.push @makeCode preface
     else
-      if @isNew then fragments.push @makeCode 'new '
+      complex = @variable.isComplex()
+      if @isNew
+        fragments.push @makeCode 'new '
+        fragments.push @makeCode '(' if complex
```
connec: This wraps complex `new` targets in parenthesis (e.g. `new (a())`) to ensure there are no errors from calling constructors without `new` (e.g. [this test](https://github.com/jashkenas/coffeescript/pull/4330/files#diff-1fc895c05f3855ef64b85685f62803acR469) was throwing that error without this change).

src/nodes.coffee
```diff
@@ -687,12 +686,20 @@ exports.Call = class Call extends Base
 
     fragments = []
     if this instanceof SuperCall
-      preface = @superReference(o) + ".call(#{@superThis(o)}"
-      if compiledArgs.length then preface += ", "
+      preface = @superReference o
+      if preface isnt 'super'
```
connec: Bare `super` (e.g. in a constructor) isn't a function reference, and needs special treatment.

src/nodes.coffee
```diff
@@ -706,19 +713,21 @@ exports.Call = class Call extends Base
   # splatArgs is an array of CodeFragments to put into the 'apply'.
   compileSplat: (o, splatArgs) ->
     if this instanceof SuperCall
-      return [].concat @makeCode("#{ @superReference o }.apply(#{@superThis(o)}, "),
-        splatArgs, @makeCode(")")
+      reference = @superReference o
+      if reference is 'super' or reference.indexOf('super.') is 0
```
connec: Uses ES2015 splats in `super` calls. Hopefully these changes will be superseded by a change to ES2015 splats.

src/nodes.coffee
```diff
@@ -706,19 +713,21 @@ exports.Call = class Call extends Base
...
-      idt = @tab + TAB
-      return [].concat @makeCode("""
-        (function(func, args, ctor) {
-        #{idt}ctor.prototype = func.prototype;
-        #{idt}var child = new ctor, result = func.apply(child, args);
-        #{idt}return Object(result) === result ? result : child;
-        #{@tab}})("""),
-        (@variable.compileToFragments o, LEVEL_LIST),
-        @makeCode(", "), splatArgs, @makeCode(", function(){})")
+      complex = @variable.isComplex()
```
connec: Compiles `new` calls with splats to ES2015 splats. Again, a changing to ES2015 splats generally would supersede this.

src/nodes.coffee
```diff
@@ -753,7 +762,11 @@ exports.SuperCall = class SuperCall extends Call
   # method.
   superReference: (o) ->
     method = o.scope.namedMethod()
-    if method?.klass
+    if method?.isMethod and method?.ctor
```
connec: `superReference` is in a bit of a state now. In order to preserve behaviours such as `super` in indirect methods and soaked super invocations, the old implementation hasn't been removed. If those were removed this would become much, much simpler.

src/nodes.coffee
```diff
@@ -1069,6 +1084,106 @@ exports.Arr = class Arr extends Base
...
+
+    if method.bound and not method.static
+      method.bound = false
+      @boundMethods.push name
+
+    # Override `method`'s `compileToFragments` to append to the class's method fragments
+    # This allows `method.compileToFragments` to be called with the correct scope whilst still
+    # being generated in the class body.
+    compileToFragments = method.compileToFragments
+    method.compileToFragments = (o) =>
```
connec: This is probably the worst hack in the PR. It's being used to allow methods to be compiled with the correct scope with respect to the class body, whilst also being hoisted into the `JsClass`. This is needed for the following to work:

```coffee
class A
  a = 1
  foo: ->
    a = 2 # changes `a` from the class scope
    b = 2 # declares a new variable

  b = 1
  bar: ->
    a = 2 # changes a from the class scope
    b = 2 # changes b from the class scope
```

See [the test](https://github.com/connec/coffeescript/blob/55d38af18822a3544a52d17bbd54dc81a9b8836c/test/classes.coffee#L409).

src/nodes.coffee
```diff
@@ -1069,6 +1084,106 @@ exports.Arr = class Arr extends Base
...
+    if not @ctor
+      @generatedCtor = true
+      ctor = new Code
+      @setConstructor null, ctor
+
+    for name in @boundMethods
+      lhs = new Value(new ThisLiteral, [name]).compile o
+      @ctor.body.unshift new Literal "#{lhs} = #{utility 'bind', o}(#{lhs}, this)"
+
+    ctor?.body.unshift new Literal 'super(...arguments)' if @parent
```
connec: This could probably just unshift `new SuperCall`

src/nodes.coffee
```diff
@@ -1069,6 +1084,106 @@ exports.Arr = class Arr extends Base
...
+
+  finalize: ->
```
connec: The second part of the method hoisting hack. Once a class body has been compiled, `finalize()` is called to inject the compiled methods into the `JsClass` body.

src/nodes.coffee
```diff
@@ -1117,50 +1232,63 @@ exports.Class = class Class extends Base
...
+  walkBody: (klass, name, o) ->
     @traverseChildren false, (child) =>
       cont = true
       return false if child instanceof Class
       if child instanceof Block
-        for node, i in exps = child.expressions
-          if node instanceof Assign and node.variable.looksStatic name
-            node.value.static = yes
+        for node, i in child.expressions
+          if node instanceof Assign and node.variable.base.value == name and node.value instanceof Code
```
connec: This no longer only detects `static` methods to ensure that `@prototype.a = ->` is captured as a class method.

src/nodes.coffee
```diff
@@ -1172,24 +1300,6 @@ exports.Class = class Class extends Base
       node instanceof Value and node.isString()
     @directives = expressions.splice 0, index
 
-  # Make sure that a constructor is defined for the class, and properly
-  # configured.
-  ensureConstructor: (name) ->
```
connec: Constructor-ensurance is now handled in `JsClass`.

src/nodes.coffee
```diff
@@ -1202,25 +1312,30 @@ exports.Class = class Class extends Base
...
     o.classScope = func.makeScope o.scope
 
-    @hoistDirectivePrologue()
     @setContext name
-    @walkBody name, o
-    @ensureConstructor name
-    @addBoundFunctions o
-    @body.spaced = yes
-    @body.expressions.push lname
+    @hoistDirectivePrologue()
```
connec: Not sure why this has been reordered...

src/nodes.coffee
```diff
@@ -1202,25 +1312,30 @@ exports.Class = class Class extends Base
...
 
     if @parent
       superClass = new IdentifierLiteral o.classScope.freeVariable 'superClass', reserve: no
-      @body.expressions.unshift new Extends lname, superClass
+      @body.expressions.unshift new Assign \
```
connec: Rather than using the `extend` utility, we are compiling an `extends` in the `JsClass`. However, old `superReference` behaviour still depends on `__super__` (unless there's a way to look that up in vanilla JS).

src/nodes.coffee
```diff
@@ -1596,15 +1711,18 @@ exports.Assign = class Assign extends Base
...
     @body        = body or new Block
     @bound       = tag is 'boundfunc'
     @isGenerator = !!@body.contains (node) ->
       (node instanceof Op and node.isYield()) or node instanceof YieldReturn
+    @isMethod    = false
 
   children: ['params', 'body']
 
-  isStatement: -> !!@ctor
+  isStatement: -> !!@method
```
connec: I guess this isn't working correctly (`@method` vs. `@isMethod`) 😨  Not sure what will be broken...

src/nodes.coffee
```diff
@@ -1666,20 +1784,30 @@ exports.Code = class Code extends Base
...
+    code = []
+    code.push @makeCode 'static ' if @static
+    code.push @makeCode 'function' unless @isMethod
+    code.push @makeCode '*' if @isGenerator
+    if @isMethod
+      name = @name.compileToFragments o
+      name.shift() if name[0].code is '.'
+      code.push name...
+
+      if @isDerivedCtor then for param in @params when param.name.this
```
connec: Derived constructors must call `super` before using `this`, which conflicts with `@` arguments going at the top of the constructor.

I think the best way to deal with this would be to automatically call `super` as the first statement in all derived constructors, disallowing it from ever being written explicitly. This would have implications with changing the arguments passed to superclasses and writing factory constructors that want to `return` other instances.

test/assignment.coffee
```diff
@@ -449,7 +449,7 @@ test "#1591, #1101: splatted expressions in destructuring assignment must be ass
 
 test "#1643: splatted accesses in destructuring assignments should not be declared as variables", ->
   nonce = {}
-  accesses = ['o.a', 'o["a"]', '(o.a)', '(o.a).a', '@o.a', 'C::a', 'C::', 'f().a', 'o?.a', 'o?.a.b', 'f?().a']
+  accesses = ['o.a', 'o["a"]', '(o.a)', '(o.a).a', '@o.a', 'C::a', 'f().a', 'o?.a', 'o?.a.b', 'f?().a']
```
connec: In ES2015 classes, `prototype` is read-only, so the `C::` case won't work.

test/classes.coffee
```diff
@@ -66,25 +68,25 @@ test "constructors with inheritance and super", ->
...
-  Function.prototype.new = -> new this arguments...
-
-  ok (TwoClass.new('three')).name is 'three'
-  ok (new OneClass).function is 'function'
-  ok OneClass.new is 'new'
-
-  delete Function.prototype.new
-
-
+# test "Overriding the static property new doesn't clobber Function::new", ->
```
connec: `delete TwoClass.new` won't work now as it's inherited from `OneClass`. I wasn't sure how to update the test properly (e.g. `delete OneClass.new` works but probably invalidates the test).

test/classes.coffee
```diff
@@ -464,7 +466,7 @@ test "ensure that constructors invoked with splats return a new object", ->
   # Ensure that constructors invoked with splats cache the function.
   called = 0
   get = -> if called++ then false else class Type
-  new get() args...
+  new (get()) args...
```
connec: `()` was necessary to ensure the `new` is called on the result of `get`, otherwise you get a "constructors must be called with new" error.

test/classes.coffee
```diff
@@ -481,6 +483,7 @@ test "`new` works against bare function", ->
 test "#1182: a subclass should be able to set its constructor to an external function", ->
   ctor = ->
     @val = 1
+    return
```
connec: Needed to avoid a "derived constructors must return an object or undefined" error.

test/classes.coffee
```diff
@@ -739,21 +742,21 @@ test "#2599: other typed constructors should be inherited", ->
...
-  ok typeof brokenArray.method is 'undefined'
-
-  class WorkingArray extends Array
-    constructor: -> super
-    method: -> 'yes!'
-
-  workingArray = new WorkingArray
-  ok workingArray instanceof WorkingArray
-  eq 'yes!', workingArray.method()
+# test "#2359: extending native objects that use other typed constructors requires defining a constructor", ->
```
connec: This is broken now, since `brokenArray instanceof BrokenArray` and `brokenArray instanceof Array` with ES2015 inheritance :D

test/classes.coffee
```diff
@@ -834,68 +837,82 @@ test "#3232: super in static methods (not object-assigned)", ->
...
-  m = m2 = 'n'
-  eq 6, B.m()
-  eq 5, b.m()
-  eq 4.5, b.m2()
-  eq 1, count
-
-  class C extends B
-    m: -> super
-  eq 5, (new C).m()
+# test "#1392 calling `super` in methods defined on namespaced classes", ->
```
connec: This is broken by the last case: `new C.a` throws a "C.a is not a constructor". ES2015 class methods (static or not) can't be used as constructors.

test/classes.coffee
```diff
@@ -834,68 +837,82 @@ test "#3232: super in static methods (not object-assigned)", ->
...
+#   eq 4, (new namespace.A).n()
+#   eq 1, count
+#
+#   class C
+#     @a: ->
+#     @a extends Base
+#     @a::m = -> super
+#   eq 5, (new C.a).m()
+
+# test "dynamic method names and super", ->
```
connec: This test would pass if the indices in dynamic class methods were cached properly, but I had some difficulty implementing that correctly. E.g.

```coffee
class A
  "#{m}": ->
```

Currently compiles to:

```js
class A {
  [m] () {}
}
```

Should be

```js
var _ref = m
class A {
  [_ref] () {}
}
```

test/classes.coffee
```diff
@@ -834,68 +837,82 @@ test "#3232: super in static methods (not object-assigned)", ->
...
+#   eq 6, B.m()
+#   eq 5, b.m()
+#   eq 4.5, b.m2()
+#   eq 1, count
+#
+#   class C extends B
+#     m: -> super
+#   eq 5, (new C).m()
+
+# test 'dynamic method names using class context', ->
```
connec: This is were executable class bodies and ES2015 classes are incompatible: when class-context variables are referenced in dynamic keys. The problem is that the class methods are all defined above the executable statements, so they're not available in the compiled class.

Possible solutions:
- Compile all dynamic keys to ye olde `.prototype` assignments.
- Attempt to detect class-scoped variables in dynamic keys and compile them to `.prototype` assignments.
- Give up on executable class bodies (sadpanda).

test/comments.coffee
```diff
@@ -321,84 +321,84 @@ test "#3132: Format hand-made (raw) jsdoc-style block-comment nicely", ->
...
-  
-    DummyClass.instance = new DummyClass();
-  
-    return DummyClass;
-  
-  })();
-  
-  """
-  eq CoffeeScript.compile(input, bare: on), result
+# test "#3132: Place block-comments nicely", ->
```
connec: Comment ordering is screwed up by class method hoisting.

test/scope.coffee
```diff
@@ -55,28 +55,28 @@ test "#2255: global leak with splatted @-params", ->
...
-  		super
-  	foo : (cb) ->
-  		dolater =>
-  			dolater =>
-  				@_i += 2
-  				super cb
-
-  b = new B
-  b.foo => eq b._i, 3
+# test "#1183: super + fat arrows", ->
```
connec: `super` in closures isn't working atm. We could compile to the old form if we detect a `super` in a closure, or else we could compile closures in methods to new-style arrow functions (but that would have implications for scope).

connec:
Some thoughts on "Dynamic method names referencing variables that change in the class body is broken."

I had hoped we might be able to detect problem cases to compile to `.prototype` assignments, but in the presence of arbitrary method indirection this is impossible.

```coffee
a = 'hello'
foo = ->
  [a, old] = ['world', a]
  a

class Blah
  # This uses a variable outside the class scope
  "#{a}": ->

  # Which gets mutated
  foo()

  "#{a}": ->

# class Blah {
#   [a] () {}
#   [a] () {}
# }
# ...not what we wanted
```

I think the only way this could work is by compiling dynamic keys to `.prototype` assignments.

vendethiel:
Wow, nice work!

I'm not actually 100% sure we **have** to drop executable class bodies. Is there a downside to declaring functions as `.prototype.`? Are functions declared inside the class bound?

connec:
@vendethiel they don't appear to be bound, I think the only difference is that they're not enumerable.

I'm hoping there's some not too complicated way of identifying 'unsafe' dynamic assignments and/or restricting ESBs such that `.prototype` assignment is an edge case/unnecessary (e.g. disallow expressions after properties, compile dynamic keys referencing variables in the class scope to `.prototype` assigns).

bjmiller:
Why aren't we changing ES6 classes *to* old-style CS classes on extend?

It seems simpler to do, and leaves open executable class bodies and the ability to implement private members if we choose to do so.

kirly-af:
Impressive work here ! Concerning the dynamic methods names, I'm tempted to go for the restriction solution (ie. no expressions after properties). However some might want to keep the ability to do:

``` coffee
class Foo
  a: ->

  if true
    bar: -> true
  else
    bar: -> false
```

So maybe there are some expressions we should allow after properties ?

Also, I don't undestand what's the hoisting issue.

GeoffreyBooth:
@connec thanks for this. I think the plan of action should probably be to implement classes per the ES spec at least at first, and discard all tricky stuff like executable class bodies. Whatever doesn’t compile fairly directly into an ES class right now, we should leave out of the CoffeeScript 2 definition of `class`, at least for now. Once we get that much working reliably, we can discuss which if any sugars to add on top of that. And maybe we add back in all of them, and somehow manage to not break backward compatibility (or not break it most of the time). But I think it’s more likely that we don’t add back any of them, because ECMA is continuing to add features to the ES `class` spec and we don’t want to get into the same situation we were in previously where we implement a sugar (like the original CS `class` itself) and then they define/implement it differently.

So maybe save this work in a branch, and branch off of it and discard all the not-supported-by-ES stuff? I would think that that should let you simplify your code greatly.

connec:
I think before that happens, we should get arrow functions and ES2015 splats done, as they're both required to adequately leverage `super`.

That said, a much easier way of resolving #4233, would be to compile the constructor into a `class` and continue doing everything else as we do now. That would also be a much simpler change.

kirly-af:
@connec definitely! Is anyone aware of some inconvenients concerning such
approach?

Sure the output would be a bit verbose because of the use of
`Object.defineProperty()`, but apart from that, that would be a quite clean
patch.

GeoffreyBooth:
> That said, a much easier way of resolving #4233, would be to compile the constructor into a class and continue doing everything else as we do now.

That’s an interesting idea. If there’s some way of knowing when a class we’re trying to extend is an ES class, then this might even be backward-compatible and eligible for inclusion in 1.x? Although I was operating under the assumption that there’s no way to interoperate with ES classes without breaking changes, hence revised classes was to be a cornerstone of CoffeeScript 2.

I’ve been working on arrow functions and splats in #4311, though I have a ways to go.

I think we’re setting ourselves up for a debugging and maintenance nightmare if we want to convert CS `class` into something with a mountain of generated JavaScript, especially as the ES `class` spec grows and changes in the future. I think we should keep our sugars to an absolute minimum, rather than trying to preserve every feature of CS `class`es that we have today. Even the one sugar I might be tempted to preserve, class properties, is probably a bad idea to implement. My inclination would be to implement them as a series of statements in the constructor, e.g. `this.someProp = someValue`. But we could also implement them as properties on some class or function prototype. And then what happens when ES adds class properties, and implements them in some slightly different way? For example they require `this` statements in a constructor to occur after a call to `super`. What if they have some other silly requirement regarding class properties, like that maybe they’re inaccessible within a constructor, or who knows what ES comes up with? We don’t want to have to suffer a breaking change _again_ because we get ahead of them on the spec and then they define it differently.

kirly-af:
@connec: it turns out super calls are going to be real issue for your suggested approach:

``` js
class Foo {
    a() {
    }
}

class Bar extends Foo {
}

Bar.prototype.b = function () {
    super.a()
}
```

When running this code I get: `SyntaxError: 'super' keyword unexpected here`.

The only workaround I can think of would be to transpile super calls to the prototype method calls by going up twice the prototype chain:

``` js
Bar.prototype.b = function() {
  Object.getPrototypeOf(Object.getPrototypeOf(this)).a.call(this)
  // or
  this.__proto__.__proto__.a.call(this) // nicer, but browser only in the spec
}
```

The worst thing about this trick is that it woudn't work for super calls inside static methods, where we only need to go up the prototype chain once.

Do you have any other suggestion ?

@GeoffreyBooth: I understand your point of view (and anyway it's more or less a consortium), however, we definitely should consider other approaches.

connec:
@kirly-af I was intending not to change the current compilation (e.g. add a `__super__` property to `Bar`.

```js
var Foo, Bar

Foo = (function () {
  class Foo {}

  Foo.prototype.a = function () {}
})()

Bar = (function (superClass) {
  class Bar extends superClass {}

  Bar.__super__ = superClass.prototype

  Bar.prototype.a = function () {
    Bar.__super__.a.call(this)
  }
})(Foo)
```

Nicer `super` is the main (only?) advantage of a full ES2015 compilation. But again it comes at the cost of features (soaked access, use in prototype-assigned methods, etc.).

I think that moving straight to ES2015 classes is too dramatic. If we can do an intermediate solution like the above, then it solves the immediate problem and shouldn't make anything *harder* to maintain (e.g. if we're only changing the constructor shouldn't add much complexity).

As for making the change on the 1.x branch, it could be done but it looks like the `class` (and therefore the whole constructor) would need to be `eval`d, which isn't too great.

```
$ node -v
v0.12.14
$ cat > test.js 
if (false) {
  class A {}
}
$ node test.js 
test.js:2
  class A {}
  ^^^^^
SyntaxError: Unexpected reserved word
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
```

kirly-af:
Makes sense now !

This might be the best way to handle this emergency. If we just have to adapt constructors of derived classes and `super` behavior, then we have: no drastic feature changes, easy implementation without ugly hacks...

One issue might be to make class methods not enumerable as ES2015 specifies. I've read somewhere it's not compatible with the way CS handles `extends` (maybe @lydell or someone else can confirm).

@lydell, @GeoffreyBooth, @rattrayalex: Thoughts ?

Concerning the class eval, isn't CS targeting modern engines ? Class support came with node v6, so if people need legacy engines support, the class transpilation should be delegated to tools like babel.

connec:
@kirly-af I should say also the plan would be to use ES2015 `extends` as well, so that whole thing will be a lot simpler (just need the `__super__` assignment).

A separate decision is whether we continue to allow stand-alone `extends` (e.g. `A extends B`). That's not something I would personally miss, but I'm not sure if it's being used in the wild. Anyway, if we drop stand-alone extends then we drop the `extend` helper and inheritance is handled entirely in the JS.

We *could* opt to use `defineProperty` for properties/methods rather than `=` to get non-enumerability. I don't mind which we chose here, but it's not something I see as a benefit.

GeoffreyBooth:
@kirly-af I’m not clear what exactly you’re proposing. Can you please spell it out, including what breaking changes if any there are? How will this adapt to future changes in the ES `class` spec?

kirly-af:
@GeoffreyBooth: Well, I was talking about @connec suggestion. I'll try to make it clear.

CS classes could be made compatible with ES classes by transpiling the CS class constructor to an ES class (have look at @connec [example](https://github.com/jashkenas/coffeescript/pull/4330#issuecomment-251339033)).

Currently CS handles it parent constructor calls with `__super__.constructor.call(this)`, but it's not valid when extending an ES class. So we need to patch constructors so that it does proper `super` calls in derived classes constructors (and also do specific treatment for @*params* in constructors).

Concerning the breaking changes, we would have:
* `this` can't be used in derived classes contructor befor calling `super` (requirement)
* prototype methods won't be enumerable anymore (minor change)

...  and I think that's all.

That's pretty simple, and once it's done nothing stops us to keep being up to date with ES further versions. Sure ES classes are likely to change at some point, but the tc39 process is so slow we'll see it coming.
Then, **if** we have to make breaking change such what we'll do with `super`, well we'll just say "hello CS 3".

GeoffreyBooth:
And what about exporting a class for ES code to extend? Are there any incompatibilities between what we’re creating and a “true” ES class definition?

Also how is his example supposed to work? I tried playing with it but didn’t get far:
image

kirly-af:
The `return` statements in the wrapper functions are missing, nothing serious.

``` js
var Foo, Bar

Foo = (function () {
    class Foo {}

    Foo.prototype.a = function () {}

    return Foo
})()

Bar = (function (superClass) {
    class Bar extends superClass {}

    Bar.__super__ = superClass.prototype

    Bar.prototype.a = function () {
        Bar.__super__.a.call(this)
    }

    return Bar
})(Foo)
```

> what about exporting a class for ES code to extend?

Can you provide an example? I'm not sure I got you.

kirly-af:
@connec: I've been thinking about your idea for a while, and seriously I like it more and more. I'd be really happy giving you a hand on this if you need help.

Can you explain this statement:
> I think before that happens, we should get arrow functions and ES2015 splats done, as they're both required to adequately leverage super.

I guess it's related `super` support outside classes bodies? In that case it's probably not required for your approach, right ?

GeoffreyBooth:
> > what about exporting a class for ES code to extend?

> Can you provide an example? I'm not sure I got you.

a.coffee:
```coffee
export default class a
  greet: -> console.log 'hello!'
```
b.js:
```js
import a from 'a.coffee';

var greeter = new a();
greeter.greet(); // 'hello!'

class b extends a {
  welcome () {
    console.log('right this way!');
  }
}
```

kirly-af:
Nothing special here.

a.js:

``` js
var a;

export default a = (function() {
  class a {}

  a.prototype.greet = function() {
    console.log('hello');
  }
  return a;
})();
```

connec:
There will certainly be no difference in the presence of Babel, since it would then become more-or-less identical to the current output with functions and prototypes.

I've had a (entirely non-comprehensive) play with extending `class`es with static and `.prototype` assignments using node v6 and everything works as expected. Static and non-static methods and properties are inherited exactly as desired.

kirly-af:
@connec: using babel doesn't really matter, it depends on which plugins you're using. I only used the [es2015 modules to commonjs plugin](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/), and I confirm everything worked as expected in node v6.

connec:
As a separate point, based on @kirly-af's observation above, we could ditch the `__super__` assignment and use `Object.getPrototypeOf` to make it a bit more ES-y.

```coffee
class A
  f: ->
    console.log 'A'

class B extends A
  f: ->
    super
    console.log 'B'
```

```js
var A, B

A = (function () {
  class A {}

  A.prototype.f = function () {
    console.log('A')
  }

  return A
})()

B = (function (superClass) {
  class B extends superClass {}

  B.prototype.f = function () {
    Object.getPrototypeOf(B).f.apply(this, arguments)
    console.log('B')
  }

  return B
})(A)
```

Also, if we chose to ban super outside of class methods we might be able to use `superClass` directly, which would be pretty neat, and would probably simplify quite a lot internally. Maybe something to consider as a next step.

```js
B.prototype.f = function () {
  superClass.prototype.f.apply(this, arguments)
  console.log('B')
}
```

connec:
Also @kirly-af to answer your last point about splats and and arrow functions, they are still needed for class constructors:

```coffee
class A

class B extends A
  constructor: ->
    super # bare super needs to forward arguments

class C extends A
  constructor: ->
    (() => super())() # this is a bit contrived, but could happen...
```

```js
B = (function (superClass) {
  class B extends superClass {
    constructor () {
      // superClass.apply(this, arguments) won't work (ReferenceError: this is not defined)
      super(...arguments)
    }
  }
})(A)

C = (function (superClass) {
  class C extends superClass {
    constructor () {
      // (function () {
      //   superClass.apply(_this, arguments)
      // })(this)
      // ...also won't work (ReferenceError: this is not defined)
      // (function () {
      //   super(...arguments)
      // }).call(this)
      // ...also won't work (SyntaxError: 'super' keyword unexpected here)
      (() => super(...arguments))()
    }
  }
})(A)
```

We could no doubt implement special cases to deal with these, but it might be better to get arrow functions and splats done first.

src/nodes.coffee
Kovensky: There's really no way to make a variadic `super` call other than argument spread.

GeoffreyBooth:
I think https://github.com/jashkenas/coffeescript/pull/4330 might be far enough along for you to start using it. It passes all the tests minus the two we agreed to remove, so hopefully it’s close to being ready to merge. I’m sure @lydell will have notes about ways to clean up the code, and he’ll probably think of some new tests we should add and that the current code wouldn’t pass; but hopefully that PR is close to done.

I haven’t dug into this PR yet, but I have a few questions:

1. Does the CoffeeScript compiler codebase compile with this PR? Do you need to make any changes to the CoffeeScript compiler codebase for it to do so?

1. Is there a test we can add that would test interoperability with React? I think that should be the primary real-world goal for `class` support, as I can’t think of a more popular JavaScript library that uses ES2015 classes. We should come up with a test that proves whatever React is looking for in classes is getting output by our compiler.

1. What would we lose by limiting the output to only what could be contained by `class { }`? A lot, I know; but I think we should have a discussion as to whether we prefer to give people a way to output a more pure class versus this wrapped, function `prototype` elaborate version.

connec:
Hi @GeoffreyBooth, I'll pick this up again at the weekend and answer your questions fully. Off the top of my head:

1. It does compile. There were a couple of minor tweaks to do with adding `super` calls to nodes that didn't have it (since `Base` has an empty constructor, but ES6 demands it in order to use `this` at all in the class). These changes would be needed regardless of pure vs. augmented classes.

2. I'll look into this and add some tests.

3. Some examples:
   - executable class bodies, meaning:
     - no 'mixin' functions*
     - no class-private variables*
     - no dynamic method generation*
   - class properties (static and non-static)

   "*" Cases could be achieved outside the class definition with JS, with the caveat that they're further removed from the class definition, e.g. compare:

```coffee
class MyComponent extends React.Component
  @propTypes = ...

  ... lots of methods ...
```

with:

```js
class MyComponent extends React.Component {
  ... lots of methods...
}

MyComponent.propTypes = ...
```

CoffeeScript has influenced ES' direction before, and since everyone seems to hope ES' classes will be improved (at least wrt. non-function properties), I think we should keep the boat pushed out to help guide them :)

carlmathisen:
What do you guys think about Public Class Fields from ES2017? It is not in the official spec yet, but the draft is finished and Babel has had it in the es2017 preset for a while.

```js
class Person {
  name = "Joe";
  speak() {
    return "Hi, I'm " + this.name
  }
}
```

* [Babel REPL](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Ces2015-loose%2Ces2017%2Cstage-2&experimental=true&loose=false&spec=false&playground=true&code=class%20Person%20%7B%0A%20%20name%20%3D%20%22Joe%22%0A%20%20speak()%20%7B%0A%20%20%20%20return%20%22Hi%2C%20I'm%20%22%20%2B%20this.name%0A%20%20%7D%0A%7D%0A%0Aconst%20p%20%3D%20new%20Person()%0Aconsole.log(p.speak()))
* [Public Class Fields TC39](https://tc39.github.io/proposal-class-public-fields/)

GeoffreyBooth:
@carlmathisen aside from the ugliness, how do public class fields compare with the implementation in this PR? What does Babel compile public class fields into?

We don’t have Babel as part of the CoffeeScript compiler or tester, and I don’t think we want to add it. That means anything we output needs to be runnable at least in the Node runtime, at least via Node nightly `--harmony`. Modules are the great exception, but I’m okay with that because I think of `import` and `export` as more of a build step than something I would expect to be evaluated by a runtime. The philosophy behind not implementing any features before runtimes support them is that we don’t want to add support for something that could change when runtimes actually implement the feature.

Kovensky:
```js
class X extends Y {
  z = 'a'
}
```

compiles to:

```js
class X extends Y {
  constructor (...args) {
    super(...args)
    this.z = 'a'
  }
}
```
.

Static fields are just added to the constructor later:

```js
class X {
  static y = 'z'
}
```

compiles to:

```js
class X {}
X.y = 'z'
```

carlmathisen:
@GeoffreyBooth Yeah, you're probably right about that, not a wise move targeting transpilers - but rather the latest Node version. Which is likely to be in the forefront anyway.

@Kovensky's explanation of Babel's compile step does seem pretty straight forward, and it doesn't bring any magic tricks to the table that could complicate things if Class Fields ever makes it into the language spec. We could easily switch from - adding properties to the constructor or directly to the object - to ES's sugar syntax without an actual change to when the variables are set/accessed in runtime.

One thing though, any insight to why the properties are set *after* `super`, and not before? One should think setting them prior allows a subclass to override properties before the parent constructor (`super`) is executed.

Kovensky:
@carlmathisen `this` does not exist until after `super` runs; attempting to access it throws a `ReferenceError: |this| used uninitialized in class constructor`.

GeoffreyBooth:
When we were initially discussing keeping support for class properties, in my head I was assuming we would take the approach Babel does here, defining as part of the constructor. Is there a reason this PR’s `prototype` pattern is better?

connec:
@GeoffreyBooth I've been thinking about that, the only way the prototype pattern is 'better' is because it is consistent with method definitions:

```coffee
class A
  property: 1
  method: ->
```

Compiling one to a prototype property and the other to an instance property could be argued to be confusing. There's no technical reason that we couldn't compile function literals to prototype properties, and anything else to a `this` assignment in the constructor.

```js
A = (function () {
  class A {
    constructor () {
      this.property = 1
    }
  }

  A.prototype.method = function () {}

  return A
})()
```

connec:
Actually, this would conflict with object literals in executable class bodies 😢 

```coffee
class A
  if something
    property: 1
    method: -> console.log 'something is true'
  else
    property: 2
    method: -> console.log 'something is false'
```

We *could* pull non-function properties in literals into the constructor...

```js
A = (function () {
  var _ref

  class A {
    constructor () {
      if (_ref) {
        this.property = 1
      } else {
        this.property = 2
      }
    }
  }

  _ref = something
  if (_ref) {
    A.prototype.method = function () {
      console.log('something is true')
    }
  } else {
    A.prototype.method = function () {
      console.log('something is false')
    }
  }

  return A
})()
```

kirly-af:
Wow, I never realized the properties defined in class bodies were assigned to prototype the same way as functions. It makes sense and is consistent, but I feel this to be completely counterintuitive:

``` coffee
class Foo
    property: 'defaultVal'
    method: ->

a = new Foo()
b = new Foo()

a.property = 'a'
b.property = 'b'

console.log a.property # 'a'
console.log b.property # 'b'

delete a.property
console.log a.property # 'defaultVal'
```

I think it's worth considering to completely remove this feature as I'd doubt most users understand what it really does. I don't see real benefits to it.

Idem for ES public class fields. I think it's quite useless feature. But that is just my opinion.

@carlmathisen: btw this proposal only reached stage 2, so it's definitely not part of ES2017 (see your babel repl example includes the [stage2 preset](https://babeljs.io/docs/plugins/preset-stage-2/)).

kirly-af:
Also, I'd say there should be at least an example on the docs with properties to show how that behave.
I still believe that we should disable this as a breaking change for 2.0

svicalifornia:
Late to the party here, but agree with the earlier comment by @bjmiller. Can CS extend ES6 classes in such a way that the CS subclasses can use executable class bodies and other CS goodness?

svicalifornia:
In addition to testing with React, you may want to consider testing with the Model class from Objection.js. Even though it's much less-known than React, it is intended as an abstract class to be subclassed, similarly to React's Component class. When using Objection.js with CoffeeScript, I had to add this boilerplate after each CS subclass definition:

    Model = require('objection').Model
    class MyModel extends Model
      ...

    Model.extend MyModel

GeoffreyBooth:
@svicalifornia Thanks for these detailed notes. Do you have time to write any tests that would prove that the new `class` supports the cases you’re describing?

I think for various reasons listed in https://github.com/coffeescript6/discuss/issues/22 there’s no way to output the `class` keyword without breaking backward compatibility, so whatever ES `class` implementation we go with will need to be part of CS2.

I don’t think ES being able to extends CS2-generated classes is an edge case that should be dismissed. Just think of any Node module written in CoffeeScript, that exports a class. We want those to work.

svicalifornia:
@geoffreybooth Yes, you're right. I didn't mean to dismiss that use case, and it should be part of the acceptance criteria for CS 2.0 compilation to ES `class` code.

But that is not a supported use case today with CS 1.x, so I think it's fair to say that the use case of CS 1.x extending ES classes properly is a higher priority than the other direction. If CS 1.x code can make better use of ES classes, that would bring a lot of benefits to existing CS 1.x that either can't or won't make the leap to 2.0 right away. And perhaps we can achieve those benefits for CS 1.x before CS 2.0 is ready for production use, just as you made import/export available in CS 1.x. (Great job, btw!)

Unfortunately, I do not currently have the bandwidth to contribute code, but I am happy that these efforts are underway, and I'll continue to offer ideas and feedback.

GeoffreyBooth:
@svicalifornia CoffeeScript 1.x is fixed at outputting ES5-compatible code, with the exceptions of modules and generators which are opt-in by typing those keywords in your code. If we wanted to be purists we would’ve limited those features to CS2, to keep things totally simple (CS1 = ES5, CS2 = ESNext) but that principle was violated with generators a few years ago and modules seemed to fall into the same category.

CoffeeScript 1.x `class`, on the other hand, is in a different category. There’s already plenty of CoffeeScript code out there that uses `class` and expects to run in ES5 environments. If we output an ES `class` keyword for the 1.x branch, then 1.x is no longer opt-in to ESNext; we’re potentially breaking websites that expect the CoffeeScript compiler’s output to work without further transpilation. Even worse, now that evergreen browsers support `class`, the developers of such websites might not even notice that their sites become broken for users with older browsers that lack support for `class`.

So we can’t change the output for `class` in the 1.x branch, unfortunately. It needs to be part of 2. And we don’t have the energy to create some third branch of CoffeeScript, that is 1.x plus ES classes, without the rest of what 2 provides. Someone else could take that task on, by forking the 1.x branch and merging this PR into it (and any follow-ups that also affect classes) but I think that will get very messy very quickly.

If there’s some way to extend an ES class without needing to use the `class` keyword, then _that_ could become part of the 1.x branch; but I don’t think anyone is working on that. If this interests you and you can prove that it’s even possible, please propose it in https://github.com/coffeescript6/discuss/issues.

svicalifornia:
@GeoffreyBooth I never said that CS 1.x should output ES6 `class` code; in fact I explicitly said, as you and others have already, that ES6 `class` is a breaking feature destined for CS 2.0.

However, CS 1.x could and should do a better job of extending ES classes. When extending a class, CS 1.x should check whether the superclass is a CS class or an ES class. It should map `super` to work correctly with ES superclasses. It should call the `extend` hook on the superclass if it exists. The resulting subclass would still be a traditional CS class compiled to ES5, with support for executable body, static properties, etc. It would just be more compatible with its ES superclass, so that the developer doesn't have to add hacky code to finish the job. CS 1.x developers shouldn't have to think much about whether they are extending a CS class or an ES class. It should just work, and because that is not the case today, many of us have had to consider switching to ES to get that compatibility.

I hope that the most prominent missing bits are simple enough that they can be added to CS 1.x while CS 2.0 is still being discussed, finalized, and implemented.

GeoffreyBooth:
> CS 1.x could and should do a better job of extending ES classes. When extending a class, CS 1.x should check whether the superclass is a CS class or an ES class. It should map `super` to work correctly with ES superclasses. It should call the `extend` hook on the superclass if it exists.

Sure. Could you write up how all that could be done? Not the code to put in the compiler, but the general approach as to how to check whether it’s extending a CS class versus an ES class, etc. Maybe post it as an issue at https://github.com/coffeescript6/discuss/issues and it can become a new effort.

bjmiller:
This was one of my biggest disagreements with the CS6 direction.  I felt that ES6 classes should be "fixed" to be more like original CS classes.

When extending them, you can feature-test whether Object.getPrototypeOf and Object.getOwnPropertyNames exists, and use that to find all of the properties of a superclass to copy onto the new CS class.  It would take some doing, but that can make all classes compatible with one another, Any CS code can extend any classes from ES6 or CS, and you can choose to improve the design thereafter with callable constructors, real private members, etc.

Actually using ES6 classes binds you to their limitations, so CS classes can never have these features that people come around looking for.

emptist:
I agree with @bjmiller. Why should a good student take after bad ones :)
People want to out put ES7. That doesn't mean support for INPUT ES7. Even browsers don't yet. Input with Babel and done.

svicalifornia:
Someone posted a table of ES class features and their support in popular JS engines. We should go down the list of features and figure out which ones are not currently handled well in CS 1.x, perhaps with priority to the features that would improve React compatibility. We don't need 100% all at once, but a plan (and some tests, as @geoffreybooth suggests) for iterative improvement.

GeoffreyBooth:
@svicalifornia Support is complete in Node: http://node.green/#class and in Edge, Firefox, and Chrome; mostly complete in Safari 9 and complete in Safari 10 and WebKit: http://kangax.github.io/compat-table/es6/#test-class.

I think the main feature that imperils interoperability is the inability to extend an ES class. (#4233) We need to be able to do that, while still trying to preserve as many of CS classes’ features as possible.

svicalifornia:
I remember a different table that just showed various features of class support, such as `extend` and others. It's not a simple binary question of class support; there are many aspects to it.

connec:
I'm working on this again now, it's getting pretty close but one thing that's a bit odd is derived constructors with `@` params, e.g.:

```coffee
class A extends B
  constructor: (@a) ->
```

With the current compilation, the `@` param will be compiled to an `a` argument and a `this.a = a` assignment at the stop of the constructor body. This will break at runtime with `this is undefined` since `super` was not called before `this` was accessed.

Even adding an explicit `super` won't help, since the `@` param assignments will be inserted at the top, above the explicit call to `super`.

Currently, my solution to this is to add an implicit bare `super` for all derived constructors with `@` params. This works fine, except it seems strange that a constructor like

```coffee
constructor: (a) ->
  this
```

Will fail, whilst one like

```coffee
constructor: (@a) ->
  this
```

Will succeed, thanks to the implicit `super`.

The two most sane solutions to this, I think, are:

1. *Always* add an implicit `super` to derived constructors, since it *must* be called for the class to be useable.
    1. An extension to this would be to allow an explicit constructor in the case where no `@` params are present to allow users to override the arguments passed to `super`.
2. Disallow `@` params in constructors.

Personally I think that option 2. is unworkable, due to the huge volume of code<sup>citation needed</sup> that uses the convenient pattern of initialising instance fields with constructor arguments.

GeoffreyBooth:
Superseded by #4354.

PR #4335, opened 2016-10-09
Define proper operator precedence for bitwise/logical operators

alangpierce:
This is an upstream port for the patch https://github.com/decaffeinate/coffeescript/pull/8

See https://github.com/decaffeinate/decaffeinate/issues/291 for the bug that this fixed.

For the most part, CoffeeScript and JavaScript have the same precedence rules,
but in some cases, the intermediate AST format didn't represent the actual
evaluation order. For example, in the expression `a or b and c`, the `and` is
evaluated first, but the parser treated the two operators with equal precedence.
This was still correct end-to-end because CoffeeScript simply emitted the result
without parens, but any intermediate tools using the CoffeeScript parser could
become confused.

Here are the JS operator precedence rules:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

For the most part, CoffeeScript already follows these. `COMPARE` operators
already behave differently due to chained comparisons, so I think we don't need
to worry about following JS precedence for those. So I think the only case where
it was behaving differently in an important way was for the binary/bitwise
operators that are being changed here.

As part of this change, I also introduced a new token tag, `BIN?`, for the
binary form of the `?` operator.


GeoffreyBooth:
Should this (and similar PRs) be merged into `master` or `2`?

lydell:
master

alangpierce:
To be clear, this isn't a breaking change, and should have no effect on the runtime behavior. It's just fixing an issue where the intermediate CoffeeScript AST was inconsistent with the runtime evaluation order. As far as I know, it never caused any issues in the output JavaScript; `and` has always been higher precedence than `or` in CoffeeScript.

jashkenas:
This is nice. Thanks, @alangpierce.

Issue #4336, opened 2016-10-10
[CS2] Failing test rejecting inconsistent whitespace in .litcoffee files


GeoffreyBooth:
@eelco Since #4313 was merged into `2`, there is one failing test:

```
failed 1 and passed 724 tests in 2.05 seconds

  test/literate.litcoffee:56:5: error: indentation mismatch
				test "tabbed code", ->
				^
```

If I delete everything above this test in `literate.litcoffee`, leaving just:

```litcoffee
Tabs work too:

				test "tabbed code", ->
					ok yes
```
Then this test passes. See [comment](https://github.com/jashkenas/coffeescript/pull/4313#issuecomment-250924753).

GeoffreyBooth:
Could #3924 help with this? That seems like a worthy PR in its own right.

GeoffreyBooth:
@jashkenas @lydell what do you want to do about this? It’s causing issues with other `2`-branch pull requests to have this broken test in the `2` branch. Should we comment it out or delete it until this issue is patched? Split the test into a separate `.litcoffee` file that only uses tabs (which doesn’t fix this issue, but rather simply makes the test pass)? Or roll back #4313 until this is fixed?

GeoffreyBooth:
Update: potential fix in #4345.

PR #4338, opened 2016-10-15
[CS2] Compile computed properties to ES2015 equivalent

connec:
This is a fairly small change that simplifies the code generation for computed properties as they're now generated in the object initializer like regular properties.

I ended up doing this whilst working on fully switching to ES2015 splats. Targets the `2` branch since the output won't compile without ES2015 support.

- [x] Tests pass (excepting #4336)
- [x] Compiles the CoffeeScript code base without modifications
- [ ] Compiles a codebase making significant use of computed keys - any suggestions?


connec:
Also, the first commit is actually just compiling some files in `2` that weren't committed for some reason.

src/nodes.coffee
```diff
@@ -981,54 +981,37 @@ exports.Obj = class Obj extends Base
...
         if prop.variable instanceof Value and prop.variable.hasProperties()
           prop.variable.error 'invalid object key'
       if prop instanceof Value and prop.this
         prop = new Assign prop.properties[0].name, prop, 'object'
-      if prop not instanceof Comment
-        if i < dynamicIndex
-          if prop not instanceof Assign
-            prop = new Assign prop, prop, 'object'
+      if prop not instanceof Comment and prop not instanceof Assign
+        if prop.isComplex()
```
connec: Previously, dynamic keys were being detected with...
```coffee
(prop.variable or prop).base instanceof Parens
```
... but this seems simpler. I don't know all the ins and outs though, so I might be missing something.

jashkenas:
Nice stuff!

GeoffreyBooth:
@connec this looks good, do you mind just giving an example of some CoffeeScript with computed properties, and what it was compiling to before and after this PR?

lydell:
@GeoffreyBooth Here's an example: https://github.com/jashkenas/coffeescript/pull/3840/commits/76c076db555c9ac7c325c3b285cd74644a9bf0d2

connec:
Sure:

Current (1.11.1):
```
★ coffee -bcs
{
  "#{computed}": 1
  "#{f()}"
  normal: 2
  "#{a.bit.more.complex = computation()}": 3
}
// Generated by CoffeeScript 1.11.1
var obj, ref;

(
  obj = {},
  obj["" + computed] = 1,
  obj[ref = "" + (f())] = ref,
  obj.normal = 2,
  obj["" + (a.bit.more.complex = computation())] = 3,
  obj
);
```

In this PR:

```
★ bin/coffee -bcs
{
  "#{computed}": 1
  "#{f()}"
  normal: 2
  "#{a.bit.more.complex = computation()}": 3
}
// Generated by CoffeeScript 1.11.0
var ref;

({
  ["" + computed]: 1,
  [ref = "" + (f())]: ref,
  normal: 2,
  ["" + (a.bit.more.complex = computation())]: 3
});
```

GeoffreyBooth:
This looks good to me. I updated your branch with the latest `2`, and the tests pass. Anyone opposed to merging this in?

lydell:
Go for it!

danielbayley:
Following on from #3597, wouldn't it now be preferable to simply allow the existing ES6 `[key]: value` syntax and pass it straight through for this one?

connec:
@danielbayley the purpose wasn't to add or change any syntax, interpolations were already allowed in object keys, this just simplifies the output by compiling to ES6. Might be worth having a separate discussion about `[key]` syntax.

Personally I prefer CS' approach to this. If "hello" is a valid key, then so too should be "#{hello}". ES' handling of template strings is needlessly restrictive imo.

danielbayley:
> interpolations were already allowed in object keys

Is this what you were referring to?
~~~ coffee
key = 'key'

obj = {}
obj[key] = 'value'
~~~
 Obviously this is nothing new, but just to clarify…
~~~ coffee
"#{key}": 'value' # works… cool! Leave it be.

[key]: 'value' # is valid ES6 but I think should also work in CS.
~~~
Maybe you're right about it being a separate discussion; is it worth opening another issue?

connec:
>> interpolations were already allowed in object keys
>
> @connec Is this what you were referring to?

Sorry, I just meant that this syntax was valid before the PR:

```coffee
key = 'key'
obj =
  "#{key}": 'value'
```

It used to compile to (roughly):

```js
var key, obj

key = 'key';

obj = (
  ref = {},
  ref['' + key] = 'value',
  ref
);
```

And now compiles to:

```js
var key, obj

key = 'key';

obj = {
  ['' + key]: 'value'
};
```

> is it worth opening another issue?

I'm not sure, @GeoffreyBooth may have an opinion on this.

GeoffreyBooth:
Personally I don’t see why `[key]: 'value'` shouldn’t be valid CoffeeScript. @lydell? I think this corresponds with our discussing allowing `...` on either side of a variable name, e.g. `...args` or `args...`, to ease the transition for people who write lots of ES.

Perhaps open a new issue requesting the new syntax. Please prefix it with `[CS2]`, as I don’t think we’re revisiting this anymore in 1.x.

lydell:
I guess we more or less have to support `[key]: 'value'` to allow stuff like `[Symbol.iterator]: foo`.

connec:
> `[Symbol.iterator]: foo`

Good point 😢

I still think it's consistent to allow interpolations in keys though.

lydell:
> I still think it's consistent to allow interpolations in keys though.

Me too.

PR #4339, opened 2016-10-16
Disallow exporting anonymous classes

GeoffreyBooth:
See [this comment](https://github.com/jashkenas/coffeescript/pull/4300#pullrequestreview-4370786).

We had a test for making sure the compiler wasn’t allowing `export class` (without a name given for the class) but it was a poorly written test, and wasn’t actually checking that the export wasn’t happening. The error message it was checking for was actually getting thrown by the Node runtime (for not understanding the `export` keyword) rather than checking that anonymous classes can’t be exported.

I added a check in `nodes.coffee` that exported classes have names, and I fixed the test.

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -1906,6 +1906,9 @@
         code.push(this.makeCode('default '));
       }
       if (!(this instanceof ExportDefaultDeclaration) && (this.clause instanceof Assign || this.clause instanceof Class)) {
+        if (this.clause instanceof Class && !this.clause.variable) {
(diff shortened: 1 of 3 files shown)


zapu:
Oh, cool, didn't notice it was fixed right after the report! Thank you very much :)

GeoffreyBooth:
Yes, it will go out with the next release. The current generated code for this edge case isn’t valid JavaScript, so the runtime should already be throwing an error; and then when the next point release goes out with this fix, the compiler will catch the error before the runtime does.

PR #4341, opened 2016-10-19
Node 6 deprecated `new Buffer` in favor of `Buffer.from` and `Buffer.…

GeoffreyBooth:
Node 6 [deprecated `new Buffer` in favor of `Buffer.from` and `Buffer.alloc`](https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe). This PR updates our calls to `Buffer` to use the new syntax. As a result of that, I also set the required version of Node per `package.json` to be >= 6.9.0, and set the overall version of the project to be 2.0.0-alpha for now.

### DIFF
--- a/lib/coffee-script/coffee-script.js
+++ b/lib/coffee-script/coffee-script.js
@@ -26,7 +26,7 @@
   base64encode = function(src) {
     switch (false) {
       case typeof Buffer !== 'function':
-        return new Buffer(src).toString('base64');
+        return Buffer.from(src).toString('base64');
       case typeof btoa !== 'function':
         return btoa(src);
       default:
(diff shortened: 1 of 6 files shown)


lydell:
LGTM. Does it make sense to do something in master as well? For instance, `(Buffer.from?(src) ? new Buffer(src)).toString('base64')`.

GeoffreyBooth:
Hmm, that’s clever. I suppose we could, though I’m not sure what benefit it would provide. `new Buffer` doesn’t throw a deprecation warning (yet), even in Node nightly.

GeoffreyBooth:
Should I maybe use a package.json version number like `2.0.0-nightly`? or `2.0.0-alpha0`? To imply that this branch will keep changing frequently without version bumps. Or keep `1.11.1`?

jashkenas:
`2.0.0-alpha` sounds about right to me.

GeoffreyBooth:
Any objections to merging this in?

lydell:
Not from me.

PR #4344, opened 2016-10-22
Properly set location for string tokens ending in a newline

alangpierce:
This is an upstream port of https://github.com/decaffeinate/coffeescript/pull/9

The existing logic for computing the end location of a string was to take the
end of the string contents, then add the delimiter length to last_column. For
example, `"""abc"""` would have an end position three characters after the `c`.
However, if a string ended in a newline, then the end location for the string
contents would be one line above the end location for the string, so the proper
fix is to move the end location to the next line, not just to shift it to the
right.

This avoids a bug where the location data would sometimes reference a
non-existent location (one past the end of its line). It fixes the AST location
data, although as far as I know, it never has caused correctness issues in the
CoffeeScript output.

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -573,7 +573,7 @@ exports.Lexer = class Lexer
 
     [firstToken, ..., lastToken] = tokens
     firstToken[2].first_column -= delimiter.length
-    if lastToken[1].substr(-1) == '\n'
+    if lastToken[1].substr(-1) is '\n'
       lastToken[2].last_line += 1
       lastToken[2].last_column = delimiter.length - 1
     else
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -573,7 +573,11 @@ exports.Lexer = class Lexer
 
     [firstToken, ..., lastToken] = tokens
     firstToken[2].first_column -= delimiter.length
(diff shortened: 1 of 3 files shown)


src/lexer.coffee
```diff
@@ -573,7 +573,11 @@ exports.Lexer = class Lexer
 
     [firstToken, ..., lastToken] = tokens
     firstToken[2].first_column -= delimiter.length
-    lastToken[2].last_column += delimiter.length
+    if lastToken[1].substr(-1) == '\n'
```
GeoffreyBooth: Please use `is` instead of `==`. In CoffeeScript `==` compiles to `===`, so we try to avoid using it in code to avoid confusion. (I know, there are a few places we should fix that use it already, but we shouldn’t add more.)

alangpierce: Thanks, I just updated it.

GeoffreyBooth:
Thanks! Looks good to me. @lydell or @jashkenas ?

GeoffreyBooth:
BTW @alangpierce are there any other similar pull requests you’re waiting on approval for?

alangpierce:
@GeoffreyBooth nope, everything else in https://github.com/decaffeinate/coffeescript is already in CoffeeScript master, thanks for checking!

PR #4346, opened 2016-10-23
Coffee Tags

shreeve:
## CoffeeTags
- tiny addition (just 27 lines total) to the native CoffeeScript compiler for rendering HTML
- most useful with projects like Vuejs or React, but can be used by many others
- no new language or syntax to learn, just use all the power of normal CoffeeScript
- resembles HTML, but is actually active/executable code
- doesn't mess with CoffeeScript syntax, takes a previously invalid syntax and makes it legal
- doesn't require tags to be closed, uses indentation level to determine
- doesn't require tags to be predeclared to use (no more `{div, table, th, tr} = React.DOM`)
- allows selector-like syntax for tags (e.g. - `<div.active@menu>` yields `<div class="active" id="menu">`
- allows naked tags (e.g. - you don't have to type `div()`, you can just use `<div>`)
- allows raw HTML using a trailing exclamation point in the tag name (e.g. - `<span.note!> "Note: #{text}"`)
- allows default tag name of `div` (i.e. - `<>`, `<.note>`, and `<.active@sidebar!>` are all valid)
- allows reserved words to be used as attributes (no more using `className: 'foo'`)
- allows attributes to be skipped (no more using `div null, "Hello"`)
- allows children to be passed as a list without having to pass them as an explicit array
- allows the first argument to indicate if the tag should be skipped (via truthy Boolean or Number)
- invokes a render function (by convention, 'h') to process tags for the target framework (React, Vue, etc.)
- a companion helper for Vuejs is found at https://github.com/vuejs/vue/pull/4016
- TODO: allows dynamic tags (e.g. - <`"h#{ level }">` can produce `<h3>`)
- TODO: allows closing tags (they are ignored, but might help some with readability)
#### CoffeeScript source

``` coffee
render: (h) ->
  <table.active.fixed@inbox>
    <tr>
      <th> 'From'
      <th> 'Subject'
      <th> 'Received'
    for item in @list
      <tr> class:item.tags,
        <td> <p> item.from
        <td> item.subj
        <td> <p> item.date
```
#### CoffeeScript output

``` javascript
render: function(h) {
  var item;
  return h('table.active.fixed@inbox', h('tr', h('th', 'From'), h('th', 'Subject'), h('th', 'Received')), (function() {
    var i, len, ref, results;
    ref = this.list;
    results = [];
      item = ref[i];
      results.push(h('tr', {
        "class": item.tags
      }, h('td', h('p', item.from)), h('td', item.subj), h('td', h('p', item.date))));
    }
    return results;
  }).call(this));
}
```
#### An example component (this one uses VueJS)

It's important to note that, just by adding a new legal token to `CoffeeScript`, this is totally valid `CoffeeScript` code (ie - there is no "transformation" step, it's just "normal" `CoffeeScript` source):

messages coffee lite 2016-10-23 00-48-13


GeoffreyBooth:
This is elegant, but if we’re going to add something like this, I wonder if it should resemble JSX? So instead of
```coffee
render: (h) ->
  <table.active.fixed@inbox>
    <tr>
      <th> 'From'
```
it would be
```coffee
render: (h) ->
  <table className="active fixed" id="inbox">
    <tr>
      <th>From</th>
```
I know, it’s uglier, but it’s closer to what people already know in JSX or HTML. If we’re going to provide something simpler, I would vote to ditch the `<`/`>` and go more Jade-like.

Also how does the implicit return factor into all of this? It seems to be simply returning the entire HTML block, when CoffeeScript syntax would seem to imply that it should only return the last line. JSX wraps the HTML in an explicit `return` block:
```jsx
render() {
  return (
    <div className="MarkdownEditor">
      <h3>Input</h3>
      <textarea
        onChange={this.handleChange}
        ref="textarea"
        defaultValue={this.state.value} />
    </div>
  );
}
```
I feel like there should be some way to assign the generated HTML to a variable, which I could choose to return (or not).

shreeve:
@GeoffreyBooth - Thank you for your thoughtful review!

The first point I wanted to stress on this PR is that it doesn't affect CoffeeScript per se, since it takes previously invalid syntax and makes it legal. Being orthogonal to the existing syntax, no existing code should be affected.

The second point is that this tiny commit (just 27 lines of code) does *not* implement a template language. All it does is provide a tiny bit of assistance from CoffeeScript's `lexer.coffee` to recognize each "CoffeeTag" as a new valid syntax token (called a `ctToken`) and allow it in the tokenized stream. Then, it makes a tiny tweak to `rewriter.coffer` that simply converts each `ctToken` into the following as follows:

```
Source:
<div> ...

Output:
h 'div', ...
```

Being normal CoffeeScript, this will in turn be understood as an implicit function call. This means that CoffeeScript will add the parentheses to make this a legit function call. Anything that is added to the line or nested below it using indentation will be passed as additional "arguments" to the `h` function. We are simply leveraging CoffeeScript's built in support for implicit function calls. If a CoffeeTag is terminal, meaning that there are no additional arguments on the same line and no nested children, then the trailing comma is not injected into the stream. It's actually super simple, but extremely powerful.

Your question about implicit returns and a how to assign generated HTML, etc. has an easy answer: CoffeeTags simply call a function, so if you want to assign it to a variable, you are free to do that. In fact, since CoffeeTags are valid syntax, you can place them anywhere you want. Just remember that each CoffeeTag is just mapped to a function call with optional arguments. In fact, you can see what final output will be generated using the following:

```
$ echo -e 'x = <div> "The time is #{new Date()}!"' | coffee-script/bin/coffee -s -p

(function() {
  var x;

  x = h('div', "The time is " + (new Date()) + "!");
}).call(this);
```

There is plenty of room for a JSX style template system and several things you mentioned are established conventions from other ecosystems, but don't seem to map cleanly to CoffeeScript. In fact, there is one already from https://github.com/jsdf/coffee-react but it is deprecated. I initially was planning on using this, but then realized that it wasn't worth trying to write a special parser and pull in many helper functions from CoffeeScript. It was just easier to make a few small tweaks to the existing CoffeeScript lexer and rewriter.

For example, the concept of closing tags in JSX doesn't seem to fit. CoffeeScript (and Python) have shown that whitespace is a perfectly valid and clean way to develop code. It seems CoffeeScript should also support a way to render HTML tags without having to close them.

Regarding inline-attributes such as `<div id='name' ...`, there is nothing to stop someone from using `Eco` templates (in fact, we also use `Eco` for templating). But to do this in a more CoffeeScript-ish way, it seems we should simply pass the attributes as such: `<div> id:'name'`.

On the point about `jade` (now called `pug`), I agree this would be great. However, that would be a templating language and not something that would easily be supported in a normal *.coffee file. What I was trying to accomplish here is a non-templating language. That is, a small tweak to enable native CoffeeScript to provide render functions out-of-the-box for today's modern web environments such as Vue and React.

Hopefully, the rationale above makes sense.

jiyinyiyong:
I have question, is this feature added specificly for Vue since the way you add `h`? In JSX it's `React.createElement` and that's a lot different from this one.

jacobbubu:
@jiyinyiyong I'm guessing the `h` comes from hyperscript and also used in Cycle.js earlier then Vue.

jiyinyiyong:
@jacobbubu I don't know that. Just checked out hypescript it seems there's no need of compilation. And I can't find a demo for Cycle. Using `h` as a variable name can be common. Let's limit the topic with `h`s generated with compilations.

GeoffreyBooth:
Yes, outputting a hard-coded `h` variable won’t do for a general-purpose language. If we’re going to add a feature like this, it’s going to need to work with React and other similar use cases. This feels like a hack to make Vue.js templating more streamlined.

The current energy in CoffeeScript development is toward building [release 2.0.0](https://github.com/coffeescript6/discuss), targeting ESNext output. I’m not sure there will be much interest in a big new (unrelated) feature like this until after 2 is out the door. It may be only 27 lines, but they’re scattered across the codebase and the impact of this new feature is dramatic, not least in what this would mean for documentation. This basically expands the scope of CoffeeScript to be a programming plus templating language like JSX, and that warrants a much bigger discussion: do we even want to add a psuedo-templating feature to CoffeeScript, if so what syntax should it have, what should it be compatible with, etc. It’s nice to know that doing so might be as easy as adding 27 lines of code, but we can’t add something like this to the language before it’s much more thoroughly designed and vetted.

src/rewriter.coffee
```diff
@@ -131,6 +141,18 @@ class exports.Rewriter
       i -= 1
     @tag(i) in tags
 
+  # Adjust token stream so that CoffeeTags look like normal function calls.
+  adjustCoffeeTags: ->
+    @scanTokens (token, i, tokens) ->
+      if token[0] is 'CT'
+        f = generate 'IDENTIFIER', 'h'; f.spaced = true
```
GeoffreyBooth: We can’t merge in this PR if what it does is replace this new coffee tag literal with calls to a mysterious `h` function that is not defined within CoffeeScript. That makes coffee tags dependant on some external library to supply this `h` function in the global scope.

GeoffreyBooth:
@shreeve please comment when the `h` is addressed and Coffee Tags are compatible with not just Vue but also React, for starters, and I’ll reopen this PR.

shreeve:
@GeoffreyBooth -- I've updated my local branch to include a new custom render function, using syntax like `<!myRenderer!>` to set the custom function to call (the default function is just `h`). I've updated my branch at https://github.com/shreeve/coffeescript/tree/coffee-tags but wasn't sure how to make this older pull request update to see the new changes.

GeoffreyBooth:
@shreeve Does this work with React? Can you show an example of how this could be used to implement a Coffee-style JSX?

shreeve:
@GeoffreyBooth - Yes, this should work with React also. The code will simply call a function (by default `h`, but you can set a custom `createElement` function by using the `<!myCustomCreateElement!>` syntax) that has the parameters (tags,data,children...).

shreeve:
Here is an example:

coffee-tags

shreeve:
In the above image, I'm actually showing the `coffee-tags` syntax (which allows the inline tags) and also something I called 'easytags' (which is a way to easily specify tags, classes, ids, etc.). So, `<@conversations>` yields `<div id="conversations">`. Also, the `<.messages>` will yield `<div class="messages">`.

GeoffreyBooth:
Can you please show how to achieve the examples in https://facebook.github.io/react/docs/react-without-jsx.html? Those are basically hello-world level.

mrmowgli:
I can see this being a powerful tool, and obviously there is a lot of great
work in this branch.   I like the idea of being able to treat HTML in a
coffee-esque way

I'm confused by a few things with this pull request, namely:

Why keep the html tag format at all?  Why not add a new keyword and
everything indented from there is assumed to be either html or another
keyword?

And then the second question is that of context, how is object context
maintained through the chain?

The last real question is how do you see this fitting in to the ecosystem?


On Feb 17, 2017 9:47 AM, "Steve Shreeve" <notifications@github.com> wrote:

In the above image, I'm actually showing the coffee-tags syntax (which
allows the inline tags) and also something I called 'easytags' (which is a
way to easily specify tags, classes, ids, etc.). So, <@conversations>
yields <div id="conversations">. Also, the <.messages> will yield <div
class="messages">.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/jashkenas/coffeescript/pull/4346#issuecomment-280718739>,
or mute the thread

jiyinyiyong:
Adding some personal opinion here as a ClojureScript programmer, who previously wrote CoffeeScript for medium-size React projects.

I think it's not very wise to add the feature into the compiler, although the idea itself is quite cool. Here's the reasons.

When I was using writing React, I can write [virtual DOM in CoffeeScript](https://www.packtpub.com/books/content/using-reactjs-without-jsx), i.e. the `React.createElement` way:

```coffee
{div = React.DOM}

render: ->
  div {},
    div {}, "text"
```

That is valid JavaScript code in CoffeeScript, nothing more.

And in ClojureScript, there is a similar situation that people want to code HTML DSLs like `div#foo`, not that JSX-style but just aliasing ids and classes, called [Hiccup](https://github.com/weavejester/hiccup#syntax):

```clojre
(html [:div#foo.bar.baz "bang"])
```

The thing is it's just normal Clojure syntax that the library can transform the DSL at runtime. So no compiler is involved.

I mean this feature is totally new syntax sugar for CoffeeScript, which involves some HTML knowledges. That will make CoffeeScript syntax depend on HTML stuffs. As it turned out it's actually a language extension, rather than core syntax. I would be worried if CoffeeScript ships such features. Well, being a ClojureScript guy.

GeoffreyBooth:
I’m not sure that this is something that should become part of the CoffeeScript compiler. On the one hand I like something that feels like an equivalent to JSX, to help prevent loss of mindshare for people who want to use React and feel like they should use JSX to do so; but on the other hand I feel like if something is going to be baked into the compiler, it needs to be abstract and able to handle _any_ templating language. Not just JSX or HTML/React or Vue, but also Handlebars and Underscore templates, or anything else people might want to use. Not all of these things will have a render function with a `(tags, data, children…)` signature.

Perhaps the real way that this should be implemented is via a plugin infrastructure, similar to how JSX is processed as a plugin for Babel. Then if @shreeve really just wants Vue templating integration, he can make just that; and someone using React or Handlebars or something else full-time on another project can tackle that integration as a plugin of its own. The compiler isn’t really designed in a very modular way, so I’m not sure how a plugin architecture would even look; or maybe there could be some kind of preprocessor, similar to how `marked` is parsing the markdown in `.litcoffee` files before leaving the code for the regular compiler to parse. Either way, the templating support becomes an optional add-on, not something baked into the compiler that requires configuration.

This also needs to be built on top of the `2` branch, not `master`. The 1.x line is closed to new features at this point.

lydell:
As far as I know, all experimental or non-standard JavaScript syntax (including JSX) is hard-coded into Babylon, and just have on/off switches (I think the same is true for "all" JavaScript parsers). There are no "parser plugins": babel/babel/issues#1351. Just sayin'.

jashkenas:
I think I've said this when it was proposed before, but...

CoffeeScript should not be a language that mixes two different alien syntaxes together. 

JSX is useful, and compelling conceptually, but hideous syntactically. It would be extremely unCoffeeScripty to munge and HTML-ish syntax into what we already have.

GeoffreyBooth:
Closing for now. I think if this can be reimplemented as a plugin that sits atop CoffeeScript, similar to how `marked` processes Markdown in `.litcoffee` files, that would be the preferred approach.

emilebosch:
@jashkenas I know what you're saying but things like typescript also have native JSX support. Its an absolute shame that implementing something like JSX as a pluginis kinda hard for none parser/grammer experts. And since CSJX is deprecated this really really forces me finally to move to babel, but i still feel completely silly writing so much babel code just because i want to use JSX. Is there really no option to include JSX?

GeoffreyBooth:
See https://facebook.github.io/react/docs/react-without-jsx.html:

```coffee
e = React.createElement

ReactDOM.render(
  e('div', null, 'Hello World'),
  document.getElementById('root')
)
```
I’ve also seen people make shortcuts for each element they want to use:
```coffee
div = (props, children) ->
  React.createElement 'div', props, children

ReactDOM.render(
  div(null, 'Hello World'),
  document.getElementById('root')
)
```

If anything, this looks _more_ Coffee-like than interspersing HTML tags in your code. And it’s fully supported by Facebook and React.

emilebosch:
I know that this exists, but its just not cutting it IMHO. CJSX is the only thing that comes close. Is there someway to hook in to the lexer/compiler as a plugin? Can’t seem to find any documentation on it.

GeoffreyBooth:
I would maybe approach this the way [Illiterate](https://github.com/billymoon/illiterate) does. Illiterate is an implementation of “literate anything”, inspired by Literate CoffeeScript. So Literate CoffeeScript treats indented code blocks as CoffeeScript and non-indented lines as Markdown comments. Illiterate takes this one step further, treating indented code blocks as _code_ (language agnostic) and non-indented lines as Markdown comments. Illiterate doesn’t try to transpile the code blocks (i.e. convert CoffeeScript code to JavaScript code); it just outputs a new file that has all the “comment” lines removed, so that some other tool can compile the “code” lines.

If you can figure out how to recognize the “JSX” lines separate from the CoffeeScript lines, you could create a fork of Illiterate that works similarly. For example, any line that begins with `<` can probably be safely assumed to be JSX. I’m sure it’s more complicated than that, that you’ll need to catch `return <` and the child lines within open/close tags, but it should be doable. Then you wrap all JSX sections in backticks, pass the revised source through the CoffeeScript compiler, then pass that output through Babel with the JSX plugin.

A dumb way to separate the JSX lines from the Coffee lines is to require open/close comments, like `# JSX START` and `# JSX END`, or `` ```jsx `` and `` ``` `` like Markdown. This is quite ugly, but would be easier to implement. Though if you’re going to do this, you should probably just use the  [triple-backtick operator](http://coffeescript.org/#embedded) that has been part of CoffeeScript since 1.12.0.

PR #4347, opened 2016-10-25
Selectively ignore CS-only keywords in ES imports and exports

GabrielRatener:
@GeoffreyBooth To address your JS interoperability concerns in #3757, this PR enables usage of CS-only keywords as local names in import lists and as any name in export lists. This will enable people to do the following

```js
// now possible
import {unless as someLocalName} from 'foo'

// but not
import {someImportedName as unless} from 'bar'

// for exports CS keywords are allowed everywhere
export {someLocalVar as unless}
export {unless, loop as until} from 'baz'
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -40,11 +40,13 @@ exports.Lexer = class Lexer
     @indebt     = 0              # The over-indentation at the current level.
     @outdebt    = 0              # The under-outdentation at the current level.
     @indents    = []             # The stack of all current indentation levels.
+    @indentLiteral = ''          # The indentation
     @ends       = []             # The stack for pairing up tokens.
     @tokens     = []             # Stream of parsed tokens in the form `['TYPE', value, location data]`.
     @seenFor    = no             # Used to recognize FORIN and FOROF tokens.
     @seenImport = no             # Used to recognize IMPORT FROM? AS? tokens.
     @seenExport = no             # Used to recognize EXPORT FROM? AS? tokens.
+    @exportList = no             # Used to identify when in an EXPORT {...} FROM? ...
 
     @chunkLine =
       opts.line or 0         # The start line for the current @chunk.
@@ -115,10 +117,18 @@ exports.Lexer = class Lexer
     if id is 'from' and @tag() is 'YIELD'
       @token 'FROM', id
       return id.length
-    if id is 'as' and @seenImport and (@tag() is 'IDENTIFIER' or @value() is '*')
-      @tokens[@tokens.length - 1][0] = 'IMPORT_ALL' if @value() is '*'
-      @token 'AS', id
-      return id.length
+    if id is 'as' and @seenImport
+      # loop hack
+      loop
+        if @value() is '*'
+          @tokens[@tokens.length - 1][0] = 'IMPORT_ALL'
+        else if @value() in COFFEE_KEYWORDS
+          @tokens[@tokens.length - 1][0] = 'IDENTIFIER'
+        else if @tag() isnt 'IDENTIFIER'
+          break
+
+        @token 'AS', id
+        return id.length
     if id is 'as' and @seenExport and @tag() is 'IDENTIFIER'
       @token 'AS', id
       return id.length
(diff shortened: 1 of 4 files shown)


test/modules.coffee
```diff
@@ -608,6 +626,30 @@ test "`as` can be used as an alias name", ->
...
+  output = """
+    import {
+      unless as bar
+    } from 'lib';
+
+    bar.barMethod();"""
+  eq toJS(input), output
+
+test "CS only keywords can't be used as unaliased names in import lists", ->
+  throws ->
```
GeoffreyBooth: This and the following test should be in `error_messages.coffee`.

src/lexer.coffee
```diff
@@ -116,10 +117,18 @@ exports.Lexer = class Lexer
     if id is 'from' and @tag() is 'YIELD'
       @token 'FROM', id
       return id.length
-    if id is 'as' and @seenImport and (@tag() is 'IDENTIFIER' or @value() is '*')
-      @tokens[@tokens.length - 1][0] = 'IMPORT_ALL' if @value() is '*'
-      @token 'AS', id
-      return id.length
+    if id is 'as' and @seenImport
+      # loop hack
```
GeoffreyBooth: “Loop hack”? Why isn’t this block:
```coffee
    if id is 'as' and @seenImport
      if @value() is '*'
        @tokens[@tokens.length - 1][0] = 'IMPORT_ALL'
      else if @value() in COFFEE_KEYWORDS
        @tokens[@tokens.length - 1][0] = 'IDENTIFIER'

      @token 'AS', id
      return id.length
```

src/lexer.coffee
```diff
@@ -137,29 +146,31 @@ exports.Lexer = class Lexer
...
-      else if tag in RELATION
-        if tag isnt 'INSTANCEOF' and @seenFor
-          tag = 'FOR' + tag
-          @seenFor = no
-        else
-          tag = 'RELATION'
-          if @value() is '!'
-            poppedToken = @tokens.pop()
-            id = '!' + id
+      unless @exportList and id in COFFEE_KEYWORDS
```
GeoffreyBooth: Combine these two lines:
```coffee
if tag is 'IDENTIFIER' and (id in JS_KEYWORDS or id in COFFEE_KEYWORDS) and
not (@exportList and id in COFFEE_KEYWORDS)
```

lydell:
If I `import {unless} from 'foo'`, I can't do anything with `unless` then, can I? What am I missing here?

GeoffreyBooth:
You would need to do `import { unless as somethingElse } from 'foo'`. This is such an edge case, I’m not sure it’s worth throwing an error warning people that they’re creating an inaccessible variable. Unless wouldn’t `import { unless } from 'foo'` be accessible via `module.unless`?

src/lexer.coffee
GabrielRatener: Because the code is more selective with the loop. In your case when encountering an `as` after an `{` on an import line, the `as` would always become an `AS` token, which would prevent its currently allowed usage as an import variable.

GeoffreyBooth: Are you sure about that?
```
coffee -ne "import { as as as } from 'lib'"
Block
  ImportDeclaration
    ImportClause
      ImportSpecifierList
        ImportSpecifier IdentifierLiteral: as IdentifierLiteral: as StringLiteral: 'lib'
```
See https://github.com/jashkenas/coffeescript/blob/master/test/modules.coffee#L587-L609

GeoffreyBooth: That’s from 1.11.1. See the link.

GabrielRatener: Yes, I haven't tried it but I think those tests would fail with the code you're suggesting.

GabrielRatener:
@lydell Right, so this PR prevents that.

src/lexer.coffee
GeoffreyBooth: Those tests pass in 1.11.1. They should still pass after your PR. I didn’t test the code I put in the comment, it was just meant to pose a question about the `loop`. I don’t like the idea of `loop`, which becomes `while (true)`. I feel like whatever you’re trying to do here, you should be able to do with a simple `if`/`else` tree. Is there a reason we need a loop?

GabrielRatener: It's not ideal, I basically wan't to reach the `@token 'AS', id` and return only for a select few token types. So basically if none of the conditionals run, I want to skip that part and not return. Since CS doesn't have `goto`, I used this as a hack.

It's possible to do it without a loop, but it would require more code.

GabrielRatener:
I moved the error tests to `error_messages.coffee`, so is this PR ready to be merged?

GeoffreyBooth:
The following is valid JavaScript, according to Babel:
```js
import { if } from 'lib';
```
I’m not sure how you would access this `if` in a runtime that natively supports modules; `module.if`/`window.if`? But regardless of that, there’s simply no way to generate this output via CoffeeScript, either before or after this PR. Maybe we don’t care, especially since even in Babel there’s no way to access this `if` unless it’s aliased?

GeoffreyBooth:
This looks good to me. @lydell, @jashkenas?

GabrielRatener:
@GeoffreyBooth I don't think we need to worry about it for now, plus Babel isn't always standards compliant anyway.

lydell:
Looks good to me too. Good work!

PR #4352, opened 2016-11-05
CS1 tagged template literals (and CS2 interpolated strings as template literals)

greghuc:
Tested implementation of tagged template literals, as defined at: https://github.com/coffeescript6/discuss/issues/28#issuecomment-247861027

For CS2, we also get interpolated strings outputting as ES6 template literals. This fell out of the tagged template literal work.


connec:
I was thinking that it would be really nice to do this as two commits, if not two PRs:

1. (hopefully) fairly simple one to compile `StringWithInterpolations` to template literals.
2. support tagged template literals (tagged string interpolations?)

greghuc:
@connec the current plan is to add tagged template literal support, as that's needed for ES6 compatibility. This feature could be published as CS 1.9.x, as it's an opt-in feature. We'd then compile `StringWithInterpolations` to template literals separately, but that has to be on the CS2 branch as it's not opt-in (all interpolated strings would start being transpiled to ES6 code).

However, your suggestion of doing `StringWithInterpolations` to template literals first might make from sense from an incremental development perspective. Today was my first day hacking on Coffeescript, so I'm definitely feeling my way on this :-)

test/error_messages.coffee
```diff
@@ -136,47 +136,48 @@ test "#1096: unexpected generated tokens", ->
...
-    [stdin]:1:2: error: unexpected string
-    a"#{b}"
-     ^^^^^^
-  '''
-  assertErrorFormat 'a"""#{b}"""', '''
-    [stdin]:1:2: error: unexpected string
-    a"""#{b}"""
-     ^^^^^^^^^^
-  '''
+# TODO: Tagged template literals impl: disabling these tests, as now valid 'tagged template literals'
```
GeoffreyBooth: Rather than commenting these out, I would move these tests into `test/tagged_template_literals.coffee` and update them accordingly. This is a pretty thorough review of all the ways tagged template literals can be written, from `a''` to `a"""b"""` and so on.

test/tagged_template_literals.coffee
```diff
@@ -0,0 +1,8 @@
+# Tagged template literals
+# ------------------------
+
+func = (text) -> "I saw: #{text}"
+
+eq 'I saw: a single line string', func'a single line string'
```
GeoffreyBooth: Don’t forget the `test` line, e.g. `test "basic tagged template literal", ->`

test/error_messages.coffee
lydell: Rather than moving and changing these tests, I would change all the `a`s to `1`s. For example, `"a''"` → `"1''"`. This is about ensuring that unexpected string errors work correctly, and is still a valid thing to test after tagged template literals are added.

test/tagged_template_literals.coffee
```diff
@@ -1,8 +1,91 @@
...
+
+  eq 'text: [multi-line double quotes] expressions: []', func"multi-line
+                                                               double quotes"
+
+  eq 'text: [multi-line\n block string] expressions: []', func"""
+                                                               multi-line
+                                                                block string
+                                                                 """
+
+test "tagged template literals: string prefix must be function", ->
```
lydell: Why not put these in test/error_messages.coffee?

greghuc: @lydell since you're around right now, are you up for a webchat about tagged template literals? I can do Skype on my github username, or Google Hangouts on ghuczynski at gmail.com

 I'm working on tagged template literals today Sunday, and I'd appreciate talking to someone with more CS compiler knowledge than me!

lydell: I’m sorry, I won’t have the time.

greghuc:
Tagged template literals status update.

So I've got tagged template literals working for pure, non-interpolated Strings (code will need tidy up before checkin).

Interpolated strings are next. This will require StringWithInterpolations in grammar.coffee to be custom-output as a JS template literal. E.g:
```
- CS input:  tagfunc"single-line double quotes #{6 + 7} interpolation"
- JS output: tagfunc`single-line double quotes ${6 + 7} interpolation`
- Not:       tagfunc"single-line double quotes " + (6 + 7) + " interpolation"
```

I'm still working out how to do this in grammar.coffee, so any input is appreciated. StringWithInterpolations is a very general class, since it's a Parens wrapping a body.

Rough rules for outputting StringWithInterpolations as a JS template literal:
* Ignore pluses between string literals and bracketed expressions
* Output bracketed expressions as ${}, not ()
- String literals have no start or end quotes
- Entire thing is wrapped in ``
- And make sure this works recursively

GeoffreyBooth:
I assume you’re not intending to support `tagfunc"a " + (6 + 7) + " b"`? That would seem to be ambiguous as to where the tagged template function ends—is it just `tagfunc"a "`, or the whole thing? I would think it should be parsed as just `tagfunc"a "`, and if people want to interpolate inside the string they need to use `#{}`, not `+`.

greghuc:
@GeoffreyBooth only ```tagfunc"a "``` would be parsed as a tagged template literal.

Using your example, here's the desired Coffeescript to Javascript compilation:
```
Coffeescript input:    tagfunc"a #{6 + 7} b"
Javascript ES6 output: tagfunc`a ${6 + 7} b`
```

At present, Coffeescript outputs ```"a #{6 + 7} b"``` to the Javascript of ```"a " + (6 + 7) + " b"```.

For tagged template literals, we need the Coffeescript to output instead as ```
`a ${6 + 7} b`
``` instead (its ES6 template literal form).

**Comment by connec** [comment]
I've had a quick look at what would need to be done to compile interpolated strings to template literals, and it's not as easy as I'd hoped since most of the work is done in the lexer.

I'd probably attempt to avoid making changes to the grammar and lexer (as far as template literals are concerned, ofc tagging will need some grammar changes). One approach could be to flesh out `StringWithInterpolations` to mirror the arguments given to tagged template literals, e.g. it could have a `strings` child for a list of `StringLiteral` nodes for the strings, and an `interpolations` child for a list of `Parens` nodes for the interpolated expressions:

```coffee
"a #{b} c"

# StringWithInterpolations
#   strings: [
#     StringLiteral 'a '
#     StringLiteral ' c'
#   ]
#   interpolations: [
#     IdentifierLiteral 'b'
#   ]
```

It should be possible to build that structure from the `body` that's currently passed to `new StringWithInterpolations` in `grammar.coffee` by flattening out the `Op +` nodes. The `compileNode` for `StringWithInterpolations` would simply iterate over the strings and interpolations, compile them and wrap them in `${...}`.

GeoffreyBooth:
Testing error messages should be in `error_messages.coffee`. I know, at this point that file is overflowing to bursting, but refactoring it to move those tests into their respective test files is a task for another PR.

The tests in that file though are only testing error messages generated by CoffeeScript. This test:
```coffee
assertErrorFormat "nofunc''", 'ReferenceError: nofunc is not defined'
```

is testing for an error thrown by Node (`ReferenceError`). We don’t include tests for Node errors; if an error is getting caught by the runtime, then we’re failing to catch it in the compiler.

This error, though, isn’t an error. `nofunc` might be undefined, but CoffeeScript doesn’t take the position that it should catch such errors; there’s no way for it to know what might be defined in the runtime environment. That’s how code like `if window.requestAnimationFrame?` works—whether `window.requestAnimationFrame` exists is evaluated by at runtime, not by the compiler. There’s no way for CoffeeScript to know whether such a line should throw an error or not.

GeoffreyBooth:
I’ve done some more cleanup, taking care of many of the items marked with `TODO`. I went through the “unexpected string” tests and confirmed that each case that is now a valid tagged template literal is covered by a test in `tagged_template_literals.coffee`, adding some new tests there. I also restored and fixed the tests that had been commented-out.

In particular note commit fe280ae47af442f07092b82cb1b201d1ed1240da: these three lines of code limit the new ES2015 syntax used for interpolated strings to only tagged template literals, so that now this PR is safe to merge into `master` and CoffeeScript 1.x. 🎉

And for CoffeeScript 2, all we need to do is delete those three lines and _all_ interpolated strings use the ES2015 syntax. So as soon as this gets merged into `master` a quick follow-up will take care of interpolated strings for CS2.

Are there any tests we can write that are more realistic? That do something similar to what a real tagged template literal is used for, which I presume is for templating? It doesn’t need to use real content, we can keep the `a`/`b` or `foo`/`bar` style, as long as what the tagged template literal function is doing is representative of how this feature gets used in the wild.

greghuc:
So tagged template literals are now working properly, and tested. And interpolated strings can now output as template literals in CS2 :-)

@GeoffreyBooth, I made some changes this morning (detailed below). I'm now happy with the code and tests, so reckon this feature this complete and ready for merging after code review. Please make any changes you see fit.

My changes:
* @GeoffreyBooth as you requested, I added a "realistic" test for tagged template literals: html templating. The `a | b` style still makes sense in other tests to check that parts of an interpolated string are correctly passed to the tag function
* I expanded the tests further, for completeness
* In `error_messages.coffee`, I moved all the old 'Unexpected string' tests into the new 'tagged template literals' section at the bottom. This makes sense, as the error messages now say "literal is not a function", not "unexpected string". Longer term, it would ideally make sense for the "literal is not a function" error message to mention tagged template literals.

GeoffreyBooth:
This looks very clean now. Great work!

So besides my question above, the only last thing I can think to check is whether this actually works with [bel](https://github.com/shama/bel), the library that [you mentioned](https://github.com/coffeescript6/discuss/issues/28) you were trying to use that required tagged template literals. Assuming you already have a project that uses bel, or can create a simple one, do you want to see if using CoffeeScript as per this PR solves the issues you were facing?

src/nodes.coffee
```diff
@@ -2233,6 +2244,42 @@ exports.Parens = class Parens extends Base
 # string concatenation inside.
 
 exports.StringWithInterpolations = class StringWithInterpolations extends Parens
+  unwrap: -> this
```
greghuc: This is a fix for a bug that appears when we compile all interpolated strings to template literals (i.e. CS2). The problem didn't show up in any tests, and I haven't successfully written a test that captures the issue.

To see the bug, write a program that just contains `"#{2}"`. Without `unwrap`, the output is the old `"" + 2;`, and not the new and expected \`2\`. I tracked the problem down to a [line in Block.compileNode](https://github.com/jashkenas/coffeescript/blob/master/src/nodes.coffee#L290), which calls `unwrapAll`, before calling `compileNode` on the result. `unwrapAll` calls `unwrap` on each descendant child node until the result doesn't change, so it seems to strip away unnecessary(?) nodes in the hierarchy, before generating code. So the bug is caused by `unwrap` on `StringWithInterpolations`, which originally returned the `body` node inside it. Block.compileNode then called compileNode on this body node (mostly an "Op + node), which returns `"" + 2;`. My investigations show that the body node can also be an "Value>StringLiteral" node in the case of empty interpolated strings. 

So it seems that before this PR, code for an interpolated string can actually be generated in two ways: by StringWithInterpolations.compileNode, or by something reaching through to the body node within it (using `unwrap`), and calling compileNode on that. By changing `unwrap` to return `this`, and not the body node, I believe the second case should not happen: all code should be generated by StringWithInterpolations.compileNode. Since StringWithInterpolations.compileNode now holds the custom logic for outputting interpolated strings as template literals, it should be the only thing responsible for code generation.

In summary:
* I believe the new `unwrap` is definitely required in CS2 when all interpolated strings become template literals. If I'm right about what `unwrap` does, then StringWithInterpolations.compileNode should be the only piece of code responsible for JS-code generation, and never a child.
* I don't believe `unwrap` is needed for CS1 tagged template literals, as I don't believe Block.compileNode is triggered: the AST always contains a function-call node, with a interpolated node as a child node. But I could be wrong!
* I would be more comfortable if there was a test for this, and someone with more CS compiler knowledge than me had a opinion about this change. I made the change, it fixed the problem, all tests still pass, so I rolled with it :-)

greghuc:
I checked the new code with a project I have which uses [yo-yo](https://github.com/maxogden/yo-yo), which in turn uses [bel](https://github.com/shama/bel). It worked fine: I was able to do html templating in a pure coffeescript file. Yay!

src/nodes.coffee
GeoffreyBooth: `"#{2}"` isn’t pointless, as it allows you to cast a value as a string. I know I’ve written the code `"" + foo` on occasion. It does come up.

GeoffreyBooth: Also, `"#{2}"` should compile in CoffeeScript 2 to `` `${2}` ``.

GeoffreyBooth:
I commented out `unwrap` for now, with a comment that it should be reenabled for CoffeeScript 2. I think this PR is probably good to go. @jashkenas or @lydell, any notes?

jashkenas:
This looks fine to me from a quick glance.

As an aside — for this sort of thing, it feels like you guys are stealing a little bit of your own thunder by maintaining the CS1 compatibility. If you just released the even-more-streamlined version for CS2 only, I bet folks would get more excited about CS2. In short: It's awfully nice of you.

GeoffreyBooth:
@jashkenas Yeah, but the code is right here for anyone to adapt as a 1.x PR. It was written originally as a 2-only PR, but it was made 1.x-friendly by commenting out about three lines of code, so it was inevitable that people would figure it out 😄 

The 2 branch has two big features that should hopefully motivate people to upgrade: classes that can extend ES classes, and async/await. Those two alone should be enough for a big chunk of people.

GeoffreyBooth:
@greghuc Congratulations! Would you like to follow this up with a PR adding documentation for this? And another PR for `2` that outputs all interpolated strings as ES template literals?

greghuc:
@GeoffreyBooth thanks! Re the PRs, I'll aim to fit them in over the next week.

Quick question: where's the right place to be doing the follow-on PRs?
* For template literal documentation, I presume we're talking about the current CS1 documentation
* For interpolated strings in CS2, what/where is the correct CS2 branch to use? This PR will be a bit more than just uncommenting the lines discussed. As I mentioned before, StringWithInterpolations should extend Base - not Parens - so will need some tweaking, and likely some input from people with more knowledge of nodes.coffee

GeoffreyBooth:
@greghuc for documentation yes, the current 1.x docs. They’re in the `documentation` folder; the `docs` folder is for the generated output. Please don’t commit anything in the `docs` folder; that only gets updated when we release. Branch off of `master`.

For interpolated strings, branch off of `2`. I’ve already merged `master` into it, so it should have your latest tagged template literals code.

lydell:
@greghuc I've done plenty of refactoring regarding strings with interpolations – all the way from the lexer to nodes.coffee. Feel free to ping me in PRs and I'll try to help out.

PR #4353, opened 2016-11-05
Compile splats in arrays and function calls to ES2015 splats

connec:
Rather than compiling splats to arrays built using `Array#concat`, splats are now compiled directly to ES2015 splats, e.g.

```coffee
f foo, arguments..., bar
[ foo, arguments..., bar ]
```

Which used to be compiled to:

```js
f.apply(null, [foo].concat(slice.call(arguments), [bar]));
[foo].concat(slice.call(arguments), [bar]);
```

Is now compiled to:

```js
f(foo, ...arguments, bar);
[ foo, ...arguments, bar ];
```

This was really easy to implement, and mostly involved deleting code. No tests were broken, and I can't think of any extra tests to add.


GeoffreyBooth:
This . . . looks awesome. Could it really be this simple? I guess this _was_ low-hanging fruit.

The changes (simplifications) to the compiled code kind of speaks for itself. I can confirm that the tests (including harmony ones) still pass. @lydell, anyone else, any other notes?

GeoffreyBooth:
Do we need any _new_ tests to cover anything related to outputting ES2015 splat syntax?

Also I’d love to make the side that the `...` go on optional, so that both `args...` and `...args` work, but that should be a separate PR.

lydell:
Implementation-wise I can’t think of anything else to do. Good job!

Testing-wise, I guess we could add a test involving generators.

```
$ node -p 'a = function*(){yield 1; yield 2; yield 3}; [...a()]'
[ 1, 2, 3 ]

$ coffee -e 'a = (-> yield 1; yield 2; yield 3); console.log [a()...]'
[]
```

With this PR, I’d expect both of the above commands to output `[ 1, 2, 3 ]`.

> Also I’d love to make the side that the `...` go on optional, so that both `args...` and `...args` work, but that should be a separate PR.

Me too! I’d even like to have that in master. I also agree that it should be a separate PR.

connec:
>> Also I’d love to make the side that the ... go on optional, so that both args... and ...args work, but that should be a separate PR.

> Me too! I’d even like to have that in master. I also agree that it should be a separate PR.

Aww :( I feel like we should have 'one way' of doing splats. I'd lament switching to ES2015's prefix notation, but that'd be better than doing both imo.

At any rate, I've added that test and it passes fine 😄

lydell:
I guess we’re good to go then!

GeoffreyBooth:
Thanks @connec, great work!

Issue #3760, opened 2014-12-15
Backtick + javascript newline escape = error: reserved word "var"


yetanotherusernamebecausegithubisbugged:
Backticked some old code to start step-by-step rewriting it in coffeescript and have discovered this confusing error message.

``` coffeescript
`
var str = "long\
string";
`
```

```
$ coffee -p test97.coffee 
Running node v0.11.13
/home/me/tst/test97.coffee:2:1: error: reserved word "var"
var str = "long\
^
```

SaranNV:
hi yetanotherusernamebecausegithubisbugged ,
  I also have this same issue on javascript.i dont know how to fix it..

yetanotherusernamebecausegithubisbugged:
Try to remove all the newline slashes. As long as I remember, it'll help.

SaranNV:
hi yetanotherusernamebecausegithubisbugged,

This is the below error line in the javascript (Gruntfile.js).it showing error as var error:reserved word error
var coffeeFiles = [
  'javascripts/hub_client/augury.js.coffee',
  'javascripts/hub_client/common/**/*.coffee',
  'javascripts/hub_client/extensions/**/*.coffee',
  'javascripts/hub_client/mixins/*.coffee',
  'javascripts/hub_client/events/**/*.coffee',
  'javascripts/hub_client/models/**/*.coffee',
  'javascripts/hub_client/collections/**/*.coffee',
  'javascripts/hub_client/routers/**/*.coffee',
  'javascripts/hub_client/templates/**/*.coffee',
  'javascripts/hub_client/views/**/*.coffee'
]

yetanotherusernamebecausegithubisbugged:
It compiles for me. Probably you have forgot to put a backticks around the javascript code?
```
`
var coffeeFiles = [
'javascripts/hub_client/augury.js.coffee',
'javascripts/hub_client/common//*.coffee',
'javascripts/hub_client/extensions//.coffee',
'javascripts/hub_client/mixins/.coffee',
'javascripts/hub_client/events//*.coffee',
'javascripts/hub_client/models//.coffee',
'javascripts/hub_client/collections//.coffee',
'javascripts/hub_client/routers//*.coffee',
'javascripts/hub_client/templates//.coffee',
'javascripts/hub_client/views//.coffee'
]
`
```
?

SaranNV:
The backticks does not help me.

GeoffreyBooth:
Fixed by #4357

PR #4357, opened 2016-11-14
Triple backticks to allow creation of JavaScript blocks

### LINKED ISSUE #1504: New escape javascript (`) token
I'm building an asset bundler that can merge both coffeescript and javascript together. It works fine placing coffeescript inside javascript -- compile it and stick it in the js file. 

The problem I'm running into is when I try to include javascript into coffeescript. My approach was to bring in the javascript, escape with **`**, and then compile the coffeescript.

The problem is _many_ libraries use **`** in comments -- including backbone. One thought is to use **```** to escape javascript like comment blocks in coffeescript.

Please let me know your thoughts on this.

**After some discussion:** The whole point of this issue is to point out that the single backtick _does not always_ work the way it's intended. Use cases and outside discussion should be irrelevant here. Currently, the backtick does not work if you try to escape a file like backbone or jquery or even a two line javascript file that has a comment that includes a backtick. 

Thanks,
Matt

GeoffreyBooth:
Closes #1504 and #3760. Per [Jeremy’s comment from 2011](https://github.com/jashkenas/coffeescript/issues/1504#issuecomment-1554215), this adds support for “heredoc” JavaScript blocks, e.g.:

    ```
    var a = 1;
    var b = 2;
    ```
    c = 3
    ```var d = 4;```
    eq a + b + c + d, 10

Escaped backticks are output as just backticks:

    ```var e = \`hello, world${'!'}\````
    eq e, 'hello, world!'


test/javascript_literals.coffee
```diff
@@ -1,10 +1,32 @@
...
+  ```
+  c = 3
+  ```var d = 4;```
+  eq a + b + c + d, 10
+
+test "block inline JavaScript containing backticks", ->
+  ```
+  // This is a comment with `backticks`
+  var a = 42;
+  var b = `foo ${'bar'}`;
```
greghuc: There's no 'eq' for variable b. Is this correct?

src/lexer.coffee
```diff
@@ -900,7 +907,7 @@ CODE       = /^[-=]>/
 
 MULTI_DENT = /^(?:\n[^\n\S]*)+/
 
-JSTOKEN    = /^`[^\\`]*(?:\\.[^\\`]*)*`/
+JSTOKEN    = /^```([\s\S]*?)```|^`[^\\`]*(?:\\.[^\\`]*)*`/
```
greghuc: I read the added regex part as "greedily capture everything between starting and end triple backquotes, which seems sane.

The original regex is rather obtuse. I read it as "greedily read everything between starting and ending single backquotes, and make sure every backslash has a character following it". I'm thinking the backslash+character part exists to stop an escaped backslash preemptively closing the Javascript block. E.g:
```
OK: `a\``
Not OK: `a\`
```

The new triple backticks regex will behave differently:

```
OK: ```a\````
Also OK: ```a````
```

I found this website useful: https://regex101.com/

jashkenas:
This is great! I'd only suggest adding some complicated tests involving nested levels of backticks / interpolation. That's usually where things get tricky to lex.

src/lexer.coffee
GeoffreyBooth: #3760 is actually complaining about just what you describe: a backslash not followed by a character (in that case, it's followed by a newline). Perhaps the original regex is incorrect here?

What would you use as the regex, if you don't mind taking the time to play with it?

GeoffreyBooth:
@jashkenas more complicated than [this](https://github.com/GeoffreyBooth/coffeescript/blob/edd36b5cdad4d42c5f902debc6e080c6c664ed8f/test/javascript_literals.coffee#L23)?

lydell:
Could somebody explain with words:

- When a `` ` `` JS block ends.
- When a ```` ``` ```` JS block ends.

Do backslashes inside JS block affect where the block ends?

jashkenas:
I'm probably not the expert in the details of it — but I'd imagine that the idea is to be sensible about escaping the backticks in the same way that we try to be sensible about escaping quotes in a heredoc.

So: **A ``` JS block ends at the first unescaped and uninterpolated ``` that follows.**

So: **A ````` JS block ends at the first unescaped and uninterpolated ````` that follows.**

GeoffreyBooth:
@jashkenas What do you mean by uninterpolated? I would’ve answered this way:

* A `` ` `` JS block ends at the first unescaped `` ` `` that follows.

* A ```` ```  ```` JS block ends at the first ```` ```  ```` that follows.

lydell:
```coffee
`\`a\n\``
```
↓
```js
`a\n`;
```

Is the above correct? Just checking if I understand the rules or not.

jashkenas:
@lydell — The above looks correct to me.

@GeoffreyBooth: By uninterpolated, I mean this:

`` `before #{"str`ing"} after` ``

The backtick inside the interpolation doesn't kill the expression halfway. Ideally the same would be true for triple backticks.

GeoffreyBooth:
@jashkenas I wasn’t aware that we supported interpolation between backticks, or wanted to.

@lydell Your example, if saved into a file that is compiled via `coffee -bp test.coffee`, outputs:
```js
\`a\n\`;
```

So clearly somehow my “escaping works!” test is flawed. I guess some escaping is happening in the test runner itself? Anyway it appears we’re not escaping backticks (that is, outputting just `` ` `` for `` \` ``), we’re just skipping over escaped backticks until we find a non-escaped one to end the token on. I can’t see any benefit to our current behavior.

If we replace `` ` `` with `` \` `` during compilation, then there becomes no way to actually output `` \` `` in the resulting JavaScript. No great loss? I don’t want to get into escaping backslashes.

jashkenas:
> @jashkenas I wasn’t aware that we supported interpolation between backticks, or wanted to.

Yes. We don't. My mistake.

connec:
Might be worth noting that the main problem in the issues you linked was for tagged template literals, which will hopefully soon be merged.

GeoffreyBooth:
@connec Yes, but this should get fixed anyway. This future-proofs us against any other weird things ECMAScript wants to use backticks for.

I figured out why my test passed when you wouldn’t expect it to: JavaScript converts `` \` `` into `` ` `` wherever it sees it:
image
So the test should compare a string against the expected JavaScript output, not the executed JavaScript.

I’ll update this PR soon, but just so we’re clear on the intended behavior:
* A `` ` `` JS block ends at the first unescaped `` ` `` that follows. Escaped backticks are converted to simple backticks (i.e. `` \` `` becomes `` ` ``).

* A ```` ```  ```` JS block ends at the first ```` ```  ```` that follows. Escaped *single* backticks are converted to simple backticks (i.e. `` \` `` becomes `` ` ``) so if you really want a triple-backtick in the generated JavaScript, you would write `` \`\`\` ``.

?

src/lexer.coffee
GeoffreyBooth: Per my last comment, I’m suggesting we convert `` \` `` to `` ` `` wherever `` \` `` appears inside a single- or triple-backtick block. That should cover this case, yes?

connec:
That sounds fine to me. I wonder if we can leverage the code paths used to handle heredocs and heregexen? This is an area of the parser I've not looked at.

GeoffreyBooth:
Okay, it took quite some doing to get ````` ```\`hello\```` ````` to match out to the last backtick, and not one before it (i.e. treating the `` \` `` as not part of the closing triple backticks). @connec or anyone else, can you please check the new regex?

I also updated the regex so that dangling backslashes at the end of a line no longer stop the capture, which explicitly fixes #3760 (as opposed to providing the new triple backticks which allow working around the issue). And escaped backticks are now replaced with just backticks in the JavaScript code, so you can write code like this:

```
`let greeting = \`hello, ${name}\``
```

I think this is everything?

lydell:
I think I'd approach this problem something like this:

```coffee
JSTOKEN      = ///^ `(?!``) ((?: [^`\\] | \\[\s\S]           )*) `   ///
HERE_JSTOKEN = ///^ ```     ((?: [^`\\] | \\[\s\S] | `(?!``) )*) ``` ///

test = (chunk) ->
  match = HERE_JSTOKEN.exec(chunk) or JSTOKEN.exec(chunk)
  console.log chunk, match
  if match
    script = match[1].replace(/\\+(`|$)/g, (string) ->
      string[-Math.ceil(string.length / 2)..]
    )
    return script
  return null

console.log([
  test('``````') # == ''
  test('`\\\\`') # == '\\'
  test('`\\\\\\`a`') # == '\\`a'
  test('```\\```') # == null
])
```

GeoffreyBooth:
@lydell I'm not sure I follow; aside from more readable regexes, what problems with the PR does your version solve?

lydell:
You should be able to output a backslash before a backtick, or at the end. For example ``` `\\\`` ``` → `` \`; `` and `` `\\` `` → `\;`. (We should add tests for even crazier numbers of backslashes as well, and tests for invalid stuff like ```` ```\``` ````.)

GeoffreyBooth:
@lydell I would think that ``` `\\\`` ``` is supposed to become `` \\`; ``, and it does in this PR. Likewise, I think `` `\\` `` should become `\\;`, as it does. It’s not our job to escape all JavaScript escaped characters; if the user types `\\` between backticks, it should just be output as `\\`. That’s the current behavior before this PR, so changing it could break backward compatibility. A backslash is normally escaped in JavaScript, so it’s probably pretty common to find `\\` in backticked code.

Technically speaking, converting `` \` `` to `` ` `` is also breaking backward compatibility, but I think this is okay since the failure to do this conversion is more of a bug than a feature. The only time a backtick is ever escaped in JavaScript is within a template literal, but people basically haven’t been able to write those yet in CoffeeScript because of the lack of any way to escape backticks.

lydell:
I only meant that backslashes should be handled specially before backticks and at the end of the block (which my code does). In other words, `\\` and `\n`, for example, are untouched. Otherwise you wouldn't be able to express _anything_ inside a JS block (which is the point of this PR, isn't it?). For example, a template literal ending in a backslash: ``` `\`a\\\`` ``` → ``` `a\`` ```

GeoffreyBooth:
@lydell can you please explain
```coffee
script = match[1].replace /\\+(`|$)/g, (string) ->
      string[-Math.ceil(string.length / 2)..]
```
? I replaced my regex with yours and adjusted the code as appropriate, but when I add this new `replace` function the first JavaScript literal test fails.

src/lexer.coffee
```diff
@@ -292,9 +292,13 @@ exports.Lexer = class Lexer
 
   # Matches JavaScript interpolated directly into the source via backticks.
   jsToken: ->
-    return 0 unless @chunk.charAt(0) is '`' and match = JSTOKEN.exec @chunk
-    @token 'JS', (script = match[0])[1...-1], 0, script.length
-    script.length
+    return 0 unless @chunk.charAt(0) is '`' and
+      (match = HERE_JSTOKEN.exec(@chunk) or JSTOKEN.exec(@chunk))
+    [js] = match
+    script = if js[0..2] is '```' then js[3...-3] else js[1...-1]
```
lydell: Why not use the capturing group instead of this trickery? `match[1]`

GeoffreyBooth: The old regex didn’t have a capturing group (see the original code that this PR revises). But yes, `match[1]` is better.

Can you please explain your `replace` function? Or rewrite it to itself use a lot less sorcery? What exactly are we trying to replace, and with what, when it comes to escaped backticks?

test/javascript_literals.coffee
```diff
@@ -1,10 +1,47 @@
...
+# JavaScript Literals
 # -------------------
 
-# TODO: refactor javascript literal tests
-# TODO: add indexing and method invocation tests: `[1]`[0] is 1, `function(){}`.call()
+test "inline JavaScript is evaluated", ->
+  eq '\\`', `
+    // Inline JS
+    "\\\`"
+  `
```
lydell: With my proposed escaping rules, this test needs to be written like this:

```coffee
eq '\\`', `
  // Inline JS
  "\\\\\`"
`
```

or like this:

```coffee
eq '\\a`', `
  // Inline JS
  "\\a\`"
`
```

src/lexer.coffee
lydell:
```
Considering the raw text inside a JSTOKEN:

\` → `
\\` → can't happen
\\\` → \`
\\\\` → can't happen
\\\\\` → \\`
...

Go from left to right. Take a backslash and the character after it.
Replace with the second of those two. That's the rule.
Or, observe the pattern: We simply end up with the second half
of a `` \\\\\\\` `` sequence!

And, for completeness, at the end of the JSTOKEN:

\ → can't happen
\\ → \
\\\ → can't happen
\\\\ → \\
\\\\\ → can't happen
...
```

GeoffreyBooth: So double backslashes are left alone _except_ when followed by an escaped backtick _or_ the end of the JS block. Seems a bit convoluted to try to explain in documentation. Can we come up with a simpler rule?

Currently the PR always converts `` \` `` to a backtick, and that's it. So there's no way to render an escaped backtick inside a template literal. One way to solve that problem is to convert the escaped backticks only in the single-backtick tokens, and leave the triple-backtick ones alone. If some needs an escaped backtick in the JS, they use triple backticks. If someone needs three consecutive backticks in their JS, they use single backticks and escape the triplet.

vendethiel: https://xkcd.com/1638/ I guess

connec: This behaviour makes sense to me, I'd describe it as:

> A backslash (`\`) can be used to escape a backtick (`` ` ``) or another backslash in a JS literal, and will otherwise be passed through.
> ```coffee
`hello`                                       # hello;
`\`hello\``                                   # `hello`;
`\`Escaping backticks in JS: \\\`hello\\\`\`` # `Escaping backticks in JS: \`hello\``;
`Single backslash: \ `                        # Single backslash: \ ;
`Double backslash: \\\\`                      # Double backslash: \\;
```

Comment (GeoffreyBooth) [inline src/lexer.coffee]: If we start requiring backslashes to be escaped, that could break existing code. I think it's better to only convert `` \` ``, only within single-backtick blocks. That preserves backward compatibility, and gives people a way to output an escaped backtick (via a triple-backtick block).

**Comment by GeoffreyBooth** [comment]
Okay, I’ll concede to the preferences of the crowd. I’ve updated the `replace` function per @lydell.

@connec, I added your examples as a new test, though I had to escape all the backslashes for them to be parsed correctly as strings. Since that made them hard to understand, I included all the before-and-afters as a big comment just above the test. To double-check it was really working, I created a `test.coffee` file and pasted your code in there:
```coffee
`hello`                                       # hello;
`\`hello\``                                   # `hello`;
`\`Escaping backticks in JS: \\\`hello\\\`\`` # `Escaping backticks in JS: \`hello\``;
`Single backslash: \ `                        # Single backslash: \ ;
`Single backslash at EOS: \\`                 # Single backslash at EOS: \;
`Double backslash at EOS: \\\\`               # Double backslash at EOS: \\;
`Double backslash: \\\ `                      # Double backslash: \\ ;
```
It becomes:
```js
hello;
`hello`;
`Escaping backticks in JS: \`hello\``;
Single backslash: \ ;
Single backslash at EOS: \;
Double backslash at EOS: \\;
Double backslash: \\\ ;
```
The only difference is the last one: you had it outputting only two backslashes, instead of the three I get. But there’s no backtick following these backslashes, so I don’t think the lexer should be affecting them.

Is everyone happy with this?

lydell:
```
`Double backslash: \\\ `                      # Double backslash: \\ ;
```

The above is nothing but wrong and misleading.

---

Do we have any tests for that `` `'\\\n'` `` becomes `'\\\n'` (in other words, making sure that we don't mess up other backslashes)?

GeoffreyBooth:
@lydell Other backslashes are unaffected. I added a test.

connec:
OK, cool.

~~I would definitely remove the `Double backslash: \\\ ;` test, or rename it to `Triple backslash: \\\ ;`, because as @lydell says it's currently misleading~~ - Whoops, looks like that's already changed.

This seems good to me. I'm still slightly dubious about just escaping `` ` ``s, but it makes sense that the literal JS should be as close as possible to what's written.

jashkenas:
I don't think that I can mentally expand these regexes well enough to comment on them materially — but if @lydell and @connec sign off on this, then I'm cool with it too.

GeoffreyBooth:
@lydell and @connec, do you have any other notes or should we merge this in?

vendethiel:
LGTM

lydell:
LGTM

PR #4369, opened 2016-11-20
Docs for triple-backticks and escaping backticks

GeoffreyBooth:
Documentation for #4357. Demonstrates the new triple-backtick syntax, and how to escape backticks and escaped escaped backticks.

### DIFF
--- a/documentation/index.html.js
+++ b/documentation/index.html.js
@@ -887,8 +887,8 @@ Block
     </p>
     <%= codeFor('embedded_escaped', 'markdown()') %>
     <p>
-      You can embed blocks of JavaScript using triple backticks. Within triple-backtick
-      blocks, single backticks are ignored.
+      You can also embed blocks of JavaScript using triple backticks. That's easier
+      than escaping backticks, if you need them inside your JavaScript block.
     </p>
     <%= codeFor('embedded_block', 'time()') %>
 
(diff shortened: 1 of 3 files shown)


documentation/index.html.js
```diff
@@ -878,6 +878,19 @@ Block
...
+      Escape backticks with backslashes: <code>\`</code> becomes <code>`</code>.
+    </p>
+    <p>
+      Escape backslashes before backticks with more backslashes: <code>\\\`</code>
+      becomes <code>\`</code>.
+    </p>
+    <%= codeFor('embedded_escaped', 'markdown()') %>
+    <p>
+      You can embed blocks of JavaScript using triple backticks. Within triple-backtick
+      blocks, single backticks are ignored.
```
lydell: > Within triple-backtick blocks, single backticks are ignored.

That sentence stood out a bit to me while first reading it. "Ignored – how? What exactly does that mean?" Then, I realized that the sentence probably exists to tell that it is a better idea to use triple-backticks than to escape backticks inside single-backtick blocks. If that's the case, perhaps change this paragraph to something like this:

> You can also embed blocks of JavaScript using triple backticks. That's easier than escaping backticks, if you need them inside your JavaScript block.

jashkenas: Yes, that would be an improvement.

GeoffreyBooth:
Sure. How about now?

PR #4375, opened 2016-11-28
Fix the browser compiler crashing on extended characters in script blocks

### LINKED ISSUE #4263: script type=text/coffeescript tag does not support unicode (utf-8) properly
Running this html file, saved with utf-8 encoding (no BOM) in Chrome:

```
<meta charset=utf-8>
<script src=/lib/coffee-script.js></script>
<script>
alert('js:おねがいします')
</script>
<script type=text/coffeescript>
alert('cs:おねがいします')
</script>
```

generates

```
coffee-script.js:6459 Uncaught InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
```

The error is generated on the lines saying

```
      case typeof btoa !== 'function':
        return btoa(src);
```

in coffee-script.js

The javascript tag has no problems with this string.

GeoffreyBooth:
Closes #4263, which is rather important because `test/test.html` has Japanese characters in a comment and therefore it doesn’t actually run in a browser without this fix 😄

### DIFF
--- a/lib/coffee-script/coffee-script.js
+++ b/lib/coffee-script/coffee-script.js
@@ -28,7 +28,9 @@
       case typeof Buffer !== 'function':
         return new Buffer(src).toString('base64');
       case typeof btoa !== 'function':
-        return btoa(src);
+        return btoa(encodeURIComponent(src).replace(/%([0-9A-F]{2})/g, function(match, p1) {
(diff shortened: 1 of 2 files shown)


jashkenas:
Super gross hack! But nice save!

PR #4378, opened 2016-11-30
Fix browser tests

GeoffreyBooth:
This PR gets `test.html` working in the browser again. It had fallen a few years behind how tests were run in `Cakefile`, and over that time many tests started using helper functions from Node’s `assert` that aren’t available in the browser; and `test.html`‘s method of checking for test passes/fails diverged at some point from `Cakefile`‘s, so there were false positives (or incorrect negatives?) in the `test.html` test results. This PR fixes all of this, and refactors `Cakefile` and `test.html` so that shared code is not duplicated. The test for UTF-8 characters in comments (#4375) has been moved from `test.html` into `test/comments.coffee`.

I didn’t commit any generated files output under `docs` because I don’t want to influence the current docs, so to see this in action after you’ve checked out this branch:
- `cake build:browser`
- `cake doc:site`
- Open `docs/v1/test.html` in a browser (a `file://` link is fine)


jashkenas:
Thanks for getting these up and running again!

PR #4383, opened 2016-12-01
Escape backticks and `${` within template literals; fixes #4380

### LINKED ISSUE #4380: [CS2] Interpolated strings containing backticks are not properly escaped.
If a backtick is found inside an interpolated string, it doesn't get escaped, causing syntax errors.

Input:
```coffeescript
console.log("This is a test: `#{process.pid}`")
```
Result:
```javascript
console.log(`This is a test: `${process.pid}``);
```
Expected result:
```javascript
console.log(`This is a test: \`${process.pid}\``);
```

GeoffreyBooth:
Closes #4380.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2274,7 +2274,11 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Parens
     fragments.push @makeCode '`'
     for element in elements
       if element instanceof StringLiteral
-        fragments.push @makeCode element.value.slice(1, -1)
+        value = element.value.slice(1, -1)
+        # Backticks or `${` inside template literals must be escaped
+        value = value.replace /`/g, '\\`'
+        value = value.replace /\${/g, '\\${'
+        fragments.push @makeCode value
       else
         fragments.push @makeCode '${'
         fragments.push element.compileToFragments(o, LEVEL_PAREN)...
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2276,8 +2276,11 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Parens
       if element instanceof StringLiteral
         value = element.value.slice(1, -1)
         # Backticks or `${` inside template literals must be escaped
-        value = value.replace /`/g, '\\`'
-        value = value.replace /\${/g, '\\${'
+        value = value.replace /(\\*)(`|\$\{)/g, (match, backslashes, toBeEscaped) ->
+          if backslashes.length % 2 is 0
+            "#{backslashes}\\#{toBeEscaped}"
+          else
+            match
         fragments.push @makeCode value
       else
         fragments.push @makeCode '${'
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2274,8 +2274,8 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Parens
     fragments.push @makeCode '`'
     for element in elements
       if element instanceof StringLiteral
-        value = element.value.slice(1, -1)
-        # Backticks or `${` inside template literals must be escaped
+        value = element.value[1...-1]
+        # Backticks and `${` inside template literals must be escaped.
         value = value.replace /(\\*)(`|\$\{)/g, (match, backslashes, toBeEscaped) ->
           if backslashes.length % 2 is 0
             "#{backslashes}\\#{toBeEscaped}"
--- a/test/tagged_template_literals.coffee
+++ b/test/tagged_template_literals.coffee
@@ -146,7 +146,7 @@ test "tagged template literal with escaped backticks", ->
   eq 'text: [ES template literals look like this: \\`foo bar\\`] expressions: []',
   func"ES template literals look like this: \\`foo bar\\`"
 
-test "tagged template literal with invalidly escaped backticks", ->
+test "tagged template literal with unnecessarily escaped backticks", ->
   eq 'text: [ES template literals look like this: `foo bar`] expressions: []',
   func"ES template literals look like this: \`foo bar\`"
 
@@ -162,7 +162,10 @@ test "tagged template literal with escaped ES interpolation", ->
   eq 'text: [ES template literals also look like this: `3 + 5 = \\${3+5}`] expressions: []',
   func"ES template literals also look like this: `3 + 5 = \\${3+5}`"
 
-test "tagged template literal with invalidly escaped ES interpolation", ->
-  eq 'text: [ES template literals also look like this: `3 + 5 = \${3+5}`] expressions: []',
+test "tagged template literal with unnecessarily escaped ES interpolation", ->
+  eq 'text: [ES template literals also look like this: `3 + 5 = ${3+5}`] expressions: []',
   func"ES template literals also look like this: `3 + 5 = \${3+5}`"
 
+test "tagged template literal special escaping", ->
+  eq 'text: [` ` \\` \\` \\\\` $ { ${ ${ \\${ \\${ \\\\${ | ` ${] expressions: [1]',
+  func"` \` \\` \\\` \\\\` $ { ${ \${ \\${ \\\${ \\\\${ #{1} ` ${"
--- a/test/tagged_template_literals.coffee
+++ b/test/tagged_template_literals.coffee
@@ -138,14 +138,31 @@ test "tagged template literal with an interpolated string that contains a tagged
   eq 'text: [inner tagged | literal] expressions: [text: [|] expressions: [template]]',
   func"inner tagged #{func"#{'template'}"} literal"
 
-test "tagged template literal with escaped backticks", ->
+test "tagged template literal with backticks", ->
   eq 'text: [ES template literals look like this: `foo bar`] expressions: []',
   func"ES template literals look like this: `foo bar`"
 
-test "tagged template literal with escaped interpolation", ->
+test "tagged template literal with escaped backticks", ->
+  eq 'text: [ES template literals look like this: \\`foo bar\\`] expressions: []',
+  func"ES template literals look like this: \\`foo bar\\`"
+
+test "tagged template literal with invalidly escaped backticks", ->
+  eq 'text: [ES template literals look like this: `foo bar`] expressions: []',
+  func"ES template literals look like this: \`foo bar\`"
+
+test "tagged template literal with ES interpolation", ->
   eq 'text: [ES template literals also look like this: `3 + 5 = ${3+5}`] expressions: []',
   func"ES template literals also look like this: `3 + 5 = ${3+5}`"
 
-test "tagged template literal with both escaped and unescaped interpolation", ->
+test "tagged template literal with both ES and CoffeeScript interpolation", ->
   eq "text: [ES template literals also look like this: `3 + 5 = ${3+5}` which equals |] expressions: [8]",
   func"ES template literals also look like this: `3 + 5 = ${3+5}` which equals #{3+5}"
+
+test "tagged template literal with escaped ES interpolation", ->
+  eq 'text: [ES template literals also look like this: `3 + 5 = \\${3+5}`] expressions: []',
+  func"ES template literals also look like this: `3 + 5 = \\${3+5}`"
+
+test "tagged template literal with invalidly escaped ES interpolation", ->
+  eq 'text: [ES template literals also look like this: `3 + 5 = \${3+5}`] expressions: []',
+  func"ES template literals also look like this: `3 + 5 = \${3+5}`"
+
--- a/test/tagged_template_literals.coffee
+++ b/test/tagged_template_literals.coffee
@@ -137,3 +137,15 @@ test "tagged template literal with an interpolated string that itself contains a
 test "tagged template literal with an interpolated string that contains a tagged template literal", ->
   eq 'text: [inner tagged | literal] expressions: [text: [|] expressions: [template]]',
   func"inner tagged #{func"#{'template'}"} literal"
+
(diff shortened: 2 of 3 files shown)


jashkenas:
You don't have to escape `}` also? I'm not quite sure how the rules work for the JS parsing of these.

GeoffreyBooth:
I don’t think so, just backtick or `${`: http://exploringjs.com/es6/ch_template-literals.html#_escaping-in-template-literals or https://hacks.mozilla.org/2015/05/es6-in-depth-template-strings-2/

This works in Chrome:
```js
foo = `this is \${ escaped}`
```

greghuc:
Fix looks good.

GeoffreyBooth:
@greghuc sorry I stole your thunder, I’m just eager to get 1.12 out 😄  and the fix was easy.

connec:
Does this not have the same issue that was debated around introducing ```` ``` ```` in that we need to also escape any escapes, e.g. would `` "\`" `` not compile to ``` `\\`` ``` (and break).

GeoffreyBooth:
@connec I don’t think so. `` \` ``, in the context of a CoffeeScript string, is invalid. It would be like typing `` " \` " `` in JavaScript—in this context, a backtick is not an escapable character. Per Chrome:
image
So if someone wants `` \` ``, they need to type `` \\` ``. We don’t need to double-escape or do anything special around escaped backticks here, because the characters that delimit the string are quotation marks.

GeoffreyBooth:
Per this [SO answer](http://stackoverflow.com/a/35804167/223225), the spec at [11.8.6 Template Literal Lexical Components][1] defines a template character as:

**$** [ lookahead ≠ { ]
**\\** _EscapeSequence_
_LineContinuation_
_LineTerminatorSequence_
_SourceCharacter_ but not one of **`** or **\\** or **$** or _LineTerminator_

The important part is that last line: that means the only characters we need to escape are a backtick and a dollar sign when followed by an open curly brace. Backslashes need escaping too, but that’s on the user to do; just as with regular strings, we don’t convert a single backslash into two backslashes.

  [1]: http://www.ecma-international.org/ecma-262/6.0/#sec-template-literal-lexical-components

lydell:
```
$ ./bin/coffee -bpe 'a"\`"'
a`\\``;
```

The above is invalid JavaScript. I'd expect the following:

```
$ ./bin/coffee -bpe 'a"\`"'
a`\``;
```

Here is my implementation of the escaping:

```coffee
replace = (string) ->
  string.replace /(\\*)(`|\$\{)/g, (match, backslashes, toBeEscaped) ->
    if backslashes.length % 2 is 0
      "#{backslashes}\\#{toBeEscaped}"
    else
      match
```

Given `first = ([x]) -> x`, here are a few test cases:

```coffee
eq '\\`', replace '`'         # ` → \`
eq '\\`', replace '\\`'       # \` → \`
eq '\\\\\\`', replace '\\\\`' # \\` → \\\`

# Add tests with the backticks replaced with ${ too ...
```

In other words, `` ` `` and `${` should only be escaped if they weren't already.

greghuc:
I don't think ```[CS] a"\`" => [JS] a`\`` => [String] "`"``` is correct.

Instead, I think we should aim to output a String that is the exact value specified at the Coffeescript level.

Which gives ```[CS] a"\`" -> [JS] a`\\\`` => [String] "\`"```

I am unclear on current Coffeescript string escaping rules, but I don't think the backtick `` ` `` character is special. As such, a CS string of ``"\`"`` should be treated literally: as a backslash followed by a backtick.

greghuc:
Actually, I'm thinking that @lydell's suggestion and my alternative are both valid. It depends on which particular behaviour we want:
* CS String has JS String semantics: ```[CS] a"\`" => [JS] a`\`` => [String] "`"```
* JS String is literal output of CS String: ```[CS] a"\`" -> [JS] a`\\\`` => [String] "\`"```

I'm still leaning towards "JS String is literal output of CS String", as backticks aren't special characters in CS, so why treat them differently. I'm also thinking literal output is more understandable+intuitive for users.

lydell:
Since ``"\`" is "`"``, then ``a"\`" is a"`"`` must be true as well.

greghuc:
The two template literals in the example above lead to different strings. From Chrome dev console:

screen shot 2016-12-02 at 09 11 49
vs
screen shot 2016-12-02 at 09 11 02

greghuc:
Ah @lydell, I see what you mean about strings and tagged template literals having the same behaviour.

In CS2, interpolated strings will also be output as tagged template literals. And non-interpolated strings and interpolated strings should behave in the same way.

As such, my "JS String is literal output of CS String" suggestion won't work, as it would cause ``"\`" `` to behave differently in non-interpolated-strings vs interpolated-strings / tagged-template-literals.

So we should be consistent with Javascript String semantics.

GeoffreyBooth:
> In other words, `` ` `` and `${` should only be escaped if they weren't already.

I don’t think this is correct. Those characters have no special meaning in CoffeeScript strings. We should treat them no differently than any other “invalid” escape sequence:
```
$ ./bin/coffee -bpe '"\!"'
"\!";
```

Certain invalid escape sequences are caught by the CoffeeScript compiler:
```
$ ./bin/coffee -bpe '"\x"'
[stdin]:1:2: error: invalid escape sequence \x
"\x"
 ^^
```
I can see the compiler behaving toward `` "\`" `` the same way it treats `"\x"`. That’s certainly better than automatically adding an extra backslash in front of `` \` ``.

Or just let it be. Is `` "\`" `` really “incorrect JavaScript”? Node and Chrome both treat such an invalid escape as just the character itself:
```
$ node -e 'console.log("\`")'
`
```
Therefore I think this PR is correct as is. Since `` ` `` and `${` are not used as delimiters in CoffeeScript strings or tagged template literals, the user shouldn’t need to escape them. If they happen to put a single backslash in front of them, it’s not our place to “correct” that. In other words, I think this is correct:
```
$ ./bin/coffee -bpe '"\`"'
"\`";
$ ./bin/coffee -bpe 'a"\`"'
a`\\``;
```
Yes, both are “invalid” JavaScript in that they both get interpreted by the runtime as `` "\`" ``, which then gets converted to `` "`" ``. But I think that’s fine. And the backtick deserves to be escaped in the latter but not the former, because the latter uses backticks as delimiters and the former doesn’t.

lydell:
I don't get we are discussing this. Tagged strings should work exactly like regular strings. I believe my suggested code does that.

GeoffreyBooth:
Per this PR, tagged strings (and CS2 interpolated strings) do behave exactly like regular strings, _to the user_. In either case, in CoffeeScript escaping  `` ` `` and `${` is not required. The output will escape them when the compiler is outputting a template literal, but not otherwise. The compiler shouldn’t be outputting an escaped backtick within a quotation-mark-delimited string, just as it shouldn’t be outputting an escaped quotation mark within a backtick-delimited string.

lydell:
> Per this PR, tagged strings (and CS2 interpolated strings) do behave exactly like regular strings, to the user.

That's what I want, but is not currently true.

```
$ ./bin/coffee -e 'a = (([x]) -> x); console.log "\`" is a"\`"'
.:10
  console.log("\`" === a`\\``);
                            ^^^
SyntaxError: Unterminated template literal
```

The above should output `true`.

GeoffreyBooth:
I see what you mean. We _should_ be honoring the user’s request and pass through `` \` `` unmolested, however wrong and stupid it is; but if we do that, we screw up the number of backticks and the JavaScript doesn’t execute. So we’re forced to do something about it.

Perhaps it’s better to throw an error on this? `` \` `` in a string/tagged template literal _is_ an invalid escape sequence, just as much as `\x` is. I don’t think we should be automatically correcting this for people. Either they intended `` \\` `` or they intended `` ` ``, and an error would force them to choose.

lydell:
> Perhaps it’s better to throw an error on this?

No. Absolutely not. There's no reason to throw an error.

> `` \` `` in a string/tagged template literal is an invalid escape sequence, just as much as `\x` is.

Are you talking about a JavaScript tagged template literal? `` \` `` is a perfectly valid escape seqence – it's the way you include a backtick in a tagged template literal!


> I don’t think we should be automatically correcting this for people.

There's nothing to correct. `"\%"` is perfectly valid JS and CS. So is `` "\`" ``. And so should `` tag"\`" `` be. It's just an _unnecessary_ escape. Why would we special-case `` \` ``, but no other unnecessary escape – and on top of that, only for a certain "type" of string?

> Either they intended `` \\` `` or they intended `` ` ``, and an error would force them to choose.

Yes, writing `` \` `` is likely a mistake. There are linters for such things.

---

All I'm asking for is consistency. There's absolutely no reason to special-case `` ` `` and `${`.

There is no implementation problem either. I've provided code that I'm quite confident is correct (we'll see when test cases are written). Writing code for throwing errors would be _more_ work.

lydell:
```
foo.coffee:30:40: error: We decided to disallow \` in tagged strings (but allowed in all other strings, and in regexps!)
result = html"Hello, accidental escape \` !"
                                       ^^

Hint: Either do something about that backslash, or replace that boring backtick
with something more fun, such as a 'π'!
```

GeoffreyBooth:
Sorry, let me try to clarify. `` \` `` _in CoffeeScript, outside of embedded JavaScript blocks,_ is always an invalid escape sequence. As in you should never type it in any of the following:
```coffee
# All wrong:
' \` '
" \` "
a' \` '
a" \` "
```
Currently, CoffeeScript 1.11.1 ”evaluates” the escape:
```
$ coffee -bpe "' \` '"
' ` ';
```
I was proposing this throw an error, like `\x` does, _for all these cases._ It wouldn’t throw an error within an embedded JavaScript block, which is the one context in CoffeeScript when one _would_ escape a backtick. But any type of CoffeeScript string: single-quote, double-quote, tagged template literal, etc.—could all get an error for an invalid escape sequence.

But I’m also fine with CoffeeScript always converting an input of `` \` `` and generating an output of `` ` ``, which _within an output of an ES backtick-delimited string/template literal_ would be escaped; so only in that context would it “remain”  `` \` `` (or be evaluated to `` ` `` and then escaped). I guess that’s what your code does, in one step rather than two. I’ve put it in and added tests. How is it now?

src/nodes.coffee
```diff
@@ -2274,7 +2274,14 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Parens
     fragments.push @makeCode '`'
     for element in elements
       if element instanceof StringLiteral
-        fragments.push @makeCode element.value.slice(1, -1)
+        value = element.value.slice(1, -1)
```
lydell: (Unrelated to this PR, but if you feel like it while at it, this could be changed to `value = element.value[1...-1]` :) )

test/tagged_template_literals.coffee
```diff
@@ -137,3 +137,32 @@ test "tagged template literal with an interpolated string that itself contains a
...
+test "tagged template literal with both ES and CoffeeScript interpolation", ->
+  eq "text: [ES template literals also look like this: `3 + 5 = ${3+5}` which equals |] expressions: [8]",
+  func"ES template literals also look like this: `3 + 5 = ${3+5}` which equals #{3+5}"
+
+test "tagged template literal with escaped ES interpolation", ->
+  eq 'text: [ES template literals also look like this: `3 + 5 = \\${3+5}`] expressions: []',
+  func"ES template literals also look like this: `3 + 5 = \\${3+5}`"
+
+test "tagged template literal with invalidly escaped ES interpolation", ->
+  eq 'text: [ES template literals also look like this: `3 + 5 = \${3+5}`] expressions: []',
```
lydell: I think this test will be clearer if the backslash in this line is removed.

test/tagged_template_literals.coffee
```diff
@@ -137,3 +137,32 @@ test "tagged template literal with an interpolated string that itself contains a
...
+
+test "tagged template literal with backticks", ->
+  eq 'text: [ES template literals look like this: `foo bar`] expressions: []',
+  func"ES template literals look like this: `foo bar`"
+
+test "tagged template literal with escaped backticks", ->
+  eq 'text: [ES template literals look like this: \\`foo bar\\`] expressions: []',
+  func"ES template literals look like this: \\`foo bar\\`"
+
+test "tagged template literal with invalidly escaped backticks", ->
```
lydell: invalidly → unnecessarily

Please also make this change in the below tests.

lydell:
```
$ coffee -bpe "' \` '"
' ` ';
```

The above is just bash string confusion. (I use the fish shell, which has different string rules, in case I've posted terminal stuff that don't make sense to you now.) Try these:

```
# Use single-quotes in bash to avoid tricky escaping rules.
$ coffee -bpe '"\`"'
"\`";

# Or put stuff in a file.
$ cat tst.coffee
'\`'
$ coffee -bp tst.coffee
'\`';

# Or go down the road of backslash hell :)
$ coffee -bpe "'\\\`'"
'\`';
```

So CoffeeScript has never done anything special with the `` \` `` escape inside strings.

---


I'd like to see this test included as well:

```coffee
test "tagged template literal special escaping", ->
  eq 'text: [` ` \\` \\` \\\\` $ { ${ ${ \\${ \\${ \\\\${ | ` ${] expressions: [1]',
  func"` \` \\` \\\` \\\\` $ { ${ \${ \\${ \\\${ \\\\${ #{1} ` ${"
```

GeoffreyBooth:
@lydell I’ve made your requested changes. BTW, you and @greghuc both have write access to my repo.

lydell:
> BTW, you and @greghuc both have write access to my repo.

Sorry for being lazy. :+1:

PR #4388, opened 2016-12-05
Allow imported module members to be named `default`

### LINKED ISSUE #4387: Importing a module member named `default` fails
Apparently [the following is valid ES2015](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=latest&code=import%20%7B%20default%20%7D%20from%20'lib'%3B%0A):
```js
import { default } from 'lib';
```
But CoffeeScript [throws an error](http://coffeescript.org/#try:import%20%7B%20default%20%7D%20from%20'lib'):
```
Error on line 1: unexpected default
```

GeoffreyBooth:
Closes #4387.

Apparently code like this is not only allowed in ES2015, [it’s common](https://github.com/kimroen/ember-cli-coffeescript/pull/124#issuecomment-263025341):
```js
import { default } from 'lib';
```
Although `import default from 'lib';` is not allowed in ES.

This PR allows destructured import (and export-import) of module members named `default`.


lydell:
I never stop learning new stuff about `import`/`export`. Good work!

GeoffreyBooth:
Too bad we didn’t slip this in before 1.12.0. Well, I guess it won’t be long before 1.12.1 . . .

lydell:
We could do 1.12.1 as soon as possible, if you feel like it.

GeoffreyBooth:
Let’s see what else people find in the next few days?

Issue #4390, opened 2016-12-05
`from` is now a reserved word in for loop declarations


IvanVergiliev:
Apparently the addition of the `for...from` loop breaks code with variables called `from`. Obviously this can't be fixed, but it should at least be documented as a breaking change in the Changelog.

Also, I had a hard time finding a set of all reserved words in CoffeeScript - this would also be a useful addition, IMO.

GeoffreyBooth:
`from` is only a keyword in the context of an import/export statement or for loop. I can use it as a variable: http://coffeescript.org/#try:from%20%3D%202

Can you post example code that worked in 1.11.1 but broke in 1.12.0?

IvanVergiliev:
```coffeescript
lines = [
  {from: 0, to: 10}
  {from: 3, to: 17}
]

for {from, to} in lines
  console.log from
```

I also noticed that I can use it as a variable name sometimes. Apparently this case breaks because it's part of a `for` loop.

GeoffreyBooth:
@atg, do you feel like looking into this?

vendethiel:
Oh, it's just a really old thing the compiler's lexer does. That's just `@seenFor` parsing stuff differently between a `for` and a `from/of/in`...

distracteddev:
@IvanVergiliev Quick question, if adding "from" support is clearly a breaking change, why did the project decide to publish it without a major version bump? Should I expect this project to respect semver in the future?

vendethiel:
@distracteddev where did it say it was "clearly a breaking change"? @GeoffreyBooth even offered to look at 1.12 regressions.

(Also, https://gist.github.com/jashkenas/cbd2b088e20279ae2c8e)

distracteddev:
My bad, assumed adding "from" support was known to be a breaking change. I see now that this wasn't intended to be the case.

Thanks for the note on semver.

GeoffreyBooth:
Fixed in #4393. We will be releasing 1.12.1 soon.

PR #4393, opened 2016-12-06
Detect when `from` in a `for` loop declaration is an identifier

GeoffreyBooth:
Per #4390, this should be valid:
```coffee
lines = [
  {from: 0, to: 10}
  {from: 3, to: 17}
]

for {from, to} in lines
  console.log from
```
But it throws an error on thinking that `from` is a keyword. Since `from` has never been a reserved word (outside the context of an `import` or `export` statement), we should try to detect when a `from` on a `for` declaration line is really a keyword or rather a variable identifier.

I’ve tried to make the detection by looking at the previous token. If the previous token is an identifier itself, like `for a from b` I think that that would mean that the present `from` is always a keyword. If the previous token is `{` or `[` or `,`, then we’re inside destructuring and `from` is always an identifier. If the previous token value is `from`, we have `for from from a` or `for a from from` and the first `from` is an identifier and the second is a keyword.

Are there any other cases I haven’t thought of?


lydell:
> Are there any other cases I haven’t thought of?

I can't think of more `for…from` cases, but `yield from from` came to mind. Not sure if it belongs in this PR, though.

Other than that this looks good to me.

GeoffreyBooth:
This bug is only because the lexer (previously) was declaring a `from` token to be `FROM` (instead of `IDENTIFIER`) if there was a `for` anywhere previously on the same line. Now the lexer is more discerning about when `from` should be considered `FROM`.

`yield from from` works already because the lexer there only declares the first `from` a `FROM` if the previous tag is `yield`.

GeoffreyBooth:
@IvanVergiliev or @vendethiel, can you think of any other cases where `from` might be a variable name and not a keyword in a `for` loop declaration, other than the ones in [the tests in this PR](https://github.com/jashkenas/coffeescript/pull/4393/files#diff-7a1439be94516e3bd8b6932a17229647R291)?

src/lexer.coffee
```diff
@@ -824,6 +825,30 @@ isUnassignable = (name, displayName = name) -> switch
 
 exports.isUnassignable = isUnassignable
 
+# `from` isn’t a CoffeeScript keyword, but it behaves like one in `import` and
+# `export` statements (handled above) and in the declaration line of a `for`
+# loop. Try to detect when `from` is a variable identifier and when it is this
+# “sometimes” keyword.
+fromIsKeywordInForDeclaration = (prev) ->
```
connec: Nit: would `isForFrom` be an acceptable, more concise name?

src/lexer.coffee
```diff
@@ -824,6 +825,30 @@ isUnassignable = (name, displayName = name) -> switch
...
+    if prev[1] is 'from'
+      prev[1][0] = 'IDENTIFIER'
+      yes
+    # `for i from iterable`
+    yes
+  # `for from…`
+  else if prev[0] is 'FOR'
+    no
+  # `for {from}…`, `for [from]…`
+  else if prev[1] in ['{', '[']
```
connec: Nit: `if prev[1] in '{['` would save an array allocation (not really sure if that's meaningful).

IvanVergiliev:
Thanks for the quick fix!

How about destructuring with aliasing (not sure if that's the right term):

```coffeescript
for {b: from} in a
  print from
```

? This is broken right now as well, and I think there might have to be a separate check for it in the lexer. (Also, all this special casing is making me a bit sad, but I guess there's no other way to detect this so it's much better than before.)

vendethiel:
> @IvanVergiliev or @vendethiel, can you think of any other cases where from might be a variable name and not a keyword in a for loop declaration, other than the ones in the tests in this PR?

```sh
./bin/coffee -bce 'for {a: from} from 1 then 1'
[stdin]:1:9: error: unexpected from
for {a: from} from 1 then 1
        ^^^^
./bin/coffee -bce 'for {a: from} from 1 then 1'
[stdin]:1:9: error: unexpected from
for {a: from} from 1 then 1
        ^^^^
```

Somewhat reminds me of https://github.com/satyr/coco/issues/195 (though we don't have this exact issue when you can have a typeless `for`).

src/lexer.coffee
GeoffreyBooth: Yes 👍

src/lexer.coffee
GeoffreyBooth: `if prev[1] in '{['` looks confusing to me. The way I wrote it seems to be the style of the codebase.

GeoffreyBooth:
Thanks, anything else?

IvanVergiliev:
I can't think of any other cases.

GeoffreyBooth:
@lydell we should probably release 1.12.1. I can prep it tonight.

PR #4396, opened 2016-12-07
Improve docs

GeoffreyBooth:
Some modernization before the fork toward the v2 docs. Now everything in the `docs/v1` folder is generated during build, for simplicity. Both v1 and v2 will share the same (updated) favicon files.

@lydell if you want, I can regenerate the 1.12.1 release with this included if you think I should slip it in. (I probably should, and bump the date.)


Cakefile
```diff
@@ -129,6 +129,11 @@ task 'build:browser', 'rebuild the merged script for inclusion in the browser',
 
 task 'doc:site', 'watch and continually rebuild the documentation for the website', ->
   # Helpers
+  css = fs.readFileSync('./documentation/css/docs.css', 'utf-8') + '\n' +
+        fs.readFileSync('./documentation/css/tomorrow.css', 'utf-8')
```
lydell: I like all the inlining stuff. But I really think we should minify the CSS first.

docs/manifest.json
```diff
@@ -0,0 +1,17 @@
...
+		},
+		{
+			"src": "\/android-chrome-512x512.png",
+			"sizes": "512x512",
+			"type": "image\/png"
+		}
+	],
+	"theme_color": "#ffffff",
+	"display": "standalone"
+}
```
lydell: When I first saw this, I thought: "Why are the `/`s escaped?" I didn't even know that `\/` is a valid JSON escape sequence before! According to this [StackOverflow answeer](http://stackoverflow.com/a/1580682):

> Allowing \/ helps when embedding JSON in a <script> tag, which doesn't allow </ inside strings

So the backslashes could be removed. It doesn't really matter, though, I just wanted to share this interesting (or not XD) piece of trivia.

(Also, this file seems to be indented with tabs, while most (all?) other files in this repo are indented with spaces. I guess it doesn't matter, either. It even saves a few bytes down the wire...))

Cakefile
GeoffreyBooth: And add another dependency? I was doing this mainly to clean up the folders, to make `v1` into just an “output” folder. With a big CoffeeScript script block, I figured tge goal was source readability over performance.

GeoffreyBooth:
The SVGs are optimized. The PNGs had file sizes like 100 bytes or 200 bytes, so I didn't think to touch those. I think we could simply remove the screen corners one if we wanted, but even that isn't _so_ big.

Cakefile
lydell: What big CoffeeScript block?

GeoffreyBooth: The big `<script type="text/coffeescript">` block in the generated index.html, line 4133 onward in the current site.

PR #4400, opened 2016-12-11
DRY up the version number

GeoffreyBooth:
We have to include the version number (e.g. 1.12.1) in `package.json`; but there’s no reason to duplicate that anywhere else. We can `require 'package.json'`.


lydell:
I like this.

GeoffreyBooth:
So . . . merge? Should I rebuild the browser compiler first? (That would change the browser compiler in use on coffeescript.org.)

lydell:
I was goind to say "Yes, I think we can merge it", but now that you mentioned rebuilding the browser compiler first I hesitated. Why would we do that (instead of at the next release as usual)?

GeoffreyBooth:
It’s a bit weird when we merge into `master`, we’re merging revised compiled .js files in `lib` but we’re not merging the revised compiled browser .js file, I guess because it’s part of the documentation which only gets rebuilt on release. But the documentation doesn’t really need to be versioned, so the browser compiler could change between releases . . .

connec:
I've not read into how this works, but could we not copy `brower.js` into the `docs` folder when docs are built, and use that version in the docs themselves?

GeoffreyBooth:
`docs/v1/browser-compiler/coffee-script.js` is a minified concatenation of all the files in `lib/coffee-script`. `lib/coffee-script/browser.js` is just the part of the compiler that is browser-specific.

jashkenas:
That's certainly one way to do it. Neat!

GeoffreyBooth:
So should I merge this as is, or rebuild the browser compiler too?

PR #4401, opened 2016-12-11
Docs improvements: content in Markdown, organization into subtemplates, fixed tests

GeoffreyBooth:
More improvements to the documentation:

- All the content is now markdown, compiled into HTML at build/template render time.
- The code examples are compiled into JavaScript at build/template render time, without a step that creates intermediate .js files. The compilation is so fast that the disk writes were slower.
- `index.html` now requires sub-templates, to divide content into relevant sub-files. The build process reassembles everything. The sub-files are mostly inside a `documentation/v1` folder, giving the 2 docs source files a logical place to go.
- `index.html` and `test.html` are now built via separate `cake` tasks, to avoid a gigantic function.
- The browser-based tests run in `test.html` now run _all_ the tests, including `error_messages.coffee`, thanks to a fixed polyfill for `@throw`.


Cakefile
```diff
@@ -161,41 +177,70 @@ task 'doc:site', 'watch and continually rebuild the documentation for the websit
...
-
-  releaseHeader = (date, version, prevVersion) -> """
-    <div class="anchor" id="#{version}"></div>
-    <b class="header">
-      #{prevVersion and "<a href=\"https://github.com/jashkenas/coffeescript/compare/#{prevVersion}...#{version}\">#{version}</a>" or version}
-      <span class="timestamp"> &mdash; <time datetime="#{date}">#{formatDate date}</time></span>
-    </b>
-  """
+  htmlFor = ->
+    marked = require 'marked'
```
lydell: This needs to be added to package.json,

documentation/sections/operators.md
```diff
@@ -0,0 +1,149 @@
...
+
+To simplify math expressions, `**` can be used for exponentiation and `//` performs integer division. `%` works just like in JavaScript, while `%%` provides [“dividend dependent modulo”](http://en.wikipedia.org/wiki/Modulo_operation):
+
+```
+codeFor('modulo')
+```
+
+All together now:
+
+<table class="definitions">
```
lydell: If we use CommonMark, then we could possibly write the tables in a more readable way. http://spec.commonmark.org/0.27/#html-blocks

What do you think?

GeoffreyBooth: Is that supported by the `marked` module? Its docs say to use this syntax: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

I just haven't gotten around to converting, the automated HTML-to-Markdown converters I tried didn't support tables.

GeoffreyBooth: Isn't that essentially treating the table as one big passthrough? But we still want markdown within the cells to be evaluated.

I would think the github style would be more idiomatic markdown than any form of HTML tags. It'll just take manual effort to convert.

lydell: I think we only have two tables in total currently, so the way it is currently is not such a big deal. But if it could be made nicer in any way, that'd be great! :+1:

Cakefile
GeoffreyBooth: Done. Are you okay with me merging this into develop whenever I'm done testing it? I'm working on getting the 2 docs built in their branch, based on this.

GeoffreyBooth: Master, I mean.

lydell: Yes, go for it!

jashkenas:
Very, very nice!

jiyinyiyong:
Then it will cost less time to translate into Chinese. Great.

PR #4407, opened 2016-12-21
Trivial: Remove trailing whitespace in test/error_messages.coffee

connec:
Trailing whitespace is generally considered 'bad style' and is often linted against or even simply removed by text editors.

One of the tests in test/error_messages.coffee depended on trailing whitespace, making the file tricky to work with for people whose editor is configured to remove trailing whitespace. The alternative is to use a literal "\n" and escape the line break.

I suppose this is personal preference, but given the proliferation of tools to remove trailing whitespace it makes sense to not depend on them when there's an alternative.

### DIFF
--- a/test/error_messages.coffee
+++ b/test/error_messages.coffee
@@ -428,7 +428,7 @@ test "#3795: invalid escapes", ->
   assertErrorFormat '''
     ///a \\u002 0 space///
   ''', '''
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
Thanks, this has been driving me nuts.

Issue #4409, opened 2016-12-23
Inconsistency of operator //=


valff:
Compiled output of operator `//=` missing parentheses for right operand.

Example:
```coffeescript
a //= b + c
```
Output:
```javascript
a = Math.floor(a / b + c);
```
However proper output should be:
```javascript
a = Math.floor(a / (b + c));
```
It would be consistent with `%%=` and other javascript assignment operators.

connec:
This has been fixed 🎉

PR #4410, opened 2016-12-23
Fixed //= operator

valff:
Resolving #4409. Not sure is it proper place for fix, first time working with coffeescript internals.


connec:
Thanks for the PR!

I think the fix has the right idea. It could be moved into `Op#compileFloorDivision`, then the conditional wouldn't be needed. Also, all nodes have an `#isComplex()` method, which could be used to determine whether parentheses are needed.

valff:
Thanks for advice! Didn't find `#isComplex()` method at first.

PR #4415, opened 2016-12-29
Allow `@` values as indices in `for` expressions

### LINKED ISSUE #4411: Cannot use @variable as index in for ... in
`for char, @srcPos in @src`

gives

> index cannot be a pattern matching expression

(background: I would like to be able to access the index variable in other functions in the same object...)

connec:
This loosens the compilation of `for` expressions to allow the index variable to be an `@` value, e.g.

```coffee
do @visit for @node, @index in nodes
```

Within `@visit`, the index of the current node (`@node`) would be available as `@index`.

Fixes #4411.

### DIFF
--- a/lib/coffee-script/nodes.js
+++ b/lib/coffee-script/nodes.js
@@ -3415,7 +3415,7 @@
       if (this.object) {
         ref3 = [this.index, this.name], this.name = ref3[0], this.index = ref3[1];
       }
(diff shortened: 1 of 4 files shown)


jashkenas:
Lovely patch!

You really know your way around the guts of the beast...

connec:
Starting to piece it together 😛

PR #4354, opened 2016-11-06
[CS2] Compile class constructors to ES2015 classes

connec:
OK, I've pushed what I believe is a more-or-less "ready" version of classes. I've adopted a similar approach to the previous PR, whereby valid methods are hoisted into an ES2015 class declaration, and other expressions are left as-is. I've also made it so that classes are compiled "bare", without an enclosing IIFE, where possible.

I've standardised the hoisting thing a little bit. `hoist` is now a method on `Base`, meaning any node can be hoisted. Calling `hoist` mutates the source node and returns a new target node, which should be put in the tree at the desired output location. When the target is compiled, it returns a fragments placeholder. When the source is compiled, it updates the target's fragments. Finally, `Block#compileRoot` will expand all placeholder fragments before returning the result.

I've not touched `super` outside constructors as that adds quite a bit of surface area and would be better handled separately, imo. The downside of this is that, for now, derived classes always need an enclosing IIFE. Handling `super` is something I'd be happy to look at as a subsequent PR, unless someone else wants to tackle it.

There is one remaining annoyance, which is the assignment. This is needed to allow `Literal` to be referenced in the second example below (and throughout the CS source). We could remove this, which would mean that classes created in expressions can only be referenced by the result of the expression (e.g. `exports.Foo = class Foo` could only be referenced as `exports.Foo`). I've erred on the side of compatibility for now, as it's easy to take this out later if we so choose.

I've not added any test, the following would be good to have. Any test contributions would be great.

- `@` params with/without `super`, including errors
- bound functions with/without `super`, including errors
- `super` in arrow functions in constructors
- `super` and external constructors
- ... any more anyone can think of

As a final note, the diff looks fairly large as I ended up rewriting the `Class` node, however a lot of the logic hasn't changed, its just been moved around to better line-up with the new usage of the node. I felt this was better than continuing to stretch the previous `Class` out of shape. Apologies to reviewers.

<table>
<tr>
<td valign='top'>
<pre lang='coffee'>
class Polygon
  constructor: (@height, @width) ->
<br>
  area: ->
    @calcArea()
<br>
  calcArea: ->
    @height * @width
</pre>
</td>
<td>
<pre lang='js'>
var Polygon;
<br>
Polygon = class Polygon {
  constructor(height, width) {
    this.height = height;
    this.width = width;
  }
<br>
  area() {
    return this.calcArea();
  }
<br>
  calcArea() {
    return this.height * this.width;
  }
<br>
};
</pre>
</td>
</tr>
<tr>
[...]

Issue #3796, opened 2015-01-16
What should `super` refer to?


lydell:
I know that `super` has been discussed many times before (#2638 and #3436 for example), but in my opinion this is something new. This issue is not about changing `super` totally, but merely about defining its edge cases. The reason I started this discussion is because of issues I bumped into when making #3785.

This issue is not about what `super` without parentheses should do etc. It is only about what `super` should refer to in different cases.

The only way I’ve ever used `super`, and the only way I’ve even _seen_ `super` be used, is like this:

```coffee
class A extends B
  method: -> super
```

That is, only in an instance method defined using the `method:` syntax inside a `class` body with `extends`.

If that was the only case `super` was allowed, compiling it would be trivial.  The above could compile to:

```js
var A;

A = (function(_super) {
  __extends(A, _super);

  function A() {
    return _super.apply(this, arguments);
  }

  A.prototype.method = function() {
    return _super.prototype.method.apply(this, arguments);
  };

})(B);
```

CoffeeScript also allows `super` in “static” methods:

```coffee
class A extends B
  @method: -> super
```

That would also be easy to compile:

```js
var A;

A = (function(_super) {
  __extends(A, _super);

  function A() {
    return _super.apply(this, arguments);
  }

  A.method = function() {
    return _super.method.apply(this, arguments);
  };

})(B);
```

If dynamic keys are implemented, that’d be straightforward too:

```coffee
class A extends B
  "#{foo()}": -> super
```

```js
var A;

A = (function(_super) {
  var _name;

  __extends(A, _super);

  function A() {
    return _super.apply(this, arguments);
  }

  A.prototype[_name = "" + foo()] = function() {
    return _super.prototype[_name].apply(this, arguments);
  };

})(B);
```

So far so good. All nice and simple.

CoffeeScript also allows a “classic” style for making “classes”:

```coffee
A = ->
A extends B
A::method = -> super
```

Note that the above isn’t equivalent to the `class` examples, since the constructor `A` doesn’t apply `B`. That could be done by writing `A = -> B.apply this, arguments` manually, but it would be much nicer to be able to write `A = -> super`, but that throws an error.

There is no way to make a static method call `super` using the “classic” style.  `A.method = -> super` throws an error.

So when _is_ it possible to use `super` “classic” style?

```coffee
A = -> super # throws
A.m = -> super # throws
A::m = -> super # works
A.prototype.m = -> super # works
A["prototype"].m = -> super # throws
[...]

Issue #2638, opened 2012-12-06
Make `super` consistent with ES6


paulmillr:
- **Current CoffeeScript** & **Ruby** `super` calls appropriate fn of the superclass with all `arguments` and `super a, b, c` for calling with other args.
- **ECMAScript 6** `super` [is just a direct reference](https://github.com/dherman/tc39-codex-wiki/blob/master/data/es6/classes/index.md) to the super _class_.
- **Coco** & **LiveScript** `super` is just a direct reference to the super function.

I’m in favour of ES6-way which will be consistent with futurejs.

So, `super` should just be a ref to superclass. This will also allow to use it in static methods etc which will close gh-1790

PR #4424, opened 2017-01-19
[CS2] Compile all super calls to ES2015 super

connec:
Super quick first pass.  This breaks using `super` in non-methods, meaning several tests are failing. Self-compilation still works.

As I mentioned in https://github.com/coffeescript6/discuss/issues/22#issuecomment-273837748, the main decision to be made here is around support for `super` in non-class methods. Opinions from @jashkenas, @GeoffreyBooth, @lydell or anyone else very welcome!

Here are a couple of the options:

1. Go all in with ES2015, drop support for `super` outside of methods in class initializers. This would clean up some code in the compiler for tracking the classes of methods, special handling of prototype assignments in class bodies, etc.
2. Why not both? If the super call is in a method in a class initializer, compile it to an ES2015 `super`, otherwise, compile it as we do now. This would just add a couple of special cases in the handling of `SuperCall` nodes.
3. Try and devise some best of both worlds approach. Compile class method to ES2015. Since we can now use ES5+ method, we could drop `__super__`, method class recording, etc. and replace it all with `Object.getPrototypeOf(this.constructor)`. This could optionally turn `super` into a fully runtime feature that will compile anywhere, and would provide clean ES2015 output.

Personally, I'm leaning towards 1. for the significant impact I expect it will have on the compiler, and the fact that, realistically, the prototype-based approach is gonna be rare in an ES2015 world. If really, really needed, `super` in a dynamic method can still be specified explicitly (e.g. `BaseClass::method.apply(this, arguments)`).

If the compatibility break of such a move is too great, I'd be interested in exploring 3. I think that 2. would not be popular as it would force us to continue using class IIFEs to ensure correct parent class caching, etc. meaning 'ugly' JS output.


mitar:
I think calling `super` inside anonymous functions inside methods can help sometimes with logic structure. So I would vote to keep existing support and not be as strict as ES2015.

lydell:
I'm all for option 1. See #3796.

kirly-af:
I personally never used `super` outside of classes, so **1** is completely fine to me.
However **3** would be an understandable choice to avoid breaking changes + nicer output.

jashkenas:
I think since this is already the big breaking change release, option #1 is the way to go.

connec:
OK, I'll go for option 1., and we can discuss about [@jashkenas' comment](https://github.com/coffeescript6/discuss/issues/22#issuecomment-273868922) as a separate new feature.

@mitar calling super within anonymous functions is possible, so long as they are arrow functions.

```js
class A {
  method () { console.log('A#method') }
}
class B extends A {
  constructor () {
    (() => super())()
  }
  method () {
    (() => super.method())()
    console.log('B#method')
  }
}
new B().method() // A#method
                 // B#method
```

GeoffreyBooth:
I also think we should go with option 1, since we can always add the more elaborate versions later.

This PR has a bug though I think. Take this input:
```coffee
class A
  foo: ->
    console.log 'woohoo!'

class B extends A
  bar: ->
    super.foo()
```

This branch produces:
```js
var A, B;

A = class A {
  foo() {
    return console.log('woohoo!');
  }

};

B = class B extends A {
  bar() {
    return super.bar(...arguments).foo();
  }

};
```

but it should produce:
```js
var A, B;

A = class A {
  foo() {
    return console.log('woohoo!');
  }

};

B = class B extends A {
  bar() {
    return super.foo();
  }

};
```

connec:
Not a bug - that's how CoffeeScript's `super` works. See https://github.com/jashkenas/coffeescript/issues/2638 for some discussion on this.

Interestingly, it seems that `super` of this form used to be planned for ES: https://esdiscuss.org/topic/referencing-super. I can't find any reference to when it was dropped.

Personally, I think being able to look up arbitrary methods on the superclass is an undesirable feature. An overridden method shouldn't need to be aware that *other* methods may be overridden. I.e. if `bar` wants to call `foo`, it should call `this.foo`; if `B#bar` wants to *specifically* call `A#foo`, it should call `A::foo`.

It also encourages people to think that `super` is an entity of its own, perhaps an object containing references to methods of the superclass, but it is not (in ES discussions it was proposed to mean `this`, which would have meant that `super.foo` and `a = super, a.foo` would target different methods!).

jashkenas:
> Not a bug - that's how CoffeeScript's super works. See #2638 for some discussion on this.
> 
> Interestingly, it seems that super of this form used to be planned for ES: https://esdiscuss.org/topic/referencing-super. I can't find any reference to when it was dropped.
> 
> Personally, I think being able to look up arbitrary methods on the superclass is an undesirable feature. An overridden method shouldn't need to be aware that other methods may be overridden. I.e. if bar wants to call foo, it should call this.foo; if B#bar wants to specifically call A#foo, it should call A::foo.
> 
> It also encourages people to think that super is an entity of its own, perhaps an object containing references to methods of the superclass, but it is not (in ES discussions it was proposed to mean this, which would have meant that super.foo and a = super, a.foo would target different methods!).

I agree with this (naturally), and I think I've made that argument in these pages before.

But! If we're switching over to ES6 semantics for classes for CS2, shouldn't we switch over to ES6 semantics for super at the same time?

GeoffreyBooth:
A `super.foo()` like my example wasn't just planned; it's implemented. Paste my "should output" example into Chrome console to see for yourself. We need to support this type of `super`.

connec:
Well, I'd argue that with classes we've changed very few semantics - we've adopted, by necessity, the `class {}` syntax (and associated baggage 😅 ), but have avoided removing any significant CS features.

@GeoffreyBooth it's the *other* form that was planned not implemented (`super()`), sorry.

I'm not sure I agree that we *need* to support it. It feel like we're just adding needless redundancy since 99%<sup>1</sup> of uses are just `super.<current function>`, and any other case is almost certainly a sign of bad design, or a contrived example for a test case. I wish ES Discuss was more searchable so I could figure out why they chose this approach.

Whilst maintaining that it would be best not to, we could make both uses live side-by-side. If we do go down that route we might want to talk about letting 'bare `super`' go, and mandating regular function syntax (i.e. `super arguments...` would replace bare `super`).

1\. Made up.

GeoffreyBooth:
Using `super` to call a different method on the parent is (one of the examples on MDN's `super` page)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super#Super-calling_static_methods], and it doesn't seem contrived. Regardless of what we do with "bare" super, we need to support this usage too.

jashkenas:
I think that once again, it'll be a step backwards — but we should make it work like JS `super` works. Keeping different semantics will just breed confusion when people try to use it, and will be a running source of new opened tickets asking for it to be changed.

connec:
OK, lets add `super.method`.

Next question is bare `super` - `super.method` makes it look like `super` is something that hosts superclass versions of methods, maybe, whereas in CS it is in fact a method call and in ES it's a syntax error. Since we're adopting `super.method`, I suggest we also adopt a `SyntaxError` for bare `super`. This eliminates the ambiguity observed by @GeoffreyBooth [above](https://github.com/jashkenas/coffeescript/pull/4424#issuecomment-273984843).

To outline the suggested compilation:

```coffee
class Base
class SuperExamples extends Base
  # Should throw a SyntaxError along the lines of "'super' must be called or accessed"
  bareSuper: -> super

  # Should compile to `super.shortSuper()`
  shortSuper: -> super()

  # Should compile to `super.esSuper()`
  esSuper: -> super.esSuper()
```

jashkenas:
Yes, we should make it work like ES `super` — in all respects.

connec:
It looks like I was hasty to say ["it's a 'syntactic form' that is only valid within a class initializer"](https://github.com/coffeescript6/discuss/issues/22#issuecomment-273848272) - it seems it's also valid within method in object initializers:

```js
({
  foo () { super.foo() } // OK
})
({
  foo: function () { super.foo() } // SyntaxError: 'super' keyword unexpected here
})
```

Of course we don't currently compile object assigns with method values to that syntax. I had planned to look into that at some point to clean up the code I had to put in `nodes.Code` for the class work, so might investigate that.

jashkenas:
But how are you supposed to get that "method" back onto the prototype from outside?

connec:
Heh, I was just writing a similar example in node, and got the same result. The MDN example shows using it via `Object.setPrototypeOf`.

```js
Object.setPrototypeOf(a, Lion)
new a().speak()
```

connec:
This works:

```js
function A () {}
A.prototype = {
  foo () { console.log('A#foo') }
}

function B () {}
B.prototype = {
  foo () {
    super.foo()
    console.log('B#foo')
  }
}

Object.setPrototypeOf(B.prototype, A.prototype)
new B().foo() // A#foo
              // B#foo
```

jashkenas:
> This works ...

No, as you demonstrate below, there's still no way to add methods to an existing prototype and make use of `super`. That's the "normal" ES3 ability that we're losing in ES6.

connec:
So it basically seems that `super` is 'half-bound' at compile time. The prototype can be changed but the 'home object' (as it's referred to in the spec) cannot.

```js
base   = { foo () { return 'base' } }
child1 = { foo () { return super.foo() + ' -> child1' } }
child2 = { foo () { return super.foo() + ' -> child2' } }

try {
  child1.foo()
} catch (e) {
  console.log(e) // TypeError: (intermediate value).foo is not a function
}

Object.setPrototypeOf(child1, base)
console.log(child1.foo()) // base -> child1

Object.setPrototypeOf(child2, base)
console.log(child2.foo()) // base -> child2

Object.setPrototypeOf(child2, child1)
console.log(child2.foo()) // base -> child1 -> child2

method = child2.foo
console.log(method()) // base -> child1 -> child2

function A () {}
A.prototype.method = child2.foo
console.log(new A().method()) // base -> child1 -> child2
```

connec:
OK, I've pushed some more updates that adds support for `super` with accessors! See [this test](https://github.com/connec/coffeescript/blob/55d78c2fb99348247f41f75c090587e87ff88f92/test/classes.coffee#L65) and [the following one](https://github.com/connec/coffeescript/blob/55d78c2fb99348247f41f75c090587e87ff88f92/test/classes.coffee#L85) for some contrived examples. @GeoffreyBooth your example now indeed compiles to:

```js
var A, B;

A = class A {
  foo() {
    return console.log('woohoo!');
  }

};

B = class B extends A {
  bar() {
    return super.foo();
  }

};
```

I've also fixed a couple of incidental bugs that were exposed by these changes, including:

1. Soaked `super` invocations. There was existing code in `Call` to deal with soaked `super` invocations, but as far as I can tell the parse tree would never actually evaluate to a soaked call due to bare `super` taking precedence (i.e. `super?()` was interpreted as `super()?()`).

2. Generating bound functions for the closures that wrap statement expressions when the expression contains `super`. This also means `super` can now be used in constructors within `try` expressions, etc. There is an edge case for statement expressions containing both `this` and `yield`, because bound functions cannot be generators.

3. Special-case handling to support `class @A extends A` when compiled as a simple class. There was a test for this that started breaking once classes with parents no longer needed an IIFE wrapper. To comply with the tests, some extra handling was put in place to wrap the class in an IIFE if the parent's name matches the class' name.

4. Throw a compile error when calling `super` with `new`. Previously, the `new` would simply not be emitted! From my experiments it seems `super` with `new` is syntactically valid JS, but it always fails at runtime with "unsupported 'super' reference", so I've made it a compile error.

GeoffreyBooth:
Awesome work! Assuming all the tests pass, is there anything else to be done for `super`?

Aside from adding support for `static`, `get` and `set`, is there anything else to be done for classes?

connec:
@GeoffreyBooth I think we need to have a discussion about `static`. CS already supports `static` with `@name [:=]` assignments, which is far more idiomatic within the language (no special syntax). From a JS dev perspective it's also a straightforward conversion (`static` => `this`).

Regarding `super`, I need to review some of the new `class` tests. At least one of them was testing that `super` in `try` blocks won't work, except I believe they should now, yet the test is still passing.

GeoffreyBooth:
We can discuss `static`, `get` and `set` in another thread. What I'm wondering is when the `2` branch is ready for release as there first alpha. It sounds like as soon as you're done checking these tests and perhaps fixing some issues, we're good to release?

jashkenas:
> @GeoffreyBooth I think we need to have a discussion about static. CS already supports static with @name [:=] assignments, which is far more idiomatic within the language (no special syntax).

I'd love to keep this sort of thing, and avoid the word `static` if possible. I think it's a terrible word for the feature, and is just legacy baggage from Java.

connec:
@GeoffreyBooth I think that would be sufficient, yeh. I'll review those tests asap and double check there's nothing else we can throw out.

Also if @lydell or @jashkenas (or anyone with more familiarity with the grammar) have any comments on the [grammar changes](https://github.com/jashkenas/coffeescript/pull/4424/files#diff-30b5df382fb36749d0d1d282b2df9a53) they'd be very welcome.

src/grammar.coffee
```diff
@@ -429,12 +436,7 @@ grammar =
     o 'Value OptFuncExist String',              -> new TaggedTemplateCall $1, $3, $2
     o 'Value OptFuncExist Arguments',           -> new Call $1, $3, $2
     o 'Invocation OptFuncExist Arguments',      -> new Call $1, $3, $2
-    o 'Super'
-  ]
-
-  Super: [
-    o 'SUPER',                                  -> new SuperCall
-    o 'SUPER Arguments',                        -> new SuperCall $2
+    o 'SUPER OptFuncExist Arguments',           -> new SuperCall LOC(1)(new Super), $3, $2
```
lydell: - Just double-ehcking: Was the `'SUPER Arguments'` case redundant because of `'SUPER OptFuncExist Arguments'`?
- What about the `'SUPER'` case?

jashkenas: Whoa there — I don't know about all of the ES6 grammar rules and wrinkles ... but wouldn't it make more sense to bring `SUPER` in as a `Value`, instead of special-casing all of the different ways you might access it?

connec: Yeh, so `SUPER Arguments` has been directly subsumed by `SUPER OptFuncExists Arguments` (which allows `super?()` to parse).

The `SUPER` case has been dropped as, in light of supporting `super` with accessors, it becomes confusing whichever way we spin it:

- Keep bare `super` as a super call to the method of the same name, forwarding arguments. This is fine until we have to deal with `super.foo`, which now looks ambiguous: is it `super.foo` as it appears, or `super().foo`? What about `(super).foo`? Tbh this use of `super` is useful enough that I'd be happy to leave it in, but it does lead to a number of ways of doing the same thing with `super`.

- Allow bare `super`, and handle it the same way `super()` is handled (compiler error in a `constructor`, and otherwise a reference to `super.<current method>`). This would be safe and consistent, except it could lead to subtle bugs and misunderstandings for people upgrading to CS2 from CS1.

SO, a bare `super` will no longer parse. Unfortunately the [error message](https://github.com/jashkenas/coffeescript/pull/4424/files#diff-55324284aab67fc5ce7570a4f82c6336R1321) isn't very friendly, and any tips to improve that would be great!

lydell: I see, makes sense. 👍

GeoffreyBooth:
@lydell did you have any further notes on this or is it ready to merge?

test/error_messages.coffee
```diff
@@ -1317,3 +1317,24 @@ test "can't use pattern matches for loop indices", ->
...
+test "bare 'super' is no longer allowed", ->
+  assertErrorFormat 'class extends A then constructor: -> super', '''
+    [stdin]:1:35: error: unexpected ->
+    class extends A then constructor: -> super
+                                      ^^
+  '''
+
+test "soaked 'super' in constructor", ->
+  assertErrorFormat 'class extends A then constructor: -> super?()', '''
+    [stdin]:1:38: error: Unsupported reference to 'super'
```
lydell: What about _illegal_ instead of _unsupported?_ (The latter almost sounds like this is a TODO comment, doesn't it?)

test/error_messages.coffee
```diff
@@ -1317,3 +1317,24 @@ test "can't use pattern matches for loop indices", ->
     a for b, {c} in d
              ^^^
   '''
+
+test "bare 'super' is no longer allowed", ->
+  assertErrorFormat 'class extends A then constructor: -> super', '''
+    [stdin]:1:35: error: unexpected ->
```
lydell: We should add a TODO comment about improving this error message..

test/regexps.coffee
```diff
@@ -110,9 +110,9 @@ test "division vs regex after a callable token", ->
     p: (regex) -> if regex then r regex else 4
   class B extends A
     p: ->
-      eq 2, super / b/g
-      eq 2, super/b/g
-      eq 2, super/ b/g
+      eq 2, super() / b/g
+      eq 2, super()/b/g
+      eq 2, super()/ b/g
       eq true, super /b/g
```
lydell: This line still has a bare `super`? Also, I'm not sure if these tests makes any sense any longer when the parentheses are needed?

connec: I'm not sure either tbh, I guess it depends whether we consider "`super()`" a 'callable token'. The test on this line is still valid, though, although I guess it's testing general parens-less calling, rather than anything regex-specific (though, I guess it's checking that a regex in that format is parsed as an arguments...).

test/error_messages.coffee
connec: 👍

test/error_messages.coffee
connec: Hrm, I copied the current node.js error message verbatim. I suspect that might be because if *is* a bit of a TODO in JS (given past discussions on ES about allowing bare `super`, amongst other things).

Since we're explicitly banning it, happy to change to "illegal" 👍

connec:
I'll aim to address those last comments and review the class super tests this evening.

test/error_messages.coffee
lydell: I see. It’s fine either way then, I guess.

GeoffreyBooth:
@connec along with making the changes discussed recently, can you please update https://github.com/jashkenas/coffeescript/wiki/%5BWIP%5D-Breaking-changes-in-CoffeeScript-2? Once the info about breaking changes for classes is in there, I think I'll move it into the docs proper.

test/error_messages.coffee
connec: I've decided to leave this as `Unsupported` for consistency with JS.

test/regexps.coffee
connec: I've not made any change for this. This test is pretty large, and seems concerned with testing the parsing of regexen vs. division. This section was added to check that things worked properly when the preceding node is a `SuperCall`, so from that point of view the test is the same (except now it could just as easily use any other function besides `super`).

connec:
@GeoffreyBooth / @lydell I've responded to the review issues. I've left some additional TODOs for the class tests that were added in #4354, rather than trying to tackle them in this PR (all the tests pass, but some should now 'fail' that were testing for broken things that now work...).

In short, I think this is ready to go.

test/regexps.coffee
lydell: I see. I think this test makes sense the way it is, then. 👍

lydell:
Good job, @connec!

PR #4427, opened 2017-01-21
[CS2] Fix handling of parameters that are function calls

### LINKED ISSUE #4406: [CS2] Destructured parameter default evaluation order is broken
Example:

```coffee
current = 0
next    = -> ++current

foo = ({ a = next() }, b = next()) -> [ a, b ]

console.log foo {}
```

Expected output: `[ 1, 2 ]`

Output on `master`: `[ 1, 2 ]`

Output on `2`: `[ 2, 1 ]`

---

This is caused by the `2` branch compiling defaults for simple parameters into the argument list, but handling complex parameters' defaults in the function body.

**Output on `master` (simplified)**

```js
function (arg, b) {
  var a
  a = arg.a != null ? arg.a : next();
  if (b == null) {
    b = next();
  }
  return [a, b];
}
```

**Output on `2` (simplified)**

```js
function (arg, b = next()) {
  var a;
  a = arg.a != null ? arg.a : next();
  return [a, b];
}
```

---

I encountered this whilst improving `super` handling in #4354 (coming soon).

I guess the way we want to fix this is to compile destructured parameters and their defaults to ES also, but I wanted to create an issue to make sure we don't forget about it!

GeoffreyBooth:
Fixes #4406. Also revises output of generator functions to be `function*` rather than `function *`, to follow [accepted idiom](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*).

@connec, any notes?

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1945,7 +1945,7 @@ exports.Code = class Code extends Base
       # encountered, add these other parameters to the list to be output in
       # the function definition.
       else
-        if param.isComplex()
+        if param.isComplex() or param.isCall()
           # This parameter is destructured. So add a statement to the function
           # body assigning it, e.g. `(arg) => { var a = arg.a; }` or with a
           # default value if it has one.
@@ -1960,7 +1960,7 @@ exports.Code = class Code extends Base
           # set by the `isComplex()` block above, define it as a statement in
           # the function body. This parameter comes after the splat parameter,
           # so we can’t define its default value in the parameter list.
-          unless param.isComplex()
+          unless param.isComplex() or param.isCall()
             ref = if param.value? then new Assign new Value(param.name), param.value, '=' else param
           # Add this parameter’s reference to the function scope
           o.scope.parameter fragmentsToText (if param.value? then param else ref).compileToFragments o
@@ -2109,6 +2109,11 @@ exports.Param = class Param extends Base
 
   isComplex: ->
     @name.isComplex()
+
+  isCall: ->
+    # We need to be careful that parameters whose default value is a function
+    # call do not call that function more than once.
+    @value? and @value instanceof Call
 
   # Iterates the name or names of a `Param`.
   # In a sense, a destructured parameter represents multiple JS parameters. This
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1945,7 +1945,7 @@ exports.Code = class Code extends Base
       # encountered, add these other parameters to the list to be output in
       # the function definition.
       else
-        if param.isComplex() or param.isCall()
+        if param.isComplex()
           # This parameter is destructured. So add a statement to the function
           # body assigning it, e.g. `(arg) => { var a = arg.a; }` or with a
           # default value if it has one.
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1945,7 +1945,7 @@ exports.Code = class Code extends Base
       # encountered, add these other parameters to the list to be output in
       # the function definition.
       else
-        if param.isComplex()
+        if param.isComplex() or param.isCall()
```
lydell: So a call does not count as "complex"? What happens if we change `.isComplex()` to also check for calls instead?

GeoffreyBooth: I just couldn’t remember what `isComplex` meant 😄 _But_ I did remember that we documented it [here](https://github.com/jashkenas/coffeescript/wiki/%5BHowTo%5D-How-parsing-works#nodescoffee), which says:
> If `node.isComplex() is false`, it is safe to use `node` more than once. Otherwise you need to store the value of `node` in a variable and output that variable several times instead. Kind of like this: `5` is not complex. `returnFive()` is complex. It’s all about “do we need to cache this expression?”

So there it is—`returnFive()` is exactly our use case. Thanks for the suggestion!

lydell: Nice!

src/nodes.coffee
```diff
@@ -2110,7 +2112,7 @@ exports.Param = class Param extends Base
     @reference = node
 
   isComplex: ->
-    @name.isComplex()
+    @name.isComplex() or (@value? and @value instanceof Call)
```
connec: Bit of a nit-pick but the check for `@value?` is redundant (`@value instanceof Call` doesn't depend on the existence of `@value`).

connec:
Neat fix 👍 made one small change request 😄

GeoffreyBooth:
Thanks @connec, I simplified it. I guess this is ready for merge?

@lydell, how would you feel about renaming `isComplex` to `needsCaching`?

lydell:
> @lydell, how would you feel about renaming `isComplex` to `needsCaching`?

That's a really good idea! However, I'm sure there's one or two places where `.isComplex()` is used to check for something else – perhaps wrongly. If you feel like it, go through each of the 40 occurrences of `isComplex` in nodes.coffee and see if it makes sense.

src/nodes.coffee
```diff
@@ -2110,7 +2112,7 @@ exports.Param = class Param extends Base
     @reference = node
 
   isComplex: ->
-    @name.isComplex()
+    @name.isComplex() or @value instanceof Call
```
connec: Ugh, pity I didn't think of this sooner: this should probably just be `@name.isComplex() or @value.isComplex()`, the following is still broken:

```coffee
i = 0
({ a = ++i }, b = ++i) ->
```

It generates:

```js
var i;

i = 0;

(function(arg, b = ++i) { // 'b' will be set first (b === 1)
  var a, ref;
  a = (ref = arg.a) != null ? ref : ++i; // 'a' will be set second (a === 2)
});
```

PR #4428, opened 2017-01-21
Fix stack trace

### LINKED ISSUE #3890: ENOENT error for missing file in stack trace
An `ENOENT` error is thrown when logging the stack trace for an error that contains a file path that is no longer on disk or fails to be read from disk.

Using CoffeeScript 1.9.1
### Steps To Reproduce
- Copy the code below to a file
- Run that file
### Code

``` coffee
fs = require 'fs'
path = require 'path'

filePath = path.join(__dirname, 'file.coffee')

# Create a file that exports a function that throws an error
fs.writeFileSync(filePath, "module.exports = -> throw new Error()")

# Require the file
throwsAnError = require(filePath)

# Delete the file
fs.unlinkSync(filePath)

# Call the exported function that throws an error and print the stack trace
try
  throwsAnError()
catch error
  console.error(error.stack)
```
### Expected Output

```
Error
  at module.exports (/Users/kevin/github/cs-test/file.coffee:1:31)
  at Object.<anonymous> (/Users/kevin/github/cs-test/a.coffee:18:3)
  at Object.<anonymous> (/Users/kevin/github/cs-test/a.coffee:1:1)
  at Module._compile (module.js:456:26)
```
### Actual Output

```
Error: ENOENT, no such file or directory '/Users/kevin/github/cs-test/file.coffee'
  at Object.fs.openSync (fs.js:427:18)
  at Object.fs.readFileSync (fs.js:284:15)
  at Object.exports._compileFile (/Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:221:14)
  at getSourceMap (/Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:340:22)
  at getSourceMapping (/Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:348:19)
  at formatSourcePosition (/Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:300:16)
  at /Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:366:33
  at Function.Error.prepareStackTrace (/Users/kevin/github/cs-test/node_modules/coffee-script/lib/coffee-script/coffee-script.js:369:7)
  at Object.<anonymous> (/Users/kevin/github/cs-test/a.coffee:20:22)
  at Object.<anonymous> (/Users/kevin/github/cs-test/a.coffee:1:1)
  at Module._compile (module.js:456:26)
```

### LINKED ISSUE #4391: Cannot read property 'readFileSync' of undefined

### LINKED ISSUE #4418: Wrong line numbers in stack trace.
Read on a few websites that just running 'coffee app.coffee' should generate source maps and print out correct debugging information.

Taken from another website, the simple example:

```
  1 test = ->
  2     people =
  3         john:
  4             first_name: 'john'
  5             last_name: 'doe'
  6         mary:
  7             first_name: 'mary'
  8             last_name: 'jane'
  9
 10     console.log("Welcome", people[p].first_name, people[p].last_name, "!!!")     for p in ['john', 'mary', 'josh']
 11
 12 process.nextTick test
```

When run with ```coffee test.coffee``` I get the following error
```
Welcome john doe !!!
Welcome mary jane !!!
/home/pi/onkyo/nodejs-onkyoserial/test.coffee:20
      results.push(console.log("Welcome", people[p].first_name, people[p].last_name, "!!!"));
                                                   ^

TypeError: Cannot read property 'first_name' of undefined
    at test (/home/pi/onkyo/nodejs-onkyoserial/test.coffee:20:52)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
    at Module.runMain (module.js:607:11)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3
```

Shows the error on test.coffee line 20, but there are only 12 lines in the file. I tried compiling it with a source map (-m) as well as with an inline map (-M) and running the resulting .js file through node, and still get errors on or around line 20.

Running coffee-script 1.12.2

GeoffreyBooth:
Fixes #4418, #4391, #3890.

So in #4399 we removed CoffeeScript’s patched `Error.prepareStackTrace`, because it was poorly implemented and threw exceptions when it shouldn’t. However, removing that patch is a major inconvenience (triggering #4418) as errors thrown by the runtime have incorrect line numbers. Just working with failing tests in the CoffeeScript codebase itself is a major pain in the ass, since the line numbers are meaningless for tests that throw errors.

So we need this patch back, but we should fix it so that it doesn’t cause the errors described in the other tickets. As I looked into it, it appears that `Error.prepareStackTrace`‘s original implementation went like this:
1. Start with the filename of the file where the exception is being thrown.
2. Reopen that file (per `getSourceMap` calling `exports._compileFile` calling `fs.readFileSync`).
3. _Recompile_ the file, generating a source map that gets cached in memory.
4. Use that source map to patch the stack trace.

The “reopen the file” part is hugely problematic, because the file might not still be available by the time the exception is thrown and this stack trace is requested; that’s what #3890 complains about. It also makes this code completely unusable in a browser, where `fs` doesn’t exist.

In #4399 @jashkenas complained that a file is compiled a second time just to generate a source map. Upon closer inspection, though, it seems to me that what’s happening is that a file gets recompiled to generate a source map _only when a stack trace is needed,_ which presumably should be a rare occurrence. I think this is the desired behavior; since the default compilation mode has source map generation off, we shouldn’t generate source maps all the time just so that we have them cached for stack traces. Recompiling only in the event of a stack trace should be faster than generating source maps all the time, whether they were requested or not. (I can see people disagreeing with me on this though; @jashkenas, what do you think?)

What we _should_ cache, then, is the code of the file we compiled. That way when a stack trace needs to recompile it to get the source map, it’s already in memory; it doesn’t need to open a file. This lets us remove the `fs.readFileSync` and its associated exceptions, which fixes #3890.

Perhaps unrelated, I patched Jison’s output to not assume that since `require` and `exports` are defined, `require(‘fs’)` must be as well. This closes #4391, and fixes edge cases in the browser compiler. I also submitted https://github.com/zaach/jison/pull/339 to try to push the fix upstream.


src/coffee-script.coffee
```diff
@@ -283,3 +295,83 @@ parser.yy.parseError = (message, {token}) ->
...
+# positions.
+Error.prepareStackTrace = (err, stack) ->
+  getSourceMapping = (filename, line, column) ->
+    sourceMap = getSourceMap filename
+    answer = sourceMap.sourceLocation [line - 1, column - 1] if sourceMap
+    if answer then [answer[0] + 1, answer[1] + 1] else null
+
+  frames = for frame in stack
+    break if frame.getFunction() is exports.run
+    "  at #{formatSourcePosition frame, getSourceMapping}"
```
lydell: While were improving all of this anyway, I think we should take the opportunity to indent the "at" lines just like Node.js does. That is, 4 spaces instead of 2.

lydell:
test.coffee:

```coffee
test = ->
  console.log a if true

do test
```

Let's try it out!

```
~/forks/coffeescript GeoffreyBooth-fix-stack-trace 
❯ ./bin/coffee test.coffee 
ReferenceError: a is not defined
  at test (/home/lydell/forks/coffeescript/test.coffee:6:26)
  ...

~/forks/coffeescript GeoffreyBooth-fix-stack-trace 
❯ coffee -c test.coffee 

~/forks/coffeescript GeoffreyBooth-fix-stack-trace % 
❯ node test.js
/home/lydell/forks/coffeescript/test.js:7
      return console.log(a);
                         ^

ReferenceError: a is not defined
    at test (/home/lydell/forks/coffeescript/test.js:7:26)
    ...
```

Observations:

1. The source-mapped line and column numbers don't seem to be correct? (deletions = actual, additions = expected)

   ```diff
   -at test (/home/lydell/forks/coffeescript/test.coffee:6:26)
   +at test (/home/lydell/forks/coffeescript/test.coffee:2:15)
   ```

   Am I missing something?

2. When running `node test.js`, I got this too:

   ```
   /home/lydell/forks/coffeescript/test.js:7
         return console.log(a);
                            ^
   ```

   I wonder why that doesn't show up in the `coffee test.coffee` call. It would be awesome if we could output this:

   ```
   /home/lydell/forks/coffeescript/test.coffee:2
     console.log a if true
                 ^
   ```

   If I change `test.coffee` to this: (note `process.nextTick`)

   ```coffee
   test = ->
     console.log a if true

   process.nextTick test
   ```

   ... I _do_ get that output, although the compiled JS version:

   ```
   ~/forks/coffeescript GeoffreyBooth-fix-stack-trace % 
   ❯ coffee test.coffee 
   /home/lydell/forks/coffeescript/test.coffee:6
         return console.log(a);
                            ^

   ReferenceError: a is not defined
       at test (/home/lydell/forks/coffeescript/test.coffee:6:26)
       ...
   ```

---

I think the approach of this PR is very good. :+1: The added code is mostly restored from the old version, with a few tweaks, right?

src/coffee-script.coffee
GeoffreyBooth: Ugh, really? I hate 4-space indentation . . .

GeoffreyBooth:
@lydell `Error.prepareStackTrace` is unchanged from when we removed it in aee27fbff03870c5479c6c33e6b1f1a32219420c. So if you’re seeing line/column number errors, those should’ve been happening back in 1.11 too. I would treat that as a separate issue/PR.

I added [new tests](https://github.com/jashkenas/coffeescript/pull/4428/files#diff-55324284aab67fc5ce7570a4f82c6336R89) to cover #3890 (file deleted between initial compilation and stack trace) and #4418 (wrong line numbers). The latter test seems to be okay to me. Technically I feel like the reported line number should be 1338 (if the first line is 1) or 1337 (if the first line is 0) but 1339 seems close enough. In my experience the line numbers in Coffee stack traces have never been overly precise, more like setting breakpoints on source maps in a browser, but they’ve been close enough to quickly find the relevant code.

src/coffee-script.coffee
lydell: I also prefer 2-space indentation in general, but I thought it might be a good idea to match the standard error formatting as closely as possible?

lydell:
> @lydell `Error.prepareStackTrace` is unchanged from when we removed it in commit aee27fb. So if you’re seeing line/column number errors, those should’ve been happening back in 1.11 too.

They were correct before:

```
~/forks/coffeescript GeoffreyBooth-fix-stack-trace % 
❯ git checkout 1.12.1
Note: checking out '1.12.1'.
...

~/forks/coffeescript 1.12.1^0 % 
❯ ./bin/coffee test.coffee 
ReferenceError: a is not defined
  at test (/home/lydell/forks/coffeescript/test.coffee:2:15)
```

GeoffreyBooth:
That’s perverse. Here’s the [diff](https://www.diffchecker.com/WrTHYIx8) between the 1.12.1 `coffee-script.coffee` and the version in this PR. I don’t see why the line numbers would be affected by these changes.

GeoffreyBooth:
@lydell you were right, the caching wasn’t working properly. I’ve made it much more robust. Stack traces should now be processing source maps correctly both for compiled files (your test) and filename-less strings compiled via `CoffeeScript.run`.

lydell:
Nice! All looks good to me now. :+1:

Regarding the "prelude":

```
/home/lydell/forks/coffeescript/test.js:7
      return console.log(a);
                         ^
```

I have no idea where that comes from, and if it is possible to do anything about it. Anyways, this PR is about restoring what was removed, so that's out of scope. But it'd be nice if we could figure out how to CoffeeScript-ify it in the future! :)

GeoffreyBooth:
@lydell It’s probably not a good idea to try to Coffeeify the “prelude.” The error that’s being thrown is a JavaScript runtime error, so the user should probably see the offending JavaScript. The original CoffeeScript might mask what the issue is, especially if the user isn’t a CoffeeScript expert and the Coffee isn’t compiling into what the user is expecting. That said, I would love to include the runtime “prelude” with our output somehow if there was a way to retrieve it.

@murrayju do you have a minute to try this branch with your project that originally led to us removing our patched `Error.prepareStackTrace`? I’d love to get some verification that this replacement doesn’t cause you any issues. I updated your [Plunker project with the new browser compiler](https://plnkr.co/edit/kzpithXAG56RlL2ijUS7), and it seems to me like it’s working as expected.

src/coffee-script.coffee
```diff
@@ -146,13 +163,13 @@ exports.run = (code, options = {}) ->
 
   # Set the filename.
   mainModule.filename = process.argv[1] =
-    if options.filename then fs.realpathSync(options.filename) else '.'
+    if options.filename then fs.realpathSync(options.filename) else '<anonymous>'
```
GeoffreyBooth: @lydell I changed this so that we have consistent “filenames” when compiling from `run` or `eval` or other non-file sources, which is important for looking up cached sources and source maps later. Do you see any issue with this? Was there anything special about a filename of `.` before?

src/coffee-script.coffee
```diff
@@ -283,3 +301,87 @@ parser.yy.parseError = (message, {token}) ->
...
+    fileLocation
+
+getSourceMap = (filename) ->
+  if sourceMaps[filename]?
+    sourceMaps[filename]
+  # CoffeeScript compiled in a browser may get compiled with `options.filename`
+  # of `<anonymous>`, but the browser may request the stack trace with the
+  # filename of the script file.
+  else if sourceMaps['<anonymous>']?
+    sourceMaps['<anonymous>']
```
GeoffreyBooth: See also.

src/coffee-script.coffee
lydell: I can't see any obvious issue, but on the other hand I have no idea if there was anything special about the `.` before either.

GeoffreyBooth: Okay. I just ran the browser tests and test.html and ran into (and fixed) some issues, which is what led me to change this. We never had a stack trace line number test that ran in the browser before, so I’m not sure our previous stack traces ever worked in the browser (I suspect they didn’t). I hope there aren’t any other edge cases I’m not thinking of.

murrayju:
:+1: Gave it a quick run through my build/tests, and it all worked fine!

GeoffreyBooth:
Thanks @murrayju!

@lydell, I guess I’ll merge this, and I guess we should do another release?

lydell:
> I guess we should do another release?

Ok, I’ll try to do that tomorrow. Remind me if I forget.

mrmowgli:
@GeoffreyBooth THIS pull request officially makes you MY hero!!  This has been broken for a while ;)

jashkenas:
Heroic of you to tackle this — nice work!

Conceptually, it feels like the right thing to do would be (while running live via `coffee`) to keep the source maps for each file cached in memory, use them when needed ... and never go back and recompile a file a second time.

That said, this works too.

GeoffreyBooth:
I haven't tested it, but I think that's what this code does. The issue is that source maps are *not* created by default, so we don't want to always create them regardless of whether they were requested, even if they're not returned. If they're created, whether in the first place or as part of a stack trace, they're cached.

src/coffee-script.coffee
gizzye: Using a tab (\t) instead of spaces would allow users to control the visual spacing in their editor/terminal preferences.

PR #4430, opened 2017-01-24
[CS2] Fix handling of parameters that are complex

GeoffreyBooth:
Fixes https://github.com/jashkenas/coffeescript/issues/4406#issuecomment-274459754. @connec?


GeoffreyBooth:
Okay, so I _think_ I’ve found a way to detect when parameter execution order might matter, and in those cases I pull the parameter out to define within the function body; but I feel like this is a losing battle. I feel like we’ll just be chasing more and more edge cases as people think of more creative ways to squeeze expressions into parameters.

connec:
@GeoffreyBooth I think the approach we should take is an all or nothing one - either all parameter defaults are compiled in the signature, or they're all compiled in the body. They are functionally equivalent, after all, so people shouldn't be too hung up about where they appear.

In terms of maximising the 'happy path', once we're compiling to ES2015 destructured parameters (with defaults), this issue will go away on its own.

I'm not sure I understand your point about

> So something as simple as `a = 1` is complex

In that case, `@value` should be a `NumberLiteral`, which is not complex. Am I misunderstanding?

lydell:
@connec So the “real” fix is not this PR, but to output destructuring as ES2015? Sounds good to me.

connec:
@lydell yes, pretty much. The test case signature (`({ a = ++i }, b = ++i)`) is valid ES2015.

```js
let i = 0
const test = ({ a = ++i }, b = ++i) => [a, b]
console.log(test({})) // [ 1, 2 ]
```

jashkenas:
> @connec So the “real” fix is not this PR, but to output destructuring as ES2015? Sounds good to me.

Amen to that.

GeoffreyBooth:
We support things in parameters that ES doesn't, like expansions or splats in non-final positions, and `this`. We already discussed dropping support for those and decided not to. We don't have the option of just limiting parameters to only what ES supports.

At the time I think I mentioned how it would be a pain to deal with `this` inside destructuring. I think we would need to check every destructured parameter to see if it contains `this`, and if so declare it in the function body. And I guess because of this execution order concern, the moment we find a first "declare in the body" parameter, all subsequent parameters also need to go in the body, like what we do now with an expansion or non-final splat.

connec:
The solution to handling `this` in parameters was partially implemented as part of the classes work:

1. Go through each name in the params (involves traversing destructured parameters) to extract a list of `this` assignments, remove `this` from the param and rename the params to temporary names if necessary.
2. Compile params as normal (all `this` references were removed).
3. Compile a list of `this` assignments in the function body (or after `super` in constructors).

Expansions and splats do make things interesting. I guess the approach we currently take there would need to continue (i.e. all parameters after a splat go in the function body).

jashkenas:
> We support things in parameters that ES doesn't, like expansions or splats in non-final positions, and this. We already discussed dropping support for those and decided not to. We don't have the option of just limiting parameters to only what ES supports.

I think maybe you should consider it again. If the mission statement for CS2 is "CoffeeScript, with all of the ES6/7 goodies", then folks are going to expect things like classes and destructuring and arrow functions to work *in the same way* that they do in JavaScript.

If it gives you more expected behavior, is easier to implement, produces cleaner output, and will lead to less future questions of "Why is my CoffeeScript destructuring doing this strange thing?", then perhaps it's the way to go.

GeoffreyBooth:
@jashkenas when there’s no choice, yes. That’s what led to the breaking change for parameter default values now applying the default only for `undefined` rather than `undefined` or `null`—we had to choose whether we were going to output ES’ default values, with their limitation of `undefined` only, or keep always assigning default values in the function body. There was no middle ground.

For something like a splat parameter, there is a middle ground: when it’s the last parameter, it gets output as ES; when it’s not, we do contortions and declare/assign variables in the function body. I think that’s the model we want to follow: when people stick to what’s possible in ES, we should output idiomatic ES; when they type something that isn’t possible in ES, but we can achieve it without too much drama, we should give them what they want. There’s no reason to ban non-default splats just because they’re not possible in ES, aside from sparing us the hassle of implementing them. Likewise with `this` in parameters, especially since there’s a lot of CoffeeScript code out there that would break if we drop support for that.

jashkenas:
> when they type something that isn’t possible in ES, but we can achieve it without too much drama, we should give them what they want

Sure, in principle. Completely sound reasoning.

> but I feel like this is a losing battle. I feel like we’ll just be chasing more and more edge cases as people think of more creative ways to squeeze expressions into parameters.

I think this might be a "too much drama" case.

GeoffreyBooth:
> I think this might be a “too much drama” case.

It might be, but in this case we could opt to keep the 1.x approach. There’s no compatibility reason to do ES’ output, unlike classes; it just pretties up the output. That improved output might not be worth the breaking changes.

GeoffreyBooth:
Getting back to this PR, I did some investigating into what was wrong with `Param::isComplex` being defined as `@name.isComplex() or (@value?.isComplex? and @value.isComplex())`. That was @connec’s suggestion that started this PR. The reason I pulled away from `isComplex` is because it started compiling
```coffee
f = (opts = {}) ->
```
into
```js
f = function(opts) {
  opts = opts != null ? opts : {};
};
```
which is not what we want. (For one thing, for ES compatibility it needs to be `!== undefined`.) So why is `{}` apparently complex?

Because all objects are complex. `Base` sets `isComplex` to true, and most classes that extend `Base` don’t redefine it. `Obj` and `Arr` are two such classes. If you think about it, it makes sense: an object can contain other objects, arrays, functions, function calls, you name it. `{}` may not be complex, but objects in general certainly could be.

So where does this leave us? Well, the safe solution would be to put back the check for `@value.isComplex()` as the way to decide whether a parameter default value can stay in the parameter list or move into the function body. I could go further and add an exception for objects and arrays, to traverse them looking for complex children; if I find none, then those objects and arrays (like `{}` or `[]`) could stay as parameter default values too. That would be the general way forward: default to defining any complex default value in the function body, and add more and more exceptions as we identify cases that are safe to put in parameter lists. What do you think?

connec:
@GeoffreyBooth I maintain that the best approach to this would be to compile all parameter defaults to ES2015, including 'complex' parameter names (i.e. destructured ones). This will resolve the ordering problem.

For the case of splats, expansions, or anything else that trips that up, we should move all default assignments *after* the interruption into the method body.

Contrived example time:

```coffee
i = 0
({ a = ++i }, b = ++i, [ c = ++i ], rest..., { d = ++i }, e = ++i, [ f = ++i ]) ->
  console.log a, b, c, d, e, f, rest
```
```js
(function ({ a = ++i }, b = ++i, [ c = ++i ], ...arg) {
  var rest, j, d, e, f;
  rest = 4 <= arg.length ? arg.slice(0, j = arg.length - 3) : (j = 0, []);
  ({ d = ++i } = arg[j++]);
  e = arg[j++];
  if (e === undefined) {
    e = ++i;
  }
  ([ f = ++i ] = arg[j++]);
  return console.log(a, b, c, d, e, f, rest);
})
console.log(foo({}, undefined, [], 'woah', {}, undefined, [])) // 1 2 3 4 5 6 [ 'woah' ]
```

connec:
**Note:** The only departure of the above from the current `2` compilation is that `a` and `c` are compiled directly as destructured parameters, and `d` and `f` are compiled as destructured assignments.

GeoffreyBooth:
@connec Yes, that is the plan. Compiling destructuring into ES syntax just hasn’t been implemented yet, and I didn’t want to tackle that in this PR. We can ship 2.0.0-alpha1 without it. It’s under discussion in https://github.com/coffeescript6/discuss/issues/69, and it’s a big enough effort that it will require its own PR.

So does this PR address everyone’s concerns with parameter ordering? I think I want to make one more change, which is to refactor `Param::isComplex` into `Param::canBeInParameterList` to make clear what this method is trying to achieve; and while I’m at it, I think a refactor of all the other `isComplex` methods is overdue. Presumably most of them could simply be renamed `mustCache` . . . @jashkenas, some insight into `isComplex` would be greatly appreciated!

But aside from unnecessary cleanup, is this PR acceptable to everyone?

src/nodes.coffee
```diff
@@ -1946,12 +1968,18 @@ exports.Code = class Code extends Base
...
-          # default value if it has one.
+          # This parameter cannot be declared or assigned in the parameter
+          # list. So add a statement to the function body assigning it, e.g.
+          # `(arg) => { var a = arg.a; }` or with a default value if it has
+          # one.
           val = ref = param.asReference o
-          val = new Op '?', ref, param.value if param.value
-          exprs.push new Assign new Value(param.name), val, '=', param: yes
+          if param.value?
+            condition = new Literal param.name.value + ' === undefined'
```
```
$ coffee -bpe 'undefined'
void 0;
```

Should we replace `undefined` with `void 0` for consistency?

src/nodes.coffee
```diff
@@ -2112,7 +2140,39 @@ exports.Param = class Param extends Base
...
+    # has a value, or else we break default values; but we want things like
+    # generator functions and iterators to evaluate in the correct order,
+    # which they might not do unless we pull them out into the function body.
+    if @name.isComplex()
+      yes
+    else if @value? and @value.isComplex()
+      if @value instanceof Value
+        unless @value.base.isComplex()
+          no
+        else
```
lydell: I'd either flip this `unless…else` around or change it into `if not…else`.

src/nodes.coffee
```diff
@@ -2112,7 +2140,39 @@ exports.Param = class Param extends Base
...
+          # Objects and arrays are always complex, but unless they contain any
+          # complex children, they’re safe to leave in the parameter list.
+          if @value.base instanceof Arr or @value.base instanceof Obj
+            hasComplexChild = no
+            @value.base.traverseChildren no, (node) ->
+              if node.isComplex()
+                # This can be further refined. An empty object will evaluate as
+                # not having any complex children, but as long as it has at least
+                # one property it will be considered complex.
```
lydell: Could you give examples on ...

- things that this code does _not_ consider complex
- things that this code _does_ consider complex, but could be improved no to in the future?

connec:
In terms of naming, perhaps flipping it around to `Node::isAtomic` (or `Node::isIdempotent`, but that's a bit long) would be clearer? An atomic/idempotent operation can be arbitrarily complex, but must not have side effects (i.e. `{}` creates a new reference, so must be cached, but `name` can appear as many time as we want, because it is an idempotent/atomic expression).

src/nodes.coffee
connec: This could be (roughly) `condition = new Op '==', param, new UndefinedLiteral`, and would then remain consistent.

jashkenas:
It's not even really about "must cache" — it's about "complicated enough that we *should* cache". A long repeated expression, even if idempotent, shouldn't be output multiple times by the CoffeeScript compiler. Cleaner to stuff the result in a temporary variable instead.

GeoffreyBooth:
@connec I think of atomic as meaning small and unified. At work I encourage atomic commits in Git, meaning each commit should do just one thing (as explained in the commit message) and nothing more. I would see `{}` and probably think it was atomic.

Idempotent is much better, but it has the same problem: why is `{}` idempotent? Thinking of idempotent as meaning “no side effects”, I would think that an empty object should qualify. It's arguably no clearer than `isComplex`.

I do like starting with `is` though. Is there a word or phrase that clearly explains what we're looking for, which is whether this node can be output repeatedly or whether we need to create a reference to it and only output the reference?

connec:
I guess from @jashkenas' comment `shouldCache` captures the compiler's perspective on it pretty well. That does make it conceptually distinct from the case here though, which is definitely param specific, and has nothing to do with caching.

In reference to `{}` - this is equivalent to `new Object` (as `[]` is to `new Array`), and so should be seen as having side-effects. 'Side-effects' in this context means within the runtime, which would perform allocations etc. Thinking about it, I probably *would* call `{}` atomic, though.

connec:
As an example of "complicated enough that we should cache", `a.b ? 'default'` will cache the intermediate `a.b`, due to `Value::isComplex` being based on the presence of accesses, even though (excepting accessor methods) `a.b` is idempotent.

jashkenas:
If you look back at old tickets, a lot of the existing `isComplex` rules in 1.X have been thought through way in the past. For example:

    a.b

Is **not idempotent**. JavaScript has getters, and `a.b` can have a side effect.

vendethiel:
@jashkenas I thought [your own stance was that getters/setters were a bad part anyway?](https://github.com/jashkenas/coffeescript/pull/2902#issuecomment-15877416)

jashkenas:
> @jashkenas I thought your own stance was that getters/setters were a bad part anyway?

It certainly is. But that doesn't make it idempotent.

GeoffreyBooth:
@jashkenas you beat me to it—yes, I need to revise this to not treat `a.b` as safe for a parameter list. Or more specifically, not safe for a parameter list _if any previous parameter has been pulled out to declare in the function body._ We could allow something like `++i` or `generator()` or `a.b` _until_ we hit a parameter that needs to be put in the function body, after which all remaining parameters would need to be declared in the function body.

Currently something like destructuring (in our 1.x mode) gets declared in the function body, without triggering the remaining parameters to also go in the function body. This would need to change. Even after we implement CS2 destructuring, there will still be the case of `this` within destructuring in a parameter, which would need to go in the function body; so this needs to be fixed regardless.

GeoffreyBooth:
And @vendethiel even if we think getters and setters are bad, we still need to support them. An object or class can be imported from a library and have a getter or setter. Using that object in a function parameter in CoffeeScript should still work.

vendethiel:
> And @vendethiel even if we think getters and setters are bad, we still need to support them. An object or class can be imported from a library and have a getter or setter. Using that object in a function parameter in CoffeeScript should still work.

We've been treating them mostly as side-effects-free for some time.

```js
$ coffee -bce "a.b? c"
if (typeof a.b === "function") {
  a.b(c);
}
```

GeoffreyBooth:
Okay, I focused on resolving the immediate issue that triggered this PR. I think I’ve got it now, but I’m tired. Can someone with fresh eyes please review this? I feel like `nodes.coffee:1970-2015` could be simplified/collapsed somehow, but my attempts kept breaking the tests. At least right now all the tests pass.

Parameter default values are preserved _until_ the first `isComplex()` param, after which all following parameters are just their reference and the default value is assigned in the function body. This preserves the correct order, while also preserving default values in the parameter list in most cases. Once we implement ES destructuring, there will be a reduction in parameters that need to be declared in the function body, and even more default values will be preserved in parameter lists. (For this, refactoring `isComplex` will go from nice-to-have to necessary, as destructuring will certainly be complex/”should cache” but not something we want to always bump into a function body rather than a parameter list.)

GeoffreyBooth:
Okay, I’ve refactored `isComplex` to `shouldCache`, including updating the relevant comments. I read through all uses of `isComplex` and this is always the intent. This new name actually fits in much better, as it’s used by other methods named `cache` and `cacheReference`. It’s all starting to make sense now.

@lydell the sections you made comments on are no longer around. After I refactored `Code` to put all declarations in the body after the first `shouldCache` parameter, it’s no longer necessary to try to determine if a parameter is idempotent or not.

I think this PR is ready for merge. Any other notes?

GeoffreyBooth:
@lydell did I satisfy your review? Do you have any further notes?

PR #4439, opened 2017-02-09
[CS2] 2.0.0-alpha1

GeoffreyBooth:
Here we go . . . separate PR to follow to update `master` with the `docs/v2` folder, so that both versions’ docs can be hosted via GitHub; and to add a note to the v1 docs that 2.0.0-alpha1 is here (after we add it to NPM).


documentation/sections/changelog.md
```diff
@@ -1,6 +1,22 @@
 ## Change Log
 
 ```
+releaseHeader('2017-02-10', '2.0.0-alpha1', '1.12.3')
+```
+*   Initial alpha release of CoffeeScript 2. The CoffeeScript compiler now outputs ES2015+ syntax whenever possible.
```
lydell: Perhaps we should say something like "The CoffeeScript compiler now _intends_ outputs ES2015+ syntax whenever possible." Currently, it sounds like CS2 is already done.

documentation/sections/changelog.md
```diff
@@ -1,6 +1,22 @@
...
+*   Added support for `async`/`await`.
+*   Bound (arrow) functions now output as `=>` functions.
+*   Function parameters with default values now use ES2015 default values syntax.
+*   Splat function parameters now use ES2015 spread syntax.
+*   Computed properties now use ES2015 syntax.
+*   Interpolated strings (template literals) now use ES2015 backtick syntax.
+*   Improved support for recognizing Markdown in Literate CoffeeScript files.
+*   Mixing tabs and spaces in indentation is now disallowed.
+*   Browser compiler is now minified using the Google Closure Compiler (JavaScript version).
+*   Node 7+ required for CoffeeScript 2.
```
lydell: Perhaps we should link to the list of breaking changes as well?

GeoffreyBooth: Good point. Added.

GeoffreyBooth:
@lydell Nothing really to review other than the changelog, and assuming you’re okay with releasing 2.0.0-alpha1 in general.

I also updated the `npm` instructions, assuming that the powers that be at NPM will let us take `coffeescript` . . .

documentation/sections/changelog.md
bdirito: package.json specifies 7.2 in engines. Does 7.[0-1] work?

GeoffreyBooth: Yes, and probably farther back than that. Basically Node 7 added support for `async`/`await` behind a flag, so that feature requires Node 7+. I haven’t tested Node 7.0 vs 7.2; generally I’ve been working in Node latest. But `async` functions aren’t used in the CoffeeScript compiler codebase, so Node 7 or the flag aren’t required to run the compiler itself. The compiler requires full support for classes, which has been around since Node 4.

So basically, to run the compiler in Node, you need Node 4+ (possibly 6+). To run all the tests, you need 7+ and the `--harmony` flag. I haven’t done tests to see exactly which versions within those ranges work.

GeoffreyBooth:
@jashkenas @lydell With the switch to `coffeescript` on NPM, at least for 2+, I thought it was confusing to still need to do `require 'coffee-script/register'`. So I went through the code and replaced `coffee-script` with `coffeescript`, including in the filenames (though not in the v1 docs). Thoughts?

lydell:
Good idea! 👍

GeoffreyBooth:
@jashkenas @lydell I’ve merged in the latest updates from the `master` and `2` branches. I think this is ready. Any last notes?

PR #4442, opened 2017-02-16
Optionally output utility function references instead of implementations

sjbarag:
Hi there!  I haven't found an open Github issue for anything like this, but I wanted to take a shot at reducing duplication in compiled CoffeeScript output.  Not how this process typically works, so please let me know if this isn't the correct way to get a change into the upstream project, especially with the CS2 alpha coming!

In large coffeescript projects with hundreds of .coffee files, the "extend", "hasProp", "bind" etc.  utilities are duplicated at the top of every file.  Using a CommonJS `require` statement, its possible to
significantly reduce compiled file sizes by referencing a common library that maintains those utility functions.  The TypeScript project implemented something similar in version 2.1 of its compiler, which
optionally references common boilerplate from the ["tslib" module](https://github.com/Microsoft/tslib).

Adding a new compiler option ("--import-helpers") allows coffeescript developers to opt into CommonJS `require`-based cost savings.

Compiled output size difference by utility function, using `--import-helpers` (assuming unminified output):
    * extend: -257 bytes/file
    * bind: -59 bytes/file
    * indexOf: -120 bytes/file
    * modulo: -34 bytes/file
    * hasProp: 0 bytes/file (unmodified; "coffeelib.hasProp" saves 0 bytes/file)
    * slice: 0 bytes/file (unmodified; "coffeelib.slice" adds 7 bytes/file)

For a project with just 10 subclasses (one per file), compiling with `--import-helpers` saves 2kB in redundant compiled output, minus some constant overhead per-file for `coffeelib = require("coffeelib"),` in
the hoisted `var` declarations.

For now, I've pushed the `coffeelib` module to [a personal repo](https://github.com/sjbarag/coffeelib), but **I haven't published anything to npm**.  In the event this PR gets accepted, I'll happily transfer ownership of that repo to whomever is running things these days, so they can publish the module when they're ready :smile:


vendethiel:
I'm interested in the numbers for minified output

jashkenas:
This would be a nice feature to have ... if perhaps a little hinkey to get implemented nicely.

sjbarag:
> I'm interested in the numbers for minified output

@vendethiel Same here! I'll run things through uglifyjs tonight and report back. If you've got a minified that you prefer, let me know and I'll test with that one too.

lydell:
A few years ago, the savings were about 200 bytes after gzip on a medium sized project: https://github.com/brunch/brunch/issues/714#issuecomment-29019351

sjbarag:
Sorry all, work got pretty crazy last night.  I'll have an update for you with `uglify-js` and `gzip` compression sometime on Monday, February 20th.

GeoffreyBooth:
Something to be aware of is that many of these helper functions are removed in CS2, as we shift to using native implementations of things like `indexOf`. Maybe check the 2 branch to see how many of these helper functions are left. And of the ones that remain, some might be there simply because we haven’t gotten around to phasing them out in favor of the native versions; something else to check.

sjbarag:
This work was originally inspired by the single-page app we ship at my day-job, which is compiled down into one multi-part JavaScript file due to architectural limitations.  The app weighs in at ~650 `.coffee` source files with >400 classes, all of which are combined via webpack and -- depending on build type -- are conditionally minified and gzip'd.  It seems like a great project to use for these tests, both because I'm familiar with it and because it's rather large (it also exposed two minor bugs that I'll be fixing in the next few days 😅). Unfortunately I can't release the source code without losing my job, so if you'd rather I use a different app let me know what you have in mind!

On to the tests.

The app has three different build configurations: "development" (webpack but no minification), "release" (webpack + `uglifyjs -cm` mangling + compression), and "production" (webpack + `uglifyjs` mangling + `gzip` on the resulting file with default settings).  I built the app with upstream `coffee-script@1.12.3` in each configuration, and did the same with `--import-helpers` enabled.  There was a reduction in filesize for all configurations.

| Implementation | Unminified Size (bytes) | Minified Size (bytes) | Minified + gzip'd Size (bytes) |
| --- | ---: | ---: | ---: |
| Upstream `coffee-script` | 6,603,867 | 1,918,488 | 456,944 |
| Using `Coffeelib`        | 6,532,835 | 1,864,034 | 454,799 |
| Delta (bytes)            |   -71,032 |   -54,454 |  -2,145 |
| Delta (percent change)   |    -1.07% |    -2.83% |  -0.47% |

@lydell's [2013 gzip test](https://github.com/brunch/brunch/issues/714#issuecomment-29019351) seems to match these results.  While `--import-helpers` doesn't make our compressed production releases significantly smaller, I'd love to skim 71kB off of the dev builds I and my colleagues load dozens of times per day.

---

> many of these helper functions are removed in CS2, as we shift to using native implementations of things like `indexOf`

@GeoffreyBooth I haven't gotten a chance to try building our app with CS2 yet, but if it's not a backwards-compatible upgrade it may take us a while to migrate. Even then, our browser targets have inconsistent and incomplete ES5 support.  I have a strong feeling that we're using the polyfill provided by the `indexOf` helper function instead of the native.  Luckily those polyfills are [still in CS2](https://github.com/jashkenas/coffeescript/blob/ff56533e0b995fbe51028dc468dd17ee41c55ab1/src/nodes.coffee#L2992-L3000) but I'm not sure if that's going to change before its final release.

That said, I'd be happy to also port these changes into CS2 if that'd make the release process easier for you all!

lydell:
Thanks for the numbers, @sjbarag!

I'd be very interested in why the dev build sizes matter to you! (Even better – _numbers_ on how a smaller size help you.) The only benefit I can think of right now is faster parsing times. That might be especially useful if you do a lot of development on "lower-end" devices. Assuming that parsing time is linear, are you looking to shave off 1% of the reload times?

I feel a bit hesitant against emitting `require()` calls from the CS1 compiler. I think we need to wait for CS2 to get rid of most of them.

GeoffreyBooth:
@sjbarag See https://rawgit.com/jashkenas/coffeescript/2/docs/v2/index.html#breaking-changes. There are very few breaking changes, and the need to convert bare `super` to `super()` or `super args` is the only one we needed to make for the CoffeeScript codebase itself. The hope is that it should be a fairly minor upgrade for most projects. As we’re nearing alpha, it would be great to have a sizeable project such as yours try it out, especially if you have tests that can verify that things still all work as expected when compiled via `2`.

As for your browser support, the plan is that you would simply pass CoffeeScript 2’s output through Babel or a similar transpiler to make it browser-safe. This also allows you to use all the ES “opt in” features like generators and modules and async/await.

You clearly are sweating bytes, so something else you may want to consider is trying the [Google Closure Compiler](https://github.com/google/closure-compiler-js). I switched CoffeeScript’s browser compiler build from Uglify to the JavaScript version of Google Closure Compiler and the final output shrank quite a bit, though the minification time is now much longer.

jashkenas:
Thanks for providing the real numbers — that's extremely helpful!

But for a **-0.47%** savings in a production app with over 650 separate CoffeeScript files, this doesn't really seem like something worth adding at this point. 

It's becoming less relevant over time, as CS2 will have fewer helper functions, and that tiny savings doesn't seem worth the conceptual complexity of adding a flag to the compiler.

In addition, it seems as though you can get the results you're looking for by concatenating some of your 650 CoffeeScript files before compiling them together.

GeoffreyBooth:
Sounds like we should close this. @sjbarag thanks for looking into this. If you’d like to help the `2` branch pare down some of its now-unnecessary helper functions, that might achieve the same goal.

PR #4444, opened 2017-02-17
Place ending heregex tokens one index earlier

alangpierce:
This is an upstream port of https://github.com/decaffeinate/coffeescript/pull/17

The lexer generates fake tokens for interpolated heregexes, and the ending
tokens were being placed where the start (inclusive) and end (inclusive) index
were one past the end of the heregex. This meant that in a case like
`[a ///#{b}///]`, the end tokens of the heregex and also the implicit function
call end were all being placed at the `]`, so the AST location data would say
that the function call ends at the end of the `]`.

To fix, I can just subtract 1 from the position of those ending heregex tokens
so that their end lines up with the end of the heregex itself. This is similar
to previous fixes that changed `OUTDENT` and `CALL_END` tokens so that the end
of the token lines up with the end of the AST node.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -403,11 +403,11 @@
             double: true
           }, this.formatHeregex);
           if (flags) {
-            this.token(',', ',', index, 0);
-            this.token('STRING', '"' + flags + '"', index, flags.length);
+            this.token(',', ',', index - 1, 0);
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
@lydell is there a reason you haven’t just merged this and #4445 in?

lydell:
I guess there isn't.

PR #4445, opened 2017-02-17
Properly update location data when setting a call to use `new`

alangpierce:
This is an upstream port of https://github.com/decaffeinate/coffeescript/pull/24

In a case like `new A().b(c)`, the jison structure ends up being different from
the resulting AST. To the jison parser, this is the `new` unary operator applied
to the expression `A().b(c)`. When the unary operator is applied, the
`Call.prototype.newInstance` function traverses into the leftmost function call
and sets the `isNew` flag to true, and the `Op` constructor returns the `Call`
node so that the call is used in place of the unary operator. However, the code
wasn't updating the node location data, so this commit fixes that.

It's sort of hard to get the location data in `newInstance`, so we set a flag on
every affected node in `newInstance` and override `updateLocationDataIfMissing`
(which is called with the location data after the fact) so that it updates just
the starting position.

### DIFF
--- a/test/parser.coffee
+++ b/test/parser.coffee
@@ -38,3 +38,43 @@ test "operator precedence for binary ? operator", ->
   eq expression.second.first.base.value, 'b'
   eq expression.second.operator, '&&'
   eq expression.second.second.base.value, 'c'
+
+test "new calls have a range including the new", ->
+  source = '''
+    a = new B().c(d)
+  '''
+  block = CoffeeScript.nodes source
+
+  assertColumnRange = (node, firstColumn, lastColumn) ->
+    eq node.locationData.first_line, 0
+    eq node.locationData.first_column, firstColumn
+    eq node.locationData.last_line, 0
+    eq node.locationData.last_column, lastColumn
+
+  [assign] = block.expressions
+  outerCall = assign.value
+  innerValue = outerCall.variable
+  innerCall = innerValue.base
+
+  assertColumnRange assign, 0, 15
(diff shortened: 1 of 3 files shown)


test/parser.coffee
```diff
@@ -38,3 +38,43 @@ test "operator precedence for binary ? operator", ->
...
+    eq node.locationData.last_column, lastColumn
+
+  [assign] = block.expressions
+  outerCall = assign.value
+  innerValue = outerCall.variable
+  innerCall = innerValue.base
+
+  assertColumnRange assign, 0, 15
+  assertColumnRange outerCall, 4, 15
+  assertColumnRange innerValue, 4, 12
```
lydell: Why 12 here? (I'm having a bit of a difficulty mapping `outerCall`, `innerValue` and `innerCall` to the corresponding parts of `source`.)

alangpierce: `outerCall` is `new B().c(d)`
`innerValue` is `new B().c`
`innerCall` is `new B()`

All three of these needed to have their locations updated by this change, since previously they ignored the `new` on the left.

lydell: Thanks, looks like I got those right after all. And when counting again, I get it to 12, too. False alarm!

GeoffreyBooth:
@alangpierce I’ve merged this into the `2` branch and it’s caused some broken tests over there. Do you mind please taking a look?

alangpierce:
@GeoffreyBooth ok, I looked into it and put up #4447.

PR #4447, opened 2017-02-18
[CS2] Fix CS2 build after merge pulled in a bare super call

alangpierce:
Looks like super calls need to explicitly specify their args now, so I added the
arg in `Call.prototype.updateLocationDataIfMissing`.

Also, the JS code hadn't updated, I guess due to the build failure, so re-run
`bin/cake build` so `lexer.js` and `nodes.js` both get updated. This fixes the
tests.

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -397,11 +397,11 @@
             double: true
           }, this.formatHeregex);
           if (flags) {
-            this.token(',', ',', index, 0);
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Thanks! P.S. You might find this of interest: https://rawgit.com/jashkenas/coffeescript/2/docs/v2/index.html#breaking-changes

alangpierce:
👍 thanks, I hadn't seen that.

PR #4448, opened 2017-02-22
2 docs on master

GeoffreyBooth:
This adds the `v2` folder (with the CS2 docs) to `master`. This doesn’t overwrite the CS1 docs.

I also added a brief note in the introduction to the CS1 docs announcing CS2 and linking to the 2 docs.


jashkenas:
I guess we still need to cherry-pick over to gh-pages though...

PR #4459, opened 2017-03-09
Allow middle mouse button scrolling

1j01:
Fixes autoscrolling the page when starting MMB dragging on a code example.

The code examples auto-expand so you don't need to scroll them. You would want to scroll in Try CoffeeScript but CodeMirror doesn't support autoscrolling anyways, it only prevents it.

Note: This likely breaks the MMB pasting found in some Linux systems, i.e. `X-selection-paste`.
If that's the case and is deemed unacceptable, we could at *least* allow autoscrolling on the readonly editors.

### DIFF
--- a/documentation/v2/docs.coffee
+++ b/documentation/v2/docs.coffee
@@ -74,6 +74,10 @@ $(document).ready ->
           editors[index + 1].setValue output
         , lastCompilationElapsedTime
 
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
This seems like a bug in CodeMirror, that should be patched there.

PR #4461, opened 2017-03-11
Fix #4422, `++` and `--` with expressions in parens

### LINKED ISSUE #4422: Lexer: Plusplus operator (++) on braced expression leads to parse error.
Please, consider:

```coffeescript
++(foo)
```

While perfectly valid as JavaScript, as CoffeeScript it leads to an "Error on line 1: unexpected end of input".  Should that happen?

Otoh, `+(foo)` is fine, as is `++foo`.

Tried for CoffeeScript `v1.12.1`.

FabianLauer:
Fixes #4422.

Also fixed the string interpolation in the 'Future Reserved Words' test (7bf9e23) ...

> Added the missing '#'s for correct string interpolation. Seems like this went unnoticed because in these test cases, the `check` function tests if the compilation fails, which it does either way.


lydell:
Thanks for the pull request!

Unfortunately, `Expression` is too permissive. There are many things in there that you can't `++` or `--`.

I'm not sure what exactly is the correct fix. Perhaps the ECMAScript specification on [Update Expressions](http://www.ecma-international.org/ecma-262/7.0/index.html#sec-update-expressions) can help.

FabianLauer:
I'll look into that, thanks for the hint 😃

GeoffreyBooth:
Closing until this is revised per the notes above. Please comment when it is fixed and I will reopen.

PR #4472, opened 2017-03-22
Write .coffee_history in $XDG_CACHE_HOME

### LINKED ISSUE #3705: Follow the XDG directory spec for .coffee_history
The coffee history file should be stored/read from the `$XDG_CACHE_HOME` directory (which defaults to `$HOME/.cache`

Suggested file: `$XDG_CACHE_HOME/coffee/history`

reference spec: http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

jessaustin:
Previously, if the `$HOME` environmental variable was set, the `.coffee_history` file was written there. This is still the case, but first we check if `$XDG_CACHE_HOME` is set, and if so we write it there instead. This allows users who want to follow the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) to do so. Actually, this helps out *all* users who want to keep their home directory a bit cleaner. It seems that using this standardized environmental variable is the cleanest way to do this, rather than involving configuration files or command line flags or anything else.

Fixes #3705.

### DIFF
--- a/src/repl.coffee
+++ b/src/repl.coffee
@@ -7,7 +7,8 @@ CoffeeScript = require './coffeescript'
 
 replDefaults =
   prompt: 'coffee> ',
-  historyFile: path.join process.env.HOME, '.coffee_history' if process.env.HOME
+  historyFile: do (p = process.env.XDG_CACHE_HOME or process.env.HOME) ->
+    path.join p, '.coffee_history' if p
   historyMaxInputSize: 10240
   eval: (input, context, filename, cb) ->
     # XXX: multiline hack.
--- a/src/repl.coffee
+++ b/src/repl.coffee
@@ -7,8 +7,9 @@ CoffeeScript = require './coffeescript'
 
 replDefaults =
   prompt: 'coffee> ',
-  historyFile: do (p = process.env.XDG_CACHE_HOME or process.env.HOME) ->
-    path.join p, '.coffee_history' if p
+  historyFile: do ->
+    historyPath = process.env.XDG_CACHE_HOME or process.env.HOME
+    path.join historyPath, '.coffee_history' if historyPath
   historyMaxInputSize: 10240
   eval: (input, context, filename, cb) ->
     # XXX: multiline hack.


src/repl.coffee
```diff
@@ -7,7 +7,8 @@ CoffeeScript = require './coffeescript'
 
 replDefaults =
   prompt: 'coffee> ',
-  historyFile: path.join process.env.HOME, '.coffee_history' if process.env.HOME
+  historyFile: do (p = process.env.XDG_CACHE_HOME or process.env.HOME) ->
+    path.join p, '.coffee_history' if p
   historyMaxInputSize: 10240
```
GeoffreyBooth: I disapprove of variable assignments in function parameter lists, and of single-character variable names other than indexes/iterators. (I know, there are examples of both in the CoffeeScript codebase, but I’ve been slowly removing them as I’ve encountered them.) Also what if neither of those environment variables is set? `historyFile` should be set to _something_.

I suggest:
```coffee
historyFile: do ->
  historyPath = process.env.XDG_CACHE_HOME or process.env.HOME or process.cwd()
  path.join historyPath, '.coffee_history'
```

jessaustin: Fair point on the variable name.

Current functionality is not to write a history file if `$HOME` is unset, which from my perspective is absolutely correct. We don't know what kind of weird environment we have if `$HOME` isn't set, and I know from playing around with other tools like `git` that they prefer not to guess about this sort of thing. One can imagine a user with no write access to `$PWD`; what a pain for them if `coffee` simply refuses to run!

GeoffreyBooth: Fair point about `$PWD`. Are you sure that the REPL works if `historyFile` is undefined? I’m not sure that’s ever been tested. Like if you change the current code to:

```coffee
historyFile: path.join process.env.HOME, '.coffee_history' if process.env.SOME_VARIABLE_THAT_DOESNT_EXIST
```

jessaustin: Further along in the file, a history file is created only if `opts.historyFile` is `true`. It's perhaps not ideal that two separate pieces of code are dealing with this, and unsetting `$HOME` is a slightly weird way to convey "don't write a history file", but I err on the side of making as small a change as possible.

```coffeescript
addHistory repl, opts.historyFile, opts.historyMaxInputSize if opts.historyFile
```

In my environment I've tested all four situations, with both `$HOME` and `$XDG_CACHE_HOME` set and unset.

jashkenas:
How widely used is `$XDG_CACHE_HOME`?

Is it a general linux thing?

lydell:
It is part of the [XDG Base Directory Specification](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support).

jessaustin:
The [support document](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support) indicates that XDG directories are widely supported. For instance, git, fish, chromium, inkscape, etc. provide "full" support. less, npm, tmux, wget, etc. provide "partial" support. This PR I would consider to provide "full" support, because it uses the standard environmental variable `$XDG_CACHE_HOME`.

jashkenas:
Groovy.

Issue #3894, opened 2015-03-10
ES: Allow deconstructing objects with splats


flying-sheep:
…for parity with ECMA7. currently we can’t do this:

``` coffeescript
x =
  a: 1
  b: 2
  c: 3

{ a, b, y... } = x
# or
{ a, b, ...y } = x

console.log a, b, y  # → 1 2 {c:3}
```

vendethiel:
Seems like #815 proposed this feature but was closed for other reasons.

lydell:
Is that really part of ES6? I've never seen it before.

flying-sheep:
ah, you seem to be right, it’s actually part of [JSX](http://facebook.github.io/react/docs/transferring-props.html#transferring-with-...-in-jsx). i assumed everything of JSX that isn’t `<ReactElement/>` syntax is a subset of ES6.

dayum, i think it’s really useful

lydell:
Cool! Thanks for those links!

flying-sheep:
no problem.

at closing of that other issue: luckily, that’s about something different (i.e. extending objects, not destructuring them)

and @jashkenas’ argument…

> Splats are for lists of things

…is also not valid anymore now that ECMAscript itself uses splats for objects.

so this should be good to go!

flying-sheep:
now it’s in the [compat-table](http://kangax.github.io/compat-table/es7/#Object_rest_properties)

bbtfr:
+1 for this, it would be very helpful to write React Apps in CoffeeScript with this feature.

reduxdj:
Hoping this is added to Coffeescript, is it going to be?

wmertens:
So I'd also love to see `{b:5, bar..., c}` as the shorthand for `merge({b: 5}, bar, {c: c})` and while we're at it `{[foo]: bar}` for `_t={},_t[foo]=bar,_t`, all very handy things slated for ES6/7.

Should these go into separate issues?

lydell:
`{[foo]: bar}` is already possible: `{"#{foo}": bar}`

wmertens:
Ooh, I didn't know about that, nicer syntax too! That just leaves the spread and rest operations...

michaelficarra:
@lydell with the caveat that it doesn't work for symbols.

vendethiel:
Well, since @jashkenas said he considered symbols a misfeature, I guess that's fine for coffee?

michaelficarra:
[...]

PR #4473, opened 2017-03-23
Spread in objects

zdenko:
Connected to [#3894](https://github.com/jashkenas/coffeescript/issues/3894).

In CS2 emit

```
obj = { x: 1, y: 2}
obj1 = {obj..., z: 3}
```
to

`obj1 = {...obj, z: 3}`

Improve destructuring assignment for object literal:

```
{a, b, x...} = {a:1, b:2, c:3, d:4, e:5}
a = 1
b = 2
x = {c:3, d:4, e:5}
```

```
{c, x..., e} = {a:1, b:2, c:3, d:4, e:5}
c = 3
e = 5
x = {a:1, b:2, d:4}
```

```
{c, x..., a:p} = {a:1, b:2, c:3, d:4, e:5}
c = 3
p = 1
x = {b:2, d:4, e:5}
```

```
futurists =
  sculptor: "Umberto Boccioni"
  painter:  "Vladimir Burliuk"
  poet:
    name:   "F.T. Marinetti"
    address: [
      "Via Roma 42R"
      "Bellagio, Italy 22021"
    ]

{poet: {name, addr1...}} = futurists
addr1 = {address: [ "Via Roma 42R",  "Bellagio, Italy 22021"]}

{poet: {addr2..., name:title}} = futurists
addr2 = {address: [ "Via Roma 42R",  "Bellagio, Italy 22021"]}
```


src/nodes.coffee
```diff
@@ -1748,6 +1748,13 @@ exports.Assign = class Assign extends Base
     vvarText = fragmentsToText vvar
     assigns  = []
     expandedIdx = false
+    # check if variable is object and containes splat, e.g. {a, b, c...}
+    hasSplat = isObject and (i for obj, i in objects when obj instanceof Splat).length == 1
```
vendethiel: I think it's better to error out if length>1?

zdenko: Since multiple splats/expansions are disallowed in an assignment, length == 1 should be ok?

test/assignment.coffee
```diff
@@ -227,7 +227,35 @@ test "destructuring assignment with objects and splats", ->
   {a: b: [y, z...]} = obj
   eq a, y
   arrayEq [b,c,d], z
-
+  
+test "destructuring assignment with objects and splats: ES2015", ->
+  obj = {a:1, b:2, c:3, d:4, e:5}
+  {a:x, b, r...} = obj
+  {a:y, c..., d} = obj
```
lydell: The rest element has to be last, according to Babel: http://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015%2Cstage-2%2Cstage-3&targets=&browsers=&builtIns=false&code=const%20%7Ba%3Ay%2C%20...c%2C%20d%7D%20%3D%20obj&experimental=false&loose=false&spec=false&playground=true

I guess CoffeeScript should support it anywhere, though, just like for arrays? I guess we just need to move the splat last in the compiled JS.

test/assignment.coffee
```diff
@@ -227,7 +227,35 @@ test "destructuring assignment with objects and splats", ->
...
+  eq b, 2
+  eq x, 1
+  eq r.d, d
+  eq c.b, b
+  eq r.e, c.e
+
+test "destructuring assignment with context (@) properties: ES2015", ->
```
lydell: I don't see any `@`s in this test?

src/nodes.coffee
lydell: No. We need to show a nice error message, like when you try to use multiple splats in array desctructuring or parameter list.

test/objects.coffee
```diff
@@ -44,6 +44,15 @@ pluck = (x) -> x.a
...
 
 
+test "Object literal spread", ->
+  obj1 =
+    a: 1
+    b: 2
+  # obj2 = "#{{obj1..., b: 3}}"
+  t1 = CoffeeScript.compile "{obj1..., b: 3}", bare: yes
+  t1 = t1.replace(/\s+/g, "")
+  eq t1, "({...obj1,b:3});"
```
lydell: The rule has been to never compare strings with the compiled output JS, unless there's no other way to assert that the compiled code works as intended. Also, CoffeeScript always strives to produce readable output. You need to add some spaces in there.

test/assignment.coffee
zdenko: That's a typo. I copied the string.

test/objects.coffee
zdenko: Error for multiple splats is already there at line 1791 in lib/nodes.coffee

lydell: Did you reply to the wrong comment? :)

Anyway, we need a test for the multiple splats error.

GeoffreyBooth:
@zdenko thank you for making this effort! Do you feel like helping with the [destructuring branch](https://github.com/GeoffreyBooth/coffeescript/tree/destructuring)? The goal there is to output ES2015 destructured syntax whenever possible. Certain Coffee destructuring arrangements aren’t possible in ES, so should stay as they are; and default values should follow the ES style of only getting applied with the variable is undefined, not undefined or null.

src/nodes.coffee
zdenko: This is the current status :

```
./bin/coffee -e "{a, x..., c, y...} = {a:1, b:2, c:3, d:4, e:5}"
[stdin]:1:14: error: multiple splats/expansions are disallowed in an assignment
{a, x..., c, y...} = {a:1, b:2, c:3, d:4, e:5}
```

zdenko:
@GeoffreyBooth sure. 

So, the output of destructing
`{a, b, x...} = {a:1, b:2, c:3, d:4}`
should be
`{a, b, ...x} = {a:1, b:2, c:3, d:4}`

And, the rest element has to be the last element.

test/objects.coffee
zdenko: Sorry. Long hours :(

Anyway, I couldn't find example how to test the output of `obj2 = {obj1..., b:5}`
CS compiles it to 
```
(function() {
  var obj2;

  obj2 = {
    ...obj1,
    b: 5
  };

}).call(this);
```

but test fails since the syntax {...obj1, b:5} is currently in stage-3 and requires Babel.

To rephrase the question, can Babel be used for tests?

GeoffreyBooth: It is our policy to only support ES features that have reached stage 4, and are supported in the Node runtime (modules are the big exception to this). Therefore the test would compile to the finalized ES syntax, and Node should be able to run it.

GeoffreyBooth:
@zdenko that syntax [doesn’t seem to work for me](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=latest&targets=&browsers=&builtIns=false&code=%7Ba%2C%20b%2C%20…x%7D%20%3D%20%7Ba%3A1%2C%20b%3A2%2C%20c%3A3%2C%20d%3A4%7D), but in general, yes, whenever there is an ES destructured syntax that is equivalent to a CoffeeScript one, we want to output the ES one. In https://github.com/jashkenas/coffeescript/pull/4311 I refactored the handling of function parameters to support expansions and splats output as ES; if the splat was the last parameter, the conversion was straightforward, otherwise a placeholder splat became the last parameter and further assignments happened in the function body. We could do something similar for destructuring.

zdenko:
@GeoffreyBooth  [this syntax](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=latest%2Cstage-3&targets=&browsers=&builtIns=false&experimental=true&loose=false&spec=false&playground=true&code=const%20%7Ba%2C%20b%2C%20...x%7D%20%3D%20%7Ba%3A1%2C%20b%3A2%2C%20c%3A3%2C%20d%3A4%7D&stage=0) should work.

The proposal for [object rest/spread properties](https://github.com/tc39/proposal-object-rest-spread) is in stage-3.

Since CS supports only stage-4, the output of object spread should be:
`n = {x, y, z...}` =>  `n = Object.assign({}, x, y, z}`
`n = { a..., { x: 1, y: 2 }... }` => `n = Object.assign({}, a, { x: 1, y: 2 })`

Is that correct?

GeoffreyBooth:
Closing in favor of #4493

PR #4475, opened 2017-03-24
v2 docs: assuming CS arrow is still `->`

tomByrer:
& not `=>`

Thanks for the V2!


GeoffreyBooth:
CoffeeScript has both `->` and `=>`. The sentence you modified was saying how CoffeeScript’s `=>` is now output as ES’s `=>`.

Issue #4477, opened 2017-03-28
[CS2] Regression in handling default for destructured param


connec:
In CS 1.x the following pattern could be used to set a default for a destructured parameter:

```coffee
({ a } = {}) ->
```

This outputs:

```js
(function(arg) {
  var a;
  a = (arg != null ? arg : {}).a;
});
```

In CS 2.0.0-alpha1 the output has changed to:

```js
(function(arg) {
  var a, arg;
  if ({
    a: a
  } === void 0) {
    a = {}.a;
  }
});
```

The undefined check is now happening on the destructuring expression, rather than the argument name.

greghuc:
@connec I just ran into this when migrating a project to CS2.

Test case:

```
test = ({ value } = {}) -> console.log "Value: #{value}"
test value: 42
```

Output:
```
CS1: Value: 42
CS2: Value: undefined
```

GeoffreyBooth:
In #4478, this:
```coffee
test = ({ value } = {}) -> console.log "Value: #{value}"
```
now outputs as this:
```js
var test;

test = function({value} = {}) {
  return console.log(`Value: ${value}`);
};
```
and `test { value: 42 }` prints `Value: 42`.

connec:
Neat! Yeh I ran across this when doing tests for destructuring.

I suspect we will have a similar issue with any non-assignable destructured parameter with a default).

```js
// ([head..., last] = []) ->
var slice = [].slice;

(function(arg) {
  var arg, head, i, last, ref;
  // this should be checking `arg`, not its default value
  if ([...head, last] === void 0) {
    ref = [], head = 2 <= ref.length ? slice.call(ref, 0, i = ref.length - 1) : (i = 0, []), last = ref[i++];
  }
});
```

Additionally, it looks like #4478 has broken expansions in destructured parameters:

```
stdin]:1:3: error: Expansion must be used inside a destructuring assignment or parameter list
([..., last] = []) ->
  ^^^
```

PR #4478, opened 2017-03-29
[CS2] Destructuring

GeoffreyBooth:
The intent of this PR is to output CoffeeScript’s destructuring as ES2015+ destructured syntax whenever possible. Migrating the task from https://github.com/coffeescript6/discuss/issues/69, this:

```coffee
numbers = [1, 2, 3]
pairs = { a: 1, b: 2, c: 3 }

[one, two, three] = numbers
{one, two, three} = pairs

foo = ->
  [@a, @b, @c] = numbers
  {@a, @b, @c} = pairs
  return
```
would compile to [this](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=latest&experimental=false&loose=false&spec=false&code=var%20foo%2C%20numbers%2C%20one%2C%20pairs%2C%20three%2C%20two%3B%0A%0Anumbers%20%3D%20%5B1%2C%202%2C%203%5D%3B%0A%0Apairs%20%3D%20%7B%0A%20%20a%3A%201%2C%0A%20%20b%3A%202%2C%0A%20%20c%3A%203%0A%7D%3B%0A%0A%5Bone%2C%20two%2C%20three%5D%20%3D%20numbers%3B%0A(%7Bone%2C%20two%2C%20three%7D%20%3D%20pairs)%3B%0A%0Afoo%20%3D%20function()%20%7B%0A%20%20%5Bthis.a%2C%20this.b%2C%20this.c%5D%20%3D%20numbers%3B%0A%20%20(%7Ba%3A%20this.a%2C%20b%3A%20this.b%2C%20c%3A%20this.c%7D%20%3D%20pairs)%3B%0A%7D%3B%0A):
```js
var foo, numbers, one, pairs, three, two;

numbers = [1, 2, 3];

pairs = {
  a: 1,
  b: 2,
  c: 3
};

[one, two, three] = numbers;
({one, two, three} = pairs);

foo = function() {
  [this.a, this.b, this.c] = numbers;
  ({a: this.a, b: this.b, c: this.c} = pairs);
};
```
The array destructuring, as you can see here, is pretty straightforward; but the object destructuring presents challenges. First, if we’re not declaring the variable at the same time (as we never do in CoffeeScript, hence the `var` line of declarations at the top of each scope) then we need to wrap the destructuring assignment in parentheses, [per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration). And if the target variable is attached to `this`, we need to use the [“assigning to new variable names” shortcut](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names) to destructure into our intended destination variable, unless people have a better suggestion for how to handle this.

There is an intentional breaking change in that CS2 destructuring, like CS2 function default parameters, will apply default values only when a variable is `undefined`, not falsy.


GeoffreyBooth:
This is really preliminary, but I thought I’d create this PR just to start the conversation. I could use a lot of help with this. 😬 @zdenko has done some great work with object destructuring; I would love if you could also help tackle the core of ES destructuring!

I didn’t want to commit any code that horribly breaks the tests, so all I’ve committed so far are various helpers that I think will be necessary in order to get `Assign` to produce destructured output at least for the simplest case, `{a} = b`.

So I can get that case to output correctly, as:
```js
var b;

({ b } = a);
```

if I change `nodes.coffee:1773-1782` to the following:
```coffee
      acc   = idx.unwrap() instanceof PropertyName
      value = new Value value
      value.properties.push new Index idx unless acc
      message = isUnassignable obj.unwrap().value
      obj.error message if message
      value = new Op '??', value, defaultValue if defaultValue
      assignOptions =
        param: @param
        wrapVariableInBraces: acc
      return new Assign(obj, value, null, assignOptions).compileToFragments o, LEVEL_TOP
```

but doing that causes 6 tests to fail, and those tests verge on incomprehensible to me.

I’d like to start small and gradually expand to more and more complicated destructuring cases, but it appears that destructuring is implemented in a very compact, recursive way that discourages piecemeal approaches. Perhaps @lydell you can help untangle it? I’m hoping to avoid the near-rewrite that I did for the function parameters. At least function parameters are a straightforward concept that I’m pretty sure I understand; destructuring can get awfully complicated . . .

GeoffreyBooth:
Okay, I’ve made a lot of progress. The tests now pass, and simple cases like `{a} = b` now output in ES syntax. Any feedback on the approach so far?

GeoffreyBooth:
Update: The branch I started this PR with has been renamed [`destructuring-geoffrey`](https://github.com/GeoffreyBooth/coffeescript/tree/destructuring-geoffrey) and @connec’s branch that he submitted #4479 with has become [my repo’s `destructuring`](https://github.com/GeoffreyBooth/coffeescript/tree/destructuring), and therefore the source for this PR.

GeoffreyBooth:
So picking up from #4479, I think @connec’s approach for destructuring is more promising than the one I was pursuing, so I want to start from his branch and cherry-pick improvements from mine. Top on my list is default values. @connec did you have an implementation in mind for those? How do you feel about [the way default values are implemented in my branch](https://github.com/GeoffreyBooth/coffeescript/blob/destructuring-geoffrey/src/nodes.coffee#L1649), and does that complement your approach?

connec:
Copied from #4479 -

I'll take another look tonight.

The main approach I'm going for is to use as much of the existing machinery as possible. If you look at the nodes for destructured expressions they look like they should just about compile verbatim, and it's only guard logic in `nodes.coffee` that's preventing that. With some small tweaks we should be able to update the guard logic to be more permissive (e.g. [here](https://github.com/jashkenas/coffeescript/pull/4479/files#diff-d149b4f007169f06a8ad463f9b2f277bL1663)).

For compiling default values, I was planning to just leverage the existing node structure, e.g.

```sh
$ echo '{ a = 1 }' | bin/coffee -bns
Block
  Assign
    Value
      Obj
        Assign
          Value IdentifierLiteral: a
          Value NumberLiteral: 1
    Value IdentifierLiteral: b
```

We should be able to tweak `Obj` so that it will compile that inner assign verbatim, rather than doing special handling for `Assign`'s whose context isn't `object`.

By destructured parameters I just mean e.g. `({a}) ->`, which on my branch compiles to

```js
(function(arg) {
  var a
  ({a} = arg);
});
```

...but could be compiled to

```js
(function({a}) {
});
```

Again, I suspect there's some small logic tweak we could make to the parameter handling to make them compile more things in-line.

zdenko:
I've merged @connec's branch into mine and made some progress with the rest element in object destructuring. I'll try to fnish a.s.a.p and push changes.

It seems to be working fine, and currently, this:

```javascript
obj = {a:1, b:2, c:3, d:4, e:5}
{a, b:z, d, r...} = obj
```

compiles to:

```javascript
var a, d, obj, z, r;

  obj = {
    a: 1,
    b: 2,
    c: 3,
    d: 4,
    e: 5
  };

  ({
    a,
    b: z,
    d
  } = obj), r =  Object.keys(obj).reduce((a,c) => (!['a','b','d'].includes(c) && (a[c] = obj[c]), a), {});
```

GeoffreyBooth:
Nice work! I feel like destructuring is finally coming along. 

One note: should your generated output be using `=>`? I know it's common style in ES, but a coffee lint rule that I tend to follow says not to use the bound/fat arrow unless you need it, i.e. unless you're using `this` inside the function.

connec:
@GeoffreyBooth I've updated [my branch] with compilation of default values. The implementation was fairly smooth. I don't seem to be able to push it here, however (no permissions?).

GeoffreyBooth:
@connec did you try to push it to GeoffreyBooth:coffeescript/destructuring? You're a collaborator there, you should be able to push. This PR is based on that branch on my repo.

connec:
Ah yes, it was my bad - I had your remote set as HTTPS but my authentication is SSH. I've updated the branch!

src/nodes.coffee
```diff
@@ -1104,47 +1115,77 @@ exports.Obj = class Obj extends Base
 
   children: ['properties']
 
+  isAssignable: ->
+    @lhs = true
```
connec: This is a bit hacky, but I wasn't sure how else to deal with `{a=1}` being valid only when destructuring. The other approach I considered was adding a "destructuring"/"assignment"/"lhs" option to `o`.

```coffee
class Assign
  ...
  compileNode: (o) ->
    o.assignment = true
    ...

class Obj
  ...
  compileNode: (o) ->
    ...
    prop.operatorToken.error "unexpected #{prop.operatorToken.value}" unless o.assignment
```

GeoffreyBooth: I'll look into this. Otherwise though, what else remains to be done?

connec: I think it's pretty close... I'd like to run it against a codebase or two and maybe add a few tests.

Would you prefer it if desrtuctured parameters went in this PR, or would you be happy for them to be tackled separately?

GeoffreyBooth: I don’t see why you think `lhs` is hacky; surely it’s better to isolate this object-specific property to the `Obj` class? Seems okay to me.

I think we should probably tackle destructured function parameters in this same PR, because basically we will need to be adding special cases to the same places you just modified in this PR to cover the things that you can normally destructure but not in a function parameter list, like `this`. I understand the appeal of merging this in and then using that as the new baseline, and only seeing the changes related to function parameters in a new PR, but since the function-parameters work is so related to this work I think it would be more useful to still see these changes. I don’t feel strongly though. There’s also https://github.com/geoffreybooth/coffeescript/pull/4 to merge in.

GeoffreyBooth:
Okay, I just cherry-picked everything useful from my alternate destructuring branch, leaving out the code related to my “expand `compilePatternMatch`” approach.

I also tried out every example from the [MDN destructuring assignment page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). The good news is that most of them are compiling as desired:

```js
// [a, b] = [1, 2]
var a, b;

[a, b] = [1, 2];


// [a, b, rest...] = [1, 2, 3, 4, 5]
var a, b, rest;

[a, b, ...rest] = [1, 2, 3, 4, 5];


// {a, b} = {a: 1, b: 2}
var a, b;

({a, b} = {
  a: 1,
  b: 2
});


// [one, two, three] = foo
var one, three, two;

[one, two, three] = foo;


// [a=5, b=7] = [1]
var a, b;

[a = 5, b = 7] = [1];


// [a, b] = [b, a]
var a, b;

[a, b] = [b, a];


// [a, b] = f()
var a, b;

[a, b] = f();


// [a, b...] = [1, 2, 3]
var a, b;

[a, ...b] = [1, 2, 3];


// {p, q} = o
var p, q;

({p, q} = o);


// ({a, b} = {a: 1, b: 2})
var a, b;

({a, b} = {
  a: 1,
  b: 2
});


// {p: foo, q: bar} = o
var bar, foo;

({
  p: foo,
  q: bar
} = o);


// {a = 10, b = 5} = {a: 3}
var a, b;

({a = 10, b = 5} = {
  a: 3
});


// {title: englishTitle, translations: [{title: localeTitle}]} = metadata
var englishTitle, localeTitle;

({
  title: englishTitle,
  translations: [
    {
      title: localeTitle
    }
  ]
} = metadata);
```

Destructured function parameters are getting output inside the function body, but _are_ using ES destructuring syntax once there:

```js
// ({a}) -> a
(function(arg) {
  var a;
  ({a} = arg);
  return a;
});


// ({a = 1}) -> a
(function(arg) {
  var a;
  ({a = 1} = arg);
  return a;
});


// userId = ({id}) -> id
var userId;

userId = function(arg) {
  var id;
  ({id} = arg);
  return id;
};


// whois = ({displayName, fullName: {firstName: name}}) -> name
var whois;

whois = function(arg) {
  var displayName, name;
  ({
    displayName,
    fullName: {
      firstName: name
    }
  } = arg);
  return name;
};
```

So is destructuring in a `for` loop, something we haven’t even discussed yet:

```js
// fn() for {name: n, family: {father: f}} from people
var f, n, x;

for (x of people) {
  ({
    name: n,
    family: {
      father: f
    }
  } = x);
  fn();
}
```

Now for the failures. The first one is expected, the destructured object splat. It’s labeled experimental by MDN, and supporting it is the goal of https://github.com/GeoffreyBooth/coffeescript/pull/4:

```coffee
# {a, b, rest...} = {a: 1, b: 2, c: 3, d: 4}
[stdin]:1:12: error: unexpected ...
{a, b, rest...} = {a: 1, b: 2, c: 3, d: 4}
           ^^^
```

One thing that ES destructuring supports that CoffeeScript doesn’t is skipped elements:

```coffee
# [a, , b] = f()
[stdin]:1:5: error: unexpected ,
[a, , b] = f()
    ^

# [,,] = f()
[stdin]:1:2: error: unexpected ,
[,,] = f()
 ^

# [, protocol, fullhost, fullpath] = parsedURL
[stdin]:1:2: error: unexpected ,
[, protocol, fullhost, fullpath] = parsedURL
 ^
```

This example throws an error because of the `= {}`:

```coffee
# drawES2015Chart = ({size = 'big', coords = {x: 0, y: 0}, radius = 25} = {}) ->
[stdin]:1:26: error: unexpected =
drawES2015Chart = ({size = 'big', coords = {x: 0, y: 0}, radius = 25} = {}) ->
                         ^
```
```js
// drawES2015Chart = ({size = 'big', coords = {x: 0, y: 0}, radius = 25}) ->
var drawES2015Chart;

drawES2015Chart = function(arg) {
  var coords, radius, size;
  ({size = 'big', coords = {
      x: 0,
      y: 0
    }, radius = 25} = arg);
};
```

And finally, we don’t support dynamic keys in destructuring:

```coffee
# {[key]: foo} = {z: 'bar'}
[stdin]:1:2: error: unexpected [
{[key]: foo} = {z: bar}
 ^
```

src/nodes.coffee
GeoffreyBooth: Wait a minute . . . is `@lhs` hacky because it’s getting _set_ via the `isAssignable` call? And you’re relying on the fact that `isAssignable` is only called when `Obj` is part of an `Assign` and therefore the left hand side of a destructuring assignment?

If that’s the case, then yeah, this is really hacky. `isAssignable` is not the name for a method that should be mutating its class. We should find a better way.

GeoffreyBooth:
Okay, I’ve gotten destructured function parameters working (at least all the cases I’ve tested):

```js
// ({a}) -> a
(function({a}) {
  return a;
});


// ({a = 1}) -> a
(function({a = 1}) {
  return a;
});


// userId = ({id}) -> id
var userId;

userId = function({id}) {
  return id;
};


// whois = ({displayName, fullName: {firstName: name}}) -> name
var whois;

whois = function({
    displayName,
    fullName: {
      firstName: name
    }
  }) {
  return name;
};


// drawES2015Chart = ({size = 'big', coords = {x: 0, y: 0}, radius = 25} = {}) ->
var drawES2015Chart;

drawES2015Chart = function({size = 'big', coords = {
      x: 0,
      y: 0
    }, radius = 25} = {}) {};
```

This needs more testing, especially functions with non-final splat or expansion parameters, since such functions push the destructuring into the function body. Though as of now all the tests pass, so hopefully everything is good.

Based on my list of examples above, the following is a potential to-do list:
- Destructured object splats; https://github.com/GeoffreyBooth/coffeescript/pull/4
- Destructuring in a `for` (or other?) loop’s initialization (the part in parentheses in JavaScript)
- Skipped values
- Destructuring with dynamic keys

I think we should finish https://github.com/GeoffreyBooth/coffeescript/pull/4, certainly, especially since it’s so close; but the other three could be goals for the future? None of them would be breaking changes if we add them post-2.0.0.

One final potential task is going through `compileDestructuring` and removing all the dead code. There are still lots of cases that fall back on that old 1.x destructuring implementation, so we can’t remove the function entirely; but surely some of its code paths are no longer used now that we handle so many destructuring cases in `Obj` and `Arr`.

src/nodes.coffee
```diff
@@ -1947,13 +2064,18 @@ exports.Code = class Code extends Base
 
         haveSplatParam = yes
         if param.splat
-          params.push ref = param.asReference o
-          splatParamName = fragmentsToText ref.compileNode o
+          if param.name instanceof Arr
```
GeoffreyBooth: Remove these lines, and the test "#4005: `([a = {}]..., b) ->` weirdness" fails. I can’t figure out why.

src/nodes.coffee
connec: Yes, precisely. Unfortunately it's the only reliable way I could think to do it without:

- Traversing the tree in `Assign` (e.g. `@variable.traverseChildren`). This is a bit awkward and probably bad for performance.
- Putting a property on `o` (e.g. `o.assignment = true`). This would be pretty seamless, but the `o` API can be hard to understand. This is probably the best alternative though.

src/nodes.coffee
```diff
@@ -1816,7 +1816,8 @@ exports.Assign = class Assign extends Base
     expandedIdx = false
     # Make vvar into a simple variable if it isn't already.
     if value.unwrap() not instanceof IdentifierLiteral or @variable.assigns(vvarText)
-      assigns.push [@makeCode("#{ ref = o.scope.freeVariable 'ref' } = "), vvar...]
+      ref = o.scope.freeVariable 'ref'
+      assigns.push [@makeCode(ref + ' = '), vvar...]
```
connec: Nit: `"#{ref}="`

src/nodes.coffee
```diff
@@ -2509,7 +2509,8 @@ exports.Op = class Op extends Base
     return @compileUnary        o if @isUnary()
     return @compileChain        o if isChain
     switch @operator
-      when '?'  then @compileExistence o
+      when '?'  then @compileExistence o      # Check if not null and not undefined
+      when '??' then @compileExistence o, yes # Check only if not undefined
```
connec: Surely this syntax addition merits its own discussion?

src/nodes.coffee
```diff
@@ -2715,10 +2717,19 @@ exports.Existence = class Existence extends Base
     code = @expression.compile o, LEVEL_OP
     if @expression.unwrap() instanceof IdentifierLiteral and not o.scope.check code
       [cmp, cnj] = if @negated then ['===', '||'] else ['!==', '&&']
-      code = "typeof #{code} #{cmp} \"undefined\" #{cnj} #{code} #{cmp} null"
+      code = "typeof #{code} #{cmp} \"undefined\" #{cnj} #{code} #{cmp} #{@comparisonTarget}"
```
connec: The second half of this expression (`#{code} #{cmp} #{@comparisonTarget}`) is redundant if `@comparisonTarget is 'undefined'` -

```js
typeof foo === 'undefined' || foo === undefined
typeof foo !== 'undefined' && foo !== undefined
```

test/functions.coffee
```diff
@@ -150,13 +150,13 @@ test "@-parameters and splats with constructors", ->
   eq b, obj.last
 
 test "destructuring in function definition", ->
-  (([{a: [b], c}]...) ->
+  (({a: [b], c}) ->
```
connec: Why did we need to remove these?

src/nodes.coffee
```diff
@@ -1121,10 +1120,17 @@ exports.Obj = class Obj extends Base
     @lhs = yes
 
     for prop in @properties
+      # Check for reserved words.
+      message = isUnassignable prop.unwrapAll().value
```
connec: I think this is already being checked in `Assign` as part of the `eachName` callback (here](https://github.com/GeoffreyBooth/coffeescript/blob/destructuring/src/nodes.coffee#L1752)?

src/nodes.coffee
```diff
@@ -1186,7 +1192,7 @@ exports.Obj = class Obj extends Base
     for prop in @properties
       prop = prop.value if prop instanceof Assign and prop.context is 'object'
       prop = prop.unwrapAll()
-      prop.eachName iterator
+      prop.eachName iterator if prop.eachName?
```
connec: I'd be inclined to leave the existence check off - the reason being that anything that's `isAssignable` should support it, and we've already checked assignability at this point.

src/nodes.coffee
```diff
@@ -2089,11 +2105,16 @@ exports.Code = class Code extends Base
...
-              ref = new Assign new Value(param.name), param.value
+              ref = new Assign new Value(param.name), param.value, null, param: yes
             else
               ref = param
-          # Add this parameter’s reference to the function scope
-          o.scope.parameter fragmentsToText (if param.value? then param else ref).compileToFragments o
+          # Add this parameter’s reference(s) to the function scope.
+          if param.name instanceof Arr or param.name instanceof Obj
+            # This parameter is destructured.
+            param.name.eachName (prop) ->
```
connec: I reckon this should check `isAssignable`, since otherwise names may appear twice if not? Not really sure though...

src/nodes.coffee
```diff
@@ -2089,11 +2105,16 @@ exports.Code = class Code extends Base
...
+              o.scope.parameter fragmentsToText prop.compileToFragments o
```
connec: This could use the simpler logic from `Assign` (e.g. `o.scope.parameter prop.value`).

src/nodes.coffee
GeoffreyBooth: Hmm, I had thought this was an internal-only thing, like the grammar tokens like `yieldfrom`; I didn’t think I was creating a new operator in the language. Though I guess now that I inadvertently _have_ created a new operator, I guess we should discuss it 😄

I’m sure there’s some other way to pass down into `class Existence` that I want a ternary that only checks against `undefined`, not `undefined` or `null` (the legacy behavior). Or . . . should we keep this? On the one hand it definitely clutters the language to have two types of checking for existence, but on the other hand it _is_ the ES spec that sometimes things “exist” only when they’re not `undefined`, so maybe it would be useful to have this “ES existence” operator. What do you think?

src/nodes.coffee
GeoffreyBooth: Good catch. 770ddcd0a409ec46177c363aa56f528b2339ba25

test/functions.coffee
GeoffreyBooth: So there was a period when I was struggling to get splatted arrays to work as function parameters, and so I modified these tests. Later on I added an [exception for this case](https://github.com/jashkenas/coffeescript/pull/4478/files#diff-d149b4f007169f06a8ad463f9b2f277bR2067) that just dumped such cases back into the legacy compilation, so now we could revert these test changes. But I think we shouldn’t, because the splat-array wrapper feels like a workaround. The version without the wrapper looks to me much more like something someone would intentionally write.

src/nodes.coffee
GeoffreyBooth: This took me perhaps the most time to figure out. It _is_ being checked in `Assign`, but `Assign` never gets invoked for function parameter destructuring like `({a}) ->`.

src/nodes.coffee
GeoffreyBooth: Without it, this test fails:
```
  #4130: unassignable in destructured param
  AssertionError: Expected [stdin]:2:12: error: keyword ‘null’ can’t be assigned
  @param : null
           ^^^^ to equal TypeError: prop.eachName is not a function
```

I guess there are checks for assignability that happen _after_ `eachName` gets run. So if we want to lose this existence check, we’ll have to make sure that validation happens before this.

src/nodes.coffee
GeoffreyBooth: Yes. 678e93146b594ef271c455cf57655f26247922d5

src/nodes.coffee
GeoffreyBooth: ? Why would we not have spaces around `=`? That’s not our style.

src/nodes.coffee
GeoffreyBooth: I’ve refactored this to use the new only-not-undefined existence check if the right-hand value has an `isDefaultValue` property set. This meant adding a new property to `Value`, but no new operator. 60f046e313d0474feb5afa9b0453723c991eb793

src/nodes.coffee
GeoffreyBooth: So I tackled this and ended up with 05c1f46a9237d6b0eb83c0cd8917f2baa8603ded. I traced how `lhs` was getting set in your version (i.e., where was the `isAssignable` call coming from) and just set the property from the parent at that point. The end result came in rather complicated, though, because of the cases of objects within arrays, and of destructuring that originates from function parameters and not `Assign`. I think the performance should be the same as your version, at least.

I’m sure you could find an even more elegant way. I would at least prefer this over an `isAssignable` method that mutates; but feel free to try to do better still 😄

GeoffreyBooth:
@connec Any other notes? I went through `compileDestructuring` and actually couldn’t find any dead code to remove. So with `isAssignable` fixed (unless you’d like to improve `lhs` handling further) and my inadvertent new operator removed, hopefully this PR is potentially ready to merge? @lydell, do you mind reviewing?

GeoffreyBooth:
Fixes #4477

test/assignment.coffee
```diff
@@ -141,9 +141,6 @@ test "#1192: assignment starting with object literals", ->
 
 # Destructuring Assignment
 
-test "empty destructuring assignment", ->
-  {} = [] = undefined
-
```
lydell: Why was this removed?

test/functions.coffee
lydell: I don't think this is acceptable. There are lots of things in programming languages that people never write, but still need to be supported for consistency.

test/functions.coffee
```diff
@@ -150,13 +150,13 @@ test "@-parameters and splats with constructors", ->
...
     eq 1, b
     eq 2, c
   ) {a: [1], c: 2}
 
   context = {}
-  (([{a: [b, c = 2], @d, e = 4}]...) ->
+  (({a: [b, c = 2], @d, e = 4}) ->
```
lydell: also here

GeoffreyBooth: These tests pass with or without this change. I just thought the revised version was a more realistic test. We could test both versions if you want, but the `[]...` syntax is still tested in the `` #4005: `([a = {}]..., b) ->` weirdness `` test, so I’m not sure we need to test it here too.

test/assignment.coffee
GeoffreyBooth: @connec removed that, but I suspect because it’s [invalid ES](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=&targets=&browsers=&builtIns=false&code=%7B%7D%20%3D%20%5B%5D%20%3D%20undefined).

lydell: It’s not. You just forgot the parentheses. https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&experimental=false&loose=false&spec=false&playground=false&code=(%7B%7D%20%3D%20%5B%5D%20%3D%20undefined)

GeoffreyBooth: Well if I restore the test, it outputs as:
```js
({} = void 0);
```

But Node doesn’t like it, or the less-mutated version:

```
node -e '({} = void 0)'
[eval]:1
({} = void 0)
    ^

TypeError: Cannot match against 'undefined' or 'null'.
    at [eval]:1:5
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at Object.runInThisContext (vm.js:95:38)
    at Module._compile (module.js:571:32)
    at evalScript (bootstrap_node.js:387:27)
    at run (bootstrap_node.js:120:11)
    at run (bootstrap_node.js:423:7)
    at startup (bootstrap_node.js:119:9)
    at bootstrap_node.js:538:3


node -e '({} = [] = undefined)'
[eval]:1
({} = [] = undefined)
           ^

TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
    at [eval]:1:12
    at ContextifyScript.Script.runInThisContext (vm.js:23:33)
    at Object.runInThisContext (vm.js:95:38)
    at Module._compile (module.js:571:32)
    at evalScript (bootstrap_node.js:387:27)
    at run (bootstrap_node.js:120:11)
    at run (bootstrap_node.js:423:7)
    at startup (bootstrap_node.js:119:9)
    at bootstrap_node.js:538:3
```

lydell: That’s just a runtime error, just like this:

```
$ node -e 'foo()'
[eval]:1
foo()
^

ReferenceError: foo is not defined
    at [eval]:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:26:33)
    at Object.exports.runInThisContext (vm.js:79:17)
    at Module._compile (module.js:571:32)
    at Immediate.<anonymous> (bootstrap_node.js:383:29)
    at runCallback (timers.js:649:20)
    at tryOnImmediate (timers.js:622:5)
    at processImmediate [as _immediateCallback] (timers.js:594:5)
```

lydell: I guess we could change the test to:

```coffee

[] = []
{} = {}
```

lydell: Or keep the CoffeeScript 1 compilation:

```
$ coffee -bpe '{} = [] = undefined'
void 0;
```

But that seems not very useful.

vendethiel: CS1 uses those for non-destructuring cases, though
```
coffee -bce> [v, []] = a
var v;

v = a[0], a[1];
```

Although it's... not very consistent...
```
coffee -bce> [[]] = a
a[0];
```

src/nodes.coffee
connec: Sorry, I forgot the space. I just meant to use interpolation 😅

src/nodes.coffee
connec: Yeh... Again I still think we're making a mistake by moving to ES' default semantics, and should have stuck with `null`/`undefined` being treated the same in CS. So I'm probably not the best person to ask 😛

src/nodes.coffee
connec: Ooh, interesting 😞

src/nodes.coffee
connec: I expect this is to do with your above point that `Param`s aren't compiled as `Assign`s by default.

src/nodes.coffee
connec: You're right, I forgot `Param` has its own `eachParamName` which is being used [here](https://github.com/jashkenas/coffeescript/pull/4478/files/1e6d910e814275e0c6bd8d781e9d216f9501aa8b..6c814a09780c5c03c7134a7eff442c0512572090#diff-d149b4f007169f06a8ad463f9b2f277bR2035) to check uniqueness.

Some of that iteration could maybe be be centralised (probably later).

test/assignment.coffee
connec: Hrm, I did remove the test because destructuring `undefined` or `null` is a runtime error. I think we should be able to restore it with an empty array/object and it should still work... I think.

test/functions.coffee
GeoffreyBooth: So @lydell, do you want me to revert these tests? They pass either way.

test/assignment.coffee
GeoffreyBooth: So the test currently fails, because it compiles to:
```js
function () {
  return ({} = void 0);
}
```
which Node doesn’t like:
```
TypeError: Cannot match against 'undefined' or 'null'.
```
Adding `= []` in there generates the other error I saw before:
```
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined
```
So I guess the question is, so what? There’s no way to make this test “pass,” since the reasonable JavaScript that a user would expect to get generated (`{} = [] = undefined`) also throws a runtime error. Having the compiler generate anything else feels like cheating.

So either we try to have the compiler catch this error, and the test becomes whether we do catch it; or we rewrite the test to test something else that wouldn’t be a runtime error in Node. I would lean toward the latter.

test/functions.coffee
lydell: They've been like this since 2010. Let's keep them that way.

GeoffreyBooth: Okay.

test/assignment.coffee
GeoffreyBooth: @lydell for now I changed the test to your suggestion:
```coffee
[] = []
{} = {}
```
and it passes. This is probably a worthwhile test on its own. Are there any other tests you’d like to add?

GeoffreyBooth:
@lydell are you okay with these revisions?

I’d like to get this merged in and tackle https://github.com/GeoffreyBooth/coffeescript/pull/4 on its own.

lydell:
Could you post two examples where we fall back to the old code generation, and what they compile to? One for objects and one for arrays. Just want to think how they differ, and if they are a potential gotcha. Otherwise LGTM.

GeoffreyBooth:
Sure. As far as I can tell the old code is only being used for splats and expansions, and for function parameters after splats or expansions:
```js
// [x,y...,z] = [a,b,c,d,e]
var i, ref, x, y, z,
  slice = [].slice;

ref = [a, b, c, d, e], x = ref[0], y = 3 <= ref.length ? slice.call(ref, 1, i = ref.length - 1) : (i = 1, []), z = ref[i++];


// f = (a, b..., {c}) ->
var f,
  slice = [].slice;

f = function(a, ...b) {
  var c, i, ref;
  ref = b, b = 2 <= ref.length ? slice.call(ref, 0, i = ref.length - 1) : (i = 0, []), ({c} = ref[i++]);
};
```

Until https://github.com/GeoffreyBooth/coffeescript/pull/4 lands there are no splats or expansions within objects, so I guess we’re not using the old code at all for any object destructuring.

lydell:
I guess the gotcha is destructuring generators, then?

```
$ node
> g = function *(){yield 1; yield 2; yield 2}
[Function: g]
> [a, ...r]=g()
{}
> a
1
> r
[ 2, 2 ]
```

I'm thinking that `[a, r...] = g()` would pull values out of the generator, while `[a, r..., b] = g()` wouldn't? If so, I'm not sure what to do about it other than docs?

GeoffreyBooth:
That isn’t really a fault of this PR, as there’s no destructuring there; but it’s something we should consider. Basically “non-final” function parameters fall back to the legacy compilation, which involves caching the result of a function call, e.g.:

```js
// [a, r…, b] = g()
var a, b, i, r, ref,
  slice = [].slice;

ref = g(), a = ref[0], r = 3 <= ref.length ? slice.call(ref, 1, i = ref.length - 1) : (i = 1, []), b = ref[i++];
```

Because of the `ref = g()`, `g` only gets called once here. In your example, I guess `…r` calls `g` until `g` is exhausted? What if `g` could generate infinitely?

I guess the “solution” would be to once again restrict ourselves to ES semantics, and only allow splats in the final parameter. But I don’t want to do that for such a fringe edge case as this . . . I guess a note in the docs, yeah.

Anyway, leaving function parameters aside, how’s the destructuring? 😄

lydell:
> In your example, I guess `...r` calls `g` until `g` is exhausted? What if `g` could generate infinitely?

`g` is called exactly once. `g()` _returns_ a generator. `...r` does take the next value of until the generator is exhausted. If the generator is infinite, you simply end up in an infinite loop. Note that other than using array destructuring, you can also use the `.next()` method on generators to pull values out of them.

I'm not really sure I understood the rest of your comment, but I think we're good to go on this PR.

GeoffreyBooth:
> I’m not really sure I understood the rest of your comment

I just mean that splats as the final parameter compile nicely into ES:
```js
// (a, b, c...) ->
(function(a, b, ...c) {});
```
But splats in any non-final position aren’t supported in ES, and so we need to do some shenanigans:
```js
// (a, b..., c) ->
var slice = [].slice;

(function(a, ...b) {
  var c, i, ref;
  ref = b, b = 2 <= ref.length ? slice.call(ref, 0, i = ref.length - 1) : (i = 0, []), c = ref[i++];
});
```
And these shenanigans aren’t equivalent to ES in the case of handling generators, as you’ve pointed out. I’m not sure there’s a way _to_ make them equivalent without a runtime check that the assignment target is a generator function as opposed to a regular function. This maybe deserves an issue of its own, even if we decide it’s too fringe of an edge case to address.

jashkenas:
You guys are really cooking on this — bravo!

connec:
> And these shenanigans aren’t equivalent to ES in the case of handling generators

Will that not actually work fine? Generator unpacking only happens at the call-site anyway, so these all work identically to ES2015 equivalents:

```coffee
g = ->
  yield i for i in [1..3]
  return

f = (a, b..., c) ->
  return [a, b..., c]

console.log f g()
# [g, undefined]

console.log f g()...
# [1, 2, 3]

console.log f 0, g()..., 4
# [0, 1, 2, 3, 4]
```

helixbass:
@GeoffreyBooth I just ran into something that broke (I assume unintentionally) as a result of this pull request: eg `({a: {b = 1}}) ->` compiled ok before this PR but gives `error: unexpected =` after. This destructuring also breaks in a destructured assignment: `{a: {b = 1}} = c` also gives `error: unexpected =`

PR #4490, opened 2017-04-06
Fix #4489: Regex octal escape sequence bug

### LINKED ISSUE #4489: Regex octal escape sequence bug
While reading `lexer.coffee` spotted a hole in the `validateEscapes()` logic: if it encounters a valid "`INVALID_ESCAPE`" (eg `\1`) inside a regex interpolation-free section then it fails to check the rest of the section for actually invalid escapes. So eg `/\1\01/` should fail but instead generates `/\1\x001/`

Will submit a fixing pull request momentarily

helixbass:
Fixes #4489

Splits `INVALID_ESCAPE` regex into separate regexes `REGEX_INVALID_ESCAPE` (which doesn't match back-references eg `\1`) and `STRING_INVALID_ESCAPE`, thus avoiding the special-case logic for regex back-references (which contained the bug)


lydell:
Nice catch, and nice fix! 👍

PR #4492, opened 2017-04-08
Fix #4491: import- and export-specific lexing should stop

### LINKED ISSUE #4491: import- and export-specific lexing should stop after import/export statement
In `lexer.coffee`, `@seenImport` and `@seenExport` only get reset by an explicit `;`. Thus allowing some devious edge cases to fail that should work, eg this works:
```
import * as lib from 'lib';
foo as       # foo(as)
3 * as 4     # 3 * as(4)
from 'foo'   # from('foo')
```
But if you remove the `;`, any of the following lines would break

Will submit what I believe to be a fixing pull request momentarily

helixbass:
Fixes #4491 (I think)

I'm not really familiar with the `import`/`export` syntax but it looks like the only the time that the import-/export-specific lexing controlled by `@seenImport`/`@seenExport` (which lexes `as`, `from`, `default` differently) needs to continue beyond a line break is if we're inside a `{...}` specifier list. So I added an `@importSpecifierList` (corresponding to the existing `@exportSpecifierList`), and unless one of these is set (indicating we're inside a specifier list) then seeing a `lineToken()` clears `@seenImport` and `@seenExport`

Not sure if I should add additional tests for any of the other various `import`/`export` syntaxes?

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -46,6 +46,7 @@ exports.Lexer = class Lexer
     @seenImport = no             # Used to recognize IMPORT FROM? AS? tokens.
     @seenExport = no             # Used to recognize EXPORT FROM? AS? tokens.
     @exportSpecifierList = no    # Used to identify when in an EXPORT {...} FROM? ...
+    @importSpecifierList = no    # Used to identify when in an IMPORT {...} FROM? ...
 
     @chunkLine =
       opts.line or 0             # The start line for the current @chunk.
@@ -365,6 +366,8 @@ exports.Lexer = class Lexer
     indent = match[0]
 
     @seenFor = no
+    @seenImport = no unless @importSpecifierList
+    @seenExport = no unless @exportSpecifierList
 
     size = indent.length - 1 - indent.lastIndexOf '\n'
     noNewlines = @unfinished()
@@ -477,6 +480,10 @@ exports.Lexer = class Lexer
       @exportSpecifierList = yes
     else if @exportSpecifierList and value is '}'
       @exportSpecifierList = no
+    else if value is '{' and @seenImport
+      @importSpecifierList = yes
+    else if @importSpecifierList and value is '}'
+      @importSpecifierList = no
 
     if value is ';'
       @seenFor = @seenImport = @seenExport = no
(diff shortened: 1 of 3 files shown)


src/lexer.coffee
```diff
@@ -46,6 +46,7 @@ exports.Lexer = class Lexer
     @seenImport = no             # Used to recognize IMPORT FROM? AS? tokens.
     @seenExport = no             # Used to recognize EXPORT FROM? AS? tokens.
     @exportSpecifierList = no    # Used to identify when in an EXPORT {...} FROM? ...
+    @importSpecifierList = no    # Used to identify when in an IMPORT {...} FROM? ...
```
GeoffreyBooth: Please keep it consistent and always put the import stuff before the export stuff.

src/lexer.coffee
```diff
@@ -477,6 +480,10 @@ exports.Lexer = class Lexer
       @exportSpecifierList = yes
     else if @exportSpecifierList and value is '}'
       @exportSpecifierList = no
+    else if value is '{' and @seenImport
```
GeoffreyBooth: Same here.

helixbass:
@GeoffreyBooth ok put imports first

Also updated previously merged (via #4490) fix for #4489 to follow style guide and use camelCase

GeoffreyBooth:
Thanks. Looks good to me, @lydell?

src/lexer.coffee
```diff
@@ -473,7 +476,11 @@ exports.Lexer = class Lexer
         @error message, origin[2] if message
       return value.length if skipToken
 
-    if value is '{' and prev?[0] is 'EXPORT'
+    if value is '{' and @seenImport
+      @importSpecifierList = yes
+    else if @importSpecifierList and value is '}'
```
vendethiel: this doesn't handle nested objects, right?

helixbass: @vendethiel right, I didn't see any cases in the documentation or tests where an `import` could contain nested-object syntax?

GeoffreyBooth: @vendethiel the “destructuring” in ES2015 `import` statements is just a [similar-looking syntax](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&code=import%20%7B%20a%3A%20b%20%7D%20from%20'lib'):
```js
import { a: b } from 'lib'
```
```
repl: ES2015 named imports do not destructure. Use another statement for destructuring after the import. (1:12)
> 1 | import { a: b } from 'lib'
    |             ^
```

vendethiel: Oh, thanks, TIL!

PR #4493, opened 2017-04-09
[WIP][CS2] Destructuring object spreads

GeoffreyBooth:
Replacing https://github.com/jashkenas/coffeescript/pull/4473 and https://github.com/GeoffreyBooth/coffeescript/pull/4. From original descriptions by @zdenko:

Connected to [#3894](https://github.com/jashkenas/coffeescript/issues/3894).

In CS2 emit

```coffee
obj = { x: 1, y: 2}
obj1 = {obj..., z: 3}
```
to

```coffee
obj1 = {...obj, z: 3}
```

Improve destructuring assignment for object literal:

```coffee
{a, b, x...} = {a:1, b:2, c:3, d:4, e:5}
a = 1
b = 2
x = {c:3, d:4, e:5}
```

```coffee
{c, x..., e} = {a:1, b:2, c:3, d:4, e:5}
c = 3
e = 5
x = {a:1, b:2, d:4}
```

```cs
{c, x..., a:p} = {a:1, b:2, c:3, d:4, e:5}
c = 3
p = 1
x = {b:2, d:4, e:5}
```

```coffee
futurists =
  sculptor: "Umberto Boccioni"
  painter:  "Vladimir Burliuk"
  poet:
    name:   "F.T. Marinetti"
    address: [
      "Via Roma 42R"
      "Bellagio, Italy 22021"
    ]

{poet: {name, addr1...}} = futurists
addr1 = {address: [ "Via Roma 42R",  "Bellagio, Italy 22021"]}

{poet: {addr2..., name:title}} = futurists
addr2 = {address: [ "Via Roma 42R",  "Bellagio, Italy 22021"]}
```

Basically, implement proposal for [rest properties for ECMAScript](https://github.com/tc39/proposal-object-rest-spread) in CS: `{a, b, c...} = x`

Since this proposal isn't yet at stage-4, CS doesn't compile to ES.
In my PR I catch rest element and assign remaining values to it:
`{a, b, c...} = x` compiles to:
`a = x.a, b = x.b, c = extractObjectWithoutKeys(x, ['a', 'b'])`

Multiple rest elements are disallowed. ES also requires the rest element to be the last, so compiler currently throws an error.

IMHO, CS should allow rest element anywhere, just like for arrays.

Later, when proposal reaches stage-4 we could implement similar conversion as you did for the function parameters, and ensure rest element is the last in the compiled JS.


GeoffreyBooth:
@zdenko Can you please merge `2` into your branch and resolve conflicts? I took a crack at it, but there are a lot of conflicts to resolve. Basically the `destructuring` branch in #4478 overlapped with much of what you were doing. You branch started from my original destructuring approach, which wasn’t nearly as clean or robust as @connec’s, so for #4478 we abandoned mine in favor of his; and his got merged into `2`. So now you need to work backwards a bit and start from this new starting point, where object destructuring is already implemented in `2`, and you only need to add the object spread destructuring. The `2` branch also already has destructured objects in function parameters.

danielbayley:
> Multiple rest elements are disallowed.

What about when merging objects like this?

~~~ js
let obj = { a:1, b:2, c:3, d:4 }
let override = { b: 'b' , d: 'd'}
const merged = {...obj, ...override}
~~~

Which [works] with the `stage-3` preset in babel…

[works]: https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-3&targets=&browsers=&builtIns=false&code=%0Alet%20obj%20%3D%20%7B%20a%3A1%2C%20b%3A2%2C%20c%3A3%2C%20d%3A4%20%7D%0A%0Alet%20override%20%3D%20%7B%20b%3A%20'b'%20%2C%20d%3A%20'd'%7D%0A%0Aconst%20merged%20%3D%20%7B...obj%2C%20...override%7D%0A%0Aconsole.log(merged)&experimental=true&loose=false&spec=false&playground=true&stage=0

zdenko:
@GeoffreyBooth I've merged `2` into the [destructuring_object branch](https://github.com/GeoffreyBooth/coffeescript/tree/destructuring_object).
With the last commit, I also covered the [object spread](https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md).

```javascript
// {a, b:z, c, r...} = {a:1, b:2, c:3, d:4, e:5}
// obj1 = {f:1, g:2, h:3}
// {f, q...} = obj1
var obj1, q, r, ref;

ref = {a: 1, b: 2, c: 3, d: 4, e: 5}, ({a, b: z, c} = ref), 
r = Object.keys(ref).filter((k) => ['a','b','c'].indexOf(k) == -1).reduce((a, c) => (a[c] = ref[c], a), {});

obj1 = {f: 1, g: 2, h: 3};
({f} = obj1), 
q = Object.keys(obj1).filter((k) => ['f'].indexOf(k) == -1).reduce((a, c) => (a[c] = obj1[c], a), {});


/* 
obj = {
    a: a1
    b: {
      c: {
        d: {
          b1
          e: c1
          f: d1
        }
      }
    }
    b2: {b1, c1}
}
{a:w, b:{c:{d:{b1:bb, r1...}}}, r2...} = obj
*/
var bb, obj, r1, r2, w;

obj = {
  a: a1,
  b: {
    c: {
      d: {
        b1,
        e: c1,
        f: d1
      }
    }
  },
  b2: {b1, c1}
};

({
  a: w,
  b: {
    c: {
      d: {
        b1: bb
      }
    }
  }
} = obj), r1 = Object.keys(obj['b']['c']['d']).reduce(function(a,c) { return !['b1'].includes(c) && (a[c] = obj['b']['c']['d'][c]), a; }, {}), r2 = Object.keys(obj).reduce(function(a,c) { return !['a','b'].includes(c) && (a[c] = obj[c]), a; }, {});


// fn = ({a, b, r...}) -> {}
var fn;

fn = function(arg) {
  var a, arg, b, r;
  ({a, b} = arg), r = Object.keys(arg).reduce(function(a,c) { return !['a','b'].includes(c) && (a[c] = arg[c]), a; }, {});
  return {};
};


// aWithOverrides = {a..., x: 1, y: 2 }
var aWithOverrides;
aWithOverrides = Object.assign({}, a, {x: 1, y: 2});

// xyWithAandB = { x: 1, a..., y: 2, b..., a... }
var xyWithAandB;
xyWithAandB = Object.assign({}, {x: 1}, a, {y: 2}, b, a);

```

GeoffreyBooth:
@zdenko does this complete your work? As in, have you covered all the cases of object spread destructuring that you plan to cover?

@connec, do you mind reviewing?

zdenko:
@GeoffreyBooth yes, as far as I'm concerned.

connec:
@GeoffreyBooth @zdenko I'll have lots of time tomorrow (Public Holidays 😄), so I'll take a look at it then!

GeoffreyBooth:
@connec how does it look?

zdenko:
@GeoffreyBooth any news on this PR?

GeoffreyBooth:
@zdenko I think @connec is reviewing it.

connec:
Yes, I plan to review it this weekend!

zdenko:
@connec anything yet?

GeoffreyBooth:
@zdenko please merge 2 into this branch and resolve conflicts.

GeoffreyBooth:
@connec I’ve merged in `2` and resolved conflicts.

test/assignment.coffee
```diff
@@ -234,7 +234,71 @@ test "destructuring assignment against an expression", ->
   [y, z] = if true then [a, b] else [b, a]
   eq a, y
   eq b, z
-
+  
+test "destructuring assignment with objects and splats: ES2015", ->
+  obj = {a:1, b:2, c:3, d:4, e:5}
+  throws (-> CoffeeScript.compile "{a, r..., s....} = x"), null, "multiple rest elements are disallowed"
```
lydell: Are there _four_ dots in `s....`?

brunovandamme: This is not compiling for me 
```
a = { b: { c: yes }}
spread = { a.b... }
```
While the following does compile
```
a = { b: { c: yes }}
spread = { a... }
```

GeoffreyBooth: `spread = { a.b... }` [works in Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=env%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&targets=&browsers=&builtIns=false&debug=false&code=var%20a%2C%20yes%2C%20spread%3B%0Aa%20%3D%20%7B%20b%3A%20%7B%20c%3A%20yes%20%7D%7D%3B%0Aspread%20%3D%20%7B%20…a.b%20%7D%3B).

brunovandamme: It works in Babel, but I meant that it did not compile with the [destructuring_object branch](https://github.com/GeoffreyBooth/coffeescript/tree/destructuring_object) 
Apologies for not making myself clear the first time.

zdenko:
@GeoffreyBooth I'm checking this error.

test/assignment.coffee
GeoffreyBooth: I meant, it works in Babel so presumably it’s correct syntax. And therefore we should support the CoffeeScript equivalent (`{ a.b... }` instead of `{ ...a.b }`).

src/nodes.coffee
```diff
@@ -1158,24 +1165,45 @@ exports.Obj = class Obj extends Base
         prop.variable
       else if prop not instanceof Comment
         prop
-
-      if key instanceof Value and key.hasProperties()
+      # Pass the splat through.
+      if key instanceof Value and key.hasProperties() and key not instanceof Splat
```
zdenko: Yep, I agree. I think I added this before your PR was merged when I tried to fix some errors.

src/nodes.coffee
```diff
@@ -1158,24 +1165,45 @@ exports.Obj = class Obj extends Base
...
       if join then answer.push @makeCode join
-    answer.push @makeCode "#{if isCompact then '' else "\n#{@tab}"}}"
+    if hasSpread
+      if spreads.length
+        spreadAnswer.push @makeCode(", "), openBracket
+        for s in spreads
+          spreadAnswer.push s.compileToFragments(o, LEVEL_TOP)..., @makeCode(", ")
+        spreadAnswer.pop()
+        spreadAnswer.push closeBracket
+      answer = [@makeCode("Object.assign({}, "), spreadAnswer..., @makeCode(")")]
```
zdenko: I'll look into it. It might even help to solve the `a.b` and `a[b]` splats.

src/nodes.coffee
```diff
@@ -1708,11 +1737,25 @@ exports.Assign = class Assign extends Base
   unfoldSoak: (o) ->
     unfoldSoak o, this, 'variable'
 
+  objectHasSplat: (o, obj) ->
```
zdenko: Yes, indeed. These checkings are all over the place and should be tamed.
The truth is, I just started to use CS and I've kinda rushed in without proper knowledge of the CS internals.

src/nodes.coffee
```diff
@@ -1728,6 +1771,21 @@ exports.Assign = class Assign extends Base
         # destructured variables.
         @variable.base.lhs = yes
         return @compileDestructuring o unless @variable.isAssignable()
+        # Find rest elements in object destructuring. Can be removed once ES proposal hits Stage 4.
```
zdenko: Actually, I started to work on a version before, where I tried to achieve something like that, but then I hit some walls and ended with this. 
I'll try to refactor this.

src/nodes.coffee
```diff
@@ -1770,10 +1828,60 @@ exports.Assign = class Assign extends Base
...
+          props.pop()
+        if obj instanceof Splat
+          obj.error "multiple rest elements are disallowed in object destructuring" if restElement
+          restElement = {key, name: obj.unwrap(), props: [props...]}
+      if restElement
+        # Remove rest element from the object.
+        objects.splice restElement.key, 1
+        # Prepare array of property keys to be excluded from the object
+        # TODO: refactor
+        excludeProps = (getPropValue(obj) for obj in objects)
```
zdenko: Agree. The current solution has too much of a "here be dragons" feeling. Hence the "refactor" reminder.

src/nodes.coffee
```diff
@@ -1770,10 +1828,60 @@ exports.Assign = class Assign extends Base
...
+        # Fix the quotes.
+        excludeProps = ((if prop[0] == "`" then prop else "'#{prop.replace /\'/g, ""}'") for prop in excludeProps)
```
zdenko: Yep. The previous comment applies here also. 
I tried to solve the computed properties, i.e. {"#{d}": z} in splats. Not good.

src/nodes.coffee
```diff
@@ -2113,6 +2221,47 @@ exports.Code = class Code extends Base
             param.name.lhs = yes
             param.name.eachName (prop) ->
               o.scope.parameter prop.value
+            # Check if object parameter has splat. Can be removed when ES proposal hits Stage 4.
```
zdenko: Agree. I'll take a look at it.

zdenko:
@connec, @GeoffreyBooth I had to chase some deadlines in the past few weeks, but my schedule is now cleared, and I can spend more time on this PR.

GeoffreyBooth:
Great to hear @zdenko! The [CSX PR](https://github.com/jashkenas/coffeescript/issues/4551) (adding Coffee-style support for JSX) is nearing completion, so this would be great to merge in shortly after it. I suggested on that PR’s thread that those interested in CSX might want to help this PR along, and that they should coordinate with you on this thread.

zdenko:
@connec I've refactored the object spread properties. Would you mind taking a look?

```
compileNode: (o) ->
    props = @properties
    ...
    hasSplat = no
    hasSplat = true for prop in props when prop instanceof Splat
    return @compileSpread o if hasSplat
    ...

  compileSpread: (o) ->
    props = @properties
    # Store object spreads.
    splatSlices = []
    propSlices = []
    slices = [new Obj [], false]
    addSlice = () -> 
      slices = [slices..., new Obj [propSlices...], false] if propSlices.length      
      slices = [slices..., splatSlices...] if splatSlices.length
    for prop in props
      addSlice()
      if prop instanceof Splat
        splatSlices.push new Value prop.name
        propSlices = []
      else
        propSlices.push prop
        splatSlices = []
    addSlice()
    (new Call new Literal('Object.assign'), slices).compileToFragments o

```

zdenko:
@connec Can you review the changes I've made. I'm onto finishing a.b.c and a[b][c] spread support.

src/grammar.coffee
```diff
@@ -268,6 +271,7 @@ grammar =
   Param: [
     o 'ParamVar',                               -> new Param $1
     o 'ParamVar ...',                           -> new Param $1, null, on
+    o '... ParamVar',                           -> new Param $2, null, on
```
xixixao: I think that allowing both ways is a pretty bad idea. It will lead to inconsistent code, the need for custom linters to tell people what to stick to and not bikeshed. Either we should switch to ES6 style or keep CS style, but not allow both. I am inclined to a switch, as it will make it easier for people to go between JS and CS, and it should be easy to codemod (I assume that CoffeeScript 2 has more subtle backwards incompatibility problems, if that's not the case, we should just stick to CS style).

src/nodes.coffee
```diff
@@ -3097,12 +3211,13 @@ exports.If = class If extends Base
 
 UTILITIES =
   modulo: -> 'function(a, b) { return (+a % (b = +b) + b) % b; }'
-
+  objectWithoutKeys: -> 'function(obj, props) { return Object.keys(obj).reduce(function(a,c) { return (![].includes.call(props, c)) && (a[c] = obj[c]), a; }, {}); }'
```
xixixao: This looks like an expensive implementation. What about
```coffee
objectWithoutKeys: "function(o, ks) {
    var res = {};
    for (var k in o) {
      indexOf.call(ks, k) < 0 && hasProp.call(o, k) && res[k] = o[k]
    }
    return res;
  }"
```
(you don't have to inline it manually, CS `"strings"` will do that for you. It's a couple characters shorter because I used shorter identifiers, but you can use longer ones.

GeoffreyBooth: We don’t have the `indexOf` or `hasProp` helpers anymore.

src/grammar.coffee
GeoffreyBooth: Allowing the dots on either side was my idea: https://github.com/coffeescript6/discuss/issues/85

We definitely can’t get rid of the dots-on-the-right syntax. One of the goals of CS2 is to introduce as few breaking changes as possible, and forcing the dots to always be on the left is a gratuitous breaking change. I suggested allowing on either side, because I and others have had the experience of going back and forth with ES and typing the dots on the wrong side; and it seemed like an easy enough quick win to just allow them on either side. There’s already precedent in CoffeeScript for such a thing, in that we allow `&&` as well as `and`, `||` as well as `or`, etc.

I think CoffeeScript recommended style will remain that the dots _should_ go on the right, just as you shouldn’t type `&&` in CoffeeScript, but I think this is one of those cases where the compiler can be flexible enough to accept either style.

If you disagree let’s please discuss it in https://github.com/coffeescript6/discuss/issues/85. @jashkenas, @lydell, your opinions would be welcomed.

GeoffreyBooth:
If you're going to implement the dots on either side as part of this PR, that will need tests, including for if people put the dots on both sides. Might be better off as its own PR, unless you need it for the object destructuring.

src/nodes.coffee
xixixao: @GeoffreyBooth How come? Even if I try /v2 I see hasProp being generated.

GeoffreyBooth: Hmm. Maybe we still have `hasProp`? I know I removed `indexOf`, in favor of native `[].indexOf`. All I meant was that you should rewrite your function to be independent, to not depend on other helpers that may or may not exist.

connec: I'm not sure if utilities can use other utilities, since they have to way of injecting them (e.g. if some code uses `objectWithoutKeys` but *doesn't* use `hasProp`, `hasProp` won't be in the source.

connec:
@zdenko Looks very nice on first pass! I'll try and have a closer look later today. I'll also echo the others with my hesitancy about leading and trailing `...` in this PR due to the test overhead etc (as well as it's an ongoing discussion 😢!).

GeoffreyBooth:
If the dots on both sides is just the one line you added to grammar.coffee, do you mind taking it out for now and that can be its own PR with tests? It deserves its own review.

zdenko:
@GeoffreyBooth dots on the left are now removed.

zdenko:
@connec I added support for deep spread properties. This is initial commit, since I still have to prepare tests.

`obj2 = {obj.c..., obj["b"]["c"]..., d: 55}`
compiles to
`obj2 = Object.assign({}, obj["b"], obj["b"]["c"], {d: 55});`

GeoffreyBooth:
@zdenko Please merge `2` back into your branch and resolve conflicts. Also please let us know when you’ve added the tests per your last comment and this is ready for another review.

zdenko:
@GeoffreyBooth I'm working on it. I had some extra work with covering object literal spread e.g. `{ {a:1, b: {c: 2}}..., a: 5 }`, and parenthetical expressions e.g. `{ a: 5, ((k) => ({a: k})(6))... }`. 
Tests and merging should be ready later today.

src/nodes.coffee
```diff
@@ -1187,6 +1197,28 @@ exports.Obj = class Obj extends Base
...
+  # Object spread properties. https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md
+  # obj2 = {a:1, obj..., c:3, d:4} => obj2 = Object.assign({}, {a:1}, obj1, {c:3, d:4})
+  compileSpread: (o) ->
+    props = @properties
+    # Store object spreads.
+    splatSlice = []
+    propSlices = []
+    slices = [new Obj [], false]
+    addSlice = () -> 
+      slices = [slices..., new Obj [propSlices...], false] if propSlices.length
```
connec: There's no need to slice `propSlices` since it is reassigned (e.g. `[propSlices...]` could just be `propSlices`).

Also, `generated = no` by default, so the second argument to `new Obj` is unnecessary.

Also, this might be a bit more readable (and performant :man_shrugging:) if it used `.push` instead of `...`, same with below. Personally I like the 'immutability' of using spreads like this, but it's the common style in the codebase to use `.push args...` instead.

src/nodes.coffee
```diff
@@ -1187,6 +1197,28 @@ exports.Obj = class Obj extends Base
...
       prop.eachName iterator if prop.eachName?
 
+  # Object spread properties. https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md
+  # obj2 = {a:1, obj..., c:3, d:4} => obj2 = Object.assign({}, {a:1}, obj1, {c:3, d:4})
+  compileSpread: (o) ->
+    props = @properties
+    # Store object spreads.
+    splatSlice = []
+    propSlices = []
+    slices = [new Obj [], false]
```
connec: Both of these parameters are optional and default to `[]` and `false` respectively, so this could just be `[new Obj]`.

Also, bit of a bigger one, but perhaps instead of starting with an empty object we could `.unshift` it at the end `unless slices[0] instanceof Obj` - that would save an empty object in case there are keys before splats (e.g. `{a, b...}` would be `Object.assign({ a }, b)` instead of `Object.assign({}, { a }, b)`).

src/nodes.coffee
```diff
@@ -1187,6 +1197,28 @@ exports.Obj = class Obj extends Base
...
+    addSlice = () ->
```
connec: I think the style in the codebase is generally to drop empty `()`.

src/nodes.coffee
```diff
@@ -1728,7 +1759,8 @@ exports.Assign = class Assign extends Base
         # destructured variables.
         @variable.base.lhs = yes
         return @compileDestructuring o unless @variable.isAssignable()
-
+        # Object destructuring. Can be removed once ES proposal hits Stage 4.
+        return @compileObjectDestruct(o) if @variable.isObject() and @variable.hasSplat()
```
connec: Gonna moan about this again (sorry) - I'd rather not make the `hasSplat` helper on `Value` when it's only used here, and `@variable.base.hasSplat()` would work just as well.

src/nodes.coffee
```diff
@@ -1717,7 +1749,6 @@ exports.Assign = class Assign extends Base
     if isValue
       # When compiling `@variable`, remember if it is part of a function parameter.
       @variable.param = @param
-
```
connec: Pedantic: unnecessary whitespace removal.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or (isValue and @variable.base instanceof Obj and not @param)
+    if o.level > LEVEL_LIST or (isValue and @variable.base instanceof Obj and (not @param or @paramWithSplat))
```
connec: `@paramWithSplat` is never assigned, so this line can be reverted.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+      if obj instanceof Assign
+        if obj.value.base instanceof IdentifierLiteral
+          newVar = obj.value.base.compile o
+        else 
+          newVar = obj.variable.base.compile o
+      else
+        newVar = obj.compile o
+      o.scope.add(newVar, 'var', true) if newVar
+    # Helper function getPropVaues() returns object compiled property value.
+    # These values are then passed as argument to helper function objectWithoutKeys which is used to assign object value to the destructuring rest variable.
```
connec: Pedantic: comment line too long.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+    getPropValue = (obj) ->
+      wrapInQutes = (obj) -> (new Literal("'#{obj.compile o}'")).compile o
+      setScopeVar obj # Declare a variable in the scope.
+      if obj instanceof Assign
+        return obj.variable.compile o if obj.variable.base instanceof StringWithInterpolations or obj.variable.base instanceof StringLiteral
```
connec: I wonder if it would be safe for `StringWithInterpolations` to `extends StringLiteral`? I guess it doesn't really qualify as a 'literal'...

Out of scope for this PR, anyway.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+          props.push getPropValue obj
+          results = traverseRest obj.value.base.objects, props
+          props.pop()
+        if obj instanceof Splat
+          obj.error "multiple rest elements are disallowed in object destructuring" if restElement
+          restElement = {
+            key,
+            name: obj.unwrap(),
+            # props: [props...].map((p) -> (new Literal "[#{p}]").compile(o)).join ""
+            props: ((new Literal "[#{p}]").compile(o) for p in [props...]).join ""
```
connec: `[props...]` is unnecessary, `props` alone is equivalent.

connec: Also, it would be nice if we could generate `a.b.c` instead of `['a']['b']['c']`. Again, this is something compiler should be able to do for us, since property variables should already be an `Access` or an `Index` (before they go through `getPropValue`).

src/nodes.coffee
```diff
@@ -2029,14 +2136,18 @@ exports.Code = class Code extends Base
...
       if node.this
         name   = node.properties[0].name.value
         name   = "_#{name}" if name in JS_FORBIDDEN
         target = new IdentifierLiteral o.scope.freeVariable name
         param.renameParam node, target
         thisAssignments.push new Assign node, target
-
+      # In case the parameter is object and contains the splat, e.g. fn({a, b, r...})
+      # Can be removed once ES proposal hits Stage 4.  
+      if param.name instanceof Obj and param.name.hasSplat()
```
connec: I'm not sure if this is necessary? Afaict splats in arrays do not get this treatment, and `.splat` on a param is used to indicate a splat *param* (e.g. `foo(...splat)`).

src/nodes.coffee
connec: Actually, the utility function is [invoked with `o`](https://github.com/GeoffreyBooth/coffeescript/blob/d59be16ab1402fc5965032a93648604d10e2f8a6/src/nodes.coffee#L3246), so this could indeed use the utilities!

test/scope.coffee
```diff
@@ -115,4 +115,4 @@ test "#3259: leak with @-params within destructured parameters", ->
 
   eq 'undefined', typeof foo
   eq 'undefined', typeof bar
-  eq 'undefined', typeof baz
+  eq 'undefined', typeof baz
```
connec: Pedantic: newline at EOF

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
       answer
 
+  # Check object destructuring variable for rest elements;
+  # can be removed once ES proposal hits Stage 4.
+  compileObjectDestruct: (o) ->
+    # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
+    # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
+    # ({a, b} = obj)
+    # Helper function setScopeVar() declares vars 'a' and 'b' at the top of the current scope.
+    setScopeVar = (obj) ->
```
connec: I think this and `getPropValue` would be a bit more readable if the variable was named, e.g., `prop`. `obj` might be thought to refer to an `Obj` instance, whereas in both methods we're always dealing with properties.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+        return wrapInQutes obj
+    # Recursive function for searching and storing rest elements in objects.
+    # Parameter props[] is used to store nested object properties,
+    # e.g. `{a: {b, c: {d, r1...}, r2...}, r3...} = obj`.
+    traverseRest = (objects, props = []) ->
+      results = []
+      restElement = no
+      for obj, key in objects
+        if obj instanceof Assign and obj.context == "object" and obj.value.base instanceof Obj
+          props.push getPropValue obj
```
connec: Rather than `push`in' an' `pop`in', this could use spreads:

```coffee
results = traverseRest obj.value.base.objects, [props..., getPropValue obj]
```

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+      setScopeVar obj # Declare a variable in the scope.
+      if obj instanceof Assign
+        return obj.variable.compile o if obj.variable.base instanceof StringWithInterpolations or obj.variable.base instanceof StringLiteral
+        return wrapInQutes obj.variable
+      else
+        return wrapInQutes obj
+    # Recursive function for searching and storing rest elements in objects.
+    # Parameter props[] is used to store nested object properties,
+    # e.g. `{a: {b, c: {d, r1...}, r2...}, r3...} = obj`.
+    traverseRest = (objects, props = []) ->
```
connec: Bit pedantic again, but maybe `properties` would be clearer. I know `Obj` has `.objects`, but I think that's just to provide a common interface with `Arr`. If we know we're dealing with an `Obj` I think it's clearer to use `properties`.

src/nodes.coffee
```diff
@@ -1770,11 +1802,86 @@ exports.Assign = class Assign extends Base
...
+      results = []
+      restElement = no
+      for obj, key in objects
+        if obj instanceof Assign and obj.context == "object" and obj.value.base instanceof Obj
```
connec: Is the `context` check necessary? I think any `Assign` that can appear as a property in an object will always have `context == 'object'`.

src/nodes.coffee
zdenko: Without this `foo({a, b, ...r})` compiles to `foo(Object.assign({}, {a, b}, r))`

zdenko: Actually your're 100% right.

src/nodes.coffee
```diff
@@ -1853,51 +1852,49 @@ exports.Assign = class Assign extends Base
...
       o.scope.add(newVar, 'var', true) if newVar
-    # Helper function getPropVaues() returns object compiled property value.
-    # These values are then passed as argument to helper function objectWithoutKeys which is used to assign object value to the destructuring rest variable.
-    getPropValue = (obj) ->
-      wrapInQutes = (obj) -> (new Literal("'#{obj.compile o}'")).compile o
-      setScopeVar obj # Declare a variable in the scope.
-      if obj instanceof Assign
-        return obj.variable.compile o if obj.variable.base instanceof StringWithInterpolations or obj.variable.base instanceof StringLiteral
-        return wrapInQutes obj.variable
+    # Helper function getPropVaues() returns compiled object property value.
```
vendethiel: Typo in the name

src/nodes.coffee
```diff
@@ -2094,7 +2200,9 @@ exports.Code = class Code extends Base
         else if param instanceof Expansion and @params.length is 1
           param.error 'an expansion parameter cannot be the only parameter in a function definition'
 
-        haveSplatParam = yes
+        # If the parameter is object and contains the splat, e.g. fn({a, b, r...}), 
+        # set haveSplatParam to false to avoid adding '...' to the function argument.
+        haveSplatParam = yes and param.name not instanceof Obj
```
vendethiel: `yes and ...` can be simplified to `...`

src/nodes.coffee
```diff
@@ -1814,6 +1847,80 @@ exports.Assign = class Assign extends Base
...
+          restElement = {
+            key,
+            name: prop.unwrap(),
+            props: ((new Literal "[#{p}]").compile(o) for p in path).join ""            
+          }  
+      if restElement
+        # Remove rest element from the object.
+        properties.splice restElement.key, 1
+        # Prepare array of compiled property keys to be excluded from the object.
+        restElement["excludeProps"] = new Literal "[#{(getPropValue(prop) for prop in properties)}]"
```
vendethiel: Minor nitpick: why is this using `["key"]`?

zdenko: Because currently, we're not forcing splat to be the last element in the object destructuring, the ["key"] is only used for storing the position of splat property in the object's properties list.
(I think it had also some other use in the previous iterations.)
Anyway, your right, it would be cleaner if a simple variable is used instead.

src/nodes.coffee
zdenko: I think this condition is not needed anymore.

zdenko:
> This is a bit of an aside, but babel will reject destructuring object spreads that aren't the last entry. ... Since object spreads aren't ordered we could optionally ban them, especially if it would make things easier.

@connec IMHO I don't think spreads should be forced as the last entry.
Currently, the code is properly transpiled, and later we can ensure the correct output:
`{a, r..., b} = obj` => `{a, b, ...r} = obj`

zdenko:
@connec 

>  compileObjectDestruct - there's a lot going on there! Any chance you could comment (on the PR, not the code) with some pseudocode?

The purpose of the `compileObjectDestruct` is to traverse recursively down the object destructuring properties, collect regular and spread properties, and use them to output code, similar to [babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=stage-3&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&playground=true&code=%0Avar%20%7Ba%3A%20%7Bb%2C%20c%3A%7Bd%2C%20...r1%7D%2C%20...r2%7D%2C%20...r3%7D%20%3D%20obj%0A%0A&stage=0).

Helpers:

- setScopeVar
Since all variables are declared at the top of the scope, the destructuring assignment must be wrapped in parentheses, e.g.`({a, b} = obj)`
For the root scope this wouldn't be necessary, but when object destructuring is used as function parameter and variables aren't declared, they will belong to the root scope which can lead into troubles:
Example:   
```
(function(arg) {
    ({a, b} = arg), r = objectWithoutKeys(arg, ['a','b']);
  })({ a: 1, b: 2, c: 3, d: 4 });

console.log(a); // a == 1;
```
- getPropValue
This functions returns the value of the property.

```
# (prop = property, quote = y/n where y = return prop in quotes)
getPropValue(prop, quote = no)
  # declare variable in the current scope
  setScopeVar(prop)
  if prop is Assign 
    return compiled prop if prop is "string"
    else return prop.variable wrapped in quotes
  else
  return prop wrapped in quotes
```

- traverseRest
Recursively traverses down the *children* properties and returns array of spread properties.
Example: 
For object properties of `{a: {b, c: {d, r1...}, r2...}, r3...}`, the returned value is
```
[
  {
     name: r1,
     path: [a, b, c],
     excludeProps: [d]
  },
  {
    name: r2,
    path: [a],
    excludeProps: [b, c]
  },
  {
    name: r3,
    path: [],
    excludeProps: [a]
  }
]
```

```
# properties = object properties, path = path to the spread element
traverseRest(properties, path = [])
  for prop in properties
    result = []
    # store spread property
    restElement = false
    if prop has child object destruct
      add prop to the path
      result = traverseRest(prop value properties, path)
      if prop is Splat
        error if there is already stored spread property at this level
        save current position (index) of the prop in the array
        restElement = {name: prop.name, path}
  # end for	
  if restElement
    remove spread property from the properties
    store remaining properties as array of Literal values in the restElement.excludeProps
    push restElement to the result array
  return result
```

- main body
The main body of the  `compileObjectDestruct`  method:
```
fragments = [] # store output result
Get properties from variable.base
Get restList from traverseRest( properties )

If the value of the Assign isn't simple variable, create one in the current scope.
# e.g. {a, b} = { a: 1, b: 2, c: 3, d: 4 } => obj = { a: 1, b: 2, c: 3, d: 4 }; {a, b} = obj)

Assign compiled object (variable) to the value and push result to fragments.

Loop thru the restList
  prepare object path
  create extractKeys as call to utility helper 'objectWithoutKeys'
  assign extractKeys to the spread property and push result to fragments

return fragments 
```

zdenko:
@GeoffreyBooth @connec 

I think I covered all cases of the spred properties:

```javascript
obj = {a:1, obj...}
// obj = Object.assign({a:1}, obj)

obj = {obj..., a:1}
// obj = Object.assign({}, obj, {a:1})

obj = {obj.b.c..., a:1}
// obj = Object.assign({}, obj.b.c, {a:1})

obj = {obj['d'].e..., a:1}
// obj = Object.assign({}, obj['d'].e, {a:1})

obj = {obj..., ((k) -> {a: k})(8)...}
// obj = Object.assign({}, obj, (function(k) {return {a: k};})(8))

obj = {obj..., foo(arg)...}
// obj = Object.assign({}, obj, foo(arg))
```

zdenko:
@connec I've added all tests and this should complete my work.
I think I covered all cases.

Sorry about the whitespace mess. I'll be more careful in the future.

GeoffreyBooth:
@zdenko Can you please merge in `2` and resolve conflicts?

helixbass:
@zdenko I was just working off of this branch and there's a lot of trailing whitespace sitting around eg in `nodes.coffee`

GeoffreyBooth:
@helixbass @zdenko I cleaned up the whitespace. @helixbass you can merge this branch back into yours.

helixbass:
@zdenko I ran into a bug: when destructuring function args, the destructured var doesn't get declared (with `var`). eg `({a, b...}) ->` compiles to:
```
var objectWithoutKeys = function(o, ks) { var res = {}; for (var k in o) ([].indexOf.call(ks, k) < 0 && {}.hasOwnProperty.call(o, k)) && (res[k] = o[k]); return res; };

(function(arg) {
  var a, arg;
  ({a} = arg), b = objectWithoutKeys(arg, ['a']);
});
```
There is no `var b` declaration, `b` just gets assigned to. Also it strikes me as bizarre/incorrect that the function param `arg` gets redeclared as `var arg` in the body, although this doesn't seem to break when running the compiled output

helixbass:
@zdenko I pushed a commit that seems to fix it (ie properly declares destructured vars and doesn't declare `arg` param): just updated `if param.name instanceof Arr or param.name instanceof Obj` to `if (param.name instanceof Arr or param.name instanceof Obj) and not param.shouldCache()` in `Code#compileNode` (line 2281)

GeoffreyBooth:
@helixbass we should add a test that covers this.

helixbass:
@GeoffreyBooth good call, writing a test exposed a case that wasn't handled so I refined my fix. I believe it's now correctly explicitly handling both the destructured param and its assigned vars but definitely should be reviewed, thanks

zdenko:
@helixbass @GeoffreyBooth Nice catch. 
I refactored this fix a little bit.
Per @connec's advise I try to keep all object destructuring logic in Obj and Assign.

Call scope.freeVariable with the argument, to avoid "arg" being declared in the `Param.asReference`:
`node = new IdentifierLiteral o.scope.freeVariable 'arg', {reserve: no}`

Declare object spread variable in the current scope in the `Assign.compileObjectDestruct`.
```
o.scope.add restElement.name.compile(o), 'var'
fragments.push new Assign(restElement.name, extractKeys ....
```

helixbass:
@zdenko I just came across another bug, default values with destructured spread params don't work eg `({a, b...} = {}) ->` compiles to:
```
var objectWithoutKeys = function(o, ks) { var res = {}; for (var k in o) ([].indexOf.call(ks, k) < 0 && {}.hasOwnProperty.call(o, k)) && (res[k] = o[k]); return res; };

(function(arg) {
  var a, b, obj;
  if (Object.assign({a}, b) === void 0) {
    obj = {}, ({a} = obj), b = objectWithoutKeys(obj, ['a']);
  }
});
```
I think it should look more like this?
```
var objectWithoutKeys = function(o, ks) { var res = {}; for (var k in o) ([].indexOf.call(ks, k) < 0 && {}.hasOwnProperty.call(o, k)) && (res[k] = o[k]); return res; };

(function(arg = {}) {
  var a, b;
  ({a} = arg), b = objectWithoutKeys(arg, ['a']);
});
```

zdenko:
@helixbass It's fixed. 

`({a, b...} = {}) ->` now compiles to
```
(function(arg = {}) {
  var a, b;
  ({a} = arg), b = objectWithoutKeys(arg, ['a']);
});
```

helixbass:
@zdenko another one: `{a: {b}, c...} = d` compiles to:
```
var c,
  objectWithoutKeys = function(o, ks) { var res = {}; for (var k in o) ([].indexOf.call(ks, k)< 0 && {}.hasOwnProperty.call(o, k)) && (res[k] = o[k]); return res; };

({
  a: {b}
} = d), c = objectWithoutKeys(d, ['a']);
```
It fails to declare `b`

zdenko:
@helixbass variable declaration is fixed.

GeoffreyBooth:
So where do we stand now with this? Have all the recently reported bugs been fixed, and covered by tests? @zdenko are you still working on other improvements?

zdenko:
@GeoffreyBooth I'm finished with the improvements.

zdenko:
As far as I know, all bugs are fixed.

connec:
I think there might be a missing traversal somewhere, this (admittedly obnoxious) example [compiles with babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-2&targets=&browsers=&builtIns=false&debug=false&code_lz=MYewdgzgLgBA3jAhgLnjARquwYF4YAmAvgDQwCmeMAZjEVXPffgOZA) but not with this branch

```
[stdin]:1:14: error: unexpected =
{ a: { b: {c = d}, e = f } = {} } = g
             ^
```

zdenko:
Also this one: 
```
[stdin]:1:10: error: unexpected =
{ a: { e = f } } = g
```

It seems the issue might be the grammar.

zdenko:
@connec I think I solved this issue.
The error occurred in the Obj at this part, where the `prop` is `Assign` and `prop.context` is `null` and `@lhs` is false.
```
key = if prop instanceof Assign and prop.context is 'object'
    prop.variable
  else if prop instanceof Assign
    prop.operatorToken.error "unexpected #{prop.operatorToken.value}" unless @lhs
    prop.variable
  else if prop not instanceof Comment
    prop
```

I solved this by setting `@lhs` to true in the `Assign` if `@variable.base` is a `PropertyName` and `@value.base` is an object.

```
if @variable.base instanceof PropertyName and @value.base instanceof Obj
  @value.base.lhs = yes
```

GeoffreyBooth:
@zdenko have all the issues been resolved?

zdenko:
@GeoffreyBooth from my side, yes.

GeoffreyBooth:
@connec @lydell I added [Babel’s example](https://babeljs.io/docs/plugins/transform-object-rest-spread/) as a test, which this passes. I also manually went through the tests in https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-object-rest-spread/test/fixtures/object-spread, and the output from this PR is the same as Babel’s expected output.

Are we ready to merge this in?

src/nodes.coffee
```diff
@@ -1157,11 +1157,21 @@ exports.Obj = class Obj extends Base
...
+    splat ? no
+
   compileNode: (o) ->
     props = @properties
     if @generated
       for node in props when node instanceof Value
         node.error 'cannot have an implicit value in an implicit object'
+
+    # Object spread properties. https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md
+    # `obj2 = {a: 1, obj..., c: 3, d: 4} => obj2 = Object.assign({}, {a: 1}, obj1, {c: 3, d: 4})`
```
lydell: This comment is duplicated a bit below. I'd say we remove this one and keep the other.

src/nodes.coffee
```diff
@@ -1236,6 +1243,29 @@ exports.Obj = class Obj extends Base
       prop = prop.unwrapAll()
       prop.eachName iterator if prop.eachName?
 
+  # Object spread properties. https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md
+  # `obj2 = {a: 1, obj..., c: 3, d: 4} => obj2 = Object.assign({}, {a: 1}, obj1, {c: 3, d: 4})`
```
```
`obj2 = {a: 1, obj..., c: 3, d: 4}` → `obj2 = Object.assign({}, {a: 1}, obj, {c: 3, d: 4})`
```

test/assignment.coffee
```diff
@@ -235,6 +235,105 @@ test "destructuring assignment against an expression", ->
...
+
+# Tests from https://babeljs.io/docs/plugins/transform-object-rest-spread/.
+test "destructuring assignment with objects and splats: Babel tests", ->
+  # What Babel calls “rest properties:”
+  { x, y, z... } = { x: 1, y: 2, a: 3, b: 4 }
+  eq x, 1
+  eq y, 2
+  eq z.a, 3
+  eq z.b, 4
+  deepEqual z, { a: 3, b: 4 }
```
lydell: Is there a reason we first manually check properties and then do a `deepEqual`?

test/assignment.coffee
```diff
@@ -235,6 +235,105 @@ test "destructuring assignment against an expression", ->
   eq a, y
   eq b, z
 
+test "destructuring assignment with objects and splats: ES2015", ->
+  obj = {a: 1, b: 2, c: 3, d: 4, e: 5}
+  throws (-> CoffeeScript.compile "{a, r..., s...} = x"), null, "multiple rest elements are disallowed"
+  throws (-> CoffeeScript.compile "{a, r..., s..., b} = x"), null, "multiple rest elements are disallowed"
```
lydell: Isn't it better to do throw tests in error_messages.coffee, so we know that we get the error location right?

test/assignment.coffee
GeoffreyBooth: Ah, this is my bad. I guess there isn’t, I was just copying Babel’s test.

connec: It does check there are no additional properties. It could probably do without the property checks though.

test/assignment.coffee
GeoffreyBooth: Perhaps, although this file is full of `throws` tests.

GeoffreyBooth:
@connec and @lydell I’ve addressed your most recent comments.

GeoffreyBooth:
FYI: https://developers.google.com/web/updates/2017/06/object-rest-spread

lydell:
@GeoffreyBooth Oh, so there are some subtle differences between `Object.assign` and the object spread syntax proposal. I had no idea. How annoying.

connec:
I did have quite a wall of text written here, but my laptop restarted on my way home and I lost it 😢 

The upshot is, I found a few more bugs (summarised below), and after playing whack-a-mole for a little while I ended up rewriting some of `traverseRest` and a few smaller things in a couple of other places. The main cases that are now handled are:

```coffee
# destructuring assignment splat with default value
# originally the properties would always be taken from obj.a, now they're taken from
# `obj.a ? c`
{ a: {b...} = c } = obj

# destructuring assignment with multiple splats in different objects
# originally `b` would be ignored
{ a: {a...}, b: {b...} } = obj

# destructuring assignment with dynamic keys and splats
# originally `foo` would be called twice
{ "#{foo()}": a, b... } = obj
```

GeoffreyBooth:
@connec so is this safe/ready to merge in now, in your opinion? Assuming @lydell and @zdenko approve of your commits?

connec:
I'm adding a final, gnarly destructuring test and then I'm done with it I think.

GeoffreyBooth:
@connec was that last commit it? Does this now have your blessing?

@lydell and @zdenko, do either of you want to final pass at this, or can I merge it in?

connec:
OK, that's definitely me done now @GeoffreyBooth. My last concern was the grammar, which didn't account for things like `{super()...}` or `{this...}`. There might still be some missing cases since using anything broader like `Value` or `Expression` leads to ambiguity (as I'm sure @zdenko can testify).

I didn't add tests for those things, and it's really late here now. Other than that, I think this is good to go (of course, as a contributor to the PR I now can't be trusted to be impartial 😄).

GeoffreyBooth:
At this point I’d rather merge it in and handle additional edge cases as further PRs. This has become an epic.

Last call!

GeoffreyBooth:
So I was writing the documentation for this, and the example I came up with doesn’t compile:
```coffee
user =
  name: 'Werner Heisenberg'
  occupation: 'theoretical physicist'

currentUser =
  user...
  status: 'Uncertain'
```
___
```
[stdin]:6:6: error: unexpected ...
	user...
	    ^^^
```
However changing `currentUser` to `currentUser = { user..., status: 'Uncertain' }` makes it work as expected. A similar example using arrays works as expected. Is this a bug?

zdenko:
@GeoffreyBooth this works
```
currentUser = {
  user...
  status: 'Uncertain'
}
```

I've tried to make it work without the `{}` in the past but always ended with ambiguity issues.

GeoffreyBooth:
Okay, well at least it just throws a compiler error for now. So if we figure out how to support that syntax later, we can add it in a later PR.

connec:
That's probably related to [`lookObjectish`](https://github.com/jashkenas/coffeescript/blob/master/src/rewriter.coffee#L110) in the rewriter, which seems to only accept implicit object literals when there's a `:` involved (which is consistent with being unable to use shorthand properties in implicit objects.

Tbh we probably want to leave it like this, or we'll need to face down ambiguities in expressions like

```coffee
f k: v, rest...
```

Which, currently, will compile to two arguments (an object literal and a splat param) but would be ambiguous if we allowed splats to exist in implicit objects.

PR #4494, opened 2017-04-09
[CS2] 2.0.0-beta1 docs

GeoffreyBooth:
https://rawgit.com/GeoffreyBooth/coffeescript/2.0.0-beta1-docs/docs/v2/index.html

I’ve updated the 2 docs to reflect:
- #4478: Destructuring
- #4485: Better parsing of Literate CoffeeScript
- #4484: Throw and error on ambiguous `get` or `set` function calls

I’ve removed Marked in favor of Markdown-It, so now we only have one Markdown parsing library across the project for both documentation and Literate CoffeeScript support. I also added syntax highlighting back in for the non-editable code blocks.

The release date for 2.0.0-beta1 is a wild guess that I expect we’ll revise as part of that actual release; I just thought these changes, especially the Markdown library swap, were significant enough to warrant their own PR and review.

On a side note, should we release beta1 now or wait for #4493 to land?


Cakefile
```diff
@@ -199,22 +199,36 @@ buildDocs = (watch = no) ->
...
+    markdownRenderer = require('markdown-it')
+      html: yes
+      typographer: yes
+      highlight: (str, lang) ->
+        # From https://github.com/markdown-it/markdown-it#syntax-highlighting
+        if lang and hljs.getLanguage(lang)
+          try
+            return hljs.highlight(lang, str).value
+          catch ex
+          return '' # No syntax highlighting
```
lydell: This line should be dedented once according to the documentation you linked to.

GeoffreyBooth: Good catch! Probably didn’t matter because highlight.js never threw an exception, but can’t hurt to get it right.

GeoffreyBooth:
So . . . release beta1 now (or rather, after I merge the 1.12.5 `master` fixes into it), or wait until #4493 is ready?

PR #4499, opened 2017-04-13
[CS2] 2.0.0-beta1

GeoffreyBooth:
Following up #4494, here’s a possible release of 2.0.0-beta1. In this PR:

- I noticed that I forgot to swap out Marked for Markdown-It in the browser build (that includes the Markdown-It minified .js file as part of its build) so this has been fixed.
- 1.12.5’s updates have been merged into `2`.

The browser build is **unminified**. This isn’t intentional; the Closure Compiler [chokes on our latest build](https://github.com/google/closure-compiler-js/issues/59). I suspect it’s because it doesn’t fully support ES destructuring syntax, and that’s an ES feature we use rather extensively in the compiler codebase.

So I guess our options are to either release it unminified, at least until Closure Compiler supports whatever is causing it to choke; find yet another minifier; or add a Babel transpilation step before minification. I’m loathe to add Babel; what do you all think? I’m hoping that maybe an unminified browser compiler is fine for the 2.0.0 beta releases, and hopefully Closure Compiler will be patched before we release 2.0.0 out of beta.

The other consideration worth discussing is whether we should wait until #4493 is merged in before releasing this. I’m leaning toward releasing beta1 now and beta2 when #4493 is ready. Any thoughts?


lydell:
Did you accidentally submit this PR against `master` instead of `2`?

GeoffreyBooth:
This is why I need you.

lydell:
I couldn't find the change log from alpha1 to beta1?

GeoffreyBooth:
https://rawgit.com/GeoffreyBooth/coffeescript/2.0.0-beta1/docs/v2/

I put 1.12.5 between 2.0.0-alpha1 and beta1 to imply that beta1 includes the changes from 1.12.5. I think it’s clear enough?

jashkenas:
No Babel addition, please — and it's fine to leave unminified temporarily, but it would be lovely to find a minifier that works for these sorts of things.

GeoffreyBooth:
@jashkenas Agreed, though remember that we switched from Uglify to Closure because Uglify didn’t support ES2015. Hopefully the Closure team fixes [the issue](https://github.com/google/closure-compiler-js/issues/59) before we release 2.0.0, or I guess the next best option is [Babili](https://babeljs.io/blog/2016/08/30/babili).

Issue #4277, opened 2016-07-01
using CoffeeScript.compile in client with webpack


Radivarig:
I am trying to use `CoffeeScript.compile ("code")` in browser but including the coffee-script library with webpack results in errors and warnings posted here #4276. The loader is not an option, because I don't have the code in advance.

There are two versions of coffee-script-min.js:

- one is bigger size 160kb which I have locally (downloaded ~1year ago) and is not available on [CDN](https://cdnjs.com). Including this file gives no errors and I'm currently using it, but would like to avoid adding it to source and to be able to update it above 1.9.3.
- the other is the one hosted on [CDN](https://cdnjs.com/libraries/coffee-script), and is 6kb. Including this one locally results in the same errors as using the coffee-script module with webpack.

That is why I am asking it here to get developers attention because I doubt that other users will have an idea on how the build was packaged.

I am kindly asking not to be referred to "google it" because it is insulting and serves no purpose. If you don't know how to solve this please do not comment or close, just pass it and let someone else take a look.

vendethiel:
Please ask questions on StackOverflow or the google groups.

akfish:
Running into the same issue here while trying to consume the compiler API in browser. 
IMHO this should be a valid feature rather a google-it-yourself-and-hack-it-together thing.

GeoffreyBooth:
Use the browser compiler:

```html
<script src="http://coffeescript.org/v1/browser-compiler/coffee-script.js"></script>
```

akfish:
@GeoffreyBooth Thanks for the information. I'm aware of the browser-compiler build.  

However I believe the issue is about consuming CoffeeScript compiler with modern toolchain like webpack, which won't take more than a browser specific entry script and a `"browser"` field in `package.json`. Babel compiler, for example, supports webpack just fine.

Anyway, I'm using a [custom fork](https://github.com/akfish/coffeescript) for my project with compiler API exposed for webpack.

Just `npm install --save github:akfish/coffeescript#webpack`.

GeoffreyBooth:
This has come up a few times before, so I’d rather solve it in the main repo than ask people to use forks that can get out of date. But I’m not sure exactly what the ask is, without an example that demonstrates what you’re trying to do.

akfish:
@GeoffreyBooth From time to time, people will want to pack the CoffeeScript source directly into his bundle with webpack, rather than using the prebuilt one. The expectations is to simply require or import coffee-script and webpack will do the rest.

[...]

Issue #4276, opened 2016-06-30
using coffee-script with webpack


Radivarig:
When I require with webpack `var CoffeeScript = require ('coffee-script')` I get the following output
```
WARNING in ./~/coffee-script/lib/coffee-script/coffee-script.js
require.extensions is not supported by webpack. Use a loader instead.

WARNING in ./~/coffee-script/lib/coffee-script/coffee-script.js
require.extensions is not supported by webpack. Use a loader instead.

WARNING in ./~/coffee-script/lib/coffee-script/coffee-script.js
require.extensions is not supported by webpack. Use a loader instead.

WARNING in ./~/coffee-script/lib/coffee-script/coffee-script.js
Critical dependencies:
175:42-49 require function is used in a way in which dependencies cannot be statically extracted
179:26-33 require function is used in a way in which dependencies cannot be statically extracted
 @ ./~/coffee-script/lib/coffee-script/coffee-script.js 175:42-49 179:26-33

WARNING in ./~/coffee-script/lib/coffee-script/register.js
require.extensions is not supported by webpack. Use a loader instead.

WARNING in ./~/coffee-script/lib/coffee-script/register.js
require.extensions is not supported by webpack. Use a loader instead.

WARNING in ./~/coffee-script/lib/coffee-script/browser.js
Critical dependencies:
8:25-32 require function is used in a way in which dependencies cannot be statically extracted
 @ ./~/coffee-script/lib/coffee-script/browser.js 8:25-32

ERROR in ./~/coffee-script/bin/coffee
Module parse failed: /path_to_repo/node_modules/coffee-script/bin/coffee Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected character '#' (1:0)
    at Parser.pp.raise (/path_to_repo/node_modules/webpack/node_modules/acorn/dist/acorn.js:923:13)
    at Parser.pp.getTokenFromCode (/path_to_repo/node_modules/webpack/node_modules/acorn/dist/acorn.js:2831:8)
    at Parser.pp.readToken (/path_to_repo/node_modules/webpack/node_modules/acorn/dist/acorn.js:2526:15)
    at Parser.pp.nextToken (/path_to_repo/node_modules/webpack/node_modules/acorn/dist/acorn.js:2518:71)
    at Parser.parse (/path_to_repo/node_modules/webpack/node_modules/acorn/dist/acorn.js:1631:10)
[...]

PR #4501, opened 2017-04-14
[WIP][CS2] Add webpack support

akfish:
Per discussions with @GeoffreyBooth, @timhuff and @connec in #4276 #4277 reported by @Radivarig, current coffee-script source cannot be packed with webpack due to references to Node.js-only modules in main script. 

Example error output of webpack:
```
Error in ./~/coffee-script/lib/coffee-script/coffee-script.js
Module not found: 'module' in project_root\node_modules\coffee-script\lib\coffee-script

 @ ./~/coffee-script/lib/coffee-script/coffee-script.js 168:23-40 208:17-34

Error in ./~/coffee-script/lib/coffee-script/register.js
Module not found: 'module' in project_root\node_modules\coffee-script\lib\coffee-script

 @ ./~/coffee-script/lib/coffee-script/register.js 25:13-30

Error in ./~/coffee-script/lib/coffee-script/register.js
Module not found: 'child_process' in project_root\node_modules\coffee-script\lib\coffee-scrip

 @ ./~/coffee-script/lib/coffee-script/register.js 7:18-42

Error in ./~/coffee-script/lib/coffee-script/command.js
Module not found: 'child_process' in project_root\node_modules\coffee-script\lib\coffee-scrip

 @ ./~/coffee-script/lib/coffee-script/command.js 16:8-32

Error in ./~/coffee-script/lib/coffee-script/grammar.js
Module not found: 'jison' in project_root\node_modules\coffee-script\lib\coffee-script

 @ ./~/coffee-script/lib/coffee-script/grammar.js 5:11-27

Error in ./~/coffee-script/lib/coffee-script/repl.js
Module not found: 'repl' in project_root\node_modules\coffee-script\lib\coffee-script

 @ ./~/coffee-script/lib/coffee-script/repl.js 11:13-28
```

## Changes
* Move Node.js-only code from `src/coffee-script.coffee` to `src/index.coffee`
* Use `lib/coffee-script/index.js` as npm package's `"main"` script
* Export `CoffeeScript` from `src/browser.coffee`
* Set `package.json`'s `"browser"` field to `lib/coffee-script/browser.js` (used by webpack as entry point)
* Use `lib/coffee-script/browser.js` as bower package's `"main"` script

## Tests
- [x] `cake build`
- [x] `cake build:browser`
- [x] `cake test`
- [x] `cake test:browser`
- [x] Manually tested packing with webpack
- [x] Manually verified compiled browser bundle works in browser (with re-built doc site)


jashkenas:
This is a lot of rejiggering for the webpack use case. CoffeeScript is primarily a Node.js compilation tool — **not** a browser tool. Our `main` should not be a `browser.js` file.

That said, a general reorganization of Node-specific parts to make webpack work sounds fine. Let's just keep the entry point for browser use as "browser".

src/coffee-script.coffee
```diff
@@ -156,101 +153,13 @@ exports.nodes = withPrettyErrors (source, options) ->
   else
     parser.parse source
 
-# Compile and execute a string of CoffeeScript (on the server), correctly
-# setting `__filename`, `__dirname`, and relative `require()`.
-exports.run = (code, options = {}) ->
-  mainModule = require.main
+# Compile and execute a string of CoffeeScript 
+exports.run = (code, options = {}) -> throw new Error 'Not implemented'
```
GeoffreyBooth: Why export this if it just throws an error?

GeoffreyBooth:
This will be tricky in 2. Take a look at `cake build:browser` in the `2` branch, where it includes the Markdown-It library.

src/coffee-script.coffee
akfish: The idea is that `coffee-script.coffee` will contain only platform-independent code. Any platform-specific code should be moved elsewhere. Throwing an error in those "abstract" method prevents developers from accidentally require `lib/coffee-script/coffee-script` and not knowing what's wrong.

akfish:
@jashkenas The npm package's main is `index.js`, which points to node.js only implementation. 

I modified bower package's main to `browser.js`, because bower describe itself as "A package manager for the web". I assume that implies any bower related usage should be browser-only. But I could be wrong because I don't use bower.

akfish:
@GeoffreyBooth The Markdown-it library seems to be added via simply concatenation in branch 2's cakefile. The coffee-script part seems to be the same as master. We could basically apply the same diff to [Line 146 of branch 2's Cakefile](https://github.com/jashkenas/coffeescript/blob/2/Cakefile#L146):
```diff
-        return require['./coffeescript'];
+        return require['./browser'];
```

akfish:
@jashkenas A little follow up on bower.json's main field.

[Bower Spec](https://github.com/bower/spec/blob/master/json.md#main) defines it as:

>The entry-point files necessary to use your package. Only one file per filetype.
... Bower does not directly use main files, they are listed with the commands bower list --json and bower list --paths, so they can be used by build tools.

It suggests that build tools determine how to use "main" field.

webpack uses it as entry point. See: https://webpack.github.io/docs/usage-with-bower.html
browserify's debowerify transform plugin use it as entry point as well. See: https://github.com/eugeneware/debowerify#notes

So it should points to a browser-specific implementation for build tools to work properly.

jashkenas:
> This will be tricky in 2. Take a look at cake build:browser in the 2 branch, where it includes the Markdown-It library.

Kind of a good argument for us trying to leave out *all* dependencies, and just do the markdown code block detection ourselves.

After all — we don't need to parse 98% of Markdown. The only thing we need to parse is code blocks.

akfish:
Just a thought: It would be trivial to setup webpack for building coffee-script browser bundle instead of `cake build:browser` after this PR is merged. Any dependencies can be easily resolved by webpack, as long as they play well with webpack. 

And I happened to be using Markdown-it in my project too. It works well with webpack.

GeoffreyBooth:
As @jashkenas implies, we’re not looking to add dependencies, so adding Webpack to our build is almost surely a nonstarter. The reason Markdown-It was added was to address false positives in detecting Markdown non-code blocks: https://github.com/jashkenas/coffeescript/pull/3924. Though the original issue didn’t have examples that failed, so I don’t know what would be involved in rolling our own implementation. That’s perhaps something we should revisit.

The docs do say:

> The [core compiler](http://coffeescript.org/v2/browser-compiler/coffeescript.js) however, does not depend on Node, and can be run in any JavaScript environment, or in the browser (see [Try CoffeeScript](http://coffeescript.org/v2/#try)).

So we should really try to honor this promise, and disentangle the Node dependencies.

akfish:
Webpack would be a devDependency, which is not the same as dependency. 
But I understand that as a compiler, the more dependencies it has, the less "meta" it feels like.

Anyway it's just a possibility and off the topic of this PR. So just ignore me. :)

lydell:
Markdown is very difficult to parse.

GeoffreyBooth:
@akfish A few questions:

1. Currently the browser compiler is built and committed into the repo in `docs/v1/browser-compiler/coffee-script.js` or `docs/v2/browser-compiler/coffeescript.js`. This isn’t included in the Node module, but what if it was? And then we added a `browser` key to `package.json` that pointed to it? Could that solve your problem all by itself?
2. Can you please post a minimal repo showing Webpack using this PR? So that we can see that it works, and also that if we flip to the released 1.12.5 it doesn’t work; and also so that I can test alternatives like the idea above.
3. Is there a test we can write that verifies this? Not a test specific to Webpack, but rather that requiring the compiler doesn’t also mean that requiring `fs` succeeds etc. Perhaps the test or the test runner can [override the `require` function](http://stackoverflow.com/a/34186494/223225), so that we have a way to fail the test if some portion of the compiler tries to `require` a Node module.

I think coming up with a good test for this is important. Years ago the compiler didn’t depend on Node, and clearly at some point that got lost; if we have a test that fails the moment someone puts `fs = require 'fs'` in the code, that will prevent such a thing from happening again.

akfish:
1. For webpack, `browser` key serves as an entry script for package resolution. They looks for calls `require` or `import` calls recursively from there and do a custom bundle each time the developer builds his project. So it doesn't matter to webpack what other files are included. The pre-built bundles will not be used by webpack, as long as they are not required explicitly by any files along its resolution path. And it is generally not a good practice to point `browser` field to pre-built bundles. It's best to only provide an entry script and let webpack figure out what needs to be packed. 
2. Will do.
3. Hook `require` could work. Although that means we would have to manually maintain a blacklist for non-compatible modules. And different tools resolves `require` calls differently (but mostly some form of static code analysis). The behavior might not be the same as runtime `require` hook. An easier and more reliable way is to run webpack build as a test. This approach requires adding webpack as devDependency. Add a webpack config file. Then simply run `webpack lib/coffee-script/browser.js dist/bundle.js` and see if there're any errors. 

I'll cook up some example repos for 2 and 3.

akfish:
Check out https://github.com/akfish/coffeescript/tree/demo/webpack-bundle. This repo demonstrates the minimal modifications (webpack as devDependency and a new `webpack.config.js` file) needed to test this PR by running a webpack build. Run `node_modules/.bin/webpack` after ran `cake build` will do the trick (or we could add a `cake test:webpack` task to run this command). I understand the the reluctance to add an additional dependency. But it's the cleanest method to test this PR that I could think of.

When running this test, I found out that generated `parser.js` contains reference to `fs` module, which messed things up. Commit 6cde4b5 fixed it by passing a NOP function to override the default `exports.main` emitted by Jison.

akfish:
Here is a minimal demo showing webpack using CoffeeScript: https://github.com/akfish/coffeescript-webpack-demo
Also commit 665eba1 removed [this line](https://github.com/jashkenas/coffeescript/blame/master/src/browser.coffee#L7) that broke the minimal demo (but oddly enough not my already-too-complex project). The commit added it 7 years ago said it's used for debugging and I didn't found any code references to it. And removing it didn't break any tests. I hope that's ok.

Cakefile
```diff
@@ -51,14 +51,8 @@ run = (args, callback) ->
...
-  parser = require('./lib/coffee-script/grammar').parser.generate()
-  # Patch Jison’s output, until https://github.com/zaach/jison/pull/339 is accepted,
-  # to ensure that require('fs') is only called where it exists.
-  parser = parser.replace "var source = require('fs')", """
-      var source = '';
-          var fs = require('fs');
-          if (typeof fs !== 'undefined' && fs !== null)
-              source = fs"""
+  # We don't need moduleMain, since the parser is unlikely to be run standalone
+  parser = require('./lib/coffee-script/grammar').parser.generate(moduleMain: ->)
```
GeoffreyBooth: This is brilliant. Thanks for solving this better than my hack.

Cakefile
```diff
@@ -135,7 +129,7 @@ task 'build:browser', 'merge the built scripts into a single file for use in a b
       var CoffeeScript = function() {
         function require(path){ return require[path]; }
         #{code}
-        return require['./coffee-script'];
+        return require['./browser'];
```
GeoffreyBooth: Can you explain this? Why `browser` here?

bower.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "coffee-script",
   "main": [
-    "lib/coffee-script/coffee-script.js"
+    "lib/coffee-script/browser.js"
```
GeoffreyBooth: This is fine. I’ve used Bower, it’s only for frontend (browser-based) JavaScript.

src/browser.coffee
```diff
@@ -4,7 +4,6 @@
 # `text/coffeescript` script tags, source maps via data-URLs, and so on.
 
 CoffeeScript = require './coffee-script'
-CoffeeScript.require = require
```
GeoffreyBooth: Thanks for catching this.

src/browser.coffee
```diff
@@ -18,6 +17,10 @@ CoffeeScript.run = (code, options = {}) ->
   options.shiftLine = on
   Function(compile code, options)()
 
+CoffeeScript.register = ->
```
GeoffreyBooth: So it seems like the strategy here is to declare this dummy method, and then in the Node version this gets overridden with the real method? This is confusing, to say the least.

GeoffreyBooth:
What if we [detected whether we’re in Node or not](https://github.com/iliakan/detect-node), and had the code branch accordingly? Or to follow the [feature detection pattern](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection), the functions that require Node modules could explicitly check for them and error out otherwise? Something like:

```coffee
try
  fs     = require 'fs'
  path   = require 'path'
  module = require 'module'
  vm     = require 'vm'
catch (e)

exports.run = (code, options = {}) ->
  unless require?.main? and process?.argv? and fs? and path? and module?
    throw new Error 'This function requires CommonJS and the process, fs, path and module modules'
  # rest of old `run` method
```

And then we don’t need abstract or dummy methods, and the refactoring doesn’t need to be so dramatic. That would make me much more comfortable that we aren’t breaking downstream projects that might require the `coffee-script` module.

Cakefile
akfish: This is for logical consistency. Since `browser.js` is already used as the entry for bower and webpack. It could avoid unnecessary confusion to use it for all browser related situations.

akfish:
@GeoffreyBooth Tools like webpack use static analysis to find `require` calls. The code is not actually executed. Conditional require and feature detection are runtime solutions. For example, the [parser's main method](https://github.com/jashkenas/coffeescript/blob/master/lib/coffee-script/parser.js#L874) is clearly not executed, but it broke webpack.

With that said, abstract method is one of the common patterns to enforce or at least document shared interfaces/behaviors, should multiple implementations are required.

src/browser.coffee
akfish: This is confusing because `register` method should not be available in browser environment, thus it should not be one of the abstract method. But doing so will break browser test. 

Because coffee's browser test is not actually ran in a browser. It runs in node environment and the first call in the `runTest` method is `CoffeeScript.register()`. I guess this might be because there's no karma, PhantomJs nor headless chromium back then.

That's why I set it to a NOP implementation to pass the test for now. I could configure a test script that actually runs in browser with one of the methods mentioned above. But I think that should be a separate PR/issue for discussion since it requires adding more devDependencies.

akfish: On a separate note, I also think `runScripts` method in the `browser.coffee` could be used as browser's `register` implementation. It's logically sound. But it will only work if the browser tests are actually running in browser. Right now it will only break the test when trying to call it in node environment.

GeoffreyBooth:
> Tools like webpack use static analysis to find `require` calls.

Hmf. So it tries to resolve all `require` calls, before the Node runtime ever gets involved? So no matter where a `require` statement is in the file, Webpack tries to resolve it? So even something like this wouldn’t work?

```coffee
# `isNode` defined above somehow

exports.run = if isNode then require('run') else ->
```

akfish:
@GeoffreyBooth Yes. IIRC, webpack performs static analysis on AST. Older tools like browserify uses simple RegExp match. The code was never run.

GeoffreyBooth:
What’s the argument for merging this into `master` instead of `2`? I’m much less worried about breaking changes in 2.

I can see how having a browser-safe module you can include that generates ES3 is desirable; is that the appeal? I assume it wouldn’t be too difficult to configure Webpack to require CoffeeScript 2 (as modified by this PR) and pipe its output through Babel, even in a browser-compatible way?

akfish:
I have no problem with that, as long as it'll will be published to npm.

I haven't had the need to generate ES3 yet so I can't say for sure. But it should be the case.

GeoffreyBooth:
CoffeeScript 2 is already on NPM: `npm install coffeescript@next`. See http://coffeescript.org/v2/

So if that's okay with you, please target the `2` branch and revise accordingly.

akfish:
OK. I'll rebase to branch `2`.

akfish:
I merged it into `2`.

* `cake build && cake test` ran without issue. 
* Built with webpack without issue.
* `cake build:browser:full` gave me this wired error. But I should be an issue with closure on my system.
```
f:\Dev\js\coffeescript>cake build:browser:full
f:\Dev\js\coffeescript\node_modules\google-closure-compiler-js\jscomp.js:4529
function FFb(b){var c;try{return b.Xh()}catch(a){a=Xgb(a);if($K(a,70)){c=a;throw Ygb(new eJ(c))}else throw Ygb(a)}}
                                                                           ^

Error: java.lang.RuntimeException: java.lang.RuntimeException: INTERNAL COMPILER ERROR. Please report this problem.  und
efined   Node(ARRAY_PATTERN): Input_0:1785:54       location = 'first_line' in options ? options : ([first_line, first_c
olumn] = this.getLineAndColumnFromChunk((ref = options.offset) != null ? ref : 0), {   Parent(ASSIGN): Input_0:1785:54
     location = 'first_line' in options ? options : ([first_line, first_column] = this.getLineAndColumnFromChunk((ref =
options.offset) != null ? ref : 0), {
```

Command line used to test webpack build is `webpack --entry ./lib/coffeescript/browser.js --output-path dist --output-filename=coffeescript.js`. This eliminated the need of adding a webpack config file.

GeoffreyBooth:
See #4499. There’s a bug with Google Closure Compiler and destructuring. For the 2.0.0-beta1 release I just did `MINIFY=false cake build:browser:full`. The hope is that the Closure Compiler team fixes [their bug](https://github.com/google/closure-compiler-js/issues/59) before the 2.0.0 release.

akfish:
Thanks for the tips. Browser build and test passed.

connec:
Looks like this needs a rebase...

GeoffreyBooth:
I made some minor fixes and brought this up to date, and fixed the target branch.

I took a project that uses Coffeeify (a Browserify plugin) and manually patched Coffeeify and that project’s `node_modules` to use this PR’s code. So Coffeeify pulls in the compiler via `var coffee = require('coffeescript')` and then runs it via `coffee.compile`. (There’s no call to `register`.) Anyway Coffeeify still worked. I gather any other plugin `require`ing CoffeeScript will continue to work in the same way.

I’d like if this had some test coverage. Is there a way to write a test that emulates [what Coffeeify is doing](https://github.com/jnordberg/coffeeify/blob/master/index.js)?

akfish:
If I understand you correctly, I don't think that this approach is a valid test for this PR. Coffeeify runs in node environment and all `require` calls are evaluated dynamically. Also it did not pack coffee-script into browser bundle. Coffeeify is used to pack user code, not the compiler itself. 

To validate this PR, the test should expect that:
* Coffee-script is used as user code (to be built into a browser bundle, not as part of the build tool)
* Coffee-script source is statistically analysed for `require` calls
* Coffee-script is packed into browser bundle without any node.js dependencies

Since static behavior cannot be reliably simulated by nodejs runtime, we have to either a) write a static analyzer, or b) use an existing one or c) just run a webpack/browserify build. All of them involving adding external dependencies (`a` requires a JS compiler to generate AST). We might as well go for the simplest approach, aka `c`, which only involves:
1. `npm install --save-dev webpack`
2. Add a task in `Cakefile` to execute `webpack --entry ./lib/coffeescript/browser.js --output-path dist --output-filename=coffeescript.js`

Or we could document it somewhere (likely in `CONTRIBUTING.md`), asking developers/contributes to run the above webpack command before committing any change, if an extra devDependency is really a problem here.

GeoffreyBooth:
@akfish My goal in testing with Coffeeify wasn’t to test that the intent of this PR was achieved; I’m assuming that it is, since you’re telling me you’ve tested it. My goal was to test that these changes didn’t break the current uses of CoffeeScript as a Node module, such as by Coffeeify. In other words, if Coffeeify or modules like it want to update to CoffeeScript 2 after this PR has been merged in, I want them to work as they do now.

akfish:
I see. In that case, if old tests are not broken, it should be fine. Since all existing test use CoffeeScript as a node module.

GeoffreyBooth:
Okay then I think it’s okay. We probably should include a test like you’re describing, so that this doesn’t get broken in the future. I can easily see someone adding a Node module in the wrong file and breaking Webpack again.

Personally I think adding Webpack as a devDependency and only running it in the context of running the browser tests is fine. @lydell? @jashkenas? Make sure its output doesn’t get checked into the repo, either `.gitignore`d or written to `os.tmpdir()` and deleted after the test completes (on either pass or fail).

timhuff:
Just wanted to say it's incredible to see what my small issue spun off into. Thank you all for the work you put into this project. I use coffeescript for pretty much all my projects and love the language.

GeoffreyBooth:
Please update with the latest `2`, especially since #4523 uses a revamped browser compiler. I assume the browser compiler changes don’t help at all for Webpack, and this PR is still necessary?

Once we have some tests around the CLI, as their own PR so we can see that they pass on `2` as it currently is and then they still pass after this PR, I think this will be ready.

akfish:
Updated. All tests passed.

That would be correct. New browser compiler does not change the situation with webpack.

GeoffreyBooth:
@akfish do you think you’ll have time for the CLI tests soon? I’d love to get this merged in, I’m planning on doing a release soon.

akfish:
Sure. 

I added a `test:webpack` task to `Cakefile` to run the test described above. It should be run after `build` task.

GeoffreyBooth:
I meant the other tests discussed above: that the Node API still works when required in another project, and that the CLI still works for the common commands. These should be submitted in a separate PR, so that we see that they pass against the Node and CLI APIs as they are today, before merging in these changes.

The Node API that needs checking is really just one exported function, `compile`. It just needs to get a string of CoffeeScript code and the test should check that JS and a source map are returned. Avoid issues with strings changing in the future by evaluating the JS, rather than comparing strings; you could send something that has no JS equivalent, like `333 // 25`, and see that the returned JS evaluates to 13.

The CLI API should be checked for `--version`, and that compilation works there too. You could do the same code as above, with `--print`, and capture `stdout` and evaluate it.

Sorry if this sounds like a big ask, but your PR is a big change.

Cakefile
```diff
@@ -434,3 +435,17 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
   (-> eval source).call result
   testResults = runTests result.CoffeeScript
   process.exit 1 unless testResults
+
+task 'test:webpack', 'run webpack test', ->
```
GeoffreyBooth: Rather than this be its own Cake task, can you create a `test/integrations.coffee` file and put this code there? That’s where the other related tests can go, that test if the `coffeescript` module as `require`d by other Node projects still works.

The top of the file should have `return if global.testingBrowser` like you see at the top of `test/repl.coffee`, as obviously a test that calls `spawnNodeProcess` requires Node. You can attach `spawnNodeProcess` to `global` to make it available in the test file.

Cakefile
```diff
@@ -434,3 +435,17 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
...
+  args = [
+    "./node_modules/webpack/bin/webpack.js"
+    "--entry=./lib/coffeescript/browser.js", 
+    "--output-path=#{os.tmpdir()}",
+    "--output-filename=coffeescript.js"
```
GeoffreyBooth: Can Webpack be configured to output to `stdout`, rather than writing a file?

Cakefile
```diff
@@ -434,3 +435,17 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
...
+  ]
+
+  console.log 'webpack', args.slice(1).join(' ')
```
GeoffreyBooth: Please avoid `console.log`s.

Cakefile
```diff
@@ -434,3 +435,17 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
...
+
+  spawnNodeProcess args, 'both', (status) ->
+    console.log "webpack exits with #{status}"
+    process.exit(status)
```
GeoffreyBooth: Please end all files with a newline.

Cakefile
```diff
@@ -434,3 +435,17 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
...
+    "./node_modules/webpack/bin/webpack.js"
+    "--entry=./lib/coffeescript/browser.js", 
+    "--output-path=#{os.tmpdir()}",
+    "--output-filename=coffeescript.js"
+  ]
+
+  console.log 'webpack', args.slice(1).join(' ')
+
+  spawnNodeProcess args, 'both', (status) ->
+    console.log "webpack exits with #{status}"
```
GeoffreyBooth: In the context of the test file, this should be `eq status, 0` (assuming you’re checking for a zero exit status).

GeoffreyBooth:
Forget what I said about the CLI tests. I was thinking of #3946. Sorry about that.

If you could please revise the Webpack test to integrate it, that would be great. It would also be good to have Webpack try to compile something, to verify that compilation works in the non-Node environment.

akfish:
Here are some of my concerns:
1. It should not be considered as a unit test. It's more like using webpack as a code linter to validate against node-only dependencies in browser-only module resolution path.
2. Webpack output is important for debugging should the test fail. It will tell the developer which line is causing the problem. Checking the exit code only is not enough. 
3. Overall it's another build process that generates a build artifact (the build report). It seems logical to me to make it a cake task.

Cakefile
akfish: See https://github.com/webpack/docs/wiki/webpack-for-browserify-users#usage.

> webpack doesn't write to stdout. You need to specify a filename. It can't write to stdout because, unlike browserify, it is ready to generate multiple output files.

akfish:
Current test is equivalent of compiling an entry script containing:
```js
require('coffee-script')
```
Since the code will not be executed, contents other than `require` calls don't really matter to webpack. It is enough to validate webpack support.

Ideally we should bundle browser tests along with CoffeeScript and run the built script in an actual browser environment to test run-time behaviors. But currently the project is not set up for that.

GeoffreyBooth:
> Ideally we should bundle browser tests along with CoffeeScript and run the built script in an actual browser environment to test run-time behaviors.

That’s a good idea. We actually have tests to run in an real browser environment: [test.html](http://coffeescript.org/v2/test.html). These tests don’t run as part of any `cake` task, though—but they could. In another project, I’m using Selenium Chromedriver via Docker to run browser-based tests. We could have a `cake` task that checks if Docker is installed, and if it is, run a Docker command that fires up the Selenium Chromedriver image and loads our test.html, and waits to see if the tests all pass.

The current test suite runs in about 2 seconds on my machine, and the Webpack test runs in about 1 second. Obviously this Docker test would take considerably longer, so it might make sense to have separate testing tasks like:
- `cake test`: same as current
- `cake test:browser`: same as current
- `cake test:slow`: run just the Webpack and Docker tests
- `cake test:full`: run all of the above

I’m not asking you to write the Docker test. But let’s put the Webpack test inside a task named `test:slow`, so that we have a new place for such more-ambitious tests.

As to your point about the Webpack output or build report: in this project, passing tests are silent and failing tests dump copious debugging info. I think the Webpack test should follow that pattern. You can add a comment explaining how to get the Webpack output in all cases (presumably just by running the Webpack command, I assume). That should be enough.

So Webpack is generating this `coffeescript.js` in a temp folder, and this new file is essentially the same as the browser compiler? It can be `require`d by another module, and it exports the `CoffeeScript` object with a `compile` method? Shouldn’t we continue the test to `require` the file that Webpack generates, and make sure it can actually compile some simple CoffeeScript into JavaScript? Or better yet, could we run the test suite against it, the way `test:browser` does?

akfish:
As for the browser test, docker is too heavy for this in my opinion. A popular choice would be [karma](http://karma-runner.github.io). I used it for my past projects. It only requires developers to have browser installed (and who doesn't). The test output is piped to console with proper exit code so it's great for automatic tests. You can run it in Travis CI too. 

Though we will need to write some adapter code (or migrate tests to an existing test framework) for it to work, since this project doesn't use an existing one. The benefit is that we can run tests in actual browser environment, while sharing the same test suites among 3 build targets. 

Apparently browser test setup is another can of worms to be opened and deserves a separate PR.

In fact we already have test coverage for webpack build from `test:browser`. Webpack build and `build:browser` both use `browser.coffee` as the entry point. They are essentially the same. I'll see if it's possible to require built bundle and run it like `test:browser` for now. But we're not completely in the blind if there're no dedicated tests for it.

akfish:
I renamed `test:webpack` to `build:webpack` and added an actual `test:webpack` task. I found a way to load the built bundle and run tests suite against it in node environment, like `test:browser` did. In this way we have the same test coverage for webpack build target as the browser one, without getting into browser test setup.

GeoffreyBooth:
What purpose does `build:webpack` serve? Do we have any reason to create that build other than to use for testing?

akfish:
Logical consistency for one. Both node and browser build target follows the 2 stage build/test process, why not for webpack? It is a build target and should be verified in both stages.
Build and test are not the same. Build verifies static behaviors and test verifies run-time behaviors. They should not be mixed. Otherwise it's like running `build:browser` as a test.
Another reason is the build report. Build task can and should output human-readable report, while tests should be silent failing by this project's convention.

GeoffreyBooth:
I’m trying to avoid creating more `cake` tasks. Each task should do something useful for people. Tasks with no clear purpose just muddy things.

It’s probably best to just have `test:browser` include what it does now, then build and test the Webpack build. The fact that the Webpack build is built in a temp folder and discarded afterward makes clear that its purpose is just for testing.

akfish:
Testing a build target is not the same thing as testing a feature. Build report is not the same as test report and should not be consumed the same way. Discarding the build result doesn't make it less of a build target. And the webpack build result is not discarded: the build report is a build artifact and it  is used.

It is common for projects to have build tasks but not to distribute the build artifacts. Many Linux packages only provide source code and let the user `make` on their own. Some npm packages don't have built js files in repo, only the source. And it's build on `prepublish` or `postinstall` hook. 

A developer who's familiar with webpack would expect it to be treated as a build target. It could be more confusing to see it not to be treated as such, than having two tasks that keeps the logical consistency and can be easily documented. The logic is very clear: **each build target would have `build:*`, `test:*` and `build:*:full` tasks**. It's not muddy.

Speaking of which, maybe I should add a `build:webpack:full` task that invoke `build:webpack` and `test:webpack`, following the convention of the other two build targets. Wouldn't that achieve the "one-command test for webpack" goal while keeping the consistency?

GeoffreyBooth:
A “webpack build” is not a feature of CoffeeScript, and I’m not interested in adding it as one. `cake build` builds a Node-based compiler, and `cake build:browser` builds a compiler that can run in browser environments; and both compilers are intended to be used by people. The compiler created by `build:webpack` is intended only for testing. I don’t want anyone to think otherwise.

Just because a test involves running a build command doesn’t mean it’s not still a test.

akfish:
It's not a feature, it's a build target. Anyone familiar with webpack will not even come close to make the mistake you are thinking of. 

Anyway, enough with the terminology stuffs, as long as the webpack build command is not run as a unit test and the build report is not silent, I'm happy with it. If that's acceptable for you, I will make the modifications so that there's only one cake task `test:webpack` that first builds a bundle then run the test suite against it.

jashkenas:
Apologies. Looks like I should have helped put this ticket out of its misery long ago...

CoffeeScript already works just fine with Webpack, and should not be restructured like this. I just walked through webpack's tutorial, using the CoffeeScript "browser" compiler, and it works perfectly.

Using the latest version of CoffeeScript, you build the browser compiler:

`cake build:browser`

Copy it into your webpack project, and then use it:

```javascript
import CoffeeScript from './coffee-script.js';

function component () {
  var element = document.createElement('div');

  element.innerHTML = CoffeeScript.compile('a = 1 + 1');

  return element;
}

document.body.appendChild(component());
```

Index.html:

```html
<html>
  <head>
    <title>webpack 2 demo</title>
  </head>
  <body>
    <script src="dist/bundle.js"></script>
  </body>
</html>
```

Then you run webpack:

    > ~/Desktop/webpack-demo webpack app/index.js dist/bundle.js

And it works:

image

@GeoffreyBooth -- if there's a change we need to make here, it's that we should be shipping the "browser" build file in a reasonable location as part of the npm install — you shouldn't have to build it yourself. Perhaps in a `dist` folder?

akfish:
It works is not the same as it works "fine". I am aware of the setup you provided above. It shows how webpack supports a legacy project like this, not the other way around. The fact that the webpack support issue has been raised over and over by multiple people on multiple occasions apparently means nothing. Because people who use webpack on daily basis are neither too lazy to read the basic document, nor too stupid to know the difference. We should know better. Thanks for your time.

Radivarig:
....unbelievable

GeoffreyBooth:
Yes I think what @akfish is getting at is that you can’t just do `CoffeeScript = require('coffeescript')` in a Webpack or Browserify project and have it work. And not only that, there’s nothing like `require('coffeescript/browser')` that works. The browser build is not distributed over NPM, so there’s no way to `require` it.

Currently the browser build is treated as just part of the documentation, but clearly that needs to change. I think it deserves enough prominence that is included in the NPM module, and `require`-able on its own.

What @akfish was accomplishing with this PR, which we might want to still consider, is that this PR enables people to do `require('coffeescript')` in Webpack/Browserify _and it just works,_ the same as if they had typed that in Node. They wouldn’t need to specify `require('coffeescript/browser')` or similar. (@akfish, please correct me if I’m wrong about this.)

Maybe that’s not enough of an upside to be worth the changes contemplated by this PR, but as far as I can tell I don’t see any downside to these changes. It’s just a lot of lines of code moving around, but no functionality or clarity lost as a result. I guess if it provides a bit of convenience for Webpack/Browserify users, especially when it comes to time saved as they can avoid needing to troubleshoot errors caused by `require('coffeescript')` not working as expected until they discover they need to do `require('coffeescript/browser')` instead, why _wouldn’t_ we make these changes?

jashkenas:
Looking through the diff a little more, that sounds fair enough — as long as we aren't contorting the file structure in any webpack-specific way. Go for it. I'm sorry if I offended.

But we should still be offering the basic "coffeescript/browser" file as a general part of our packaging — so that folks who want to load it under a script loader of any stripe, or just vanilla JS, can get at it.

carlsmith:
Clientside CoffeeScript is useful in some circumstances, even if it's normally a mistake to use it that way. If you're using CoffeeScript in the browser for the right reasons, then you may not even have Node set up in your development environment. This PR is more to do with how the compiler is modularised for browsers, but having first-class support for clientside CoffeeScript is worthwhile.

GeoffreyBooth:
Agreed. @akfish, is my explanation of this PR’s value correct? Leaving aside the discussion of tests.

GeoffreyBooth:
@jashkenas can you please reopen?

connec:
Looks like the repo/branch has been removed so the PR can't be reopened.

akfish:
Sorry for that. I should not have deleted the branch before the decision is final. I understand that the compiler usually don't mingle with webpack in this specific way thus there are some difficulties in communication. @GeoffreyBooth have put a lot of efforts into this and from his last comments I can see that we are on the same page. 
I'll resend a PR.

PR #4508, opened 2017-04-16
Replaces meta elements' content key-value pair separator ; -> ,

lexjacobs:
Chrome 57.0 returns console error:
`Error parsing a meta element's content: ';' is not a valid key-value pair separator. Please use ',' instead.`
This PR fixes the syntax to resolve this error.


GeoffreyBooth:
All those files are auto-generated by [Docco](https://github.com/jashkenas/docco). You need to fix it there, or else the semicolons will keep getting reintroduced every time the annotated source is regenerated.

lexjacobs:
Thanks. I'll close this and investigate at Docco.

PR #4520, opened 2017-04-20
[CS2] Unicode code point escapes on 2, fixes #4248

### LINKED ISSUE #4248: Unicode code point escapes

helixbass:
Fixes #4248: contains the Unicode code point escape handling specific to `2` branch (as [suggested](https://github.com/jashkenas/coffeescript/pull/4498#issuecomment-293797878) by @lydell)

This currently also includes some commits from `master` (including my code point escape stuff targeted against `master` ie #4498) that haven't been merged into `2` yet. If it's preferable to separate that out, I can open a pull request for my `2_merged_master` branch (against `2` branch), which just contains `master` merged into `2`

But otherwise [here](https://github.com/helixbass/copheescript/compare/2_merged_master...helixbass:iss4248_unicode_code_point_escapes_on_2) is the diff of just my `2`-specific changes. Basically we just have to pass around the (he)regex `flags` in order to be able to distinguish non-`u`-regexes (whose code point escapes *should* be rewritten to normal Unicode escapes) from strings/`u`-regexes (whose code point escapes *should not* be rewritten)


helixbass:
@lydell sorry I wrongly assumed that once `master` was merged into `2` the diff would clean up, I re-merged `2` and now it's correct

src/lexer.coffee
```diff
@@ -807,8 +810,9 @@ exports.Lexer = class Lexer
     low = (codePoint - 0x10000) % 0x400 + 0xDC00
     "#{toUnicodeEscape(high)}#{toUnicodeEscape(low)}"
 
-  # Replace \u{...} with \uxxxx[\uxxxx] in strings and regexes
+  # Replace \u{...} with \uxxxx[\uxxxx] in regexes without `u` flag
   replaceUnicodeCodePointEscapes: (str, options) ->
+    shouldReplace = options.flags? and options.flags.indexOf('u') is -1
```
lydell: `'u' not in options.flags`?

helixbass: @lydell as currently written, `options` won't always contain a `flags` key (only does if this is for a regex, in which case `flags` will be set to the regex flags string) and `not in` seems to break with an `undefined` arg. Does seem like there should be a simpler way to write this though - I guess you could do something like `options.flags ?= 'u'` before `'u' not in options.flags` to handle the string case but that seems confusing semantically. Or could refactor to pass in eg `dontReplace` option instead of `flags` but it seemed very clean to just pass around existing `flags` and then have the logic for interpreting it with regards to code point escape replacement be contained in this method

lydell: I meant only replacing the right side of the `and`.

helixbass: @lydell ok yup that reads well, updated

@GeoffreyBooth [requested](https://github.com/jashkenas/coffeescript/pull/4498/files#r112801329) that I refactor `toJS()` into a shared test helper so I opened a pull request #4522 for that (against `master`). Perhaps wait to merge this one until we've gone through the cycle of merging that into `master`, merging `master` into `2`, and re-merging `2` here (or it may not matter if that refactor doesn't cause any merge conflicts here)?

lydell:
Unfortunately, there are merge conflicts now :(

helixbass:
@lydell ok I re-merged `2` and resolved that merge conflict. `master` still hasn't been merged into `2` since I refactored `toJS()` to a shared helper via #4522 - it doesn't look like that will cause a merge conflict with this so I guess there's no need to get that merged into `2` before merging this?

PR #4525, opened 2017-04-23
[CS2] Fix non-bare `get` `set` when called with `@` argument

### LINKED ISSUE #4524: [CS2] `get` as method should not require parentheses
This should not throw an error:

    location = App.locationCollection.get @model.get('location')
                                      ^^^
    ParseError: 'get' cannot be used as a keyword, or as a function call without parentheses

We should find a way to detect when `get` is used as `.get`, rather than bare `get`; and restrict the error to only the latter.

GeoffreyBooth:
Fixes #4524.

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -135,7 +135,7 @@
           this.error(`'${prev[1]}' cannot be used as a keyword, or as a function call without parentheses`, prev[2]);
         } else {
           prevprev = this.tokens[this.tokens.length - 2];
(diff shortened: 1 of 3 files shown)


vendethiel:
I think we could use three more tests (I don't think they're present yet):

```
get 5 # get without parentheses in an unambiguous
get {a: 5} # same
@get 5 # with `@`
```

Nicely done.

GeoffreyBooth:
@vendethiel I’ve added some more tests. How does it look?

vendethiel:
That looks good :).

GeoffreyBooth:
Thanks!

PR #4526, opened 2017-04-23
[CS2] Remove unnecessary utility helper functions

GeoffreyBooth:
Inspired by @connec’s [comment](https://github.com/coffeescript6/discuss/issues/65#issuecomment-267344468):

- [x] Remove polyfill for `indexOf`, leaving the shortcut to the native function.
- [x] Remove utility function for `bind` entirely, replacing its one usage with a call to the native `bind`.
- [x] Unrelated but I slipped it in: updated the modules tests that were waiting on `class` to be supported

The one outlier is the utility function for `extends`, which remains. Now that we compile to native `class` and `extends`, there’s only one very narrow case where this helper still gets used:

```coffee
    o 'SimpleAssignable EXTENDS Expression',    -> new Extends $1, $3
```

which pairs with:

```coffee
#### Extends

# Node to extend an object's prototype with an ancestor object.
# After `goog.inherits` from the
# [Closure Library](https://github.com/google/closure-library/blob/master/closure/goog/base.js).
exports.Extends = class Extends extends Base
  constructor: (@child, @parent) ->
    super()

  children: ['child', 'parent']

  # Hooks one constructor into another's prototype chain.
  compileToFragments: (o) ->
    new Call(new Value(new Literal utility 'extend', o), [@child, @parent]).compileToFragments o
```

We have zero tests covering this. I suppose the hundreds of `class` tests used to cover it in the 1.x codebase, but now that this only applies to straight prototypal inheritance we don’t have any tests that refer to that as opposed to classes. This is [mentioned](http://coffeescript.org/v2/#prototypal-inheritance) in the docs, but there’s no example, just a single half-sentence: “The `extends` operator can be used to create an inheritance chain between any pair of constructor functions”.

Anyway for the purposes of this PR we can just leave it as is. Later on though we should probably decide whether or not we want to keep it (might as well, I suppose) and if we do keep it, it should get some tests of its own and probably an example in the docs.

**EDIT**

- [x] Removed `extends`.
- [x] Added helper for `splice` just like `slice`.


connec:
Neat!

On `extends` - I think we should seriously consider removing it. It doesn't make much sense as a language feature any more, especially since it behaves differently from `class extends`, and there's no ES equivalent for a standalone `extends`.

GeoffreyBooth:
The place I see it being used is with frameworks. Consider Backbone, with lines like `class AppView extends View` where `View` is just a vanilla function. Wouldn't we still need this old `extends` in cases like that?

src/nodes.coffee
```diff
@@ -3118,7 +3118,7 @@ UTILITIES =
 
   # Correctly set up a prototype chain for inheritance, including a reference
   # to the superclass for `super()` calls, and copies of any static properties.
-  extend: (o) -> "
+  extend: (o) -> """
```
lydell: This will make the helper function take several lines in the output JS. Is this change intended?

GeoffreyBooth: Oh, no, I was just figuring that multiline strings should always be block-quoted as a matter of style.

jashkenas:
> Neat!
> 
> On extends - I think we should seriously consider removing it. It doesn't make much sense as a language feature any more, especially since it behaves differently from class extends, and there's no ES equivalent for a standalone extends.

I agree. I think it should be removed for CS2. 

The reason for its existence was CS1's "classes via JavaScript prototypes" helpers.

GeoffreyBooth:
@jashkenas you would be the one to confirm: does removing `extends` pose any issues for compatibility with Backbone? I guess not, since `class AppView extends View` is an ES class extending a function, which I guess ES classes can do?

jashkenas:
Even if it did — that shouldn't be a consideration that CoffeeScript needs to worry about.

But it doesn't: http://backbonejs.org/#Model-extend

GeoffreyBooth:
Okay, I’m convinced.

- [x] Removed `extends`.
- [x] Added helper for `splice` just like `slice`.

Do these shortcuts, like `slice` for `[].slice`, really make a speed difference? I guess they help in minification, though I question the value of minification nowadays with gzip compression enabled.

GeoffreyBooth:
Regardless of whether we should remove the shortcuts, does this PR look okay?

PR #4527, opened 2017-04-23
[CS2] JavaScript keywords used as properties no longer require quoting

### LINKED ISSUE #4513: [CS2] ES5+ don't require keyword property names to be escaped
In other words, compiling `object.class` to `object["class"]` and `class: "foo"` to `{"class": "foo"}` is no longer necessary, and they're also idiomatically unquoted in modern JS.

So could this redundancy be removed? It aids both in readability and emitted code size.

GeoffreyBooth:
Closes #4513.

### DIFF
--- a/lib/coffeescript/browser.js
+++ b/lib/coffeescript/browser.js
@@ -9,7 +9,7 @@
 
   compile = CoffeeScript.compile;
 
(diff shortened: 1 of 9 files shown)


jashkenas:
LGTM.

Issue #4512, opened 2017-04-18
Proposal: Deriving a function from an instance method that is bound/closed-over the instance  –  instance:method


blitmap:
Hello -

I was hoping to propose a syntax that would allow you to create a function with a bound/closed-over "this".

```coffeescript
instance.method
instance:method
instance:method(...)
```

Would be:
```javascript
class::method
class::method.bind(instance)
class::method.bind(instance, ...)
```

Please forgive me if I got something wrong here... I usually wind up submitting proposals when I'm too tired and there are often typos.

I believe Babel has something like this called the Bind operator:

https://babeljs.io/docs/plugins/transform-function-bind/

@vendethiel: You closed my other issue from yesterday but I corrected some typos :(  Please go back and have another look?

vendethiel:
FTR, this feature [is available in Coco/LS](https://github.com/satyr/coco/wiki/additions#binding-access-).

blitmap:
@vendethiel, respectfully, I use Coffeescript because I think it strikes the right balance between being concise and expressive... :>

Looks like this is related to #3634

```coffeescript
@@method # may be a bit clearer in intent
```

vendethiel:
> @vendethiel, respectfully, I use Coffeescript because I think it strikes the right balance between being concise and expressive... :>

I just wanted to point out a case in the wild, and your issue link is great for that.

---


> Looks like this is related to #3634

Well, not completly. The suggestion wants `@@method` as a shortcut to `=> @method ...`. Your suggestion here would cover more use cases, like `console~bind`, which can't be done with `@@`.

blitmap:
Using it within the definition of a class would look like this:

```coffeescript
class Handler
   constructor: (node) ->
     $(node)
       .on('click', @:on_click)
       .on("mouseover", @:on_mouserover)

   on_click: -> #...
   on_mouseover: -> # ...
```

This is how it would be used confusingly:

```coffeescript
instance:a.b.d
```
```javascript
instance.a.b.d.bind(instance)
```
What's on the left of the `:` is the bind target.
```coffeescript
instance.a.b:d
```
```javascript
instance.a.b.d.bind(instance.a.b)
```
And then from the global context...
```coffeescript
:console.log
```
```javascript
window.console.log.bind(window)
```

The special cases are having `@` or (nothing) on the left of `:`, which get replaced by `this` or the global/`window`.

GeoffreyBooth:
[...]

Issue #3093, opened 2013-07-29
Function stubs?


00dani:
In MoonScript, it's possible to get a reference to an object method that remains bound to its original object. This feature is called a "function stub" and works like this:

``` lua
higherOrderFunction obj\method
-- compiles to
return higherOrderFunction((function()
  local _base_0 = obj
  local _fn_0 = _base_0.method
  return function(...)
    return _fn_0(_base_0, ...)
  end
end)())
```

Given the way JavaScript's objects work and that `Function::bind` isn't reliably available on all browsers, a similar functionality in CoffeeScript would be valuable.

We don't use `\` to denote method calls in CoffeeScript, so simply copying MoonScript's notation for the concept wouldn't work so well; perhaps reusing the bound-function symbol `=>` in the contexts we use `.` would, like so?

``` coffeescript
higherOrderFunction obj=>method
```

Would compile to:

``` javascript
higherOrderFunction(function () {
  return obj.method.apply(obj, arguments);
});
```

The above CoffeeScript is currently a syntax error, so adding this use for `=>` shouldn't break backwards-compat with anything. I'm open to other operator suggestions, of course.

Issue #4497, opened 2017-04-12
[CS2] Bound methods are unsafe to use with inheritance


lluchs:
Bound methods are now only bound after calling the super constructor. If the super constructor calls any methods in the class, you end up with unbound references to functions.

Example:

```coffee
class SomeView extends Backbone.View
  initialize: ->
    # Can't use @callback here - initialize is called from the super constructor
  someFunc: ->
    # It's *probably* safe to use @callback here - but it would suddenly get unsafe 
    # if we (or the super class or a derived class) decided to call @someFunc() in 
    # initialize later on.
  callback: => # ...
```

While it's not possible to bind the functions before calling the super constructor, there are some ways CS2 could make bound methods "safer" to use. Some ideas:

 - In classes with inheritance, always bind references to bound methods where they're used. In the example above, `@callback` would always compile to `this.callback.bind(this)`. While this solves the issue for local uses, it doesn't help for external access.
 - Don't allow bound methods in classes with inheritance. Safest option, but not great.
 - Rename the bound method and only set the proper name when binding `this`. Doesn't solve the issue, but would make errors like in the example above more visible. Would need additional thought to avoid naming conflicts and for interaction with non-CoffeeScript classes in the inheritance chain.

Related issue #4239 still stands, but now it's also an issue for the base class.

connec:
I'd like to throw in a 4th option:

- Remove bound methods entirely.

The use-case for them is pretty small, and the behaviour is very implicit and has a lot of caveats. Explicitly binding methods at call time or in a constructor makes things much easier to follow, and lets users choose their own tradeoffs.

GeoffreyBooth:
If they're not supported in ES classes, I vote for removing them.

connec:
> If they're not supported in ES classes

They're sort of supported through public class properties, [for example](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=stage-2&targets=&browsers=&builtIns=false&code=class%20Foo%20%7B%0A%20%20f%20%3D%20()%20%3D%3E%20this%0A%7D&experimental=false&loose=false&spec=false&playground=true).

In general such properties cannot currently be represented in CoffeeScript due to the executable class body:

```coffee
class Foo
  a = 1  # this is a plain assignment in the class body scope, not a public property initialiser
  a: 1   # this is a prototype assignment, not a public initialiser
  @a = 1 # this is an assignment to the class, equivalent to a public static property
```

[...]

PR #4530, opened 2017-04-26
[CS2] Remove support for bound instance methods

connec:
Bound methods are implemented as assignments to `this` in the constructor. In derived classes, where `this` is unavailable until after `super` has been called, the binding is applied and assigned after the `super` call. This means that any references to ‘bound’ methods reachable from the parent constructor will actually point to the unbound prototype methods.

This can lead to very subtle bugs where a method that is thought to be bound is handed off and later called with an incorrect context, and the only defence is for users to be vigilant about referencing bound methods in constructors.

___

This is probably the most extreme solution to the problem, and would leave heavy users of bound methods with an unpleasant refactoring task, however the alternatives aren’t terribly appetising either:

- Restrict the error to derived classes only. A common use of bound methods is for event handlers in frontend frameworks such as Backbone and React, both of which rely on extending base classes. Choosing this approach would not help these users, and would make bound instance methods even more of a fringe feature.

- Hoist the entire method definition into the constructor, ensuring it isn’t assigned at all until `super` has been called. This would at least make any bugs easier to spot, and also looks very similar to the [class public fields ECMA Stage 2 proposal](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=stage-2&targets=&browsers=&builtIns=false&debug=false&code=class%20A%20%7B%0A%20%20bound%20%3D%20()%20%3D%3E%20%7B%7D%0A%7D&experimental=false&loose=false&spec=false&playground=false). Matching some ES equivalent is probably the best strategy long term, but as this proposal is only Stage 2 we can’t rely on it not being changed.

- Leave everything as it is. We would could simply document the caveats of bound method interactions with `super` and let everyone figure it out.

*Edit:* #4512, #3093, and others propose a binding property access (as - @vendethiel would be quick to add - is [available Coco/LS](https://github.com/satyr/coco/wiki/additions#binding-access-)). This would make the necessary refactoring for this or option 1 slightly easier, and is also an [ES proposal](https://github.com/tc39/proposal-bind-operator).

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -1741,16 +1741,13 @@
 
       compileClassDeclaration(o) {
         var ref1, result;
-        if (this.externalCtor || this.boundMethods.length) {
+        if (this.externalCtor) {
           if (this.ctor == null) {
             this.ctor = this.makeDefaultConstructor();
           }
         }
         if ((ref1 = this.ctor) != null) {
           ref1.noReturn = true;
-        }
-        if (this.boundMethods.length) {
-          this.proxyBoundMethods(o);
         }
         o.indent += TAB;
         result = [];
@@ -1799,7 +1796,6 @@
       walkBody() {
         var assign, end, executableBody, expression, expressions, exprs, i, initializer, initializerExpression, j, k, len1, len2, method, properties, pushSlice, ref1, start;
         this.ctor = null;
-        this.boundMethods = [];
         executableBody = null;
         initializer = [];
         ({expressions} = this.body);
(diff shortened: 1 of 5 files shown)


GeoffreyBooth:
This looks good to me. @lydell, did you have any feedback? I assume you agree with this solution to #4497?

lydell:
I think this looks good to me, too.

xixixao:
Wait, shouldn't this compile to a property initializer?

```coffee
class A
  b: => this.x
```

```es6
class A {
  b = () => this.x;
}
```
There is no way to get the above output or any equivalent now?

xixixao:
@connec @lydell

connec:
@xixixao property initializers are just a Stage 2 proposal, so there is no support for them currently.

The equivalent would be:

```coffee
class A
  constructor: ->
    @b = => @x
```

PR #4531, opened 2017-04-27
[CS2] Named functions

### LINKED ISSUE #15: Named Functions
For debugging purposes, it would be better if CoffeeScript generated only named functions, ie: 

```
square: x => x * x.
```

Compiles into:

```
var square = function square(x) { return x * x; };
```

Questions:

Are there any namespace problems with this, if function names in different scopes collide? If so, what are the consequences (debuggers getting confused, perhaps)? Is it safe to take the final portion of the id as the function name when defining a function on a nested object?

GeoffreyBooth:
Closes #15, the oldest open issue (from 2009!). With this PR, this:

```coffee
square = (x) -> x * x
```

now compiles to this:

```js
var square;

square = function square(x) {
  return x * x;
};
```

Note the `function square(x)`, instead of just `function(x)`. This mimics [how Babel converts methods into ES5 functions](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=env&targets=&browsers=&builtIns=false&debug=false&code=var%20obj%20%3D%20%7B%0A%20%20square%20(x)%20%7B%0A%20%20%20%20return%20x%20*%20x%3B%0A%20%20%7D%0A%7D).

The name is _not_ output for bound functions, which per ES spec must be anonymous; or for any function name that contains a period (`helpers.eq = ->`), bracket (`method[name] = ->`) or quotes (`'not': ->`). Those cases just stay anonymous like they are now.

The only breaking change caused by this is when a function returns a reference to itself, as illustrated by the test that needed updating:

```coffee
  changeMe = ->
    changeMe = 2

  changeMe() # This line was changed to `changeMe = changeMe()`
  eq changeMe, 2
```

Since the first line now compiles to `changeMe = function changeMe() {`, the inner `return changeMe = 2` no longer overrides the variable name `changeMe` one level up. Hopefully this is a rare enough edge case that it shouldn’t come up too often?

And there’s the question of whether this is doing at all at this point, now that debuggers can often infer function names from the names of the variables they’re assigned to. My instinct is that it is still worth doing, since it _does_ add some clarity to the output (presumably that’s by Babel does it) and there are probably some debugging tools that are helped by this.

### DIFF
--- a/lib/coffeescript/browser.js
+++ b/lib/coffeescript/browser.js
@@ -27,7 +27,7 @@
   }
 
   if ((typeof btoa !== "undefined" && btoa !== null) && (typeof JSON !== "undefined" && JSON !== null)) {
-    compile = function(code, options = {}) {
+    compile = function compile(code, options = {}) {
       options.inlineMap = true;
       return CoffeeScript.compile(code, options);
     };
@@ -60,7 +60,7 @@
     return xhr.send(null);
   };
 
-  runScripts = function() {
+  runScripts = function runScripts() {
     var coffees, coffeetypes, execute, fn, i, index, j, len, s, script, scripts;
     scripts = window.document.getElementsByTagName('script');
     coffeetypes = ['text/coffeescript', 'text/literate-coffeescript'];
@@ -76,7 +76,7 @@
       return results;
     })();
     index = 0;
-    execute = function() {
+    execute = function execute() {
       var param;
       param = coffees[index];
       if (param instanceof Array) {
@@ -85,7 +85,7 @@
         return execute();
       }
     };
-    fn = function(script, i) {
+    fn = function fn(script, i) {
       var options, source;
       options = {
         literate: script.type === coffeetypes[1]
--- a/lib/coffeescript/cake.js
+++ b/lib/coffeescript/cake.js
@@ -23,16 +23,16 @@
   oparse = null;
 
   helpers.extend(global, {
-    task: function(name, description, action) {
+    task: function task(name, description, action) {
       if (!action) {
         [action, description] = [description, action];
       }
       return tasks[name] = {name, description, action};
     },
-    option: function(letter, flag, description) {
+    option: function option(letter, flag, description) {
       return switches.push([letter, flag, description]);
     },
-    invoke: function(name) {
+    invoke: function invoke(name) {
       if (!tasks[name]) {
         missingTask(name);
       }
@@ -67,7 +67,7 @@
     return results;
   };
 
-  printTasks = function() {
+  printTasks = function printTasks() {
     var cakefilePath, desc, name, relative, spaces, task;
     relative = path.relative || path.resolve;
     cakefilePath = path.join(relative(__originalDirname, process.cwd()), 'Cakefile');
@@ -84,17 +84,17 @@
     }
   };
 
-  fatalError = function(message) {
+  fatalError = function fatalError(message) {
     console.error(message + '\n');
     console.log('To see a list of all tasks/options, run "cake"');
     return process.exit(1);
   };
 
-  missingTask = function(task) {
+  missingTask = function missingTask(task) {
     return fatalError(`No such task: ${task}`);
   };
 
-  cakefileDirectory = function(dir) {
+  cakefileDirectory = function cakefileDirectory(dir) {
     var parent;
     if (fs.existsSync(path.join(dir, 'Cakefile'))) {
       return dir;
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -25,7 +25,7 @@
 
   exports.helpers = helpers;
 
-  base64encode = function(src) {
+  base64encode = function base64encode(src) {
     switch (false) {
       case typeof Buffer !== 'function':
         return Buffer.from(src).toString('base64');
@@ -38,7 +38,7 @@
     }
   };
 
-  withPrettyErrors = function(fn) {
+  withPrettyErrors = function withPrettyErrors(fn) {
     return function(code, options = {}) {
       var err;
       try {
@@ -198,7 +198,7 @@
       if (!(sandbox !== global || sandbox.module || sandbox.require)) {
         Module = require('module');
         sandbox.module = _module = new Module(options.modulename || 'eval');
-        sandbox.require = _require = function(path) {
+        sandbox.require = _require = function _require(path) {
           return Module._load(path, _module, true);
         };
         _module.filename = sandbox.__filename;
@@ -236,7 +236,7 @@
 
   if (require.extensions) {
     ref = this.FILE_EXTENSIONS;
-    fn1 = function(ext) {
+    fn1 = function fn1(ext) {
       var base;
       return (base = require.extensions)[ext] != null ? base[ext] : base[ext] = function() {
         throw new Error(`Use CoffeeScript.register() or require the coffeescript/register module to require ${ext} files.`);
@@ -270,7 +270,7 @@
   lexer = new Lexer;
 
   parser.lexer = {
-    lex: function() {
+    lex: function lex() {
       var tag, token;
       token = parser.tokens[this.pos++];
       if (token) {
@@ -282,11 +282,11 @@
       }
       return tag;
     },
-    setInput: function(tokens) {
+    setInput: function setInput(tokens) {
       parser.tokens = tokens;
       return this.pos = 0;
     },
-    upcomingInput: function() {
+    upcomingInput: function upcomingInput() {
       return "";
     }
   };
@@ -312,7 +312,7 @@
     return helpers.throwSyntaxError(`unexpected ${errorText}`, errorLoc);
   };
 
-  formatSourcePosition = function(frame, getSourceMapping) {
+  formatSourcePosition = function formatSourcePosition(frame, getSourceMapping) {
     var as, column, fileLocation, filename, functionName, isConstructor, isMethodCall, line, methodName, source, tp, typeName;
     filename = void 0;
     fileLocation = '';
@@ -360,7 +360,7 @@
     }
   };
 
-  getSourceMap = function(filename) {
+  getSourceMap = function getSourceMap(filename) {
     var answer;
     if (sourceMaps[filename] != null) {
       return sourceMaps[filename];
@@ -380,7 +380,7 @@
 
   Error.prepareStackTrace = function(err, stack) {
     var frame, frames, getSourceMapping;
-    getSourceMapping = function(filename, line, column) {
+    getSourceMapping = function getSourceMapping(filename, line, column) {
       var answer, sourceMap;
       sourceMap = getSourceMap(filename);
       if (sourceMap != null) {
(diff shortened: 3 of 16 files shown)


vendethiel:
> Since the first line now compiles to changeMe = function changeMe() {, the inner return changeMe = 2 no longer overrides the variable name changeMe one level up. Ho

Is there a reason we don't just compile it to `function changeMe()` then?

```
> function fn() { fn = 5; }
> fn()
undefined
> fn
5
```

vs

```
> var fn = function fn() { fn = 5; }
> fn()
undefined
> fn
function fn() { fn = 5; }
```

jashkenas:
To note — this is the same way that CoffeeScript briefly compiled functions (probably in 2010).

We removed it because it broke some versions of Internet Explorer at the time: the function declaration would clobber variables in places that it shouldn't have.

I'd assume that versions of IE currently in use don't have that problem anymore.

GeoffreyBooth:
@vendethiel Compiling to function declarations (your suggestion) as opposed to function expressions (current behavior) changes the scope of the functions. It's also not true to the input. How would we compile `a = b = ->`?

vendethiel:
Oh, I completly forgot `var fn = function fn(){};` kept `var` hoisting without function hoisting.

lydell:
With this PR, the following would cause an infinite loop, wouldn’t it?

```coffee
foo = -> console.log('foo')

obj =
  foo: -> foo()

obj.foo()
```

Babel [renames the `foo` variable](http://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015%2Creact%2Cstage-2%2Cstage-3&targets=&browsers=&builtIns=false&debug=false&code=const%20foo%20%3D%20()%20%3D%3E%20console.log(%27foo%27)%0A%0Aconst%20obj%20%3D%20%7B%0A%20%20foo%3A%20()%20%3D%3E%20foo()%2C%0A%7D%0A%0Aobj.foo()&experimental=false&loose=false&spec=false&playground=true) to avoid an infinite loop.

The ES2015 specification made the `.name` property on functions more useful. That’s why Babel adds functions names: It tries to add this ES2015 feature to ES5 runtimes. See http://2ality.com/2015/09/function-names-es6.html for more information on `.name` in ES2015.

CoffeeScript emits ES2015 code, so I don’t think it makes sense to do this in CoffeeScript.

GeoffreyBooth:
> CoffeeScript emits ES2015 code, so I don’t think it makes sense to do this in CoffeeScript.

Which part, the function renaming or this PR in general?

lydell:
This PR in general.

jashkenas:
Yeah, it seems like for modern JS, we really don't need to add this:

> Variable declarations and assignments  #

> Functions pick up names if they are created via variable declarations:

    let func1 = function () {};
    console.log(func1.name); // func1

    const func2 = function () {};
    console.log(func2.name); // func2

    var func3 = function () {};
    console.log(func3.name); // func3

> But even with a normal assignment, name is set up properly:

    let func4;
    func4 = function () {};
    console.log(func4.name); // func4

    var func5;
    func5 = function () {};
    console.log(func5.name); // func5

> With regard to names, arrow functions are like anonymous function expressions:

    const func = () => {};
    console.log(func.name); // func

GeoffreyBooth:
Well based on the recent comments in #15, “modern JS” is only V8—Node and Chrome. Apparently debugging in Firefox and other runtimes is still aided by having the names.

That said, presumably they’ll catch up, so maybe it’s not worth introducing a breaking change (however minor) for this. I don’t feel strongly one way or the other. At least we’ll be able to definitively close that issue and settle the question about whether we will ever support named functions.

@jashkenas and others, do you agree with @lydell? Or is there some reason to pursue this?

jashkenas:
I agree with @lydell. Named functions have always been kind of hacky — subject to cross-browser incompatibilities, etc.

With better support for regular variables in debugging, I don't think we need to do this.

GeoffreyBooth:
Okay, so if we’re not going to add this, then there’s no reason to keep #15 open. It’s settled.

Can someone write a few sentences explaining why CoffeeScript deliberately lacks named functions? So I can add it to the docs near where we explain why we don’t support `let` or `const` (or `var`).

connec:
[The MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) is probably a good reference point - the summary would be that:

CoffeeScript relies on the function naming capabilities of the runtime. ES2015 includes rules for inferring function names from assignments. If you wish to give a name to a function in an expression, simply assign it and the name will be inferred by the runtime.

[Example](http://coffeescript.org/v2/#try:logName%20%3D%20(f)%20-%3E%20console.log%20f.name%0A%0AlogName%20-%3E%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%23%20''%0AlogName%20foo%20%3D%20-%3E%20%20%20%20%20%20%20%20%20%20%20%20%20%20%23%20'foo'%0AlogName%20foo.bind%20this%20%20%20%20%20%20%20%20%20%23%20'bound%20foo'%0AlogName%20%7B%20method%3A%20-%3E%20%7D.method%20%23%20'foo')
```coffee
logName = (f) -> console.log f.name

logName ->                    # ''
logName foo = ->              # 'foo'
logName foo.bind this         # 'bound foo'
logName { method: -> }.method # 'method'
```

Bear in mind that function names may change if using a name-mangling minifier.

GeoffreyBooth:
@connec That’s a good explanation for why users don’t need named functions, but it doesn’t answer the question of why we made the design decision to not support them in the first place. What would be the answer to the reasonable question “Why can’t I get CoffeeScript to somehow output `function foo` instead of `foo = function`?”

I suppose the answer could be “because you don’t need them,” but that’s not very satisfying. Another answer could be that CoffeeScript simply doesn’t have declarations of variables, and ultimately `function foo` is a function declaration just like `var foo` is a variable declaration, and the reason we don’t have function declarations is the same as the reason we don’t have variable declarations. Is this closer to the answer?

carlsmith:
You don't need to explain *why* CoffeeScript doesn't have them. It's on those that do want them to explain why. You just need to state that they're redundant.

GeoffreyBooth:
@carlsmith It’s for this section in the docs: http://coffeescript.org/v2/#unsupported. I think it’s nice to explain our design decisions.

carlsmith:
Sorry. It seemed like you were retrofitting a rational without really meaning to.

I would just say that named functions were implemented in an early version of CoffeeScript, as the feature was often requested, but it created problems in IE, so was quickly removed again. It has been asked for many times since, but to date, no one has provided an actual use case. There are benefits to functions being referred to by name in stack traces, but browsers already do a good job of identifying anonymous functions when they are assigned a name.

Until a compelling case for named functions is provided, they will remain unsupported, no matter how many times they are asked for :)

GeoffreyBooth:
Archived this branch at https://github.com/GeoffreyBooth/coffeescript-archived-branches/tree/named-functions

dotnetCarpenter:
I'm gonna chime in with the same example as I did in #15.

I am/was cleaning up [an old coffeescript project](https://github.com/geejs/gulp-tap/) and came across `arguments.callee`, because coffeescript does not have support for named functions. Obviously it's desirable  to run in `'use strict '` mode, which can not be done with `arguments.callee`.

```coffee
if obj instanceof baseStream && !obj._readableState.ended
  obj.on('end', next)
  obj.on('data', ->
    obj.removeListener('end', next)
    obj.removeListener('data', arguments.callee)
    next()
  )
else
  next()
```

In es5 you would do:

```js
if (obj instanceof baseStream && !obj._readableState.ended) {
  obj.on('end', next)
  obj.on('data', function data() {
    obj.removeListener('end', next)
    obj.removeListener('data', data)
    next()
  })
} else
  next()
```

**If coffeescript had support for named functions**, you could do:

```coffee
if obj instanceof baseStream && !obj._readableState.ended
  obj.on('end', next)
  obj.on('data', \data ->
    obj.removeListener('end', next)
    obj.removeListener('data', data)
    next()
  )
else
  next()
```

Which is a nicer than:
```coffee
if obj instanceof baseStream && !obj._readableState.ended

  # don't inline event handlers in coffeescript (no support for named functions)
  # arguments.callee is deprecated so don't use it to remove event handler
  data = () ->
    obj.removeListener('end', next)
    obj.removeListener('data', data)
    next()

  obj.on('end', next)
  obj.on('data', data)
else
  next()
```

The **`\name`** is the anti _Lambda Abstractions_ hehe ;). Joke aside, I don't know how the syntax should be, but **\\** is an anonymous function in Haskell, so perhaps it could be a named function in coffee.

connec:
As I mentioned above:

> If you wish to give a name to a function in an expression, simply assign it

Your lambda example can be rewritten as

```coffee
if obj instanceof baseStream && !obj._readableState.ended
  obj.on('end', next)
  obj.on('data', data = ->
    obj.removeListener('end', next)
    obj.removeListener('data', data)
    next()
  )
else
  next()
```

dotnetCarpenter:
@connec You're right. Sorry for my coffee ignorance. Perhaps you can also look through the following. I wrote it before I saw your answer and perhaps the argument is mute.

Another reason to only have named functions, is that it makes for a nicely readable program.

Example (es2015):

```js
const world = getWorld()

let action = compose(x => x.doAction(), filter(compose(is, dot('property'))))
let action2 = compose(x => x.doAction(), filter(compose(is, dot('another')))) 

if(condition) action(world)
else action2(world)

function curry(f) {}
function compose(...fs) {}
function filter(predicate, data) {}
function is(a) {}
function dot(property, obj) {}
```
_above: all functions with more than 1 argument are curried_

1. You read the first 3 LOC and can understand the structure of the program.
1. Next you read the actions, 2 LOC with the if statement and can understand the intention of the program.
1. Finally, you can dive into the function definitions themselves and see how it's implemented.

Since function definition are parsed (need a better word) first in a js program, this structure is possible. With function expression, this is not the case.

Not even sure how to express that in coffee. But if all functions are function definition then anonymous function could be `\x -> x * x`, like in Haskell.

connec:
That's a legitimate point @dotnetCarpenter - CoffeeScript drops supports for ES' function hoisting in favour of having a single way of defining any variable/function (via regular assignment).

Whilst this does come up occasionally, I generally find that most of the code I write (in CoffeeScript and in general) is 'lazy' in that the main work in a script is exported as a function making this a non-issue. To paraphrase your example:

```coffee
exports.run = ->
  world    = getWorld()

  action  = compose ((x) -> x.doAction()), filter compose _is, dot 'property'
  action2 = compose ((x) -> x.doAction()), filter compose _is, dot 'another'

  if condition
    action world
  else
    action2 world

curry = (f) -> ...
compose = (fs...) ->
filter = (predicate, data) -> ...
_is = (a) -> ...
dot = (property, obj) -> ...
```

In other cases, such functions are anyway imported from a separate module:

```coffee
{ curry, compose, filter, is: _is, dot } = require './helpers'

world = ...

action = ...

...
```

Obviously that's a bit of a cop-out answer, and the honest one as above is that the hoisting behaviour has been sacrificed in the name of a unified way of defining variables and functions.

mrmowgli:
I just realized that the AirBnB style guide (
https://github.com/airbnb/jabascript#functions) explicitly uses named
functions.

This styleguide is heavily influential and their linting rules will
definitely be copied.

PR #4532, opened 2017-04-30
[CS2] `return` and `export default` can now accept implicit objects

### LINKED ISSUE #1263: Braceless object `return`
I'd like to be able to return an object literal from a function without typing any braces:

```
makeAwesome = (what, ever) ->
  stuff = doStuff what, ever
  return
    foo: what
    bar: doMore stuff
    baz: stuff
```

Right now, you have to do this:

```
makeAwesome = (what, ever) ->
  stuff = doStuff what, ever
  return {
    foo: what
    bar: doMore stuff
    baz: stuff
  }
```

Not terrible, but come on... We don't wanna type no stinking braces! LOL

Actually, this is legal but I think it looks confusing:

```
makeAwesome = (what, ever) ->
  stuff = doStuff what, ever
  foo: what
  bar: doMore stuff
  baz: stuff
```

GeoffreyBooth:
Fixes #1263, probably others. This now compiles:

```coffee
return
  a: 1
  b: 2
```

As does the same for `export default`.


vendethiel:
Very nice :+1:

GeoffreyBooth:
Well, it was previously a compiler error, so it’s not a breaking change. I guess the question is, are we opening ourselves up to allowing the compiler to output too much? Are there any edge cases where something indented on a new line after a `return` wouldn’t be returnable, that we should error on?

src/lexer.coffee
```diff
@@ -790,7 +790,7 @@ exports.Lexer = class Lexer
     LINE_CONTINUER.test(@chunk) or
     @tag() in ['\\', '.', '?.', '?::', 'UNARY', 'MATH', 'UNARY_MATH', '+', '-',
                '**', 'SHIFT', 'RELATION', 'COMPARE', '&', '^', '|', '&&', '||',
-               'BIN?', 'THROW', 'EXTENDS']
+               'BIN?', 'THROW', 'EXTENDS', 'DEFAULT']
```
connec: Out of interest, but would adding 'RETURN" here achieve the same thing as https://github.com/jashkenas/coffeescript/pull/4532/commits/75f6912529f3f8f546dbdb9e5210bfa22fd61cd0?

GeoffreyBooth: That was my first thought, but it's too permissive. This causes a return line continuation even if the line following the return is the same indentation, which we definitely don't want.

connec: Ah, good point, that makes sense!

GeoffreyBooth:
Also merged into `master`: 26cb24acc8d33affe9d91b888cca9841384fb7fc

PR #4534, opened 2017-05-01
Fix #4533: chained calls incorrectly wrapping enclosing implicit objects

### LINKED ISSUE #4533: Lexer: Chaining: Method chaining wraps enclosing object incorrectly
Input:
```coffee
x =
  a: f 3
  b: f 4
    .something()
```
Output:
```js
x = {
  a: f(3),
  b: f(4)
}.something();
```
But consider that
```coffee
x =
  a: 3 # <- no function call here
  b: f 4
    .something()
```
compiles to
```js
x = {
  a: 3,
  b: f(4).something()
};
```
I have a fix coming.

xixixao:
Fixes #4533.

I looked through the other open and closed method chaining issues and this doesn't seem to fix any of them. It's possible it fixes some other issues I'm not aware of (I kinda feel like it should change some behavior, but we don't have any existing tests covering it).

### DIFF
--- a/test/formatting.coffee
+++ b/test/formatting.coffee
@@ -128,6 +128,9 @@ test "indented heredoc", ->
 #   * single line arguments
 #   * inline function literal
 #   * inline object literal
+#
+# * chaining inside
+#   * implicit object literal
 
 test "chaining after outdent", ->
   id = (x) -> x
@@ -220,6 +223,14 @@ test "chaining should work within spilling ternary", ->
       a: 3
       .a
   eq 3, result.h
+
+test "method call chaining inside objects", ->
+  f = (x) -> c: 42
+  result =
+    a: f 1
+    b: f a: 1
+      .c
+  eq 42, result.b
 
 # Nested blocks caused by paren unwrapping
 test "#1492: Nested blocks don't cause double semicolons", ->
--- a/lib/coffee-script/rewriter.js
+++ b/lib/coffee-script/rewriter.js
@@ -171,7 +171,7 @@
       stack = [];
       start = null;
       return this.scanTokens(function(token, i, tokens) {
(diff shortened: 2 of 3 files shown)


GeoffreyBooth:
LGTM. Do you think it’s enough of a breaking-change risk that we should make this only on `2`? (I’m thinking probably not, but just wanted to make sure.)

xixixao:
I think this should be fixed in `1` (alas it is a breaking change, so it should go into the "next" version in our brave no-semver world).

GeoffreyBooth:
Can you please explain how this would be a breaking change?

test/formatting.coffee
```diff
@@ -198,6 +201,14 @@ test "#1495, method call chaining", ->
   ).join ', '
   eq 'a, b, c', result
 
+test "method call chaining inside objects", ->
+  id = (x) -> x
+  result =
+    a: id 1
+    b: id {a: 42}
+      .a
+  eq 42, result.b
```
xixixao: Good catch, let me fix that.

xixixao:
This is a breaking change in the sense that code which previously compiled now compiles to something else. Luckily the previous compilation was not really meaningful, so no one should have hopefully relied on it. Changes like these are rather common here.

GeoffreyBooth:
FYI @xixixao I reviewed all the open issues. I put “Chaining:” in the titles for ones involving chaining or leading dots: https://github.com/jashkenas/coffeescript/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20Chaining

Do you mind reviewing those after you’re done with this? I bet most of them are related. I also put “Lexer:” in the title for anything that seemed like it had to do with the code in `lexer.coffee`, which you seem to be quite expert on.

xixixao:
I've already done that, I'll put the summary here later. I'm definitely no expert in any part of the compiler.

xixixao:
Rebased and fixed tests.

GeoffreyBooth:
@lydell what do you think?

PR #4535, opened 2017-05-01
Fix #4150: Correctly outdent ternary followed by method call

### LINKED ISSUE #4150: Possible bug with dot at beginning of line, ternary and object
1.7.0 added "Leading . now closes all open calls, allowing for simpler chaining syntax."

This is great, but I may have just stumbled on a bug with that when using the ternary operator and an object argument.

``` coffee
gulp.task 'stylus', ->
    gulp.src [
        './src/stylus/vars.styl'
        './src/stylus/**/*.styl'
    ]
        .pipe if argv.sourcemaps then require('gulp-sourcemaps').init() else gutil.noop()
        .pipe require('gulp-concat') 'styles.styl'
        .pipe require('gulp-stylus')
            use: require('nib')()
        .pipe if isLocal() then gutil.noop() else require('gulp-minify-css')
            compatibility: '' # Empty string is IE9+ (this is default)
        .pipe if argv.sourcemaps then require('gulp-sourcemaps').write() else gutil.noop()
        .pipe gulp.dest './dist/css'
```

This is coming out as

``` js
// Generated by CoffeeScript 1.9.2
gulp.task('stylus', function() {
  return gulp.src(['./src/stylus/vars.styl', './src/stylus/**/*.styl']).pipe(argv.sourcemaps ? require('gulp-sourcemaps').init() : gutil.noop()).pipe(require('gulp-concat')('styles.styl')).pipe(require('gulp-stylus')({
    use: require('nib')()
  })).pipe(isLocal() ? gutil.noop() : require('gulp-minify-css')({
    compatilibilty: ''
  }).pipe(argv.sourcemaps ? require('gulp-sourcemaps').write() : gutil.noop())).pipe(gulp.dest('./dist/css'));
});
```

I was expecting
- `compatibility: ''` object to be an argument to `require('gulp-minify-css')` -- in this case the result is as expected.
- the dot at the start of the next line to close both the ternary and the `pipe()` from the preceding line -- in this case the result is not as expected: the next `.pipe()` is chained on the end of the `gulp-minify-css` object.

Note that the `gulp-stylus` pipe above also receives an object as an argument but does not have a ternary. This one compiles as I expect.

Note also that the `gulp-sourcemaps` pipes have ternaries but no object argument, and these ones also compile as I expect.

So it seems it's a combination of the tenary and the object argument on a new line.

Is this a bug or am I understanding something incorrectly?

xixixao:
I think the code in #4150 is pretty horrible, and the fact that an inline ternary can continue after a method call in else just seems bizarre. But I do think that:
```coffee
if 0 then 1 else g
  a: 1
.h()
```
should not compile (certainly not as `.h()` being called on the inline object inside the ternary).

So this PR makes it cause a syntax error, as it should. I would guess that implementation-wise this is not the right fix, but it works and I'm not an expert on the `rewriter` logic.

### DIFF
--- a/test/formatting.coffee
+++ b/test/formatting.coffee
@@ -198,6 +198,29 @@ test "#1495, method call chaining", ->
   ).join ', '
   eq 'a, b, c', result
 
+test "chaining should not wrap spilling ternary", ->
+  throws -> CoffeeScript.compile """
+    if 0 then 1 else g
+      a: 42
+    .h()
+  """
+
+test "chaining should wrap calls containing spilling ternary", ->
+  f = (x) -> h: x
+  id = (x) -> x
+  result = f if true then 42 else id
+      a: 2
+  .h
+  eq 42, result
+
+test "chaining should work within spilling ternary", ->
+  f = (x) -> h: x
+  id = (x) -> x
+  result = f if false then 1 else id
+      a: 3
+      .a
+  eq 3, result.h
+
 # Nested blocks caused by paren unwrapping
 test "#1492: Nested blocks don't cause double semicolons", ->
   js = CoffeeScript.compile '(0;0)'
--- a/lib/coffee-script/rewriter.js
+++ b/lib/coffee-script/rewriter.js
@@ -395,7 +395,7 @@
       starter = indent = outdent = null;
       condition = function(token, i) {
         var ref, ref1, ref2, ref3;
-        return token[1] !== ';' && (ref = token[0], indexOf.call(SINGLE_CLOSERS, ref) >= 0) && !(token[0] === 'TERMINATOR' && (ref1 = this.tag(i + 1), indexOf.call(EXPRESSION_CLOSE, ref1) >= 0)) && !(token[0] === 'ELSE' && starter !== 'THEN') && !(((ref2 = token[0]) === 'CATCH' || ref2 === 'FINALLY') && (starter === '->' || starter === '=>')) || (ref3 = token[0], indexOf.call(CALL_CLOSERS, ref3) >= 0) && this.tokens[i - 1].newLine;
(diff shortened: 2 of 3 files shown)


test/formatting.coffee
```diff
@@ -198,6 +198,27 @@ test "#1495, method call chaining", ->
...
   eq 'a, b, c', result
 
+test "chaining after inline ternary", ->
+  throws -> CoffeeScript.compile """
+    if 0 then 1 else g
+      a: 42
+    .h()
+  """
+
+  # Wrap calls containing spilling ternary
```
GeoffreyBooth: Please put a new line like `test "wrap calls containing spilling ternary", ->` to start a new test.

Currently if anything inside this `test` function fails, the whole test fails and it’s hard to know what part of this failed, since you’re testing multiple things here (the `throws` and the two `eq`s). It would be nice if each of those were separate `test` functions, so that if one of them fails we know exactly which one.

GeoffreyBooth:
While I agree that your example above should throw an error, this shouldn’t:

```coffee
gulp.pipe if 0 then 1 else g
  a: 1
.h()
```

I haven’t tested this, so I don’t know if your version does or doesn’t allow this. We should have a test though that makes sure the above still works as expected (i.e. `gulp` should call `pipe` and then `h`).

test/formatting.coffee
```diff
@@ -198,6 +198,27 @@ test "#1495, method call chaining", ->
...
+  f = (x) -> h: x
+  id = (x) -> x
+  result = f if true then 42 else id
+      a: 2
+  .h
```
xixixao: @GeoffreyBooth I think you missed this test, it's exactly what you asked for.

I will split them each into a separate test call.

GeoffreyBooth: It’s it different between the `gulp.pipe` and `result = f` lines? I agree they’re very similar, and should both work, but I was wondering if the assignment made a difference.

xixixao: Nope, there is no difference. I can add a separate test if you want, but I think as long as this works it should be ok (it's hard to make this pass but your case fail).

xixixao:
@GeoffreyBooth I split up the tests.

GeoffreyBooth:
Thanks. LGTM. @lydell?

lydell:
LGTM

Issue #3551, opened 2014-07-23
JS API


jonathanong:
can't implement anything with coffeescript without a public API, and i'm not going to look through the annotated source

unless you wanna help maintain some processors in https://github.com/normalize/transforms.js, @Nami-Doc :D

GeoffreyBooth:
Done in #4536.

Issue #3281, opened 2013-12-10
Docs: `if item not in items`


Soviut:
I discovered that

```
if not item in items
    items.push item
```

Doesn't work. I discovered that the syntax was

```
if item not in items
```

But shouldn't my original expression still work? Shouldn't the `not` still negate the `item in items` boolean result? I realize my original may not read as nicely, but both make logical sense.

GeoffreyBooth:
Done in #4536.

Issue #2360, opened 2012-05-30
Docs: Lexical Scoping and Variable Safety


Kerrick:
I was just checking out CoffeeScript for the first time, and I came to the section "[**Lexical Scoping and Variable Safety**](http://coffeescript.org/#lexical_scope)" at [coffeescript.org](http://coffeescript.org/). It says:

> This behavior is effectively identical to Ruby's scope for local variables.

However, it's _not_.

The following script in Ruby is _not_ effectively identical to its CoffeeScript counterpart below, namely the value of `outer` after all is said and done.

``` ruby
outer = 1
def change_numbers
  inner = -1
  outer = 10
end
inner = change_numbers
puts inner # => 10
puts outer # => 1
```

``` coffeescript
outer = 1
changeNumbers = ->
  inner = -1
  outer = 10
inner = changeNumbers()
alert inner # => 10
alert outer # => 10
```

While I'm not necessarily asking this _behavior_ be changed in CoffeeScript (as I'm sure it's there for good reason), I do wish the documentation wasn't misleading.

GeoffreyBooth:
Done in #4536.

PR #4536, opened 2017-05-02
[CS2] Docs updates

### LINKED ISSUE #2039: Docs: Better explanations of comprehensions
There are a couple open issues on nested comprehensions, where folks ask whether CoffeeScript should be more like Python and Haskell.  However those issues get resolved, we can probably start by documenting the current behavior more precisely.

First, without getting into abstract definitions, it's pretty easy to call out the two main differences between CoffeeScript and Python:
1. Nested comprehensions are not automatically flattened (map instead of concatMap/flatMap)
2. The index of the comprehension nested deepest varies fastest

Second, it's important to call out the distinction between these lines in coffeescript:

```
eat(x) for x in a # loop w/side effects
squares = (square(x) for x in a) # comprehension
```

Third, it's important to distinguish these two syntactical constructs (which produce the same code):

```
squares = (square(x) for x in a) # comprehension
```

```
squares = (for x in a 
  square(x)) # array expression
```

The two pieces of code are identical semantically, but they are different syntactically.  IMHO we should use a precise definition of "comprehension" that refers specifically to the first syntax.  Wikipedia and Python's PEP both provide precedence for calling out comprehensions as a specific syntactical notation for producing lists.  Wikipedia emphasizes the analogy to set builder notation, which specifically puts the output function first in the expression.

I propose these definitions:
1. Loops - classic loops used to produce side effects
2. Postfix loops -- classic loops where the one-statement body goes first
3. List expressions -- loops in CS that are evaluated as expressions, where "for" precedes the body
4. Comprehensions -- syntax in CS that specifically places the output expression before the "for"

I predict these definitions will be up for extensive debate.  See the discussion between me and @michaelficarra at [issue 2030](https://github.com/jashkenas/coffee-script/issues/2030).

Once the definitions get nailed down, I propose that we reorganize the docs to show all four forms separately, even if we can't agree to come up with four different terms for the forms.

Other related issues:
1. [issue 1191](https://github.com/jashkenas/coffee-script/issues/1191)
2. [issue 2038](https://github.com/jashkenas/coffee-script/issues/2038)

GeoffreyBooth:
I’m trying to close all the open tickets regarding documentation, and make general docs improvements in preparation for 2.0.0-beta2. See live: https://rawgit.com/GeoffreyBooth/coffeescript/docs/docs/v2/index.html

- [x] Rewrote “CoffeeScript 2” section to introduce CS2, explain what’s new and why 2 is important; moved practical info about transpilation into “Usage”
- [x] #3551: Node module API
- [x] #2360: Remove incorrect reference to Ruby variable scoping
- [x] #3281: Add example `if item not in items`
- [x] Better documentation of regular operators, including adding `from` and the JS equivalent for `in`
- [x] #1631: Much more comprehensive documentation of the existential operator
- [x] #4478: Fix destructuring’s comment about default values
- [x] Move the chaining function calls documentation out of the changelog into its own section
- [x] Spinoff separate sections for generator and async functions (from section on bound/fat arrow functions)
- [x] #4530: Removed support for bound instance methods
- [x] #4526: No more `extends` keyword for function prototypes
- [x] #15, #4531: Named functions
- [x] Add separate sidebar links for the sub-sections of “Unsupported ECMAScript Features” and “Breaking Changes from 1.x”

I closed #2039, about comprehensions, as the docs as written matched it almost word for word. I guess someone forgot to close it after the docs were updated sometime long ago.


xixixao:
One thing I noticed while trying the new docs, while it's awesome that the examples are editable! the page is quite slow and janky. It might be because of all these editors or for many other reasons. Compare scrolling through the current page vs the new one.

Thanks for all your hard work 👍

GeoffreyBooth:
Yeah I’m pretty sure it’s because of all the editors. It’s a sacrifice I’m willing to make 😄 though maybe we could make them read-only on mobile sizes. Pull requests welcome 😉

GeoffreyBooth:
Any other notes? @lydell or @jashkenas? @connec I updated the classes docs to reflect that bound methods are no longer allowed. 

Anyone familiar with the Node API who can verify that I've documented it correctly?

PR #4543, opened 2017-05-08
[CS2] Un-prefer global

### LINKED ISSUE #4049: CLI: `npm install` should not warn that CoffeeScript prefers global installation
In order to publish a usable npm package with CoffeeScript, one must compile it first.  In order to do that, you need coffee-script as a `devDependency`, so that anyone contributing to the package will end up with the same version of the coffee-script package for development and deployment.

Their `npm install`, however, will then result in a warning about global installation, when no global installation is either desirable or necessary (if one has used another build tool or an npm `script` entry in `package.json`, which will both use a local install in _preference_ to a global one).  The warning then becomes noise, either irritating people or teaching them to ignore warnings of this type in general.  (Especially since local installs of the coffee-script package are more frequent than global installs.)

In other words, beginning with npm 2, PR #3415 introduced a spurious warning for all non-commandline uses of the coffee-script package, since it is not really true that the package is "primarily a command-line application".  (The npm shipping with node 0.10 didn't do anything with `preferGlobal`, so the warning only shows up for most people at node 0.12.)

At one point, it was perhaps arguable that the preferred usage of CoffeeScript was as a globally-installed command-line tool.  But for node-related development _today_. it is anything but: best practice for development of node modules is to build via the npm `prepublish` script option (so that downstream developers can build a package from source), which doesn't require a global install even if you use a `coffee` command line in it.  (And of course, different packages may in fact be built with different versions of CoffeeScript, making a global install counterproductive.)

So, the warning is not only unnecessary, it actually advises _against_ what it should be advocating _for_: defining a project-local `devDependency` on the specific version required for that project.  Meanwhile, unless there were lots of people getting confused about global installs before #3415 who _aren't_ getting confused now, it's not clear that the warning is actually benefiting anyone.

(Indeed, other command-line tools such as `gulp` do not include this warning.  Best practice for global command-line development tools nowadays is for the global cli to specifically always invoke the _local_ installation of the same package when run, to ensure the correct version is used.  I don't know if the `coffee` tool should go so far as to do that, but it's an illustration of how far out-of-sync the use of `preferGlobal` is, here.  `preferGlobal` is for things that are _just_ command-line tools, not cli-plus-library things.)

GeoffreyBooth:
Closes #4049.

- [x] Installing CoffeeScript no longer triggers NPM’s “prefer global” warning
- [x] The `coffee` and `cake` commands now first check if you have a locally-installed CoffeeScript module under the current path, and execute that version if so, similar to how Gulp works; and fall back to the global version otherwise.

This enables using different versions of the CoffeeScript CLI on a per-project basis:

```bash
/tmp ▶ coffee -v
CoffeeScript version 1.12.5
/tmp ▶ mkdir coffeescript1.11
/tmp ▶ cd coffeescript1.11
/tmp/coffeescript1.11 ▶ npm install coffeescript@1.11.0
/private/tmp/coffeescript1.11
└── coffeescript@1.11.0

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/coffeescript1.11/package.json'
npm WARN coffeescript1.11 No description
npm WARN coffeescript1.11 No repository field.
npm WARN coffeescript1.11 No README data
npm WARN coffeescript1.11 No license field.
/tmp/coffeescript1.11 ▶ coffee -v
CoffeeScript version 1.11.0
/tmp/coffeescript1.11 ▶ cd ..
/tmp ▶ mkdir coffee-script1.11
/tmp ▶ cd coffee-script1.11
/tmp/coffee-script1.11 ▶ npm install coffee-script@1.11.0
/private/tmp/coffee-script1.11
└── coffee-script@1.11.0

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/coffee-script1.11/package.json'
npm WARN coffee-script1.11 No description
npm WARN coffee-script1.11 No repository field.
npm WARN coffee-script1.11 No README data
npm WARN coffee-script1.11 No license field.
/tmp/coffee-script1.11 ▶ coffee -v
CoffeeScript version 1.11.0
/tmp/coffee-script1.11 ▶ cd ..
/tmp ▶ mkdir coffeescript2
/tmp ▶ cd coffeescript2
/tmp/coffeescript2 ▶ npm install coffeescript@next
/private/tmp/coffeescript2
└─┬ coffeescript@2.0.0-beta1
  └─┬ markdown-it@8.3.1
    ├─┬ argparse@1.0.9
    │ └── sprintf-js@1.0.3
    ├── entities@1.1.1
    ├── linkify-it@2.0.3
    ├── mdurl@1.0.1
    └── uc.micro@1.0.3

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/coffeescript2/package.json'
npm WARN coffeescript2 No description
npm WARN coffeescript2 No repository field.
npm WARN coffeescript2 No README data
npm WARN coffeescript2 No license field.
/tmp/coffeescript2 ▶ coffee -v
CoffeeScript version 2.0.0-beta1
```

If you don’t have a globally-installed version, the `coffee` or `cake` commands don’t do anything; you either have to use the locally-installed module via Node, like a normal module, or access the CLI via `./node_modules/coffeescript/bin/coffee`.

### DIFF
--- a/bin/cake
+++ b/bin/cake
@@ -2,6 +2,17 @@
 
 var path = require('path');
 var fs   = require('fs');
-var lib  = path.join(path.dirname(fs.realpathSync(__filename)), '../lib');
 
-require(lib + '/coffeescript/cake').run();
+var potentialPaths = [
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
+  path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
+  path.join(path.dirname(fs.realpathSync(__filename)), '../lib/coffeescript')
+];
+
+for (var i = 0; i < 4; i++) {
+  if (fs.existsSync(potentialPaths[i])) {
+    require(potentialPaths[i] + '/cake').run();
+    break;
+  }
+}
--- a/bin/cake
+++ b/bin/cake
@@ -7,10 +7,10 @@ var potentialPaths = [
   path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
   path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
   path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
-  path.join(path.dirname(fs.realpathSync(__filename)), '../lib/coffeescript')
+  path.join(__dirname, '../lib/coffeescript')
 ];
 
-for (var i = 0; i < 4; i++) {
+for (var i = 0, len = potentialPaths.length; i < len; i++) {
   if (fs.existsSync(potentialPaths[i])) {
     require(potentialPaths[i] + '/cake').run();
     break;
--- a/bin/coffee
+++ b/bin/coffee
@@ -2,6 +2,17 @@
 
 var path = require('path');
 var fs   = require('fs');
-var lib  = path.join(path.dirname(fs.realpathSync(__filename)), '../lib');
 
-require(lib + '/coffeescript/command').run();
+var potentialPaths = [
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
+  path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
+  path.join(path.dirname(fs.realpathSync(__filename)), '../lib/coffeescript')
+];
+
+for (var i = 0; i < 4; i++) {
+  if (fs.existsSync(potentialPaths[i])) {
+    require(potentialPaths[i] + '/command').run();
+    break;
+  }
+}
--- a/bin/coffee
+++ b/bin/coffee
@@ -7,10 +7,10 @@ var potentialPaths = [
   path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
   path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
   path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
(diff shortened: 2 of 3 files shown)


bin/cake
```diff
@@ -2,6 +2,17 @@
...
 
-require(lib + '/coffeescript/cake').run();
+var potentialPaths = [
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
+  path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
+  path.join(path.dirname(fs.realpathSync(__filename)), '../lib/coffeescript')
+];
+
+for (var i = 0; i < 4; i++) {
```
lydell: Why a hard coded `4` rather than `potentialPaths.length`?

bin/cake
```diff
@@ -2,6 +2,17 @@
...
 var path = require('path');
 var fs   = require('fs');
-var lib  = path.join(path.dirname(fs.realpathSync(__filename)), '../lib');
 
-require(lib + '/coffeescript/cake').run();
+var potentialPaths = [
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffeescript'),
+  path.join(process.cwd(), 'node_modules/coffeescript/lib/coffee-script'),
+  path.join(process.cwd(), 'node_modules/coffee-script/lib/coffee-script'),
+  path.join(path.dirname(fs.realpathSync(__filename)), '../lib/coffeescript')
```
lydell: I know it was like this before, but why not use `__dirname`?

GeoffreyBooth: Dunno. Was there a reason it was how it was before? Windows?

bin/cake
GeoffreyBooth: I was trying to be optimized, though I can see how this comes across as overkill. I toyed with a series of if statements to avoid even the unnecessary path manipulation lines, but testing didn't show a difference. I can make this more idiomatic if you think it's silly to hard-code the length to save 1ms.

GeoffreyBooth:
@lydell how about now?

PR #4546, opened 2017-05-11
[WIP][CS2] Add webpack support

akfish:
Resend #4501


Cakefile
```diff
@@ -440,3 +435,30 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
   (-> eval source).call result
   testResults = runTests result.CoffeeScript
   process.exit 1 unless testResults
+
+task 'build:webpack', 'build webpack test bundle', ->
+  args = [
+    "./node_modules/webpack/bin/webpack.js"
+    "--entry=./test/webpack/entry.js",
```
GeoffreyBooth: Why this entry, and not browser.js as above?

Cakefile
```diff
@@ -440,3 +435,30 @@ task 'test:browser', 'run the test suite against the merged browser script', ->
...
+    "--output-path=test/webpack",
+    "--output-filename=coffeescript.js"
+  ]
+
+  spawnNodeProcess args, 'both', (status) ->
+    process.exit(status)
+
+task 'test:webpack', 'run test suite against the webpack test bundle', ->
+  loader = """
+  var __exports__ = {};
```
GeoffreyBooth: What is this variable? Is it distinct from CommonJS `exports`?

akfish: Webpack will scope known globals. It is not possible to access them directly. This, along with the custom entry file and the loader script in test:webpack task, allows the bundled CoffeeScript to escape thus it can be required in node environment.

akfish: There might be a better way to do this. I'll get back to you later.

akfish:
Turns out there's an webpack option for that and no additional configuration files nor plugins required. That makes more sense. Don't know what I was thinking back then.
I changed the entry point to `./` instead of explicitly specify one. This let webpack to use `browser` field in `package.json`.
I also removed the `build:webpack` task and have the build run in `test:webpack` task.

GeoffreyBooth:
@lydell or @jashkenas did you have any further notes? I’m leaving the `test:integrations` task as a place for future tests like loading `test.html` via Selenium Chromedriver.

Somehow `cake test:browser` fails both here and on `2`, so something we did recently broke the browser build; but it’s unrelated to this PR.

jashkenas:
Looks great!

PR #4550, opened 2017-05-15
[CS2] 2.0.0-beta2

GeoffreyBooth:
Updated docs at http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/. From the [changelog](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#changelog):

*   This release includes [all the changes from 1.12.5 to 1.12.6](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#1.12.6).
*   Bound (fat arrow) methods in classes must be declared in the class constructor, after `super()` if the class is extending a parent class. See [breaking changes for classes](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#breaking-changes-classes).
*   All unnecessary utility helper functions have been removed, including the polyfills for `indexOf` and `bind`.
*   The `extends` keyword now only works in the context of classes; it cannot be used to extend a function prototype. See [breaking changes for `extends`](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#breaking-changes-super-extends).
*   Literate CoffeeScript is now parsed entirely based on indentation, similar to the 1.x implementation; there is no longer a dependency for parsing Markdown. See [breaking changes for Literate CoffeeScript parsing](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#breaking-changes-literate-coffeescript).
*   JavaScript reserved words used as properties are no longer wrapped in quotes.
*   `require('coffeescript')` should now work in non-Node environments such as the builds created by Webpack or Browserify. This provides a more convenient way to include the browser compiler in builds intending to run in a browser environment.
*   Unreachable `break` statements are no longer added after `switch` cases that `throw` exceptions.
*   The browser compiler is now compiled using Babili and transpiled down to Babel’s `env` preset (should be safe for use in all browsers in current use, not just evergreen versions).
*   Calling functions `@get` or `@set` no longer throws an error about required parentheses. (Bare `get` or `set`, not attached to an object or `@`, [still intentionally throws a compiler error](http://rawgit.com/geoffreybooth/coffeescript/2.0.0-beta2/docs/v2/#unsupported-get-set).)
*   If `$XDG_CACHE_HOME` is set, the REPL `.coffee_history` file is saved there.

I added a note about the bound class methods to the classes breaking changes section. I also renamed “Change Log” to “Changelog” since [that’s how Wikipedia spells it](https://en.wikipedia.org/wiki/Changelog).


GeoffreyBooth:
I was hoping to include #4493 in this but I guess that’s not going to happen. @connec, whenever that’s ready I would be happy to release a beta3. There’s no reason we can’t release beta versions more often; though as #4493 isn’t a breaking change, it could also come post-2.0.0.

There seems to be something wrong with the `cake release` command building or testing the regular and browser builds concurrently, and triggering erroneously failing tests. If we go back to the old style of separately running `cake build:full` and then `cake build:browser:full` to build and test each in turn, all the tests pass. Ditto `test.html`. I want to refactor this anyway to add a Selenium Chromedriver-based test, so I’ll try to fix `cake release` then.

The other item on my to-do list is to try to implement some kind of “comments anywhere” support, so that block comments can go anywhere, to enable things like JSX support through a pre- and post-compilation plugin. That would be a mild breaking change, since currently block comments get moved around in placement during compilation (and they would still get moved around in the new version, though not necessarily to the same places). This might be mild enough of a breaking change that it could also come post-2.0.0.

So I guess if this release seems stable and we don’t get any major bugs to fix from it, hopefully this can be the release candidate for 2.0.0.

GeoffreyBooth:
@lydell do you have any notes on this? Nothing new here except some documentation.

PR #4551, opened 2017-05-16
[CS2] Support for CSX - equivalent of JSX

### LINKED ISSUE #4529: Meta: Support JSX similar to Literate CoffeeScript?
I’ve been thinking about JSX, and a way that CoffeeScript could perhaps support it. In particular, I’ve been thinking about how it’s actually somewhat similar to Literate CoffeeScript:

- Literate CoffeeScript is essentially a Markdown file that uses indentation to define CoffeeScript code blocks.
- JSX is JavaScript that uses XML open and closing tags to define HTML blocks (that themselves can contain JavaScript, within braces).

I haven’t used JSX in a project, so if I’m oversimplifying here, please correct me.

**What if the CoffeeScript compiler detected XML open and closing tags and treated them the same as backticks?** In other words, everything from the first `<` of the first opening tag through to the last `>` of the last closing tag would be passed through unmodified, similar to embedded JavaScript. The compiler would therefore be outputting JSX, and then the next step in the build chain would be Babel with the JSX transform.

This wouldn’t even be a breaking change. Currently `<div>` throws `unexpected <`, and `</div>` throws `missing / (unclosed regex)`.

We also wouldn’t be favoring one framework over another, because we’re just outputting the straight HTML, not converting it into code like `React.createElement`. [Vue has added support for JSX](https://vuejs.org/v2/guide/render-function.html#JSX), and the [Babel transform is now generic](http://revelry.co/using-jsx-with-other-frameworks-than-react/) to allow future frameworks to adopt it, so it’s not a React-specific thing anymore.

How is this different from [coffee-react or coffee-react-transform or CJSX](https://github.com/jsdf/coffee-react), you might ask?
- That project is deprecated, for starters, and is pinned to version 1.9.1 of the compiler.
- It didn’t rely on Babel, compiling straight to the final JavaScript `React.createElement` output, and therefore was dependant on React; the last version supports “React >= 0.14.x”.
- It aimed to compile CoffeeScript _within_ the XML blocks.

That last point is important. In my proposal, within an XML block you would need to use `this.foo`, not `@foo`. You couldn’t write code like coffee-react‘s example `{<p key={n}>This line has been printed {n} times</p> for n in [1..5]}`—you would need to write the JavaScript equivalent. But I don’t think this is so terrible. In my experience, most of the JavaScript inside a template is very minimal, generally just references to objects or basic constructs like `if` statements or loops. (Again, please correct me if I’m wrong.) It would be a formidable challenge to parse CoffeeScript _within_ XML tags, which is probably why `coffee-react` was deprecated; but I don’t think we need to take on that task.

I can certainly see people objecting to this proposal on the grounds that people shouldn’t be forced to use JavaScript, or even snippets of it, within a CoffeeScript file. Fair enough. Another perfectly valid objection is that [JSX](https://facebook.github.io/react/docs/introducing-jsx.html#jsx-is-an-expression-too) isn’t hard to achieve right now using plain backticks:

```coffee
getGreeting = (user) ->
  if user
    `<h1>Hello, {formatName(user)}!</h1>`
  else
    `<h1>Hello, Stranger.</h1>`
```
___
```jsx
var getGreeting;

getGreeting = function(user) {
  if (user) {
    return <h1>Hello, {formatName(user)}!</h1>;
  } else {
    return <h1>Hello, Stranger.</h1>;
  }
};
```
See [CoffeeScript](http://coffeescript.org/v2/#try:getGreeting%20%3D%20(user)%20-%3E%0A%20%20if%20user%0A%20%20%20%20%60%3Ch1%3EHello%2C%20%7BformatName(user)%7D!%3C%2Fh1%3E%60%0A%20%20else%0A%20%20%20%20%60%3Ch1%3EHello%2C%20Stranger.%3C%2Fh1%3E%60%0A) and [Babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=env%2Creact&targets=&browsers=&builtIns=false&debug=false&code=var%20getGreeting%3B%0A%0AgetGreeting%20%3D%20function(user)%20%7B%0A%20%20if%20(user)%20%7B%0A%20%20%20%20return%20%3Ch1%3EHello%2C%20%7BformatName(user)%7D!%3C%2Fh1%3E%3B%0A%20%20%7D%20else%20%7B%0A%20%20%20%20return%20%3Ch1%3EHello%2C%20Stranger.%3C%2Fh1%3E%3B%0A%20%20%7D%0A%7D%3B) conversions. If we don’t implement this proposal, at the very least we should perhaps at this example to the docs, either in the backticks section or in a new section discussing JSX.

But anyway, now that JSX is growing into a framework-independent standard, and we can support it in a minimal, framework-independent way, should we? And if we should, does this seem like the way to add support for it?

xixixao:
So here's a fully functional implementation of what I call CSX - the JSX equivalent for CoffeeScript. I tried to adhere as much as possible to the syntactical rules of JSX (as opposed to doing possible fancy stuff).

First, let me address why I think it theoretically makes sense to include this in CoffeeScript:

- To implement CSX properly, you need a fully functioning CoffeeScript parser. This is so difficult to do without using the actual implementation as it's practically impossible. Consider that https://github.com/jsdf/coffee-react-transform took over two years of development to get to its current state, with over 1200 lines of CS source code. This PR took me two nights and is adding 170 LOC to the compiler, while avoiding many issues that plagued the transform, f.e. https://github.com/jsdf/coffee-react-transform/issues/56
- JSX is immensely popular. Although I personally would not pick XML as role-model for anything, it has took off and is now used widely outside of React. It basically allows second compilation mode for some expressions. If CoffeeScript cannot support it, it's gonna be a major driver for people moving to ES6/7 with Webpack etc. toolchain (the only other really big thing CoffeeScript is missing is static typing, which might eventually be its final nail in the coffin).
- JSX has real merits over custom runtime replacements in terms of syntax, mainly ease of textual content surrounding other expressions.

Second, I implemented this on top of `2`, but the implementation would make it trivial to support it in original CoffeeScript too, provided we would have a way (option, pragma) to determine what target the CSX tags should compile to.

So how is this implemented and why: JSX is a super simple transform, provided you know you are looking at an expression in the surrounding language. Conceptually it works like this:
```js
<div /> → div()
<div x="a" /> → div({x: 'a'})
<div>A</div> → div({}, 'A')
```
and so on. Of course `React` for example actually uses `Something('div')` instead of `div()`, but that doesn't really matter for the parsing phase. So I actually turn CSX into valid CoffeeScript tokens in the lexer. Note that some of the work could be done in the rewriter, the split between the lexer and the rewriter is very arbitrary, and I just found it easier to do this in the lexer, instead of having to pass bunch of information to the rewriter. But I think it would be doable to move most stuff outside of `csxToken` method there.

Once I do this conversion, it's actually tricky to tell which nodes were originally CSX. The Jison parser does not make it easy at all to pass adhoc info from the lexer to `nodes` (which compile the tree to JS), so I created a new token for the tag, and seed the info from there when compiling. The tagging of children nodes is not super clean, but I haven't done anything that wasn't at least a bit used before.

Now, for differences between JSX, coffee-react-transform and CSX:

1. CSX doesn't support inline comments inside interpolations, same as JSX
2. bare expressions as attribute values are not supported, same as JSX
3. there is no support for custom pragmas, the compiler outputs JSX
4. **BREAKING CHANGE** `<x` **with leading space** is no longer supported. Supporting this leads to highly ambiguous syntax:
```coffee
x = a<b>c; # allowed
x = a <b> c; # syntax error, missing closing tag
```

both JSX and the transform handle this "correctly", allowing this code to be treated as comparison. I'm not sure that there is a way to allow this in CoffeeScript without making the resulting compilation ambiguous. JSX can do this easily, because `a <b/>` is not a valid JS epxression, but in CS it is simply `a(<b/>)`. I think the transform looks ahead to determine whether there is a closing tag, and we could do something similar, but I don't know how to do that without introducing incorrect edge cases. ~~Ideas welcome.~~ _(Update: I think restricting this to only unbalanced whitespace should alleviate most concerns)_

You can see these in the tests I added. Note that I removed some original coffee-react-transform tests, which were mainly testing the CoffeeScript parsing, not the CSX part.

What's left?

Spreads, blocked by #3894
One case which is broken, which boils down to this being invalid CoffeeScript:
```
article({number: 2,
  range: 1,
},"\n")
```
(there is a corresponding CSX test with a TODO).

Fixes #4529.


GeoffreyBooth:
This is very impressive. Thanks for making this effort!

If I could try to head off some expected critiques, if readers have an objection to whether JSX should be supported in the CoffeeScript ecosystem at all, please discuss that in #4529. For the purposes of this PR let’s please assume that CoffeeScript should support JSX, and let’s keep this thread to the specifics of this implementation.

Reading through what you’ve done, it seems obvious that the approach I had in mind in #4529, of trying to treat `<`s and `>`s essentially as backticks and then finding a way to update the compiler to allow backticks anywhere, would never work. It also seems obvious that the thought behind #4540, of building hooks for plugins, would be far too insufficient to achieve what you’ve done here. At first glance, at least, this appears to be the “full-assed” implementation that @jashkenas was [referring to](https://github.com/jashkenas/coffeescript/issues/4529#issuecomment-297105753). Well done.

A few general questions before getting into specifics:

1. How does this PR avoid the problems that doomed `coffee-react-transform`? Is it basically that this outputs JSX, whereas that project attempted to do what Babel’s React transform does and compile down to the `React.createClass` etc.?
2. Is the only breaking change that now a space is required on either side of the `<` and `>` operators? (I can live with that.)
3. How do these new tags interact with CoffeeScript’s comprehensions? Our comprehensions remain a big selling point over ES, and CSX working well with them would be a big reason to use CSX over JSX.
4. How does indentation affect these tags? Or is it more or less irrelevant, since you know the hierarchy from the open and closing tags? Are we going to allow any indentation around the tags, and if so, how does that affect the code within them?
5. How important are the TODOs? I assume we can live without the spread operator until #3894, as it’s not really part of JSX at all but rather a proposed ES feature that happens to be especially convenient in JSX; but are there any important parts of JSX that are unachievable until the TODOs are completed? Are there any future breaking changes that will happen when the TODOs get addressed? I’m wondering if this PR can be merged in as is, or if we need to treat it as work-in-progress until some of the TODOs are addressed, and if so, which ones.

lib/coffeescript/parser.js
```diff
@@ -1,4 +1,4 @@
-/* parser generated by jison 0.4.17 */
+/* parser generated by jison 0.4.13 */
```
GeoffreyBooth: Somehow you don’t have the latest version of Jison installed. Please `npm install`.

src/lexer.coffee
```diff
@@ -48,6 +48,7 @@ exports.Lexer = class Lexer
     @seenExport = no             # Used to recognize EXPORT FROM? AS? tokens.
     @importSpecifierList = no    # Used to identify when in an IMPORT {...} FROM? ...
     @exportSpecifierList = no    # Used to identify when in an EXPORT {...} FROM? ...
+    @includesCSX = no            # Used to optimize CSX checks
```
GeoffreyBooth: In general, all comments should be complete sentences and end in periods. This makes the annotated source feel, well, annotated, like there is normal written commentary accompanying the code.

test/csx.coffee
```diff
@@ -0,0 +1,616 @@
+# We usually do not check the actual JS output from the compiler, but since
+# CSX is not readily supported by Node, we do it in this case
+eqCSX = (cs, js) ->
```
GeoffreyBooth: There’s already a `toJS` function in global scope. Can we please just use it, and follow the pattern of the modules tests?

src/nodes.coffee
```diff
@@ -535,6 +538,17 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
 
 exports.StringLiteral = class StringLiteral extends Literal
 
+  compileNode: (o) ->
+    res = if @csx then [@makeCode @unquote yes] else super()
+
+  unquote: (literal) ->
+    unquoted = @value[1...-1]
+    if literal
+      unquoted.replace /\\n/g, '\n'
```
lydell: Could you explain what's special about the newlines here?

src/nodes.coffee
```diff
@@ -535,6 +538,17 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
...
+        .replace /\\"/g, '"'
```
lydell: Dealing with backslashes is never this simple. What if the backslash is already escaped? `\\"`

test/csx.coffee
```diff
@@ -0,0 +1,616 @@
...
+test 'regex in interpolation', ->
+  eqCSX '''
+    <div x={/>asds/}><div />{/>asdsad</}</div>
+  ''', '''
+    <div x={/>asds/}><div />{/>asdsad</}</div>;
+  '''
+
+# Unlike in coffee-react-transform
+test 'bare numbers not allowed', ->
+  throws -> CoffeeScript.compile '<div x=3 />'
```
lydell: I would like to see what the error messages look like for these, and move the tests to error_messages.coffee

lydell:
Random thoughts:

- Would coffeelint be able to create CSX lint rules, without bending over backwards trying to differentiate CSX from function calls or something?
- What are error messages for mistakes in CSX like?
- Could any problems arise when combining literate CS and CSX?
- I think JSX has special string escaping rules which we need to test, but I can't remember them right now.

greghuc:
For the future of coffeescript, I think having a clear solution for handling jsx is good. However, I'm not convinced that adding _compiler-level support_ is the way to go: it increases complexity, makes future change to the compiler harder, and biases coffeescript towards jsx over other templating approaches.

As a pointer for another direction, a clean jsx solution might be possible using tagged template strings. Consider:
* hyperx: https://github.com/substack/hyperx
* bel (built on top): https://github.com/shama/bel

With bel/hyperx an "html template" is specified as the tagged template "string", which is supplied to the tagged template "function" for processing into a dom (or react?) element. I don't use jsx, but I do use [bel](https://github.com/shama/bel) to construct html dom elements like this in Coffeescript:

```
    element = bel"""
      <div>
        <h3>Contact Us</h3>
        <form>
            <label for="email">From</label>
            <input type="text" oninput=#{handleEmailInput} value="#{model.email}">
            <textarea placeholder="Some text" oninput=#{handleContentInput}>{model.content}</textarea>
        </form>
      </div>
    """
```

I find this a very clean approach: the html template is specified as a templated-string, which is post-processed into a dom element tree using the prefixing template-function (bel in this case). This is what motivated me to [propose and add tagged template string support to CS2](https://github.com/jashkenas/coffeescript/pull/4352).

So I would suggest that a cleaner solution to adding JSX support to Coffeescript is to find/create a small library that uses tagged template strings ala hyperx/bel. There might even be a library that does this already. Given the current popularity of JSX, the coffeescript docs could then have a section showing use of JSX with Coffeescript using this library. But we wouldn't add first-class support at the compiler level.

greghuc:
Quick addendum: here's a hyperx issue discussing using it with jsx: https://github.com/substack/hyperx/issues/9

GeoffreyBooth:
@greghuc how would you implement @lydell’s [JSX example](https://github.com/jashkenas/coffeescript/issues/4529#issuecomment-297071257) using tagged template literals?

```jsx
<Table>
  {items.map(item => {
    const itemId = item.is_accessory ? item.parentId : item.id;
    const hidden = !allHidden && removedItemIds.indexOf(itemId) >= 0;
    return (
      <ItemRow
        key={item.id}
        item={item}
        quantityOverride={itemQuantityOverrides[itemId] || null}
        className={hidden ? 'js-Hiding is-hidden' : undefined}
      />
    );
  })}
</Table>
```

greghuc:
@GeoffreyBooth the example would probably look like:

```
    bel"""
      <Table>
        #{items.map (item) ->
            itemId = if item.is_accessory then item.parentId else item.id
            hidden = !allHidden && removedItemIds.indexOf(itemId) >= 0
            bel"""
              <ItemRow key=#{item.id}
                       item=#{item}
                       quantityOverride=#{itemQuantityOverrides[itemId] || null}
                       className=#{if hidden then 'js-Hiding is-hidden' else undefined}
              />
            """
         }
      </Table>
    """
```

I've never done this in jsx, but I do regularly use bel with nested (real) dom components to create complicated dom trees. It works very nicely:
* The tagged template literal means you can nest code blocks inside the "html", ad infinitum
* If a code block returns a dom element (or array of), hyperx/bel is clever enough to merge this in when assembling the whole dom tree.

GeoffreyBooth:
Interesting. Do you mind please shifting this to #4529 and we can discuss it there? I can delete the comments from this thread. I'd like to keep this thread focused on this PR and not a free-for-all about JSX (which the other thread is).

greghuc:
@GeoffreyBooth done re moving comments :-)

xixixao:
@GeoffreyBooth 

> ... full-assed ...

Correct on all points.

1. > How does this PR avoid the problems that doomed coffee-react-transform? Is it basically that this outputs JSX, whereas that project attempted to do what Babel’s React transform does and compile down to the React.createClass etc.?

No, I think the main troubles came from a custom parser that was independent of CoffeeScript. Outputting JSX mainly simplifies the job of escaping stuff (since the escaping/unescaping will be done by the JSX transform).

2. > Is the only breaking change that now a space is required on either side of the < and > operators? (I can live with that.)

Yes, I think so. To this change: at least it's quite unlikely that you will have code that previously compiled compile to something else. So it is a breaking change, but most people will be able to easily fix it.

3. > How do these new tags interact with CoffeeScript’s comprehensions? Our comprehensions remain a big selling point over ES, and CSX working well with them would be a big reason to use CSX over JSX.

There is nothing special in regards to comprehensions. You can use attribute strings without wrapping (**currently broken, will fix and add a test**).

4. > How does indentation affect these tags? Or is it more or less irrelevant, since you know the hierarchy from the open and closing tags? Are we going to allow any indentation around the tags, and if so, how does that affect the code within them?

You can see from the first TODO that indentation matter. After all, the tags are treated as object literal arguments. There might be a way to filter the indentation out, so it doesn't cause problems, but I don't think it's important to allow any formatting. CoffeeScript is an indentation based syntax, you should not expect CSX to suddenly ignore whitespace.

5. > How important are the TODOs? I assume we can live without the spread operator until #3894, as it’s not really part of JSX at all but rather a proposed ES feature that happens to be especially convenient in JSX; but are there any important parts of JSX that are unachievable until the TODOs are completed? Are there any future breaking changes that will happen when the TODOs get addressed? I’m wondering if this PR can be merged in as is, or if we need to treat it as work-in-progress until some of the TODOs are addressed, and if so, which ones.

Spread is part of the JSX syntax and requires special support. Note that `<div a="x" {...props}>` will compile via `div({a: "x", ...props})`, the curlies need to be accounted for and stripped. It's also pretty widely used and hence important. But I don't want to add it until there is support for object spreads in CoffeeScript AST. The only other TODO is the indentation bug, otherwise this is complete JSX as far as I am aware (there might be some missing edge cases, help welcome). This PR should be mergeable as is, if we want this in the language.

@lydell

1. > Would coffeelint be able to create CSX lint rules, without bending over backwards trying to differentiate CSX from function calls or something?

There's a linter for ES with JSX too, right? Should be doable.

2. > What are error messages for mistakes in CSX like?

I will add this to the tests, but in short, they are pretty:
```
error: expected corresponding CSX closing tag for bad
<div><bad></div>
      ^^^
```
I modelled the new errors I added after JSX errors, although the attribute one is prettier than JSX:
```
error: expected wrapped or quoted CSX attribute
<div x=f(3) />
       ^
```

3. > Could any problems arise when combining literate CS and CSX?

Literate CS uses indentation no? Shouldn't be a problem.

4. > I think JSX has special string escaping rules which we need to test, but I can't remember them right now.

There are some tests for this ported over from coffee-react-transform. I am essentially passing everything through and allowing JSX to deal with this.

src/lexer.coffee
xixixao: The two lines above need fixing then .)

test/csx.coffee
xixixao: @GeoffreyBooth is that boilerplate useful? I find the tests easier to read without it.

src/nodes.coffee
xixixao: We format strings to JS strings, which means escaping new lines and quotes. This function removes it, so the string's content is printed out literally.

"b\\"a" throws syntax error, so that can never happen.

jashkenas:
This is indeed, the sort of "full-assed" support I was hoping for. Congrats!

For what it's worth, I still think that CoffeeScript, as a concept, is made weaker by including JSX than leaving it out.

But that said, if you fine folks want to put it in 2.0, this sort of patch is the way to do it.

One note: Wouldn't it be better (and more harmonious) to have CSX leave out closing tags, and simply use CoffeeScript's significant whitespace to close all CSX tags?

test/csx.coffee
GeoffreyBooth: Just keeps things DRY, and ultimately we’re trying to keep the codebase internally consistent. Not everything in there is done the way I’d prefer either.

src/lexer.coffee
GeoffreyBooth: Yes they do. They should also get some backticks, as they’re treated as Markdown when getting converted into the annotated source. I’ve generally been fixing comments as I come across them, rather than taking the time to go through all the code from top to bottom, so there are plenty of bad examples waiting to be fixed. All I’m asking you is to not add to the list of comments to correct 😄

GeoffreyBooth:
> Wouldn’t it be better (and more harmonious) to have CSX leave out closing tags, and simply use CoffeeScript’s significant whitespace to close all CSX tags?

I like this idea. If we’re going to support JSX output, we might as well try to improve it (safely) however we can, just as we try to improve some of JavaScript’s rough edges. And I really want CSX to fit in with CoffeeScript’s overall use of significant whitespace, and not be another feature like chaining that creates a whitespace-insignificant zone in the middle of otherwise sensible code.

I wonder if we could also find an alternate syntax that doesn’t use `<` or `>`, like Jade, but that’s probably going too far to save us from two difficult-to-type characters.

xixixao:
Updates to code: Fixed comment. Used `toJS`. Added tests to `errorMessages`. Fixed error in identifier regex that cause `<2` to be valid tag name (so this limits the breaking change to only `<idenfitier`, `<42` is still valid). Fixed interpolation inside attribute values, added test case for it.

test/error_messages.coffee
```diff
@@ -1545,3 +1545,43 @@ test "#4248: Unicode code point escapes", ->
...
+  assertErrorFormat '''
+    <Person><<</Person>
+  ''',
+  '''
+    [stdin]:1:9: error: unexpected <<
+    <Person><<</Person>
+            ^^
+  '''
+
+test "csx error: ambigious tag-like expression", ->
```
lydell: ambigous

test/csx.coffee
lydell: There's no reason to have these test both here and in error_messages.coffee, is it?

test/csx.coffee
```diff
@@ -0,0 +1,622 @@
...
+    <div popular="yes" />;
+  '''
+
+# The following tests were adopted from James Friend's
+# https://github.com/jsdf/coffee-react-transform
+
+test 'ambigious tag-like expression', ->
+  throws -> CoffeeScript.compile 'x = a <b > c'
+
+test 'ambigious tag', ->
```
lydell: ambigous

xixixao:
@jashkenas I considered using whitespace. But there are theoretical and practical reasons not too. Theoretical:
1. JSX is verbose even for JS. Is there a point in trying to make it less verbose? It's verbosity is kinda the point of it (i.e. "designers" can use it, it's 1:1 with HTML, etc.)
2. CS2 is trying to match ES7+. JSX is a feature of the ES7+ ecosystem. With this approach you can tell people: You can now use CSX just like you would use JSX, just swap CS for JS and rest is the same.

Practical:
1. It just doesn't work well. Consider the following very simple snippet:
```es6
<div>
  <div>Title</div>
  <div className="bla">
    Hello
  </div>
  world
</div>
```
Now based on indent:
```es6
<div>
  <div>Title</div>
  <div className="bla">
    Hello
  world
```
Now you could play with the rules and come up with some variations, but they will all be confusing/ambiguous/inconsistent.

Think of JSX as heredocs (which is also, coincidentally, how I implemented it). You wouldn't want to use indentation there either, you also use delimiters
```coffee
"""Hello #{
  f(x)
} something"""
```

test/error_messages.coffee
xixixao: You mean ambiguous? 😆

lydell: Yes! (Looks like I edited by comment too late 😀)

GeoffreyBooth:
@xixixao the next time you commit, can you please not force-push or amend? It’s easier to review this if each revision is a separate commit. All the commits will get squashed together when we merge the PR into `2`, so the history on your branch doesn’t matter.

jashkenas:
@xixixao — While I'm not persuaded in the least by your two theoretical reasons, I am moved by your practical one:

> It just doesn't work well.

I would love to see a sampler of 3-5 real-world examples of reasonable JSX templates, done in closed-tag and significant-whitespace styles. You may very well be right, but it would be worth evaluating.

xixixao:
@GeoffreyBooth Github n00b, will do

xixixao:
@GeoffreyBooth fixed, sorry about that.

helixbass:
@xixixao @jashkenas @GeoffreyBooth @lydell I didn't see this until now, I've also been working on JSX-in-Coffeescript this week. I'll take a look at @xixixao's implementation but seems like I took a different approach: in addition to normal JSX syntax, there are whitespace-indented syntaxes (basically HAML) and it lexes differently inside of JSX elements but creates tokens and has grammar rules for them. To me that's more Coffeescript-y (it's certainly what I want for the same reasons I love Coffeescript) and I'm not surprised by @jashkenas' comment to that effect. I'll open a pull request as well (though I wasn't sure if merging it into Coffeescript proper (1.x or 2) would be the way to go, whether b/c of breaking changes, philosophical objections, ...)

GeoffreyBooth:
@helixbass thanks! It’s a shame that people are duplicating effort, but this makes clear just how in-demand this feature is. I’m more convinced than ever that we need to support JSX in a non-hacky way.

Please target the `2` branch. I’m trying to consider the 1.x line as more-or-less retired except for important bugfixes. I don’t want the burden of trying to maintain two branches. Also we need more reasons for people to upgrade 😄

@helixbass and @xixixao it might be worth looking at the [Coffee Tags PR](https://github.com/jashkenas/coffeescript/pull/4346). It was an attempt to support a JSX-like syntax that used significant whitespace, and compiled out to Vue or React.

xixixao:
@GeoffreyBooth That syntax didn't support easily specifying text content, which is what imposes the syntactical restrictions of JSX, imho. @helixbass pointed out some of the implementation constraints in #4553 too.

helixbass:
@xixixao after reading through this PR and some discussion with @GeoffreyBooth in #4553, I'm of the opinion that your version of JSX support is what belongs in Coffeescript proper. It's more lightweight (very elegant!), less breaking change-y, and less opinionated/controversial syntax-wise -- I think having support for JSX in Coffeescript proper is wise (so that there's a simple answer "yes you can use JSX in Coffeescript") and I think your approach is right - Compile directly to JSX, support JSX syntax as-is (just with Coffeescript expressions). And then my (or any other) whitespace-indented version can live side-by-side (@GeoffreyBooth suggested trying to implement/use a plugin architecture, I'll certainly ponder that but I have the feeling the hooks required would be too subtle/specialized (and @lydell's comment in #4346 about Babel not supporting plugins is duly noted) so I'm planning on maintaining my version separately) and then any folks who want a whitespace-indented syntax can use that instead of Coffeescript proper.

To be clear for all, the core difference between the two approaches (beyond the syntactic sugar) is whether you treat the body of a JSX element as a (specially-)interpolated string or as structured/indented content. Both make sense - markup is both unstructured-text-by-default and structured/nested content - but by taking the latter approach you have to basically go into a different lexing mode (ie not fall back to the core loop in `tokenize()`) so that you can keep track of eg indentation structure while treating things by default as text content (at least if you want to have text content be "first-class" inside JSX elements, @xixixao this relates to your preceding comment and is part of why I prefer a Haml-style syntax (where text content is the default inside elements) to the approach in #4346). So you end up with a much more heavyweight implementation, which again coupled with more breaking changes and a more opinionated syntax makes me feel that my version doesn't belong in Coffeescript proper. 

So +1 to this being merged

src/lexer.coffee
```diff
@@ -105,7 +107,9 @@ exports.Lexer = class Lexer
   # referenced as property names here, so you can still do `jQuery.is()` even
   # though `is` means `===` otherwise.
   identifierToken: ->
-    return 0 unless match = IDENTIFIER.exec @chunk
+    inCSXTag = @atCSXTag()
+    regex = if inCSXTag then CSX_IDENTIFIER else IDENTIFIER
```
xixixao: Actually, there is a difference, JSX doesn't allow `'.'`s in attribute names.

src/lexer.coffee
```diff
@@ -205,8 +209,11 @@ exports.Lexer = class Lexer
       [tagToken[2].first_line, tagToken[2].first_column] =
         [poppedToken[2].first_line, poppedToken[2].first_column]
     if colon
-      colonOffset = input.lastIndexOf ':'
-      @token ':', ':', colonOffset, colon.length
+      colonOffset = input.lastIndexOf if inCSXTag then '=' else ':'
+      colonToken = @token ':', ':', colonOffset, colon.length
+      colonToken.csxColon = yes if inCSXTag # used by rewriter
+    if inCSXTag and tag is 'IDENTIFIER' and prev[0] isnt ':'
```
xixixao: This is for attributes following bare attributes.
```jsx
<Component scrollable x="this would blow up" />
```
I'll see if I'm missing this test case.

src/lexer.coffee
```diff
@@ -475,6 +485,70 @@ exports.Lexer = class Lexer
...
+      else if firstChar is '>'
+        # Ignore terminators inside a tag
+        @pair '/>' # As if the current tag was self-closing
+        origin = @token '}', '}'
+        @token ',', ','
+        {tokens, index: end} =
+          @matchWithInterpolations INSIDE_CSX, '>', '</', CSX_INTERPOLATION
+        @mergeInterpolationTokens tokens, {delimiter: '"'}, (value, i) =>
+          @formatString value, delimiter: '>'
+        match = CSX_IDENTIFIER.exec @chunk[end...]
```
xixixao: Good catch.

GeoffreyBooth:
@xixixao or @helixbass, if one of you feels like resolving the notes above, I’d love to get this merged into `2` soon and released as the next beta. I don’t think this should wait on the destructuring object spreads PR. As much as destructuring object spreads feels almost essential to this because idiomatic JSX uses it extensively, technically it’s not required; and adding that can happen post-2.0.0, because it’s not a breaking change, whereas this PR _is_ a breaking change.

So the sooner we get this merged in, the sooner we can potentially release 2.0.0 out of beta. Destructuring object spreads can follow whenever it’s ready (and that PR isn’t especially close to merging in at the moment).

xixixao:
@GeoffreyBooth Done. I agree that this can go before spreads, but it needs them badly. Is anyone working on them? We could add them to CSX first before the general support, but I don't know which way we want the dots to go.
`<div {...props} />`
or
`<div {props...} />` CS style? I guess depends on the eventual object syntax.

GeoffreyBooth:
Thanks @xixixao. Destructuring object spreads are in progress at #4493, and its author @zdenko just posted today that he’s resumed working on it. That said, his first attempt got fairly substantial feedback, such that major chunks of it need to be redone. I suspect he could use some help. The branch he’s working in is hosted at my repo, and I’ve just invited you as a collaborator. Maybe coordinate your efforts at #4493?

For now we’re keeping the CoffeeScript spread syntax of the dots to the right. It’s been in the back of my mind to get around to submitting a PR that makes it optional which side they’re on, so that people coming from ES can type them that way too. I’ve treated that as a much lower priority though.

src/lexer.coffee
```diff
@@ -48,6 +48,7 @@ exports.Lexer = class Lexer
     @seenExport = no             # Used to recognize EXPORT FROM? AS? tokens.
     @importSpecifierList = no    # Used to identify when in an IMPORT {...} FROM? ...
     @exportSpecifierList = no    # Used to identify when in an EXPORT {...} FROM? ...
+    @includesCSX = no            # Used to optimize CSX checks.
```
GeoffreyBooth: What does this mean exactly? That we’re inside a CSX block? That we’ve seen a CSX opening delimiter?

src/nodes.coffee
GeoffreyBooth: @xixixao Could you please reply to this, and also explain what’s going on with the escaping here. Per @lydell’s earlier comment, why is the escaping so simple? Why can we ignore things like escaped backslashes?

GeoffreyBooth:
@xixixao I pushed some commits to do the refactor I had mentioned earlier, of DRYing up the `toJS`/`eqJS` test helper. I also did a pass for style.

I noticed that `lexer.coffee`’s `addCSXToken` method was unused, so I removed it; and all the tests still pass. Is this surprising?

I think there are just two comments (noted above) that need responses to. If nothing needs to change as a result of those notes, we should at least put your explanations into the code as comments. Aside from those notes, though, this looks just about ready.

BTW @jashkenas @lydell and others: this PR puts us at 1,001 tests. 🎉

GeoffreyBooth:
General question: this PR introduces a breaking change where people will now need to add spaces around `<` and `>` operators, to disambiguate those operators from CSX delimiters. Do we want to improve the detection or error messages for the old, previously allowed syntax?

```bash
$ ./bin/coffee -bpe 'less = a<b'
[stdin]:1:10: error: missing />
less = a<b
         ^
```

DanielRamosAcosta:
@GeoffreyBooth I would recommend adding better error messages.

src/lexer.coffee
xixixao: I think it's pretty clear from the code.

src/nodes.coffee
xixixao: See the comment below.

src/lexer.coffee
xixixao: Maybe the only thing to clarify, although it should be pretty obvious: This is set for the whole program when we first encounter CSX. It is used so that programs without CSX tags don't get the perf hit from CSX checks.

xixixao:
I am thinking of lessening the requirement for files that don't include CSX, which should alleviate most problems people would run into. Give me a sec.

xixixao:
@GeoffreyBooth Also please do not remove the helper for the tests, it will clutter the file (unless you are just renaming it and moving it elsewhere, which would be good).

src/nodes.coffee
```diff
@@ -2871,19 +2870,17 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
...
 
   isNestedTag: (element) ->
-    @csx and
-    (exprs = element?.body?.expressions) and
-    exprs.length is 1 and
-    (call = exprs?[0]) instanceof Call and
-    call.csx
+    exprs = element?.body?.expressions
+    call = exprs?[0]
+    @csx and exprs and exprs.length is 1 and call instanceof Call and call.csx
```
xixixao: @GeoffreyBooth I'd prefer to revert this. It has different semantics, and although it should be ok, the point was that this is only happening for `@csx`. If you want to keep the style, add an early return.

xixixao:
@GeoffreyBooth I updated the original post at the top with the new semantics. I think this should be good now (we can still amend the error message, but I think it's gonna be more likely that you messed up the CSX, and mentioning comparison might be confusing).

src/lexer.coffee
GeoffreyBooth: So this is just a flag that gets set to true the first time a CSX tag is encountered? How are we saving any processing, if we have to check for CSX tags in order to determine that we can skip CSX processing?

src/nodes.coffee
GeoffreyBooth: Ah okay, we can put `return unless @csx` at the top. What I was trying to achieve was to move the assignments out of the conditionals. As a general style rule I’m trying to avoid assignments inside conditionals (like `if (a = 1)`).

GeoffreyBooth:
@xixixao Thanks. So what is the rule for when a space is required before or after `<` or `>`? For the documentation I’m thinking of saying something like this:

> A space is required before or after the `<` and `>` operators unless you are comparing a number or the compiler can determine that you aren’t intending to type an XML tag; but in general, it’s safest to just always put a space between the `<` and `>` operators and whatever they’re comparing.

Is that correct? What is the exact rule per the parsing now, if we wanted to spell it out?

xixixao:
@GeoffreyBooth For the documentation purposes, I would say:
> A space is now required after the `<` operator, to disambiguate it from a CSX opening tag.

And the details are:

> Because a CSX opening tag starts an expression, the first tag in the file cannot be directly preceded by another expression, so `y<x`, `1<x` and `(y)<x` are still valid uses of the `<` operator if CSX is not used. The compiler does not allow mixing these non-spaced `<` operators after CSX is used.

Which makes me realize I missed `y[key]<x` case, so let me fix that before merging. Too bad I have to do this in the parser, where it's quite hacky (later on I could just test the node type).

GeoffreyBooth:
While you're at it, can you please look into `includesCSX`. I'm wondering why it's not named`seenCSX` like the others, first of all; and why we need it at all. If it's only a performance optimization, is it really speeding anything up?

xixixao:
I can rename it. The check for csx is expensive and run often, that's why I want to avoid it for code that is not using CSX.

xixixao:
@GeoffreyBooth I fixed the indexing and used a counter, which actually allows the unspaced comparison everywhere, which makes the complete documentation much simpler:

> A space is now required after the `<` operator if it is preceded by a space, to disambiguate it from a CSX opening tag.

`y<x`, `1<x`, `(y)<x` and `y[0]<x` are still allowed everywhere (including inside CSX interpolations).

GeoffreyBooth:
@xixixao That’s a great improvement. Sorry to give you so many notes, but I think the final product has gotten better. This is looking pretty great!

Is there anything left to address? Are we ready to merge? @lydell?

xixixao:
The only other thing in my mind is that this will make it harder (heresy incoming) to support type annotations with generics. The current approach would have to be completely redone, I'm afraid (essentially going back to running JSX as a separate pass). There isn't much interest in Flow/TypeScript typing here though, and so for now I think this is fine.

GeoffreyBooth:
I’d say let’s tackle that challenge when we have some idea of what form of static types we want, and then a rough plan for how to implement it. No reason to hold up this PR for some potential future feature that we may decide we don’t even want.

@lydell, any last notes before we merge this?

jashkenas:
> No reason to hold up this PR for some potential future feature that we may decide we don’t even want.

FWIW — which is perhaps not very much — I feel **very, very strongly** about not adding "static" types to CoffeeScript. Period.

We already have TypeScript, and forks like this: https://www.npmjs.com/package/typed-coffee-script

GeoffreyBooth:
I was being diplomatic. The obstacles to static typing are significant. Let's please not get into it here.

Any last notes before we merge?

vendethiel:
> We already have TypeScript, and forks like this: https://www.npmjs.com/package/typed-coffee-script

I agree in general, but I'm not sure a fork of Redux that hasn't been updated since 2014 counts.

lydell:
I don’t have much more to say here. I’d say we merge it and sort out things as we find them (if any).

GeoffreyBooth:
It’s done. JSX is in CoffeeScript. Somewhere, an angel just broke a nail.

GeoffreyBooth:
Time to write the documentation for this. Anyone want to volunteer what it should be? Or at least what the example in the docs should be? Ideally it would be something that shows moderately complicated JSX, but takes advantage of one or more of CoffeeScript’s unique features such as array comprehensions.

DanielRamosAcosta:
I have this [component](https://github.com/DanielRamosAcosta/mondieu/blob/d37a615228779756ac43e49e2f21d6dacf60a330/app/components/Rate/Rate.cjsx) in my experimental project, if it's useful to you. It renders a rating like this:

image

DanielRamosAcosta:
I'm only getting empty stars with that code... I'm looking into it

lydell:
(A lesser known feature of CoffeeScript is that you can leave out the `foo in` part of `for` loops if you don't need it:

```coffee
for [1..3]
  do knock
```
)

DanielRamosAcosta:
@GeoffreyBooth  I've written some [unit tests](https://github.com/DanielRamosAcosta/mondieu/blob/coffescript/app/components/Rate/Rate.test.coffee) for my component and [refactor with some of your improvements](https://github.com/DanielRamosAcosta/mondieu/blob/coffescript/app/components/Rate/Rate.cjsx). The component you proposed I think it should be something like this:

```cjsx
renderStarRating = ({ rating, maxStars }) ->
  <aside title={"Rating: #{rating} of #{maxStars} stars"}>
    {for wholeStar in [0...Math.floor(rating)]
      <Star className="wholeStar" key={wholeStar} />}
    {if rating % 1 isnt 0
      <Star className="halfStar" />}
    {for emptyStar in [Math.ceil(rating)...maxStars]
      <Star className="emptyStar" key={emptyStar} />}
  </aside>
```

GeoffreyBooth:
I see you use many more braces, which I was hoping to avoid. Are they necessary to get it to work?

DanielRamosAcosta:
Yes 😢 , React only renders the return value of each statement inside the braces...

GeoffreyBooth:
Okay, I’ve updated my example: https://rawgit.com/GeoffreyBooth/coffeescript/2-docs/docs/v2/#jsx

```coffee
renderStarRating = ({ rating, maxStars }) ->
  <aside title={"Rating: #{rating} of #{maxStars} stars"}>
    {for [0…Math.floor(rating)]
      <Star className="wholeStar" />}
    {if rating % 1 isnt 0
      <Star className="halfStar" />}
    {for [Math.ceil(rating)…maxStars]
      <Star className="emptyStar" />}
  </aside>
```

Does this work?

DanielRamosAcosta:
That works, but you'll get React warnings becouse when you render arrays [it is required to asign the "key" prop to the component](https://facebook.github.io/react/docs/lists-and-keys.html). This is the warning:

`Warning: Each child in an array or iterator should have a unique "key" prop. Check the top-level render call using <aside>. See https://fb.me/react-warning-keys for more information.`

GeoffreyBooth:
Man, this is becoming a messy example then. I was hoping to use a loop comprehension with JSX, because loop comprehensions are something that ES lack (though `.forEach` or `.map` come close) and therefore we could demonstrate an advantage of CSX over JSX. But this is looking overly complicated. Any ideas for anything simpler?

askucher:
Did you try? https://hackernoon.com/react-livescript-livescript-2-0-164d35ca5373

microdou:
@GeoffreyBooth
Great that `CSX` and `destructuring object spreads` finally get merged into beta3!
However, spreads inside csx brakets (`<div a="x" {...props}></div>`) are not working yet, still throwing error.

zdenko:
@xixixao @GeoffreyBooth 
Not sure how should this compile.
In JSX this
```javascript
// props = {id: "a", title: "some title"}
<div {...props}></div>
```
compiles to
```javascript
<div id="a" title="some title"></div>
```

I don't have much experience with JSX, but it seems that `<div  {Object.assign({},props)}></div>` will not work ([example](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=react%2Cstage-3&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&playground=true&code_lz=G4QwTgBADmD2UGcIF4IG8C-AoUkAmKEAPHgJbDoDyARgFYCmAxgC4B0ICCpA5gHYAUmADTQ4iAJQYAfEQD0ZYFKA&stage=0)).

GeoffreyBooth:
Could we just compile
```coffee
<div a="x" {props...}></div>
```
to
```js
<div a="x" {...props}></div>
```
? Is that really an object spread in there, as opposed to generic spread syntax like we’d find in function parameters or an array?

zdenko:
I was thinking the same. 
In React (and probably other libraries) ` <div {...props}></div>` is compiled to `React.createElement("div", props);`.

microdou:
@GeoffreyBooth 
It seems like `{...props}` inside `<>` in `JSX` is a special notation different from the ES6 generic spread syntax. `{...props, a: "x"}` or `{...props1, ...props2}` inside `<>`don't work.
You are right: a special treatment to `{props...}` inside `<>` of `CSX` is the way to go.

microdou:
Works in `JSX`:
```javascript
<div {...props} />
```
```javascript
<div {...props1} {...props2} />
```
```javascript
newProps = {...props, a: "x"}  // when generic spread syntax is supported
<div {...newProps} />
```
```javascript 
<div { ...{...props, a: "x"} } />  // when generic spread syntax is supported
```
Don't work in `JSX`:
```javascript
<div {...props, a: "x"} />
<div {...props1, ...props2} />
```


Accordingly, `CSX` should be able to support the following:
```javascript
<div {props...} />
```
```javascript
<div {props1...} {props2...} />
```
```javascript
newProps = {props..., a: "x"}  // when generic spread syntax is supported
<div {newProps...} />
```
```javascript 
<div { {props..., a: "x"}... } />  // when generic spread syntax is supported
```

GeoffreyBooth:
Implemented via #4607.

askucher:
Did you try https://github.com/askucher/lsxc

Pug/Jade like syntax

GeoffreyBooth:
Looks like this really works: https://iainhouston.com/blog/exploring-react-with-coffeescript-2.html

Issue #2051, opened 2012-01-18
Lexer: Incorrect JS generated for function called with an object with a conditional property


barshow:
# working

```coffee
alert
  val: "hello"
  val2: "all good"
  val3: "works" if true
```

compiles correctly to

``` javascript

alert({
  val: "hello",
  val2: "all good",
  val3: true ? "works" : void 0
});
```
# incorrect compilation

```coffee
alert
  val3: "works" if true
  val: "hello"
  val2: "all good"
```

compiles incorrectly to 

``` javascript

alert(true ? {
  val3: "works"
} : void 0, {
  val: "hello",
  val2: "all good"
});
```

GeoffreyBooth:
As of 2.0.0-beta4 this compiles as expected:

```js
alert({
  val3: true ? "works" : void 0,
  val: "hello",
  val2: "all good"
});
```

PR #4554, opened 2017-05-24
Fix #4544: postfix conditional on first line of implicit object

### LINKED ISSUE #4544: Lexer: Postfix conditionals in object literals
*This is a repost of #3534, with additional detail since @GeoffreyBooth closed that issue as unclear. 
 Some valuable discussion, including potential workarounds, can be found on the original issue.*

There seems to be some inconsistency in the handling of postfix conditionals within nested object literals. For instance:

```coffeescript
# this compiles correctly
one =
  foo:
    baz: 1337
    bar: 42 if qux
```

The output, as one would expect, is:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var one;

  one = {
    foo: {
      baz: 1337,
      bar: qux ? 42 : void 0
    }
  };

}).call(this);
```

However, this (apparently equivalent code) fails to compile at all (`error: unexpected newline`):

```coffeescript
# this does not compile
two =
  foo:
    bar: 42 if qux
    baz: 1337
```

The expected behavior is that this would compile to the following:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var two;

  two = {
    foo: {
      bar: qux ? 42 : void 0,
      baz: 1337
    }
  };

}).call(this);
```

The general principle applied here is that swapping two lines of input at the same nesting level of an object should result in those lines simply swapped in the output.

When adding intervening function calls, the results are even more confusing.  For example:

```coffeescript
# this compiles correctly
three =
  foo: call
    baz: 1337
    bar: 42 if qux
```

This compiles to the expected:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var three;

  three = {
    foo: call({
      baz: 1337,
      bar: qux ? 42 : void 0
    })
  };

}).call(this);
```

However, the very similar

```coffeescript
# this compiles incorrectly
four =
  foo: call
    bar: 42 if qux
    baz: 1337
```

Compiles to the strange:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var four;

  four = {
    foo: call(qux ? {
      bar: 42
    } : void 0, {
      baz: 1337
    })
  };

}).call(this);
```

Whereas one would expect it to likewise result in a simple swap of the nested object lines:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var four;

  four = {
    foo: call({
      bar: qux ? 42 : void 0,
      baz: 1337
    })
  };

}).call(this);
```

In particular, the results as generated seem to break the expectation that keys at the same nesting level in the source will be put into a single object of the output.

Finally, this similar example, which seems to have pretty obviously wrong semantics:

```coffeescript
five =
  foo: call
    bar: 1337
  baz: 42 if qux
```

This compiles to:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var five;

  five = qux ? {
    foo: call({
      bar: 1337
    }),
    baz: 42
  } : void 0;

}).call(this);
```

But one would expect it to compile to:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var five;

  five = {
    foo: call({
      bar: 1337
    }),
    baz: qux ? 42  : void 0
  };

}).call(this);
```

One might try to get the expected semantics by reversing the order of the `foo` and `baz` keys, like so:

```coffeescript
six =
  baz: 42 if qux
  foo: call
    bar: 1337
```

But of course, as with example two above, this results in an `error: unexpected newline`.  In case it wasn't clear from the previous examples, this should compile to:

```js
// Generated by CoffeeScript 1.12.5
(function() {
  var five;

  five = {
    baz: qux ? 42  : void 0,
    foo: call({
      bar: 1337
    })
  };

}).call(this);
```

Hopefully this provides enough details to make clear what the issue is as well as the expected behavior, but if not, feel free to ask any clarifying questions!

helixbass:
Fixes #4544. Also includes some little `rewriter.coffee` cleanups

The reason for the behavior @couchand describes in #4544 is that postfix `if`s are currently treated as implicit-object-literal-closing if they're on the first line of the object literal so that things like `x = a: b if c` work correctly. But it's assuming that it's a single-line object literal, so if the expected compilation described by @couchand is desired (which I think it is, ie you should be able to swap two lines of an object literal without anything weird happening) then this fixes it by sniffing ahead to see if we appear to be on the first line of a multi-line object literal

Added the examples given by @couchand as well as a few more cases to `test/objects.coffee`

@GeoffreyBooth I figured you'd want this targeted against `2`?

@GeoffreyBooth @lydell let me know if I should justify the correctness of any/all of the little cleanup refactorings

### DIFF
--- a/src/rewriter.coffee
+++ b/src/rewriter.coffee
@@ -45,16 +45,18 @@ exports.Rewriter = class Rewriter
     i += block.call this, token, i, tokens while token = tokens[i]
     true
 
-  detectEnd: (i, condition, action) ->
+  detectEnd: (i, condition, action, opts = {}) ->
     {tokens} = this
     levels = 0
     while token = tokens[i]
-      return action.call this, token, i     if levels is 0 and condition.call this, token, i
-      return action.call this, token, i - 1 if not token or levels < 0
+      return action.call this, token, i if levels is 0 and condition.call this, token, i
       if token[0] in EXPRESSION_START
         levels += 1
       else if token[0] in EXPRESSION_END
         levels -= 1
+      if levels < 0
+        return if opts.returnOnNegativeLevel
+        return action.call this, token, i
       i += 1
     i - 1
 
@@ -65,25 +67,23 @@ exports.Rewriter = class Rewriter
     @tokens.splice 0, i if i
 
   # The lexer has tagged the opening parenthesis of a method call. Match it with
-  # its paired close. We have the mis-nested outdent case included here for
-  # calls that close on the same line, just before their outdent.
+  # its paired close.
   closeOpenCalls: ->
     condition = (token, i) ->
-      token[0] in [')', 'CALL_END'] or
-      token[0] is 'OUTDENT' and @tag(i - 1) is ')'
+      token[0] in [')', 'CALL_END']
 
     action = (token, i) ->
-      @tokens[if token[0] is 'OUTDENT' then i - 1 else i][0] = 'CALL_END'
+      token[0] = 'CALL_END'
 
     @scanTokens (token, i) ->
       @detectEnd i + 1, condition, action if token[0] is 'CALL_START'
       1
 
-  # The lexer has tagged the opening parenthesis of an indexing operation call.
+  # The lexer has tagged the opening bracket of an indexing operation call.
   # Match it with its paired close.
   closeOpenIndexes: ->
     condition = (token, i) ->
-      token[0] in [']', 'INDEX_END']
+      token[0] is [']', 'INDEX_END']
 
     action = (token, i) ->
       token[0] = 'INDEX_END'
@@ -140,7 +140,7 @@ exports.Rewriter = class Rewriter
     @scanTokens (token, i, tokens) ->
       [tag]     = token
       [prevTag] = prevToken = if i > 0 then tokens[i - 1] else []
-      [nextTag] = if i < tokens.length - 1 then tokens[i + 1] else []
+      [nextTag] = nextToken = if i < tokens.length - 1 then tokens[i + 1] else []
       stackTop  = -> stack[stack.length - 1]
       startIdx  = i
 
@@ -159,30 +159,35 @@ exports.Rewriter = class Rewriter
       # class declaration or if-conditionals)
       inImplicitControl = -> inImplicit() and stackTop()?[0] is 'CONTROL'
 
-      startImplicitCall = (j) ->
-        idx = j ? i
+      startImplicitCall = (idx) ->
         stack.push ['(', idx, ours: yes]
         tokens.splice idx, 0, generate 'CALL_START', '('
-        i += 1 if not j?
 
       endImplicitCall = ->
         stack.pop()
         tokens.splice i, 0, generate 'CALL_END', ')', ['', 'end of input', token[2]]
         i += 1
 
-      startImplicitObject = (j, startsLine = yes) ->
-        idx = j ? i
+      startImplicitObject = (idx, startsLine = yes) ->
         stack.push ['{', idx, sameLine: yes, startsLine: startsLine, ours: yes]
         val = new String '{'
         val.generated = yes
         tokens.splice idx, 0, generate '{', val, token
-        i += 1 if not j?
 
       endImplicitObject = (j) ->
         j = j ? i
         stack.pop()
         tokens.splice j, 0, generate '}', '}', token
         i += 1
+
+      implicitObjectContinues = (j) =>
+        nextTerminatorIdx = null
+        @detectEnd j,
+          (token) -> token[0] is 'TERMINATOR'
+          (token, i) -> nextTerminatorIdx = i
+          returnOnNegativeLevel: yes
+        return no unless nextTerminatorIdx?
+        @looksObjectish nextTerminatorIdx + 1
 
       # Don't end an implicit call on next indent if any of these are in an argument
       if inImplicitCall() and tag in ['IF', 'TRY', 'FINALLY', 'CATCH',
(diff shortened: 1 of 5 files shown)


GeoffreyBooth:
`2` is great, thanks. This looks quite good. Yes thanks for the cleanups, `rewriter.coffee` is a beast and can use all the cleanup it can get 😄

Does this fix any of the other lexer or related open issues? Thanks for tackling this!

lydell:
Wow! Just glancing over the diff I have no idea what's going on, but it all looks sensible. Good job! (I guess 😉 )

GeoffreyBooth:
Yes, @lydell echoes my thoughts exactly ;)

Just merge this?

helixbass:
@GeoffreyBooth yup figured this was a good excuse to jump into the belly of the beast and understand how `addImplicitBracesAndParens()` works. Took a brief look through open issues, this also closes #2051. Thanks for organizing and tagging them "Lexer:", I'll start going through those more closely as the lexer is what I'm most familiar with at this point. On a somewhat related note, I'd like to start following all activity in the Coffeescript repo (and perhaps `coffeescript6/discuss`), is Github notification emails the best way to get notified of stuff that happens? I've found the Slack Github integration feed easy to browse/scroll but looks like you need to own the repos you want to follow

@lydell @GeoffreyBooth I'll go through now and leave specific comments in case you do want to vet anything more closely but I appreciate your trust that I'm not doing anything nonsensical

GeoffreyBooth:
@helixbass thanks! Yes, that was the idea behind the tagging. You might also want to skim through the “Invalid JS”-tagged ones, they might be related to the lexer/rewriter stuff.

I don’t know how to get better notifications. I watch both repos but I only get emails when I’m mentioned or I’ve already participated in a thread. If you find a better way please let me know. On a related note, the plan is to merge `discuss` into this repo when 2.0.0 is released.

I’ve never worked in the rewriter, so you know it far better than I do. Reading your code there’s nothing sloppy and nothing obviously wrong, so if you say it fixes the bug and doesn’t break any tests, I’m kind of taking your word for it 😄

src/rewriter.coffee
```diff
@@ -197,7 +202,7 @@ exports.Rewriter = class Rewriter
         #  1. We have seen a `CONTROL` argument on the line.
         #  2. The last token before the indent is part of the list below
         #
-        if prevTag not in ['=>', '->', '[', '(', ',', '{', 'TRY', 'ELSE', '=']
+        if prevTag not in ['=>', '->', '[', '(', ',', '{', 'ELSE', '=']
```
helixbass: safe to remove `TRY` from this list b/c an implicit call could never be at the top of the stack (ie `inImplicitCall()` could never be true) when we've just seen `TRY` b/c the preceding clause would've caused a `CONTROL` to get pushed onto the stack (in the previous `scanTokens()` iteration). Effectively, this list can safely be mutually exclusive from the list above (`['IF', 'TRY', ...]`)

src/rewriter.coffee
```diff
@@ -236,13 +241,6 @@ exports.Rewriter = class Rewriter
       #       a: b
       #       c: d
       #
-      # and
```
helixbass: this comment seems to reference an old syntax that no longer compiles (now the *first* arg has to be an implicit object in order for indentation to imply an implicit call?)

src/rewriter.coffee
```diff
@@ -316,7 +314,8 @@ exports.Rewriter = class Rewriter
           # Close implicit objects such as:
           # return a: 1, b: 2 unless true
           else if inImplicitObject() and not @insideForDeclaration and sameLine and
-                  tag isnt 'TERMINATOR' and prevTag isnt ':'
+                  tag isnt 'TERMINATOR' and prevTag isnt ':' and
+                  not (tag is 'POST_IF' and startsLine and implicitObjectContinues(i + 1))
```
helixbass: the actual fix. If we see a `POST_IF` on the first line of an implicit object, check if it looks like that implicit object is actually multi-line rather than assuming it's an inline-implicit-object-ending `POST_IF`

src/rewriter.coffee
```diff
@@ -159,31 +159,36 @@ exports.Rewriter = class Rewriter
...
         tokens.splice idx, 0, generate '{', val, token
-        i += 1 if not j?
 
       endImplicitObject = (j) ->
         j = j ? i
         stack.pop()
         tokens.splice j, 0, generate '}', '}', token
         i += 1
 
+      implicitObjectContinues = (j) =>
```
helixbass: looks ahead (from the `POST_IF` in question) for a following line that should be treated as part of the same implicit object (ie is at the same indentation level, is the next line (after any properly balanced intervening nested stuff), and is itself looking like an object literal)

test/objects.coffee
```diff
@@ -439,6 +439,12 @@ test "#3216: For loop declaration as a value of an implicit object", ->
   arrayEq ob.b, test
   arrayEq ob.c, test
   arrayEq ob.d, test
+  byFirstKey =
```
helixbass: added a couple tests since the `@insideForDeclaration`/#3216 stuff really only kicks in on the first line of an implicit object

lydell:
Well, I’m convinced! I’d say let’s just merge it!

src/rewriter.coffee
```diff
@@ -159,31 +159,36 @@ exports.Rewriter = class Rewriter
...
+        nextTerminatorIdx = null
+        @detectEnd j,
```
vendethiel: How costly is that lookahead here? It's called every single time we rewrite an implicit object literal, right? Is there a perf. impact?

helixbass: @vendethiel nope only gets called when it sees a `POST_IF` on the first line of an implicit object literal

vendethiel: Thank you for the explanation.

Issue #4528, opened 2017-04-24
REPL: Warning about underscore


xixixao:
I can't find an existing issue for this:
```
$ node -v
v7.9.0
$ coffee -v
CoffeeScript version 1.12.5
$ coffee
coffee> 4
Expression assignment to _ now disabled.
4
coffee>
```
Node REPL is constantly warning that assignment to underscore is now disabled. Latest Node, latest CS.

vendethiel:
See https://github.com/nodejs/node/issues/5431 for rationale

xixixao:
@vendethiel Might explaining specifically for CoffeeScript? I am not using any explicit _ assignment I'm aware of.

vendethiel:
Look at the repl code. We generate an assignment to `_` to force it to be an expression.

lydell:
Couldn't we assign to some other name, then?

vendethiel:
Yes, we need to do this. My comment didn't mean we should close the issue.

GeoffreyBooth:
We do this because the Node REPL itself treats `_` as a special case, assigning it to the last returned value:

```bash
$ node
> a = 1
1
> _
1
```

As @vendethiel noted above, see https://github.com/nodejs/node/issues/5431.

xixixao:
Can we just assign to `__` ? This is super annoying in the REPL.

GeoffreyBooth:
Resolved by #4559.

PR #4559, opened 2017-06-04
Fix REPL assignment

xixixao:
Is no one ever using the REPL? The variable was never documented, but Node does this on its own, so you can access the last expression with both `_` and `__`. No more annoying messages!!!

### DIFF
--- a/lib/coffee-script/repl.js
+++ b/lib/coffee-script/repl.js
@@ -38,7 +38,7 @@
           return results;
         })();
         ast = CoffeeScript.nodes(tokens);
-        ast = new Block([new Assign(new Value(new Literal('_')), ast, '=')]);
+        ast = new Block([new Assign(new Value(new Literal('__')), ast, '=')]);
         js = ast.compile({
           bare: true,
           locals: Object.keys(context),
--- a/src/repl.coffee
+++ b/src/repl.coffee
@@ -33,7 +33,7 @@ replDefaults =
       ast = CoffeeScript.nodes tokens
       # Add assignment to `_` variable to force the input to be an expression.
       ast = new Block [
-        new Assign (new Value new Literal '_'), ast, '='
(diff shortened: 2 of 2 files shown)


GeoffreyBooth:
I don’t understand what’s going on here. See https://github.com/jashkenas/coffeescript/issues/4528, which you were a part of too.

We’re assigning to `_` explicitly because Node does, in order to override it. Why would we assign to `__`?

lydell:
So … the Node.js REPL has this feature that `_` always refers to the value of the last expression, kind of like the “Ans” feature on some calculators (if you’ve ever used one of those).

As mentioned in a comment in the code, we add `_ =` before the inputted CoffeeScript code to force it to be an expression. So a variable name had to be chosen there. I guess `_` was chosen since the Node.js REPL would clobber that variable anyway. However, newer versions of Node.js added the warning message whenever `_` was assigned to.

So this PR simply uses another name (`__`) for the force-expression-assignment.

Everything works as before, with the exception that `__` will now always be clobbered and we finally get rid of that annoying message.

I just tried this PR and it seems to be working fine:

```
$ ./bin/coffee
coffee> 1 + 1
2
coffee> _
2
coffee> __
2
coffee> 
```

GeoffreyBooth:
Okay, @lydell if you think it’s okay then I’m fine with it. What about tests?

lydell:
I’m not sure there very much to test? And `cake test` passes locally on this branch.

GeoffreyBooth:
A test that `__` does, in fact, pull the last value. Or maybe we don’t care about this enough to bother?

lydell:
If @xixixao feels like writing a test then go for it, but if not it shouldn’t hold up this PR. There was no test before either.

xixixao:
I'm not bothered.

nigon135:
I'm not bothered

Issue #4436, opened 2017-02-06
[CS2] Nodes: Compilation of named classes in expressions is broken


connec:
```coffee
new class
# new class {}

new class A
# new (A = )class A {};
```

iAmMaxmaxmaximus:
error to
```coffeescript
new class Cat
```

connec:
That's definitely less esoteric.

GeoffreyBooth:
Fixed by #4569.

PR #4561, opened 2017-06-06
[CS2] Restore bound class methods via runtime check to avoid premature calling of bound method before binding

helixbass:
As discussed in coffeescript6/discuss#84 by @ryansolid @connec @GeoffreyBooth: restore bound methods with a runtime check to avoid silent breakage of code that under `2` would call the method before it was bound (ie a base class constructor calling a derived class bound method in a non-`this`-preserving way eg as a callback)

Much of the code and tests here are restored from pre-#4530, I'll comment in the code on new stuff I've added

As I [commented](https://github.com/coffeescript6/discuss/issues/84#issuecomment-306605288) in coffeescript/discuss#84, there seems to be a question of how to handle anonymous classes (with bound methods). If my suggestion (3) (give anonymous classes with bound methods a name) is deemed the best option, I can try and implement that behavior (instead of currently (2) (forgo the runtime check for anonymous classes))


src/nodes.coffee
```diff
@@ -1362,6 +1365,9 @@ exports.Class = class Class extends Base
         @ctor = method
       else if method.isStatic and method.bound
         method.context = @name
+      else if method.bound
+        @boundMethods.push method.name
+        method.constructorName = @name
```
helixbass: Attach the class name (so that it can be passed as an argument to the runtime check in the method body), not sure if this is already available to the `method` somehow?

Removed `method.bound = false`, as now it needs to know about its bound-ness when compiling itself

src/nodes.coffee
```diff
@@ -1399,7 +1405,8 @@ exports.Class = class Class extends Base
       method.name = new (if methodName.shouldCache() then Index else Access) methodName
       method.name.updateLocationDataIfMissing methodName.locationData
       method.ctor = (if @parent then 'derived' else 'base') if methodName.value is 'constructor'
-      method.error 'Methods cannot be bound functions' if method.bound
+      method.error 'Cannot define a constructor as a bound function' if method.bound and method.ctor
+      # method.error 'Cannot define a bound method in an anonymous class' if method.bound and not @name
```
helixbass: For illustration, this would be option (1) (disallow bound methods in anonymous classes) of the three options I [outlined](https://github.com/coffeescript6/discuss/issues/84#issuecomment-306605288) for dealing with anonymous classes with bound methods

src/nodes.coffee
```diff
@@ -2140,6 +2154,9 @@ exports.Code = class Code extends Base
     wasEmpty = @body.isEmpty()
     @body.expressions.unshift thisAssignments... unless @expandCtorSuper thisAssignments
     @body.expressions.unshift exprs...
+    if @isMethod and @bound and not @isStatic and @constructorName
```
helixbass: The addition of the actual runtime check (using new `_boundMethodCheck` helper). Currently deals with anonymous classes via option (2) (forgo the runtime check for anonymous classes) with `and @constructorName`

src/nodes.coffee
```diff
@@ -3097,6 +3114,13 @@ exports.If = class If extends Base
 
 UTILITIES =
   modulo: -> 'function(a, b) { return (+a % (b = +b) + b) % b; }'
+  _boundMethodCheck: -> "
```
helixbass: helper that throws runtime error if bound method called with a `this` that isn't an instance of its own class

as discussed in coffeescript6/discuss#84, this results in a stack trace where the offending method is in the second line

test/classes.coffee
```diff
@@ -1574,3 +1670,76 @@ test "CS6 Class extends a CS1 compiled class with super()", ->
   eq B.className(), 'ExtendedCS1'
   b = new B('three')
   eq b.make(), "making a cafe ole with caramel and three shots of espresso"
+
+test 'Bound method called as callback before binding throws runtime error', ->
```
helixbass: all the new tests are here, including tests that the runtime error actually occurs for the edge case and that bound methods work ok in other situations. Quite possible there are some other scenarios that should be tested, if explained I'll add more tests

helixbass:
@GeoffreyBooth ok updated (based on your suggestion) to run the check against the parent class (and only run it for child class bound methods). Had to get a little fancy with making sure that bound methods can properly reference the base class (which could be an arbitrary `Expression`) so I'll go through and make some more code comments

src/nodes.coffee
```diff
@@ -1362,6 +1380,10 @@ exports.Class = class Class extends Base
         @ctor = method
       else if method.isStatic and method.bound
         method.context = @name
+      else if method.bound
+        @boundMethods.push method.name
+        @setParentRef(o)
```
helixbass: if we have any bound methods, see if the parent class expression should be cached to a variable ref and save the (possibly updated) parent on `method` so that it can be used in the runtime check

src/nodes.coffee
```diff
@@ -1279,12 +1279,18 @@ exports.Class = class Class extends Base
...
-    @ctor ?= @makeDefaultConstructor() if @externalCtor
+    @ctor ?= @makeDefaultConstructor() if @externalCtor or @boundMethods.length
     @ctor?.noReturn = true
 
+    @proxyBoundMethods o if @boundMethods.length
+
     o.indent += TAB
 
     result = []
+    if @assignParentRef
```
helixbass: if we have bound methods and the parent class expression should be cached (as determined by `setParentRef()`), turn the class expression into a `(ref = parent.class().expression, class Child extends ref {...})`-style expression

src/nodes.coffee
GeoffreyBooth: Is there a reason this helper’s name starts with an underscore? Our other helpers’ names don’t.

src/nodes.coffee
```diff
@@ -2140,6 +2169,9 @@ exports.Code = class Code extends Base
     wasEmpty = @body.isEmpty()
     @body.expressions.unshift thisAssignments... unless @expandCtorSuper thisAssignments
     @body.expressions.unshift exprs...
+    if @isMethod and @bound and not @isStatic and @parentClass
```
helixbass: the runtime check, basically `_boundMethodCheck(this, parentClassRef)`. Will only be added for child class bound methods

src/nodes.coffee
helixbass: @GeoffreyBooth I think @connec was mimicking babel's `_classCallCheck`. Updated to `boundMethodCheck`

src/nodes.coffee
```diff
@@ -1399,7 +1421,7 @@ exports.Class = class Class extends Base
       method.name = new (if methodName.shouldCache() then Index else Access) methodName
       method.name.updateLocationDataIfMissing methodName.locationData
       method.ctor = (if @parent then 'derived' else 'base') if methodName.value is 'constructor'
-      method.error 'Methods cannot be bound functions' if method.bound
+      method.error 'Cannot define a constructor as a bound function' if method.bound and method.ctor
```
GeoffreyBooth: In error messages, let’s use the phrasing “bound (fat arrow) function”. [MDN calls `=>` functions “arrow functions,”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) so that’s what most people will probably know them as. I think we should be explicit about “fat” arrow to distinguish from `->`.

src/nodes.coffee
```diff
@@ -1418,6 +1440,13 @@ exports.Class = class Class extends Base
 
     ctor
 
+  proxyBoundMethods: (o) ->
+    @ctor.thisAssignments = for name in @boundMethods by -1
+      name = new Value(new ThisLiteral, [ name ]).compile o
+      new Literal "#{name} = #{name}.bind(this)"
```
GeoffreyBooth: Rather than `new Literal`, shouldn’t this be `new Assign` etc.?

src/nodes.coffee
```diff
@@ -3097,6 +3129,13 @@ exports.If = class If extends Base
 
 UTILITIES =
   modulo: -> 'function(a, b) { return (+a % (b = +b) + b) % b; }'
+  boundMethodCheck: -> "
+    function(instance, Constructor) {
+      if (!(instance instanceof Constructor)) {
+        throw new Error('Bound instance method accessed before binding')
```
GeoffreyBooth: Add semicolon.

test/classes.coffee
GeoffreyBooth: Checking for errors goes in `error_messages.coffee`. Not sure why, we should probably refactor them out of there into the relevant test files, but for now that’s where all such tests go.

GeoffreyBooth:
@connec and @ryansolid, do you mind reviewing?

helixbass:
@GeoffreyBooth ok pushed changes based on your comments

GeoffreyBooth:
@helixbass Please merge `2` into your branch and resolve conflicts.

helixbass:
@GeoffreyBooth ok merged `2`

GeoffreyBooth:
@ryansolid, does this PR solve your issue?

@connec, do you mind reviewing? I think as soon as this gets merged in we can release a new beta, unless you think #4493 is close enough to be worth waiting for. All tests pass.

connec:
@GeoffreyBooth I'll take a look this evening, all going well. I think it would be worth waiting for #4493, once the tests are in it should be pretty much there.

src/nodes.coffee
connec: Whilst this is a nice way of doing (so I'm not suggesting changing it), but the [`.cache`](https://github.com/jashkenas/coffeescript/blob/master/src/nodes.coffee#L105) mechanism could be used to instead generate code like:

```
class Child extends (ref = parent.class().expression) {}
```

GeoffreyBooth:
@connec You raise a good point. Wouldn’t checking just for `null` also resolve the case of a grandchild class (a class that `extend`s a class that itself `extend`s a class)? Seems simpler and cleaner. If someone wanted to check specifically that they were calling via the correct instance, they can add that check explicitly.

I thought there is no anonymous class issue, since it’s not possible to `extend` an anonymous class?

connec:
> I thought there is no anonymous class issue, since it’s not possible to `extend` an anonymous class?

Honestly I didn't follow that discussion very well. My understanding was that was why the check switch to the parent class (because there's no usable reference for the `constructor`)? I guess the issue is with code like:

```coffee
class Base
  constructor: ->
    @boundMethod()

new class extends Base
  boundMethod: =>
    console.log 'bound!'
```

In that case there's no reference to use for the `instanceof` check, hence the decision to use the parent reference.

GeoffreyBooth:
Well switching to checking for `null` should solve that, right?

connec:
Yup!

helixbass:
@connec so if a not-yet-bound method is run in a callback style (in any environment), `this` will always be `undefined`? As opposed to how `this` is the `window` object if a normal function which references `this` is called (normally) in the browser?

If that's true, then your suggestion and thinking make sense to me, as the `!= null` check would presumably be less expensive than an `instanceof` check, and as you say it'll simplify the implementation and sort of bypass this issue of "what kind of guarantee are we making about `this`?" (which is what I was talking about in my [last point](https://github.com/coffeescript6/discuss/issues/84#issuecomment-306624310))

helixbass:
Optimistically updated the PR to use @connec's suggestion (to simplify the `boundMethodCheck()` to just throw if `this == null`)

Preserved the logic to only add the `boundMethodCheck()` on child class bound methods (still based on the thinking that our breaking cases only involve child class bound methods being called in a parent class constructor), but if we think that all bound methods should have the check that's easy to change too

ryansolid:
I originally thought I could perhaps get away with null/undefined. However sometimes `this` can be something else which is why I added the instanceof check. Like in the case of the addEventListener example we were throwing around `this` will actually be the dom element if you don't bind it. I would not be surprised if there were other exceptions.

helixbass:
@ryansolid ok yup so reverted to `instanceof` check against parent class. @connec your understanding of the parent class/anonymous class stuff seems quite clear - basically that using the parent class as the `Constructor` in the `instanceof` check avoids the issue of always needing a reference to the child class (ie the one defining the bound method) - which gets tricky for an anonymous child class - but provides a weaker guarantee. If it's easy enough to figure out how to force a reference to be created for an anonymous class (that's a child class with bound methods), then I see no reason not to do it that way (ie use the child class for the `instanceof` check) and provide the stronger guarantee. But if it's not easy/worth it, I'd think the parent class `instanceof` check is satisfactory since it should cover all of the edge cases we were concerned about where there's not some very specific `.call()`/`.apply()` manipulation happening?

helixbass:
@connec I guess the other aspect that you're discussing is that in theory since the bound version of the method is placed on the instance rather than the prototype you could directly access the prototype method and do stuff with it in an unbound way. And that by adding these `boundMethodCheck()`s we're limiting the ability to do that b/c the `boundMethodCheck()` is inside the prototype method. To me it seems like a very questionable thing to do since it goes against the intuitive understanding of a bound method. But I guess in theory there could be 1.x code out there that does this that would then break by introducing the `boundMethodCheck()`? Eg in 1.x you can do this:
```
class A
  b: => console.log(this)

a = new A

# from an instance
a.__proto__.b.call(window)
# or the class
A::b.call(window)
```
I don't think any of the proposed bound-method compilations would allow this? The only way I can think of to allow it would be in the constructor to, on the prototype, replace the `boundMethodCheck()`ing version of the method with a non-checking version of the method (I guess you could do it right before calling `.bind()` for the method?). But that seems insane? My instinct is that it's ok to effectively disallow bypassing the bound-ness of a bound method and that probably noone's been doing that in practice

connec:
Good point @ryansolid! Sorry for the misdirection.

@helixbass I agree, I don't think there's any need to remove the bound method checks, I was mostly building an argument for removing the `instanceof` check (which it now turns out is necessary :cry:).

It should be possible to force a reference around [here](https://github.com/helixbass/copheescript/blob/0c914111fc9b04a11e19b774d6caa4ee3f6b3fb2/src/nodes.coffee#L1314) - if the class has a parent and any bound methods and *doesn't* have a `@name` (no stable reference), we can inject a `freeVariable` assignment for the constructor and use that as the name in the instanceof check. To use the example from my above comment, the desired output would be:

```js
var Base, ref

class Base {
  constructor () {
    this.boundMethod()
  }
}

new (ref = class extends Base {
  constructor () {
    super(...arguments)
    this.boundMethod = this.boundMethod.bind(this)
  }

  boundMethod () {
    boundMethodCheck(this, ref)
    console.log 'bound!'
  }
})
```

Perhaps a viable way of doing it would be to `?=` assign `@variable` in [`Class::proxyBoundMethods`](https://github.com/helixbass/copheescript/blob/0c914111fc9b04a11e19b774d6caa4ee3f6b3fb2/src/nodes.coffee#L1481), something like:

```coffee
proxyBoundMethods: (o) ->
  @variable ?= new IdentifierLiteral o.scope.freeVariable 'class'
```

Also, and this might be a bit controversial, we could avoid attaching more things (e.g. `parentClass`) to `Code` by injecting the method in, e.g., `proxyBoundMethods` or another function from `Class` (e.g. `Class` injects the `boundMethodCheck`, rather than `Code` compiling with it).

connec:
@helixbass something like this might work:

```coffee
# nodes.coffee:1421
else if method.bound
  @boundMethods.push method

...

# nodes.coffee:1481
proxyBoundMethods: ->
  @variable ?= new IdentifierLiteral o.scope.freeVariable 'class'
  for method in @boundMethods by -1
    method.body.unshift new Call(boundMethodCheck, [new Value(new ThisLiteral), @variable])
    ...
```

helixbass:
@connec thanks for the suggestions, here's a couple things I ran into:
- injecting the `boundMethodCheck()` `Call` from eg `proxyBoundMethods()` seems problematic b/c there are other things that get `unshift()`ed inside `Code#compileNode`, including `@`-params, so eg `f: (@g) =>` would compile to `f(g) { this.g = g; boundMethodCheck(...); }`, when we want the `boundMethodCheck()` to run first
- the `@variable ?= new IdentifierLiteral o.scope.freeVariable '_class'` inside `proxyBoundMethods()` seems to mostly work, but two issues that I saw related to this:
  - your example where you're immediately instantiating an anonymous class actually pointed out a current bug on `2` - `x = new class A` compiles to `x = new (A = )class A {};`. This looks to be a result of the dummy `Assign` to `''` (inside `if @variable` in `Class#compileNode`). I'm not sure what the right way to resolve this bug is, it seems like it'd be avoided if it used a real `Assign` to the class but that looks like some refactoring
  - since the class `@variable` is a `SimpleAssignable`, it looks like we'd have to cache a reference to the `@variable` in some cases now that we'd be referencing it multiple times (in `boundMethodCheck()`s). eg `class a().b` is allowed

xixixao:
What are the perf implications of this runtime check?

(also it seems like the instanceof check is not bulletproof, leading to even more subtle bugs?)

helixbass:
@xixixao can you give an example of a bug under the instanceof check? From my understanding, the only way a bound method could get called with a `this` other than its instance would be if:
  1. inside the constructor of a base class, a child class bound method gets `.call()`/`.apply()`'d supplying a `this` that's a different instance of the (child or parent, pending discussion) class
  2. at any time (ie before or after the method is bound to the instance), you explicitly pluck the method off the prototype and `.call()`/`.apply()` it, again with "related" `this`

Given that both of these require a conscious modification of `this` to something specific, I don't see it as ripe for bugs from existing or new code. Certainly the main case we've been trying to address is when bound methods could get called in a callback style (where `this` might typically be `undefined`, a DOM element, ...) before being bound

As far as perf, I don't know how to gauge its impact but you'd basically be incurring the cost of an `instanceof` check (inside a call to a helper utility function) each time a bound method of a child (ie `extend`-ing) class gets called

helixbass:
@connec ok so I updated to basically use your suggestion, with caveats based on my previous reply:
- kept the `boundMethodCheck()` call generation inside `Code#compile_node` for the reason described above
- added logic (inside `proxyBoundMethods()`) to get a reference to the class-name-expression (using `.cache()`, thanks for the pointer) since as explained above it could need caching
- I guess the bug I found (broken `new class A` compilation) was already reported as #4436 - once that's fixed, the corresponding `new <anonymous child class w/ bound methods>` case here should presumably work as well - I added a commented-out test to `test/classes.coffee` with instructions to uncomment once #4436 is fixed

So ya doing it this way seems cleaner (got rid of all the parent-ref stuff) and provides the tighter guarantee, thanks for helping me figure out how to handle the anonymous class case

GeoffreyBooth:
So where does this stand now? @helixbass do you consider all edge cases resolved, and this PR is ready to merge per one final review by @connec?

helixbass:
@GeoffreyBooth yup the broken case of `new <anonymous child class w/ bound methods>` is equivalent to #4436 (ie should work once #4436 is fixed) so if @connec thinks that the updated implementation looks good I'd think it's ready

connec:
I put in a PR for #4436 over in #4569.

src/nodes.coffee
```diff
@@ -1353,8 +1355,9 @@ exports.Class = class Class extends Base
       @variable.error message if message
     if name in JS_FORBIDDEN then "_#{name}" else name
 
-  walkBody: ->
+  walkBody: (o) ->
```
connec: Correct me if I'm wrong, but this doesn't seem to be needed any more

src/nodes.coffee
```diff
@@ -1456,6 +1461,18 @@ exports.Class = class Class extends Base
 
     ctor
 
+  proxyBoundMethods: (o) ->
+    @variable ?= new IdentifierLiteral o.scope.freeVariable '_class' if @parent
+    [@variable, @variableRef] = @variable.cache o unless @variableRef?
```
connec: This needs to account for `@variable` being null if there's no `@parent`, e.g.

```coffee
class
  a: =>
```

currently throws "Cannot read property 'cache' of null".

src/nodes.coffee
```diff
@@ -1456,6 +1461,18 @@ exports.Class = class Class extends Base
...
+
+    @ctor.thisAssignments = for method in @boundMethods by -1
```
connec: Nitpick: I know the `by -1` was restored from the old version, but I suspect it must be from a time when the expressions were being `unshift`ed directly into the function body. Currently they are being assigned in reverse source order (not that it matters, besides aesthetics). Would be nice to fix this :smile: 

```coffee
class extends Foo
  a: =>
  b: =>
```
```js
_class = (class extends B {
  constructor () {
    super(...arguments)
    this.b = ...
    this.a = ...
  }
})
```

src/nodes.coffee
connec: My suggestion would be to move both lines into an `if @parent`, e.g.

```coffee
if @parent
  @variable ?= ...
  [@variable, @variableRef] = ... unless ...
```

connec:
Think this is pretty much there, apart from the small points above. If #4569 gets merged first we can uncomment that test too and be done with it :smile:

GeoffreyBooth:
#4569 is merged, you can uncomment that test.

helixbass:
@GeoffreyBooth @connec ok merged in `2` (post-#4569-merge), uncommented test, and made fixes. The one thing I had to adjust b/c of #4569 was to pull the class `@variable` caching out of `proxyBoundMethods()` (into the body of `Class#compileNode`) so that it runs before the `Assign`  to `@variable`. If you'd prefer it stylistically I can put that into its own method so eg `@cacheVariable() if @boundMethods.length and @parent`

@connec thanks for reading so closely no need to apologize for nitpicking/being pedantic

connec:
Looks great @helixbass, and thanks for being such a gracious contributor 😄  This LGTM now 👍

We could consider dropping the `@parent` check in the future if we want to add the `boundMethodCheck` to base classes (for consistency). Not sure what you, @ryansolid, or @GeoffreyBooth think about that?

GeoffreyBooth:
So we would output the `boundMethodCheck` helper for all classes that have a bound method? Is this because we need the check on a parent class too, not just the `extend`ing class? And since a class could be extended in another file, we don’t necessarily know if a parent class is extended or not?

If we only need the check on classes that `extend`, I think it just adds noise to output it for all classes as opposed to only the classes that need it.

jashkenas:
Ooh — I *really* don't like adding a new "utility" helper function for this. Especially when it's being called at runtime.

(FWIW, which ain't much)

GeoffreyBooth:
@jashkenas The alternative is removing the feature entirely, which was our first approach (and released as 2.0.0-beta2). But I’ve come around to a runtime utility helper being a lesser evil than removing this feature, as clearly it’s being used. The utility helper is only output when the feature is used, so if you never use it, it’s the same as if we removed it 😉

connec:
> If we only need the check on classes that extend, I think it just adds noise to output it for all classes as opposed to only the classes that need it.

Only `extend`ed classes could encounter the original issue. If people start depending on bound methods being uncallable from other contexts (for some reason), it might be worth making them consistent across the board.

> I really don't like adding a new "utility" helper function for this. Especially when it's being called at runtime.

I'm not sure how much of https://github.com/coffeescript6/discuss/issues/84 you've read @jashkenas but a runtime check is the only way (we can think of) to guard against bound methods being called unbound through parent class constructors. The only other options we know of are:

- Leave it as is. Will work most of the time, but bound methods called in parent constructors won't be bound (the original issue reported in https://github.com/jashkenas/coffeescript/issues/4497).

- Remove bound methods. This was actually merged in https://github.com/jashkenas/coffeescript/pull/4530, but it has since been regarded as a bad move, as it is a feature that some users depend on quite heavily.

jashkenas:
Please feel free to ignore me, but:

My understanding of the gist of the problem is this: `"if the super constructor tries to use a bound function, it will get the unbound one"`

Instead of doing a utility function and a runtime check, can we not, at compile time, look at all method calls within the super constructor, and if they appear to be calling the name of a function we know is bound, throw the error as you try to compile?

It wouldn't catch tricky uses like `this['meth' + 'od']()` within the constructor, but would get the point across. No?

GeoffreyBooth:
> Instead of doing a utility function and a runtime check, can we not, at compile time, look at all method calls within the super constructor, and if they appear to be calling the name of a function we know is bound, throw the error as you try to compile?

There’s no way to know a method is bound if it’s from an `extend`ed class that’s imported:
```coffee
import parent from 'somewhere'

class child extends parent
  constructor: ->
    @foo()
```

jashkenas:
Yeah, I get that — it just seems like you get 80-95% percent of the mistakes, and educate people that **"In CoffeeScript 2, you can't call bound methods from class constructors"** with the compile time check. More complex cases would slip through the cracks. But do what you think is best!

connec:
It would fail to catch the original mistake though (from #4497) wherein the bound method is called from `initialize`, which is an unbound method called from the constructor of the super class (which is imported). I don't think we have the facilities to track that kinda thing statically.

Is your concern around the runtime check mostly the performance impact?

GeoffreyBooth:
> If people start depending on bound methods being uncallable from other contexts (for some reason), it might be worth making them consistent across the board.

This sounds very speculative. I think unless you can come up with a realistic example, we should probably limit it to the scenario in question. We can always expand it later.

jashkenas:
> Is your concern around the runtime check mostly the performance impact?

My concern is mostly about the inherent ugliness of the thing. If we had always had to insert such a check — we never would have added bound methods in the first place. CS2 is an opportunity to take them out. Leaving them in (most of the time) with a nasty runtime check to make sure that you're not running into a catch-22 seems like a half-assed implementation.

Seems like we can do better. But if this is the best option...

helixbass:
If the runtime check is deemed inappropriate on philosophical/perf/... grounds, I'll argue again that leaving bound methods as-is (with documentation of the breaking change) is better than removing bound methods. I still think it's theoretically equivalent to the `@`-constructor args breaking change in `2`, eg this breaks in `2` for basically the same reason:
```
class B
  constructor: ->
    @f()

class A extends B
  constructor: (@f) ->
    super()

a = new A -> console.log 'called'
```
So in theory I'd say just document it the same way as however you're documenting the `@`-constructor args breaking change

However I think the use case that started the discussion (things like Backbone `initialize()`) is much more realistic and has merited trying to find a graceful solution. But as @connec says there's no way it could be caught at compile-time - I'd be down to try and add some compile-time checks but think that actually most wouldn't be catchable if we'd basically be looking for normal calls made directly inside a parent class constructor that's defined in the same file as the child class

Removing bound methods would be a hugely-breaking change for the sake of a little elegance. I'll again invoke Python 3 as a very clear warning against doing this. While I respect the feeling that bound methods might've never been added in the first place had this been the initial situation, CS2 is not a new "initial situation", it bears the weight of all the code written under 1.x

So for bound methods that leaves me weighing the runtime check vs just warning about possibly subtly breaking code (hand-in-hand with a similar warning about `@`-constructor args). Personally I can easily stomach the runtime check from a perf and ugliness standpoint but I also deeply respect @jashkenas' staunch stance

connec:
> We can always expand it later.

Sorry for not being clear, @GeoffreyBooth, that was my suggestion all along :sweat_smile: 

---

I had another read of https://github.com/coffeescript6/discuss/issues/84 and there was a proposal that would avoid runtime checks by compiling `class Child extends Parent then method: => console.log @` to:

```js
class Child extends Parent {
  constructor () {
    super(...arguments)
    this.method = () => {
      return console.log(this)
    }
  }

  method () {
    throw new Error('Bound instance method accessed before binding')
  }
}
```

GeoffreyBooth:
That’s remarkably elegant. I feel like there must be a catch.

connec:
> there must be a catch

It [doesn't work with `super`](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=stage-2&targets=&browsers=&builtIns=false&debug=false&code=class%20Parent%20%7B%0A%20%20method%20()%20%7B%0A%20%20%20%20console.log('Parent!')%0A%20%20%7D%0A%7D%0A%0Aclass%20Child%20extends%20Parent%20%7B%0A%20%20method%20%3D%20()%20%3D%3E%20%7B%0A%20%20%20%20super().method()%0A%20%20%20%20console.log('Child!')%0A%20%20%7D%0A%7D&experimental=false&loose=false&spec=false&playground=true) is a pretty big one.

connec:
Yet another option could be to rename the prototype method (leaving it `undefined` when accessed before binding):

```js
class Child extends Parent {
  constructor () {
    super(...arguments)
    this.method = this.method$.bind(this)
  }

  method$ () {
    return console.log(this)
  }
}
```

This is a simplified version of [this proposal](https://github.com/coffeescript6/discuss/issues/84#issuecomment-302854336) without the runtime check. This behaviour would be equivalent to that of `@`-params in the same circumstances.

GeoffreyBooth:
> Yet another option could be to rename the prototype method

That looks ridiculous. If I saw that as my output I would wonder what the hell had happened. I suppose instead of `method$` you could rename it something like `method__temporaryNameToFoilAttemptAtCallsBeforeBinding` but I’m not sure that’s better than a runtime check. And either way, if I tried to call that method before binding, i.e. what the runtime check checks for, the runtime would throw an error because I tried to access an undefined variable. Which would be confusing, because in my original code, the variable was, in fact, defined. At least the runtime check tells me exactly what’s wrong.

Okay, so unless there’s anything else, can we merge this in? @jashkenas I’m sorry for the inelegance, but this appears to be the least offensive solution to the problem. If a better solution can be found, it can become a new PR. I don’t think backward compatibility should be sacrificed for the sake of elegance.

ryansolid:
I'd add the method renaming does havok to trying to resolve prototype chains. That proposal only worked because the properly named method is on the prototype too and all it was accomplishing was removing calling the check once the method was bound. However once you call super on an inherited bound method it ends up calling that function up the prototype chain which essentially lands us at this PR. I'm sure there is a way for this check to disappear post binding for inheritance chains that are less than 3 deep but I doubt it's worth the complexity.

So while I'm in full support of this PR (or leaving it largely as was pre #4530), if there was a directional precedence that made that impossible, @connec proposal pulling them into the constructor (breaking inheritance but keeping class instance bound methods) is by far the better approach out of all alternatives and is consistent with Typescript and possibly that new ES standard. But that would still be a significant breaking change which we are trying to avoid. So obviously very happy to just merge this PR and deal with that separately if needed.

jashkenas:
I think I agree with @helixbass that just leaving bound methods in, and documenting the breakage is a bit better than a nasty runtime check — but I'll defer to @GeoffreyBooth and @connec's call on this.

xixixao:
I'd add that ES6 compilation (whatever it's downsides) would seem like a better way to go for CS2 than this workaround.

CS1 > ES6 > this PR

GeoffreyBooth:
@jashkenas The problem is that we’ve already received bug reports for the “just leave them in” approach, which was released in 2.0.0-beta1. Granted we hadn’t documented the pitfalls, but I think it’s much safer for the community at large if we add the check; I would rather get complaints about “what’s this random helper I didn’t want in my code” than “my classes don’t work in CS2.” I can’t imagine too many CoffeeScript users really understand the ins and outs of prototypes and class compilation, and the differences in how CS2 classes are generated versus CS1. We can always remove the check later if we feel confident that it’s not needed.

@xixixao I don’t understand your comment. In CS2 we are outputting ES classes. ES classes don’t support bound methods, hence the question of what to output in their place.

@connec Aside from philosophical debate, is this PR ready to merge in?

xixixao:
@GeoffreyBooth I meant ESWhateverVersionThatSupportsThis, Babel:
```es6
var A = function A() {
  var _this = this;

  this.f = function () {
    return _this.x;
  };
};
```
Also you are still gonna keep getting "my classes don't work" tickets with this PR, it's just gonna be clear why they don't work. This PR doesn't change what you can and cannot do, it just makes the error obvious.

ryansolid:
@xixixao There is an idiomatic approach of writing every non-constructor or non-static class method as fat arrow bound so that regardless of how it was used you could set and forget and have it always bound to your class.  CS1 let you do this, and I'm sure it's a practice used religiously by atleast a few dev shops since it aligns with how most OOP languages work when it comes to class methods which was really the only thing to serve as basis if you go back a few years.

From my understanding that ES8 or maybe never proposal doesn't support both inheriting from bound methods or binding inherited methods and calling super.  Both really change things.  The first for library/framework writers since they would need to remove all bound methods from their inheritable classes.  Any inherited class code that was expecting this behavior could then also be broken. The second definitely breaks any of that application code that wants to bind an inherited method. It's arguable whether most of these methods actually need to be bound. It's possible to wrap any inherited method inside these new bound methods, but they essentially become something different as they are no longer part of the prototype chain.

@jashkenas 
The difficulty I have with the original issue that caused all this is twofold. Firstly the pattern that made it apparent I believe isn't really a class thinking forward. I have to assume methods like Backbone's initialize exists since there was no standard way of doing constructors. The code you put in initialize you could put after you call super in your child class constructor. It's such a small piece of the puzzle that we shouldn't restrict bound methods in anyway that you wouldn't restrict unbound methods, with the exception of them not being bound as expected. Unfortunately every compiler detection method mentioned doesn't seem to be so non-invasive.

Secondly, having it silently fail I can see being pretty bad. Not bad for those aware I suppose but because it spans over multiple files/classes people might not get what's happening. I mean all the control is really in the child inherited class arguably so it is definitely on the child class writer to do the right thing. But they won't know until it causes weird errors on execution.

connec:
@GeoffreyBooth yes, the PR looks good to me.

On 14 Jun 2017 8:00 pm, "Ryan" <notifications@github.com> wrote:

@xixixao <https://github.com/xixixao> There is an idiomatic approach of
writing every non-constructor or non-static class method as fat arrow bound
so that regardless of how it was used you could set and forget and have it
always bound to your class. CS1 let you do this, and I'm sure it's a
practice used religiously by atleast a few dev shops since it aligns with
how most OOP languages work when it comes to class methods which was really
the only thing to serve as basis if you go back a few years.

From my understanding that ES8 or maybe never proposal doesn't support both
inheriting from bound methods or binding inherited methods and calling
super. Both really change things. The first for library/framework writers
since they would need to remove all bound methods from their inheritable
classes. Any inherited class code that was expecting this behavior could
then also be broken. The second definitely breaks any of that application
code that wants to bind an inherited method. It's arguable whether most of
these methods actually need to be bound. It's possible to wrap any
inherited method inside these new bound methods, but they essentially
become something different as they are no longer part of the prototype
chain.

@jashkenas <https://github.com/jashkenas>
The difficulty I have with the original issue that caused all this is
twofold. Firstly the pattern that made it apparent I believe isn't really a
class thinking forward. I have to assume methods like Backbone's initialize
exists since there was no standard way of doing constructors. The code you
put in initialize you could put after you call super in your child class
constructor. It's such a small piece of the puzzle that we shouldn't
restrict bound methods in anyway that you wouldn't restrict unbound
methods, with the exception of them not being bound as expected.
Unfortunately every compiler detection method mentioned doesn't seem to be
so non-invasive.

Secondly, having it silently fail I can see being pretty bad. Not bad for
those aware I suppose but because it spans over multiple files/classes
people might not get what's happening. I mean all the control is really in
the child inherited class arguably so it is definitely on the child class
writer to do the right thing. But they won't know until it causes weird
errors on execution.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jashkenas/coffeescript/pull/4561#issuecomment-308526954>,
or mute the thread

GeoffreyBooth:
Thanks @helixbass!

@connec, I think there’s just #4493 left before we release 2.0.0-beta3. That PR looks possibly ready.

GeoffreyBooth:
@ryansolid Do you want to try compiling your project using the `2` branch?

ryansolid:
Yeah just tried it with a couple sample project, everything looks good from server through client. Even running through a tight loop rendering benchmark where the check method was present in a few locations the performance loss wasn't that bad about .15 ms per loop. So all in all I'm pretty content with this. Thank you @GeoffreyBooth, @connec, and @helixbass for making this happen.

helixbass:
Report from the wild: I just had the bound method runtime check kick in on some (new) code where I was dereferencing a bound method off `this` in code called by the parent constructor. So just had to postpone dereferencing, definitely saved me some debugging headaches

@ryansolid no problem, thanks for working through the alternatives and coming up with a good one. If you're curious about the performance difference if the bound check was inlined, would want to turn off the bound method check for code that you feel sure doesn't need it (eg in production), ... it shouldn't be that hard to roll your own slightly modified build, I'd be happy to help make that happen

GeoffreyBooth:
@helixbass I’m trying to write the documentation for this. This is the example I’m thinking of using, based on [this test](https://github.com/helixbass/copheescript/blob/b0fb8052d447ced35003a54476a8c2403c97ddbd/test/error_messages.coffee#L1589):

```coffee
class Base
  constructor: ->
    @onClick()      # This works
    clickHandler = @onClick
    clickHandler()  # This throws a runtime error

class Component extends Base
  onClick: =>
    console.log 'Clicked!', @
```

I’m testing this via `c = new Component()`. Looking at this, I have to wonder: shouldn’t the `@onClick()` line also throw an error? Why would that line work and not `clickHandler()`?

helixbass:
@GeoffreyBooth it works because it's being called normally (thus with expected `this`). It's only if calling a method "non-directly" eg as a callback (like `clickHandler()`) that it won't have the expected `this` and thus will fail the runtime check

GeoffreyBooth:
Ah, okay. So is the documentation correct or would you change anything?

helixbass:
@GeoffreyBooth ya that example illustrates the basic idea nicely and concisely

Goues:
Hello, is there a way to turn this off? I am migrating an older codebase to version 2 of coffescript that is using a `bare` transpilation to have smaller output and then it gets compiled by google closure library so the no files are wrapped in extra functions. However, this means that `boundMethodCheck` exists in global space and fails for all classes except the last one. I don't want to wrap everything in a wrapper function, that's redundant, and I trust our code enough to let it crash inszead of having this runtime check. Could there be a config option to skip injecting it?

GeoffreyBooth:
No, sorry, there are no config options for changing the behavior of the compiler (other than the few we already have, like `bare`). I suggest you refactor the classes in question to avoid needing the helper.

PR #4569, opened 2017-06-12
[CS2] Nodes: Fix immediately invoked named classes

### LINKED ISSUE #4436: [CS2] Nodes: Compilation of named classes in expressions is broken

connec:
Fixes #4436.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1300,21 +1300,22 @@ exports.Class = class Class extends Base
     parentName    = @parent.base.value if @parent instanceof Value and not @parent.hasProperties()
     @hasNameClash = @name? and @name is parentName
 
-    if executableBody or @hasNameClash
-      @compileNode = @compileClassDeclaration
-      result = new ExecutableClassBody(@, executableBody).compileToFragments o
-      @compileNode = @constructor::compileNode
-    else
-      result = @compileClassDeclaration o
+    node = @
 
+    if executableBody or @hasNameClash
+      node = new ExecutableClassBody node, executableBody
+    else if not @name? and o.level is LEVEL_TOP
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1301,20 +1301,19 @@ exports.Class = class Class extends Base
...
     if @variable
-      assign = new Assign @variable, new Literal(''), null, { @moduleDeclaration }
-      [ assign.compileToFragments(o)..., result... ]
-    else
-      result
+      node = new Assign @variable, node, null, { @moduleDeclaration }
+
+    @compileNode = @compileClassDeclaration
+    try
+      return node.compileToFragments o
```
GeoffreyBooth: Why is this in a `try` block? When would it ever throw an exception, and in what cases would we want to catch such an exception?

connec: This is just ensure the override to `compileNode` is undone. We don't want to catch it all (hence no `catch`).

connec:
Sorry for the force pushes @GeoffreyBooth, had to update authorship (reinstalled my OS :cry:).

src/nodes.coffee
connec: It's possible that we should have a separate `ClassDeclaration` node, which would avoid these shenanigans. That feels like a bit too much of a refactor for this issue though.

GeoffreyBooth: Okay, I trust you.

GeoffreyBooth:
@lydell, any notes on this? Looks good to me.

lydell:
I guess it’s fine.

Issue #4568, opened 2017-06-12
Incompatible code generated with 1.12.6 using function call followed by if statement.


jo-soft:
Until version 1.12.5 the code `foo bar if baz` transpiled to `if(baz){foo(bar)}` 
since 1.12.6 this code transpiles to `foo(baz ? bar: void 0)` which breaks old code.

But when i use `mood = greatlyImproved if singing` (from docs) it compiles to `if (singing) {mood = greatlyImproved;}` 
Is this indented? Maybe by 'Bugfixes for incorrect code generated by the ? operator within a termary if statement`? I cant find any Bug report for that.

bg,
Johannes

GeoffreyBooth:
[Are you sure?](http://coffeescript.org/#try:foo%20bar%20if%20baz)

connec:
Perhaps you somehow triggered it being parsed as `foo(bar if baz)`? It would help if you could paste a full example that's reproducible in [Try CoffeeScript](http://coffeescript.org/#try:foo%20bar%20if%20baz) (or a link).

jo-soft:
bigger example of code failing in 1.12.6 bug working in 1.12.5 and lower:
```
test:
  bam: () ->
    fn bar: {
      foo: 123
    } if not condition
```
results in 1.12.6 (example generated by: http://coffeescript.org/#try)
```
({
  test: {
    bam: function() {
      return fn({
        bar: !condition ? {
          foo: 123
        } : void 0
      });
    }
  }
});
```

results in 1.12.5. (example generated by [js2.coffee](http://js2.coffee):

```
({
  test: {
    bam: function() {
      if (!condition) {
        return fn({
          bar: {
            foo: 123
          }
        });
      }
    }
  }
});
```

both examples can also locally be reproduced by transpiling locally

vendethiel:
Paging @helixbass since it might be from #4554 (?).

helixbass:
@vendethiel not a bad guess but #4554 was only merged into `2`

GeoffreyBooth:
It’s from #4532, allowing implicit braces after a `return` or `export default`. Basically, in this code:
```coffee
test:
  bam: () ->
    fn bar: {
      foo: 123
    } if not condition
```
[...]

PR #4570, opened 2017-06-13
Fix #4564: indent closes implicit object

### LINKED ISSUE #4564: Lexer: Ending a “for of” or “for in” loop with a function in which you provide an object, causes the next line to result in an error
The following [code](http://coffeescript.org/#try:for%20key%20of%20test%20a%3A%205%0A%20add%5Bkey%5D%20%3D%20true) fails to compile:

```coffeescript
for key of test a: 5
 add[key] = true
```

`Error on line 2: unexpected [`

Adding parentheses [works](http://coffeescript.org/#try:for%20key%20of%20test(a%3A%205)%0A%20add%5Bkey%5D%20%3D%20true):
```coffeescript
for key of test(a: 5)
 add[key] = true
```
And the one-liner [works too](http://coffeescript.org/#try:add%5Bkey%5D%20%3D%20true%20for%20key%20of%20test%20a%3A%205):
```coffeescript
add[key] = true for key of test a: 5
```
I feel this is a bug or is this expected behaviour?

helixbass:
Fixes #4564 

The breaking example from that issue was:
```
for key of fn k: v
  a
```
The general case is when an implicit object is followed by an indent, eg this also breaks:
```
if k: v
  1
```

The solution was to make implicit objects act more like implicit calls as far as how they behave when they see an `INDENT`. Basically the exception to them behaving the same is when the `INDENT` follows the `:` of an object key (at which point clearly an `INDENT` should not close the implicit object). Otherwise they should behave the same and this allows us to close multiple open implicit calls and/or implicit objects when we see an `INDENT` (like in the first example)

To make implicit objects act more like implicit calls, we need to push `CONTROL`s (ie suppressors of `INDENT`-closes-implicits behavior) onto the `stack` not just when we're (immediately) inside an implicit call (ie `inImplicitCall()`) but also when we're (immediately) inside an implicit object (ie `inImplicitObject()`)

This led to a slightly cleaner solution to handle things like `a: for x in y then x` than what @xixixao did in #3328. Basically if we see a `FOR` right after a `:`, we explicitly push a `CONTROL` (rather than having a special case represented by `@insideForDeclaration` that thwarts attempted `endImplicitObject()`s, that's what `CONTROL` is there to do). So I was able to get rid of all the `@insideForDeclaration` stuff from that PR

@GeoffreyBooth if this is deemed potentially-breaking I can retarget it to `2`


GeoffreyBooth:
After #4568 I’m wondering if all fixes like this should just go on `2`. I don’t want to keep getting surprised by unexpected breaking changes.

GeoffreyBooth:
Assuming all the tests still pass, this looks good to me. I think we should probably target `2` though. I think we should put a stop to bugfixes for 1.x unless the bug in question is really egregious.

helixbass:
@GeoffreyBooth sure, retargeted to `2`

xixixao:
But this is a pretty bad bug in 1 no?

GeoffreyBooth:
> But this is a pretty bad bug in 1 no?

If it throws a compiler error, then by definition it isn’t. No one is relying on this bug for their current output, so there’s no 1.x code that needs this fix.

Issue #4563, opened 2017-06-07
Meta: Support Some Form of Static Typing?


xixixao:
Let's preface this with:

> @jashkenas: FWIW [...] I feel very, very strongly about not adding "static" types to CoffeeScript. Period.

Keeping that in mind, a scenario in which I think we could support some type system:

1) CoffeeScript does not have a type system. Neither static nor dynamic, besides what JS has already.
2) There should be minimal perf impact.
3) We should not have to add any more dependencies to the compiler.
4) If you don't want to use types, we will never force you, and we will not compromise the syntax of CoffeeScript for this support.

What I think we could do is support a syntax that could be used by an existing type system. Afaik there are two popular existing choices in ES6 ecosystem: TypeScript and Flow (add more if you know of any).

No reason to beat around the bush: TypeScript will be much harder to support. It's got more extra syntax to JS.

I imagine that the following code would be possible:

```coffee
foo = (x: ?number): string ->
  x ? "default string"
```

Which with Flow would report an error.

I think it might be interesting to figure out the upsides and downsides here and if there is any interest in having this supported. At this point it's very theoretical, as I'm not even sure how much effort it would take to support all of the typing syntax.

GeoffreyBooth:
I’m working on a PR to improve support for comments, to allow comments to “follow” the next or previous token into the final output. So something like this:
```coffee
foo = (x###: ?number###)###: string### ->
  x ? "default string"
```
___
```js
foo = function (x/*: ?number*/)/*: string*/ {
  return x || "default string";
}
```
The same logic could apply to backticks (embedded JS).

There are many problems to solve to get this to work. For one thing, it’s more or less straightforward to attach a comment (or embedded JS block) to a ”literal” token like `)`, or to an identifier like `x`, as these are output pretty much as is. The issue is what to do when attaching to a token like `->`; should the comment be next to the `{` or next to the `function`? Or a comment in the middle of something like `for item in items by 3`, where the source tokens have no direct corollaries in the output. This may not be solvable.

[...]

PR #4572, opened 2017-06-14
[CS2] Comments

### LINKED ISSUE #4290: Lexer: Block comment in array literal unexpectedly throws ‘unexpected newline’

GeoffreyBooth:
This PR aims to reimplement how we compile comments—both block (`### ###` or `/* */`) and end-of-line (`#` or `//`). This would close #4290 per the discussion in #4541. It would also help provide better support for plugins, as discussed in #4540, #4529 and #4563. Once we get this working with comments, the same techniques could apply to embedded JavaScript, which will also help plugins or pre-/postprocessors.

The basic idea is that when the lexer consumes a comment, instead of making a token for it, the comment data is attached to a nearby token as a property like our current `locationData`. The comments property hitches a ride on that token through the parser and rewriter, possibly getting moved onto a different token by the rewriter as appropriate; and then in `nodes.coffee` the comment gets output before or after the token it stowed away on.

Because the comment never becomes a token, the grammar never touches it; and so you can put block comments anywhere: `foo = 2 ### yo! ### + 3`. The tricky part is that not all tokens make it through the parser; and some that do get jumbled. The rewriter has significant work to do to make sure that the comment makes it across attached to the nearest surviving token, with enough information preserved that the comment can be output sensibly.

So, have you ever wondered what magic awaits in the rewriter, where streams of tokens are mutated while being looped over? Ever glimpsed `parser.js` and pondered the mysterious `yy` variable and gargantuan `switch` statement? Read `grammar.coffee` and considered how the hell the function at the top somehow gets rendered into JavaScript in `parser.js`, and what it does there? Well, this PR is for you!


vendethiel:
PR best viewed with `?w=1`..

jashkenas:
Amazing! Heroic work. Bravo.

This is so much of a better architectural approach to block comments than what we were doing previously — that I think it should definitely be merged once y'all have signed off on the details.

If it works well, should we start passing through line comments and printing them as well?

GeoffreyBooth:
> If it works well, should we start passing through line comments and printing them as well?

Yes, that’s the plan. Line comments are already getting forwarded along to `nodes.coffee`, I just haven’t yet written the code to output them.

GeoffreyBooth:
So I’ve gotten this to the point where only one test fails, and it’s because I can’t seem to get the indentation to stay correct across the code fragments. Anyone have any thoughts? It’s the commented-out test in `test/comments.coffee`.

#3132: Place block-comments nicely
Expected generated JavaScript to be:
```js
/**
 * A dummy class definition
 *
 * @class
 */
var DummyClass;

DummyClass = (function() {
  class DummyClass {

    /**
     * @constructor
     */

    constructor() {}

  };


  /**
   * Singleton reference
   *
   * @type {DummyClass}
   */

  DummyClass.instance = new DummyClass();

  return DummyClass;

})();
```
but instead it was:
```js
/**
 * A dummy class definition
 *
 * @class
 */
var DummyClass;

DummyClass = (function() {
  class DummyClass {
      /**
       * @constructor
       */
    constructor() {}

  };

  /**
   * Singleton reference
   *
   * @type {DummyClass}
   */
DummyClass.instance = new DummyClass();

  return DummyClass;

})();
```

I’m okay with declaring that block comments will no longer automatically have blank lines on either side of them; but:

1. the `@constructor` comment should really be at the same indentation as `constructor()`; and
2. `DummyClass.instance` should be indented (the same as `return` on the following nonblank line).

GeoffreyBooth:
Okay, I’ve gotten that last test to pass. The handling of indentation for comments is now much improved, though I’ve discovered that inserting `@tab` before a node is handled in several places around `nodes.coffee` (like in `Code` and in `Return`) so there might be more places left that require adjusting indentation. If I’ve missed any, though, they just get block comments with jagged indentation; there should be no impact on code execution. I’ve also gone through the grammar and expanded the list of tokens that I think don’t survive contact with the parser, and improved the handling of moving comments off such tokens.

This _might_ be all that needs to be done for block comments. The CoffeeScript codebase has very few such `###` comments, so it’s not the best corpus to use to ensure that this PR is safe for general projects. If anyone has a big project with lots of block comments in it, please compile your code with this PR and let me know if anything breaks 😉

Next up: line comments.

GeoffreyBooth:
Line comments are now output! With the tests all passing. You can see what this looks like by looking at the generated `lib/` JavaScript files in this PR.

The line comments aren’t output particularly _well_ just yet; they’re unindented, and they occasionally hug the end of the previous line when they should be starting a new line. But they don’t get output within strings, at least, so they shouldn’t break anything (I think).

Aside from fixing the formatting of line comments’ output, other remaining tasks include ensuring that all comments are, in fact, output; and adding lots more tests which I’m sure will expose edge cases I haven’t addressed yet. But this is getting close.

jashkenas:
Very exciting!

There are probably a bunch of old related dead tickets you could go comment on and make folks very happy about.

srb-:
Great stuff Geoffrey! When this makes it into the next beta I'll test it out with Flow using its comment type syntax. Could be interesting.

GeoffreyBooth:
Thanks @srb-. Can you post an example of what that might look like? It would make a good test.

GeoffreyBooth:
This PR already fixes #4290, though indentation isn’t correct:

```coffee
[
  ###  ###
  'foobar'
]
```
___
```js
[
  /*  */
'foobar'
];
```

So perhaps it’s already worth merging in, and improving the indentation can be a follow-on PR. Though I’m going to keep working on the indentation for awhile first.

GeoffreyBooth:
Line comments are indented now, and reading through the new compiled output it appears they’re always indented correctly. :crossed_fingers: Still to do:

- [ ] Sometimes line comments “trail” (start after a line of code, like `; // trailing comment`) when they should start a new line.
- [ ] Block comments in array literals, like the previous example, still screw up indentation. I’m sure there are other cases where indentation is messed up, I need to write more tests to find them.

GeoffreyBooth:
Okay, I’ve cleaned up the output of comments within array literals.

```coffee
  arr = [
    # How many stooges are there?
    3
    # Who’s on first?
    'Who'
    # Who?
    {
      firstBase: 'Who'
      secondBase: 'What'
      leftField: 'Why'
    }
  ]
```
___
```js
  var arr;

  arr = [
    // How many stooges are there?
    3,
    // Who’s on first?
    'Who',
    {
      // Who?
      firstBase: 'Who',
      secondBase: 'What',
      leftField: 'Why'
    }
  ];
```
I know, I’m mixing the Three Stooges with Abbott and Costello. Whatever.

At this point I think the only work left is just improving formatting, like indentation and making sure comments output in the correct order etc. But there’s a point of diminishing returns, where we’re never going to get perfect output. I’m not sure how to handle line comments within heregexes, for example. I think this is probably at the point where it’s ready for review, and I’m sure people will have constructive notes to improve this, and hopefully after your notes are addressed we can merge it in and leave “but _this_ could be formatted better” edge cases for later PRs.

All the tests pass. I got to remove lots of old code about the `Comment` node type, though I also had to add lots of new code for the `HereComment` and `LineComment` fragment types. Overall this is a huge PR. @jashkenas @srb- @vendethiel @lydell your notes are welcome!

src/helpers.coffee
```diff
@@ -111,12 +111,34 @@ buildLocationData = (first, last) ->
 # This returns a function which takes an object as a parameter, and if that
 # object is an AST node, updates that object's locationData.
 # The object is returned either way.
-exports.addLocationDataFn = (first, last) ->
+exports.addDataToNode = (parserState, first, last) ->
   (obj) ->
-    if ((typeof obj) is 'object') and (!!obj['updateLocationDataIfMissing'])
+    # Add location data
+    if obj?.updateLocationDataIfMissing? and first?
```
vendethiel: nitpick:

```coffee
if first?
  obj?.updateLocationDataIfMissing? buildLocationData(first, last)
```

src/lexer.coffee
```diff
@@ -301,17 +301,59 @@ exports.Lexer = class Lexer
...
+      matchIllegal = HERECOMMENT_ILLEGAL.exec comment
+      if matchIllegal
+        @error "block comments cannot contain #{matchIllegal[0]}",
+          offset: matchIllegal.index, length: matchIllegal[0].length
+
+      # Parse indentation or outdentation as if this block comment didn’t exist.
+      chunk = @chunk.replace "####{here}###", ''
+      # Remove leading newlines, like `Rewriter::removeLeadingNewlines`, to
+      # avoid the creation of unwanted `TERMINATOR` tokens.
+      chunk = @chunk.replace /^\n+/, ''
```
vendethiel: we use `@chunk` here, instead of the `chunk` we created 2 lines earlier. doesn't seem correct.

vendethiel: also, `newLine` matches whitespace-only lines, but not here. Does this have any incidence on anything?

src/lexer.coffee
```diff
@@ -301,17 +301,59 @@ exports.Lexer = class Lexer
...
+      @lineToken chunk
+
+      # Pull out the ###-style comment’s content, and format it.
+      content = here
+      content = content.replace /^(\s*)###/, ''
```
vendethiel: if `here` contains the `###` prefix/suffix, doesn't it mean `chunk = @chunk.replace "####{here}###", ''` L320 is effectively doing nothing? (since it'd try to replace `######...######`

vendethiel:
Tough one. Very good work. I can't try yet, so I wanted to know: what happens to:

```
///
# x
///
```

GeoffreyBooth:
> Tough one. Very good work. I can’t try yet, so I wanted to know: what happens to comments in heregexes?

They just get swallowed up and disappear. Your example generates the same output in this PR as it does in 1: `/(?:)/;`. I suppose they might be rescuable, similar to how I’m rescuing line comments inside code blocks that are within interpolated strings, but I’d rather push that off to a separate PR. Line comments within heregexes are an edge case, in my opinion. In the docs we should mention that the compiler will only make a best effort to output all comments.

src/helpers.coffee
```diff
@@ -111,12 +111,34 @@ buildLocationData = (first, last) ->
...
       obj.updateLocationDataIfMissing buildLocationData(first, last)
 
-    return obj
+    # Add comments data
+    unless parserState.tokenComments
+      parserState.tokenComments = {}
+      for token in parserState.parser.tokens when token.comments
+        tokenHash = "#{token[2].first_line}x#{token[2].first_column}-#{token[2].last_line}x#{token[2].last_column}"
+        unless parserState.tokenComments[tokenHash]?
```
GeoffreyBooth: I try to put the most likely condition first. In this case, it’s much more likely that we haven’t yet encountered this hash before, so that’s the primary flow and the `else` is the exception.

src/helpers.coffee
GeoffreyBooth: I’m not really a fan of the `fn? arg` notation. That goes a little too far for me, and as far as I can tell it’s undocumented.

src/lexer.coffee
GeoffreyBooth: Good catch. Actually `here` doesn’t include the `###` delimiters, so these two `replace` lines are redundant.

GeoffreyBooth:
Thanks @vendethiel. Any other notes?

vendethiel:
Just as a note:

```coffee
$ echo -n "->\n# lol\nx" | ./bin/coffee -bcs
// Generated by CoffeeScript 2.0.0-beta3
(function() {});

// lol
$  echo -n "->\n  # lol\n  x" | ./bin/coffee -bcs
// Generated by CoffeeScript 2.0.0-beta3
(function() {
  // lol
  return x;
});

```

I don't think it's too bad. I just was unsure.

vendethiel:
Erm, there's something I'd like to have explained, because I don't understand what it's doing...


```js
# Commented indented by two space, `x` indented by one:
echo -n "->\n  # y\n x" | ./bin/coffee -bcs
// Generated by CoffeeScript 2.0.0-beta3
(function() {
  // y
  return x;
});

# `y` indented by two spaces, `x` indented by one:
$ echo -n "->\n  y\n x" | ./bin/coffee -bcs 
// Generated by CoffeeScript 2.0.0-beta3
(function() {
  return y;
});

x;

```

src/nodes.coffee
```diff
@@ -497,6 +523,103 @@ exports.Block = class Block extends Base
...
+    for fragment, fragmentIndex in fragments
+      # Insert comments into the output at the next or previous newline.
+      # If there are no newlines at which to place comments, create them.
+      if fragment.precedingComments
+        # Determine the indentation level of the fragment that we are about
+        # to insert comments before, and use that indentation level for our
+        # inserted comments. Because some fragments begin with a space, e.g.
+        # ` + 1`, we look for either a tab character or at least two spaces
+        # to detect indentation. If someone is actually using one-space
+        # indentation, well, they’ll get unindented line comments.
```
vendethiel:
```js
$ echo -n "->\n # y\n x" | ./bin/coffee -bcs 
// Generated by CoffeeScript 2.0.0-beta3
(function() {
  // y
  return x;
});
```

Seems like the line comment *is* indented correctly, or am I misunderstanding something?

src/nodes.coffee
```diff
@@ -497,6 +523,103 @@ exports.Block = class Block extends Base
...
+        fragmentIndent = ''
+        for pastFragment in fragments[0...(fragmentIndex + 1)] by -1
+          indent = /\t+| {2,}/.exec pastFragment.code
```
vendethiel: doesn't that want to be `/^` instead, to only catch tabs/two spaces at the beginning?

vendethiel:
I commented on two things I'm not sure I understood. otherwise LGTM.

src/nodes.coffee
GeoffreyBooth: I think you’re correct on this. Thank you for looking so closely!

GeoffreyBooth:
@vendethiel Seems like you forgot to copy the last line of the output?

```js
echo -n "->\n# lol\nx" | ./bin/coffee -bcs
// Generated by CoffeeScript 2.0.0-beta3
(function() {});

// lol
x;
```

But I think that output is correct, you get the same thing if you take out the comment. (Minus the comment line, of course.)

src/nodes.coffee
GeoffreyBooth: It’s looking at the indentation of the previous token’s _generated_ code. In this case, the comment is attached to the `x` fragment, so it’s looking at the previous token’s code, which is `"  return "`—note the two spaces. (For some reason GitHub is showing this to me as one space, but it’s supposed to be `(quote)(space)(space)return(space)(quote)`). That’s how the `// y` line gets two spaces prepended to it.

This is inside `precedingComments`, so then it looks farther back to find the next previous newline, and inserts the `// y` line after it.

Does the CoffeeScript compiler always output JavaScript indented with two spaces? I had assumed it would follow whatever whitespace the original Coffee source had. If it’s always two spaces, that simplifies things considerably.

GeoffreyBooth:
Does anyone have an example of Flow with comments that we can test this with?

Or any projects that have copious comments, that we can try compiling with this? Especially a project with tests.

src/nodes.coffee
vendethiel: > Does the CoffeeScript compiler always output JavaScript indented with two spaces?

Yes.

GeoffreyBooth: Any particular reason for that? I’m wondering why it doesn’t automatically output tab characters instead. That would automatically give us smaller file sizes, though it makes things more difficult to read in debuggers and GitHub etc. Though I guess the compiler’s output obviously isn’t intended to be a minimal file size, since it’s outputting comments etc.

src/nodes.coffee
```diff
@@ -497,6 +523,103 @@ exports.Block = class Block extends Base
...
+      if fragment.precedingComments
+        # Determine the indentation level of the fragment that we are about
+        # to insert comments before, and use that indentation level for our
+        # inserted comments. Because some fragments begin with a space, e.g.
+        # ` + 1`, we look for either a tab character or at least two spaces
+        # to detect indentation. If someone is actually using one-space
+        # indentation, well, they’ll get unindented line comments.
+        fragmentIndent = ''
+        for pastFragment in fragments[0...(fragmentIndex + 1)] by -1
+          indent = /^\t+| {2,}/.exec pastFragment.code
```
vendethiel: Sorry I wasn't explicit enough – the `^` is needed in both alternations.

GeoffreyBooth: You mean `/^\t+|^ {2,}/`?

Couldn’t I change this to `/^ {2,}/`? Since if CoffeeScript always outputs as indented by two spaces, that should be all I need to look for.

danielbayley: Maybe `/^(\t+| {2,})/`?

> CoffeeScript always outputs as indented by two spaces

Shouldn't/couldn't it match the indentation of the input?

vendethiel: > Couldn’t I change this to /^ {2,}/? Since if CoffeeScript always outputs as indented by two spaces, that should be all I need to look for.
  
If you're parsing output, then yes, definitely.

danielbayley: Maybe the parser should detect the indentation from the first few lines right at the outset, and assign it to an `indentation` variable which then gets used everywhere else, including here?

vendethiel: There's really no need to bikeshed this at this point in Coffee's life, I think.

danielbayley: I don't actually care much personally since I just use 2 spaces for everything now anyway… but wouldn't the mentioned approach simplify/DRY up the codebase a little and cut down on some regexps?

GeoffreyBooth: At the very least it’s a separate PR. @danielbayley you might want to propose it in https://github.com/coffeescript6/discuss for feedback first.

GeoffreyBooth: Sometimes an indent appears not at the beginning of a token’s `code` property, such as after a comma for an array: `,\n\s\s` (using `\s` because GitHub swallows the spaces). 

`/^ {2,}/m` catches everything as far as I can tell.

GeoffreyBooth:
So I discovered an interesting issue when I tried to update the browser compiler: one should never rely on `fragment.type`, as `type` is defined by `parent?.constructor?.name` which can return values like `a` or `e` in the minified compiler. So my `fragment.type is 'HereComment'` lines weren’t working in the minified browser compiler. Rather than adjust the minifier options, I updated the code to set and test for an explicit flag, rather than relying on `type`. I also updated Babel and Babili.

GeoffreyBooth:
An updated version of the docs and browser compiler are here:
https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/

You can see all the docs examples now with line comments output, and you can play around with [Try CoffeeScript](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20Hello%2C%20world!%0Aalert%20'Hello%20CoffeeScript!'%0A%0Aif%20yes%20%23%20A%20true%20optimist!%0A%09%23%20Foo%3F%0A%09foo()%0A%09%23%20Bar!%0A%09bar()%20%23%20Baz!%0A%0Aarr%20%3D%20%5B%0A%20%20%23%23%23%20How%20many%20stooges%20are%20there%3F%20%23%23%23%0A%20%203%0A%20%20%23%20Who%E2%80%99s%20on%20first%3F%0A%20%20'Who'%0A%20%20%23%20Who%3F%0A%20%20%7B%0A%20%20%20%20firstBase%3A%20'Who'%0A%20%20%20%20secondBase%3A%20'What'%0A%20%20%20%20leftField%3A%20'Why'%0A%20%20%7D%0A%5D) to see this PR’s output interactively. Please let me know if you find any cases of egregiously bad output (or worse, output where the comments interfere with the code). @jashkenas @lydell @vendethiel @danielbayley

lydell:
I played around quickly with the browser compiler, and the only thing I've found so far is that line comments after a backslash line continuation seem to be dropped. Nice work!

https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:a%20%3D%201%20%2B%20%5C%20%23%20comment%0A%20%202

lydell:
This one is pretty fun: https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%7B%0A%20%20%23%20comment%0A%20%20%22%23%7B1%7D%22%3A%202%0A%7D

lydell:
Lost comment in splicing syntax: https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20comment%0Aa%5B1..2%5D%20%3D%20%5B1%2C%202%2C%203%5D

lydell:
These comments end up in an unexpected order: https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20one%0A%23%23%23%20two%20%23%23%23%0A%23%20three

connec:
This is probably OK, but comments at the end of an object literal get realigned: https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:o%20%3D%0A%20%20k%3A%201%0A%20%20%23%20some%20thing

srb-:
Sorry for the delay Geoffrey. And thanks for the in-browser way of testing this build. 

Here's the comment-style syntax of Flow that (ideally) we'd be able to support:

https://flow.org/en/docs/types/comments/

There's a couple of examples on the page... but for example, using the new compiler, I couldn't get this to work:

```
`// @flow` # needed at top of file to trigger flow

method = (str ###: string ### ) ->
  console.log str

method(44)  # should give Flow error
```

I haven't followed the discussion closely on this PR, but it looks like preserving the comment directly next to the parameter is not in the cards? Ideally it would output: `function method(str: /*: string */) { `

danielbayley:
@GeoffreyBooth Yeah great work so far 👍

Similat to @connec's example around object literals, these seem to behave oppositely from what might be expected…

https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:obj%20%3D%20%7B%0A%20%20one%3A%201%0A%20%20two%3A%202%0A%7D%20%23%20comment%0A%0Aobj%20%3D%0A%20%20one%3A%203%0A%20%20two%3A%204%0A%20%20%23%20comment%0A

GeoffreyBooth:
Thank you all for the examples! I’ll make new tests out of the ones I think I can improve.

@srb-: I tackled the Flow example first, since clearly that’s our highest priority. It took some shenanigans to get a block comment to output between the `)` that ends function parameters and the `{` that starts the function block. We now have a new `FuncGlyph` node type to hold such comments, and `->` and `=>` tokens now go through the parser (as opposed to turning into flags). This is the example I tested with, and what it turns into:

```coffee
# @flow

fn = (str ###: string ###, num ###: number ###) ###: string ### ->
  str + num
```
___
```js
// @flow
var fn;

fn = function(str/*: string */, num/*: number */)/*: string */ {
  return str + num;
};
```

Are there more complicated Flow examples I should try, or that are broken when you try in the updated [browser compiler](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20%40flow%0A%0Afn%20%3D%20(str%20%23%23%23%3A%20string%20%23%23%23%2C%20num%20%23%23%23%3A%20number%20%23%23%23)%20%23%23%23%3A%20string%20%23%23%23%20-%3E%0A%20%20str%20%2B%20num)?

As part of getting this to work, I also made sure that initial comments (`# @flow`) output above `var` declaration lines. @lydell, this solved the [splice example](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20comment%0Aa%5B1..2%5D%20%3D%20%5B1%2C%202%2C%203%5D), though its indentation is odd. None of the others are fixed yet.

GeoffreyBooth:
@lydell Comments after backslashes work in Node; what you’re seeing appears to be a bug in the browser compiler or CodeMirror. If you open the console for the browser and type `CoffeeScript.compile('1 + \ # comment\n2', {bare: true})` (note just a single lone backslash, not an escaped backslash), it works. I’m not sure how to replace two backslashes (i.e. an escaped backslash) with just one.

lydell:
@GeoffreyBooth You seem to have made a mistake in backslash hell :)

Pasting this in the console:

```js
CoffeeScript.compile('1 + \ # comment\n2', {bare: true})
```

Is equivalent to pasting this:

```js
CoffeeScript.compile('1 +  # comment\n2', {bare: true})
```

In other words, escaping the space there is completely unnecessary. This is what you want:

```js
CoffeeScript.compile('1 + \\ # comment\n2', {bare: true})
```

That will give CoffeeScript a single backslash to parse. And the output is still missing the comment.

GeoffreyBooth:
@lydell yes, but the same thing outputs the comment from the compiler in Node. The bug is specific to the browser compiler.

GeoffreyBooth:
Ah, wait, I probably wasn't escaping the backslash in the Node test.

GeoffreyBooth:
Hey @lydell @vendethiel @danielbayley @srb- I’ve fixed most of the cases you mentioned:

- [Line continuations](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%20%201%20%2B%20%5C%20%23%20comment%0A%20%20%20%202)
- [Comments before `Existence` node (like `return unless window?`)
](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%20%20%23%201%0A%20%202%0A%0A%20%20%23%203%0A%20%20return%20unless%20window%3F%0A%0A%20%20%23%23%23%204%20%23%23%23%0A%20%20return%20if%20global%3F)
- [Comments before `super` calls or method definition](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%20%20class%20A%20extends%20B%0A%20%20%20%20%23%20This%20is%20a%20fine%20class.%0A%20%20%20%20%23%20I%20could%20tell%20you%20all%20about%20it%2C%20but%20what%20else%20do%20you%20need%20to%20know%3F%0A%20%20%20%20constructor%3A%20-%3E%0A%20%20%20%20%20%20%23%20Something%20before%20%60super%60%0A%20%20%20%20%20%20super()%0A%0A%20%20%20%20%23%20This%20next%20method%20is%20a%20doozy!%0A%20%20%20%20%23%20A%20doozy%2C%20I%20tell%20ya!%0A%20%20%20%20method%3A%20-%3E%0A%20%20%20%20%20%20%23%20Whoa.%0A%20%20%20%20%20%20%23%20Can%20you%20believe%20it%3F%0A%20%20%20%20%20%20no%0A%0A%20%20%20%20%23%23%23%20Look%20out%2C%20incoming!%20%23%23%23%0A%20%20%20%20anotherMethod%3A%20-%3E%0A%20%20%20%20%20%20%23%23%23%20Ha!%20%23%23%23%0A%20%20%20%20%20%20off)
- [Comments before and after interpolated strings, and object key interpolated strings](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%7B%0A%20%20%23%20comment%0A%20%20%22%23%7B1%7D%22%3A%202%0A%7D)

I haven’t yet tackled the [“comments out of order” case](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20one%0A%23%23%23%20two%20%23%23%23%0A%23%20three) (which is a real edge case, because in that example there are no other non-comment tokens for the comments to attach to) or the splice[ case](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:if%20indented%0A%20%20%23%20comment%0A%20%20a%5B1..2%5D%20%3D%20%5B1%2C%202%2C%203%5D). I’m not sure if I’ll be able to do much better for comments at the end of a block (before or after `}`) because I’m not sure how to track which side of the brace the comment belongs on.

I’ve updated the [browser compiler](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try), if you’d like to give me some more edge cases that need fixing. As a suggestion, try putting `if indented` as your first line, and then the code you want to test indented inside the `if` block. If a code string _begins_ (or ends) with a comment, that’s a special case and I add a `JS` token to hold it; whereas anywhere else, the comment is attached to the nearest surviving token. Long story short, comments that begin or end scripts are special cases and will likely work when they otherwise shouldn’t, so put something above a leading comment (or below a trailing one) to give the compiler a more thorough test.

src/nodes.coffee
```diff
@@ -23,6 +23,10 @@ NO      = -> no
 THIS    = -> this
 NEGATE  = -> @negated = not @negated; this
 
+# Track comments that have been compiled into fragments, to avoid outputting
+# them twice.
+compiledComments = []
```
GeoffreyBooth: I think this variable is local to this file, not global. We need to track what comments have been output because they get matched with tokens based on location data, and many tokens share the same location data hashes; so the same comment can appear on several tokens.

GeoffreyBooth:
Okay @lydell @vendethiel @danielbayley @srb-, I’ve fixed all of the cases you mentioned, aside from comments at the end of a block outputting outside the block. So the basic logic for how this works that comments are attached to tokens, like location data, and get output following those tokens. If a comment starts a new line (i.e. there’s nothing but whitespace between the start of the comment and the previous newline) the comment is instead shifted to be attached to the _next_ token, with a flag that the comment should be output preceding this token. So in the [example](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:if%20yes%0A%20%20foo()%0A%20%20%23%20Comment%20at%20end%20of%20block%0A%0A%23%20Comment%20preceding%20new%20block%0Aif%20no%0A%20%20bar()) below, both tokens are attached to the second `if` token:

```coffee
if yes
  foo()
  # Comment at end of block

# Comment preceding new block
if no
  bar()
```

It’s far more common to have comments preceding the lines they reference, so most comments are output in this “precede the token the comments are attached to” style, and I favor the indentation of the token to be preceded (`if` in this case) rather than the indentation of the previous token (`foo()` here). I suppose we could track what token a comment was moved _from,_ so that the “end of block” token above could have some record that it was shifted to `if` from `foo()` and therefore we know to indent based on `foo()`‘s indentation; but then I’d also have to deal with shifting it back to be inside the first block’s braces, which is basically like the rewriter in reverse. That would be far too complicated to implement. I tried simply removing the code from the rewriter that shifts comments forward when they precede line breaks, but that added a lot of complexity to Nodes to know when to output the comments preceded by newlines. There might be an elegant way to solve this, but at the moment I’m leaning toward leaving it for a future PR, especially since comments at the ends of blocks don’t seem all that common. Your thoughts?

I’ve fixed the [“comments out of order” case](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:%23%20one%0A%23%23%23%20two%20%23%23%23%0A%23%20three) and the [splice case](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#try:if%20indented%0A%20%20%23%20comment%0A%20%20a%5B1..2%5D%20%3D%20%5B1%2C%202%2C%203%5D). For “comments out of order,” we’re now outputting all non-inline comments (i.e., all comments other than inline block comments like `1 + /* comment */ 2`) in the same method, which is an makes this cleaner. Can you think of any other cases this PR is currently handling poorly?

This PR is the last breaking change addition to 2.0.0, so the sooner this gets merged in, the sooner we can release beta4 and then 2.0.0.

src/nodes.coffee
GeoffreyBooth: I’ve moved this variable to be a property of `Block` just in case. I tested with this code and got the same output both times:

```coffee
CoffeeScript = require './lib/coffeescript'

coffee = '''
  # 1
  return unless window?

  ### 2 ###
  return if global?
  '''

console.log CoffeeScript.compile coffee
console.log CoffeeScript.compile coffee
```

GeoffreyBooth:
@jashkenas @lydell @vendethiel @danielbayley @srb- @connec:
I’ve rewritten the [comments](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#comments) documentation and added a new section for [type annotations](https://rawgit.com/GeoffreyBooth/coffeescript/comments/docs/v2/#type-annotations) using Flow.

Any other notes?

vendethiel:
Nope. Moving the global to a properties is a great touch as well. 👍

GeoffreyBooth:
Okay folks. I wrote some code to compare the CoffeeScript source and its generated output, including tests, and I’ve verified that (as of 33e3ad701562dd393671ba0a83182bdc87c320bc) _every comment in the source_ is getting output in JavaScript. This helped me identify lots of edge cases to fix, including:

- Comments before various types of `for` declaration lines, including fixing indentation
- Comments before a method that calls `super`
- Comments inside heregexes
- Comments inside “assignment if” e.g. `foo = if`
- Comments before unary tokens, such as `not` in `if not foo`
- Comments before static method or “this”-something (`@foo =`) assignments
- Comments before a function where the first parameter is a splat
- Comments before a soak
- Comments before object destructuring

And finally, now if there are blank lines between successive line comments, those blank lines are output.

I think this is it. A fair bit of code was added to tackle the above list, so it’s probably worthwhile for @vendethiel or @lydell or anyone else who’s up to it to review this again. But I think (pray) that this is done. If only I used type annotations, it might’ve been worth it.

src/helpers.coffee
connec: There is an occurrence of it in the second example [here](http://coffeescript.org/#existential-operator).

src/helpers.coffee
```diff
@@ -111,12 +111,35 @@ buildLocationData = (first, last) ->
...
+    # Add location data
+    if obj?.updateLocationDataIfMissing? and first?
       obj.updateLocationDataIfMissing buildLocationData(first, last)
 
-    return obj
+    # Add comments data
+    unless parserState.tokenComments
+      parserState.tokenComments = {}
+      for token in parserState.parser.tokens when token.comments
+        tokenHash = "#{token[2].first_line}x#{token[2].first_column}-#{token[2].last_line}x#{token[2].last_column}"
```
connec: Nit: you could make a function here e.g. `locationHash`, and use it below as well:

```coffee
locationHash = (loc) ->
  "#{loc.first_line}x#{loc.first_column}-#{loc.last_line}x#{loc.last_column}"
…
tokenHash = locationHash token[2]
…
objHash = locationHash obj.locationData
```

That way the hashes can't diverge.

src/helpers.coffee
```diff
@@ -111,12 +111,35 @@ buildLocationData = (first, last) ->
...
+exports.addDataToNode = (parserState, first, last) ->
   (obj) ->
-    if ((typeof obj) is 'object') and (!!obj['updateLocationDataIfMissing'])
+    # Add location data
+    if obj?.updateLocationDataIfMissing? and first?
       obj.updateLocationDataIfMissing buildLocationData(first, last)
 
-    return obj
+    # Add comments data
+    unless parserState.tokenComments
```
connec: Nit: might not help, but `parserState.tokenComments` is repeated a lot in this block. It could be a bit more readable if we pull that out first:

```coffee
{ tokenComments } = parserState
…
unless tokenComments[tokenHash]?
  tokenComments[tokenHash] = token.comments
else
  tokenComments[tokenHash] = tokenComments[tokenHash].concat token.comments
…
```

src/helpers.coffee
connec: Since I'm nitpicking in this area, I'll throw this in the ring as well:

```coffee
tokenComments[tokenHash] ?= []
tokenComments[tokenHash].push token.comments... # or .concat
```

src/helpers.coffee
```diff
@@ -111,12 +111,35 @@ buildLocationData = (first, last) ->
...
+
+    if obj.locationData?
+      objHash = "#{obj.locationData.first_line}x#{obj.locationData.first_column}-#{obj.locationData.last_line}x#{obj.locationData.last_column}"
+      if parserState.tokenComments[objHash]?
+        attachCommentsToNode parserState.tokenComments[objHash], obj
+
+    obj
+
+exports.attachCommentsToNode = attachCommentsToNode = (comments, node) ->
+  return unless comments?.length isnt 0
```
connec: This might read clearer as `return if comments?.length is 0`.

src/helpers.coffee
```diff
@@ -111,12 +111,35 @@ buildLocationData = (first, last) ->
...
+  node.comments = if node.comments
```
connec: Nit: might be clearer as a conditional initialisation + update:

```coffee
node.comments ?= []
node.comments.push comments...
```

I don't feel especially strongly about this, but I think it's a bit easier to digest than a whole multiline `if` statement.

src/lexer.coffee
```diff
@@ -301,17 +301,62 @@ exports.Lexer = class Lexer
...
+      contents = [content]
+    else
+      # The `COMMENT` regex captures successive line comments as one token.
+      # Remove any leading newlines before the first comment, but preserve
+      # blank lines between line comments.
+      content = comment.replace /^(\n*)/, ''
+      content = content.replace /^([ |\t]*)#/gm, ''
+      contents = content.split '\n'
+
+    commentAttachments = []
```
connec: Good candidate for a comprehension maybe?

```coffee
commentAttachments = for content, i in contents
  …
```

src/lexer.coffee
```diff
@@ -337,6 +382,8 @@ exports.Lexer = class Lexer
           offset: match.index + match[1].length
       when match = @matchWithInterpolations HEREGEX, '///'
         {tokens, index} = match
+        comments = @chunk[0...index].match /\s+(#(?!{).*)/g
+        @commentToken comment for comment in comments if comments
```
connec: I'm not sure I understand the loop here. Probably missing something obvious, but since `comments` is the result of `RegExp#match`, will this not add `comment` twice, once with leading whitespace and once without?

```
coffee> console.log i for i in ' a'.match(/ (a)/)
 a
a
```

src/lexer.coffee
```diff
@@ -482,7 +529,12 @@ exports.Lexer = class Lexer
   # Use a `\` at a line-ending to suppress the newline.
   # The slash is removed here once its job is done.
   suppressNewlines: ->
-    @tokens.pop() if @value() is '\\'
+    prev = @prev()
+    if prev[1] is '\\'
+      if prev.comments
+        # By definition, there must be a token before a `\` token.
```
connec: [Unfortunately not](http://coffeescript.org/v2/#try:%5C%0A%23%20wow%0Awow) 😢 

```coffee
\
# wow
a
```

connec:
```
TypeError: Cannot read property 'comments' of undefined
    at attachCommentsToNode (coffeescript/lib/coffeescript/helpers.js:205:32)
    at Lexer.suppressNewlines (coffeescript/lib/coffeescript/lexer.js:711:11)
    at Lexer.lineToken (coffeescript/lib/coffeescript/lexer.js:598:16)
    at Lexer.tokenize (coffeescript/lib/coffeescript/lexer.js:64:100)
    at Object.<anonymous> (coffeescript/lib/coffeescript/coffeescript.js:100:20)
    at Object.compile (coffeescript/lib/coffeescript/coffeescript.js:53:19)
    at compileScript (coffeescript/lib/coffeescript/command.js:264:33)
    at ReadStream.<anonymous> (coffeescript/lib/coffeescript/command.js:305:14)
    at emitNone (events.js:110:20)
    at ReadStream.emit (events.js:207:7)
    at endReadableNT (_stream_readable.js:1047:12)
    at _combinedTickCallback (internal/process/next_tick.js:102:11)
    at process._tickCallback (internal/process/next_tick.js:161:9)
```

src/lexer.coffee
```diff
@@ -793,6 +845,7 @@ exports.Lexer = class Lexer
         last_line:    lastToken[2].last_line
         last_column:  lastToken[2].last_column
       ]
+      lparen[2] = lparen.origin[2]
```
connec: Just curious: what does this fix?

src/nodes.coffee
```diff
@@ -57,6 +60,34 @@ exports.Base = class Base
...
     fragmentsToText @compileToFragments o, lvl
 
+  # Occasionally a node is compiled multiple times, for example to get the name
+  # of a variable to add to scope tracking. When we know that a “premature”
+  # compilation won’t result in comments being output, set those comments aside
+  # so that they’re preserved for a later `compile` call that will result in
+  # the comments being included in the output.
+  compileWithoutComments: (o, lvl, method = 'compile') ->
+    if @comments
+      @ignoreTheseCommentsTemporarily = @comments
```
connec: This is a bit unpleasant. Could we get away with a flag such as `@ignoreComments = yes or no`, or would that result in peppering the code with `if @ignoreComments …`?

Alternatively, do they have to be assigned to `@` since they're always reassigned at the end of the method? I'd probably be less bothered if it was just a local variable.

connec: *Edit:* I see it affects `unwrapped` as well, but we could still keep two local variables (e.g. `comments` and `unwrappedComments`).

src/nodes.coffee
```diff
@@ -102,6 +136,54 @@ exports.Base = class Base
...
+      else
+        if fragments.length isnt 0
+          precedingFragment = fragments[fragments.length - 1]
+          if commentFragment.newLine and precedingFragment.code isnt '' and
+             not /\n\s*$/.test precedingFragment.code
+            commentFragment.code = "\n#{commentFragment.code}"
+        fragments.push commentFragment
+
+    for comment in node.comments when comment not in @compiledComments
+      @compiledComments.push comment # Don’t output this comment twice.
```
connec: Is this necessary when combined with the `ignoreTheseCommentsTemporarily` above? What case does it cover?

Also, since `compiledComments` is a prototype property it will be the same reference for all nodes, is that the intention?

```coffee
class Base
  foo: []
class A extends Base
class B extends Base
a = new A
a.foo.push 1
console.log new B().foo # [ 1 ]
```

src/lexer.coffee
vendethiel: Not due to the `g` flag.

connec: TIL 😅

src/helpers.coffee
vendethiel: can't really do, at least not here, due to the following line

src/nodes.coffee
```diff
@@ -102,6 +136,54 @@ exports.Base = class Base
         parts.push    @makeCode ")"
     parts
 
+  compileCommentFragments: (o, node, fragments) ->
+    return fragments unless node.comments
+    # This is where comments, that are attached to nodes as a `comments`
+    # property, become `CodeFragment`s. “Inline block comments,” e.g.
```
vendethiel: minor: the comma is in the quotes

src/lexer.coffee
GeoffreyBooth: Damn, talk about edge case. There’s no sensible reason to start a file with a line continuation, is there? Or even to start a line with a line continuation? I’m thinking about just discarding the comments in such a case, as it seems foolish to handle.

src/lexer.coffee
GeoffreyBooth: The `lparen` token was missing location data, though the data was right there on its `origin` property, so I copy the data up. I need the location data on `lparen` so that `addDataToNode` can match it up with the comments that belong on it.

src/nodes.coffee
GeoffreyBooth: The `comments` property already exists on the node before its `compile` method is called. (It’s created by `addDataToNode`, called from `parser.js`.) If I tried to use a flag, that would mean updating the nodes that use `compileWithoutComments` so that each such node’s `compile` or `compileNode` methods reimplemented the logic you see here. There would be a lot more duplication.

The point of this method is that sometimes nodes compile children while discarding side effects. Search for places where I call `compileWithoutComments` and you’ll see. This happens e.g. to get the name of a variable, but then that name is used as a string and the `Value` or `IdentifierLiteral` node never gets properly compiled itself later on. In many such cases I move the comments of a node’s children to the node itself in the node’s constructor (see the constructor for `For`, for example) to avoid the issue; but in some cases it’s unavoidable. (Function parameters that are splats are a particularly nasty case, and I couldn’t figure out how to use this method and instead had to implement a variation of it inside `Code`.) You could argue that cases where `compileWithoutComments` is necessary are cases that are poorly implemented in the first place, and we should refactor those nodes’ `compile` or `compileToFragments` methods . . . and you’d probably be right, at least in most cases. But this PR is big enough as it is, without taking on that additional scope.

src/nodes.coffee
GeoffreyBooth: It’s necessary because `compileWithoutComments` is only called in a handful of cases.

There are lots and lots of nodes that end up with `comments` properties that are duplicative. Like an `Assign` node where it and its children, `variable` and `value`, all have `comments` properties with the same comments. I’m not sure why this happens; it could be that all those nodes have the same location data, so the comments get attached to all of them; or when tokens go through the parser, things bubble up (like `Identifier` inside `Expression` inside `Assign`) the comments get duplicated with each pass. If there were a clean way to prevent duplication in the first place I’d prefer that, but I couldn’t figure out how to achieve that. Even if we could cut down (or eliminate) duplicated comments, this two-line check would still be a good failsafe to ensure we didn’t miss anything.

src/nodes.coffee
GeoffreyBooth: The CoffeeScript docs are written in U.S. English: https://english.stackexchange.com/a/114403.

GeoffreyBooth:
@connec I’ve revised per (or responded to) all your notes above.

GeoffreyBooth:
(Merging in per offline conversation. If @connec finds any further issues, I’ll start a new PR.)

GeoffreyBooth:
Good god, this resolved a lot of issues.

srb-:
Great work everyone. I finally got home and was able to try Flow with Beta4. It was brilliant to be able to compile a function and have Flow spit out a type error!

Now I'm stuck trying to declare a variable of a certain type, due to CS separating var declaration and use. Try this code:

```
# @flow

myNum ###: number ### = 55     # Flow errors out here

nyNum = "string"               # but I want an error here
```
Perhaps I'm missing something?

GeoffreyBooth:
@srb- can you please open a new issue? And please include what JavaScript you’re aiming for. I don’t see anything on the [Flow Comment Types page](https://flow.org/en/docs/types/comments/) that corresponds with variable declarations, only function parameters and return values.

connec:
See the docs on variable types: https://flow.org/en/docs/types/variables/

It looks like it's to do with the separation of assignment and declaration - all the examples type the `var/let/const` statement, and the problem doesn't occur if the `var myNum` line is typed. That's a more awkward thing to fix I suspect (might be easier to add an enhancement to Flow...).

PR #4574, opened 2017-06-14
[CS2] destructured nested defaults [fixes #4566]

### LINKED ISSUE #4566: [CS2] Lexer: Destructuring with nested default values
@GeoffreyBooth I just ran into something that broke (I assume unintentionally) as a result of #4478: eg `({a: {b = 1}}) ->` compiled ok before this PR but gives `error: unexpected =` after. This destructuring also breaks in a destructured assignment: `{a: {b = 1}} = c` also gives `error: unexpected =`

helixbass:
Fixes #4566

Added recursive assignment of `@lhs` to destructured objects (corresponding to the recursive assignment of `@lhs` that was already there for arrays)

Looks like this won't cause a merge conflict with #4493

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1171,7 +1171,10 @@ exports.Obj = class Obj extends Base
       for prop in props when prop instanceof Assign
         {value} = prop
         unwrappedVal = value.unwrapAll()
-        unwrappedVal.lhs = yes if unwrappedVal instanceof Arr or unwrappedVal instanceof Obj
+        if unwrappedVal instanceof Arr or unwrappedVal instanceof Obj
+          unwrappedVal.lhs = yes
+        else if unwrappedVal instanceof Assign
+          unwrappedVal.nestedLhs = yes
 
     isCompact = yes
     for prop in @properties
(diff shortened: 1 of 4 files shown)


GeoffreyBooth:
This is very elegant. Looks good to me. @lydell or @connec, do you care to review?

helixbass:
@GeoffreyBooth I actually just stumbled on another similar nested destructuring bug eg `{a: {b} = {}} = c` was generating:
```
({
  a: ({b} = {})
} = c);
```
So I added another commit to suppress the `Assign`-wrapping parentheses when it's nested like that

src/nodes.coffee
```diff
@@ -1810,7 +1821,7 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or (isValue and @variable.base instanceof Obj and not @param)
+    if o.level > LEVEL_LIST or (isValue and @variable.base instanceof Obj and not @nestedLhs and not @param)
```
connec: I take it `@variable.base.lhs` is not adequate (and/or desirable) here?

helixbass: @connec nope it seems to interfere w/ the wrapping in parens of top-level destructured object assigns. Not crystal clear on the different nested `Obj`/`Assign` structures so there may be a way do it without introducing `nestedLhs` but that's the way I found to distinguish the top-level `Assign`s from the nested ones

Issue #4580, opened 2017-06-17
Lexer: two-value postfix for in implicit object


helixbass:
Came across this while testing fix for #4579 (doesn't block):

This breaks:
```
a: b
c: d for d, e of f
```
As does two-value `for...in`:
```
a: b
c: d for d, e in f
```
whereas the corresponding single-value `for`s don't:
```
a: b
c: d for d of f

a: b
c: d for d in f
```

PR #4581, opened 2017-06-17
Fix #4579: postfix for in object literals

### LINKED ISSUE #4579: Lexer: postfix for in first line of implicit object
This is very similar to #4544/#4554, basically the corresponding case (to that issue's postfix `if`/`unless`) where the behavior of a postfix `for` is different depending on whether you're in the first line of an implicit object or not. Eg:
```
a: 1
b: x for x in y
```
compiles to the expected:
```
var x;

({
  a: 1,
  b: (function() {
    var i, len, results;
    results = [];
    for (i = 0, len = y.length; i < len; i++) {
      x = y[i];
      results.push(x);
    }
    return results;
  })()
});
```
while this:
```
b: x for x in y
a: 1
```
compiles to:
```
var i, len, x;

for (i = 0, len = y.length; i < len; i++) {
  x = y[i];
  ({
    b: x
  });
}

({
  a: 1
});
```

Will submit a fixing PR momentarily

As opposed to postfix conditionals, where they make sense with a single-line implicit object (eg `return a: 1 if b`), I'm not sure why the "`for` applies to the entire implicit object rather than just the value" interpretation is preferable even for one-line implicit objects? Eg `a: x for x in y` gets treated as `{a: x} for x in y` (which creates a list of single-key objects with the same key, not seemingly very useful) rather than `a: (x for x in y)` (really what that syntax reads as is an object comprehension, but that's a different story). So I'm wondering if `a: x for x in y` could be interpreted as `a: (x for x in y)` in all cases, not just when one line of a multi-line implicit object? Or is it possible that people have been using it to express `{a: x} for x in y` so should stay as is?

helixbass:
Fixes #4579 

In a multi-line implicit object, treats postfix `for` the same regardless of whether it's on the first line

While writing tests, came across an orthogonal bug that I submitted as #4580. There's one test line that I marked as being able to be uncommented once #4580 is fixed

Looks like `command.coffee` wasn't recompiled after a previous change? But tests pass so I'm assuming the recompiled `lib/coffeescript/command.js` should stay in the PR


lydell:
What about postfix `while` and `until`?

helixbass:
@lydell sure, makes sense, updated and added tests

PR #4586, opened 2017-06-25
Fix #3906: safe missing end token error

### LINKED ISSUE #3906: Lexer: Chaining: Variable indents throw TypeError during compilation
``` coffeescript
a
    .b ->
        c
    .d

e(
    f
)

g
```

```
...\node_modules\coffee-script\lib\coffee-script\coffee-script.js:37
        throw helpers.updateSyntaxError(err, code, options.filename);
                      ^
TypeError: Cannot read property '2' of undefined
  at Lexer.exports.Lexer.Lexer.tokenize (...\node_modules\coffee-script\lib\c
offee-script\lexer.js:43:52)
  at Object.<anonymous> (...\node_modules\coffee-script\lib\coffee-script\cof
fee-script.js:49:20)
  at Object.compile (...\node_modules\coffee-script\lib\coffee-script\coffee-
script.js:34:19)
  ...
```

I get this error with CoffeeScript 1.9.1 on Node 0.12.0, io.js 1.6.1, and in the browser. The in-browser compiler on http://coffeescript.org gives the message `i.origin is undefined`.

helixbass:
Fixes #3906 

This fixes the TypeError thrown by the compiler when trying to compile the examples from #3906. All of the examples from that issue except @GeoffreyBooth's should really successfully compile though - I fixed that as well, but figure that the compiler-throwing-an-error part should be merged into `master` while the fix-wonky-indentation-behavior part should be merged into `2`? So I'll open a second pull request against `2` that makes those examples compile successfully

### DIFF
--- a/lib/coffee-script/lexer.js
+++ b/lib/coffee-script/lexer.js
@@ -12,7 +12,7 @@
     function Lexer() {}
 
     Lexer.prototype.tokenize = function(code, opts) {
-      var consumed, end, i, ref2;
+      var consumed, end, i, ref2, ref3;
       if (opts == null) {
         opts = {};
       }
@@ -46,7 +46,7 @@
       }
       this.closeIndentation();
       if (end = this.ends.pop()) {
-        this.error("missing " + end.tag, end.origin[2]);
+        this.error("missing " + end.tag, ((ref3 = end.origin) != null ? ref3 : end)[2]);
       }
       if (opts.rewrite === false) {
         return this.tokens;
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -78,7 +78,7 @@ exports.Lexer = class Lexer
       return {@tokens, index: i} if opts.untilBalanced and @ends.length is 0
 
     @closeIndentation()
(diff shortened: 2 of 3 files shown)


GeoffreyBooth:
I think at this point, the only fixes that we should be merging into `master` are fixes for recent regressions. #3906 has been an open issue for over two years, so it shouldn’t be affecting anyone at the moment. I think it can stay unresolved in 1.x and only get fixed in 2. I’m much more concerned that fixing it could create another unintended regression like #4568 (which we _should_ fix in 1.x, if we can find a way to do so without allowing syntax that isn’t possible in 2; @helixbass you’re probably the best one to tackle that if you don’t mind).

@lydell or @jashkenas, do you agree?

lydell:
Yes, I agree. :+1:

GeoffreyBooth:
@helixbass Can you please retarget this at `2` and/or combine it with #4587?

helixbass:
@GeoffreyBooth sure, merged this into #4587. And ya I can take a look at #4568

PR #4587, opened 2017-06-25
[CS2] Fix #3906: handle indentation inside indented chain

### LINKED ISSUE #3906: Lexer: Chaining: Variable indents throw TypeError during compilation

helixbass:
As explained in #4586, beyond just fixing the TypeError thrown by the compiler when trying to compile the examples in #3906, the examples (except @GeoffreyBooth's) should all actually successfully compile, which this fixes

The issue was with `@outdebt`. The common characteristic of the examples from that issue was an "indented" chained call followed by a further-indented block followed by another chained call, I added tests for all three of the examples. I could only reproduce the issue by having the code at the top level so used `CoffeeScript.eval()` in the tests. This led to something weird where a test in `test/repl.coffee` was failing b/c I assigned `a` inside my `CoffeeScript.eval()`'d test and it expected `a` to be `undefined` but clearly my scope bled into theirs and caused `a` to be defined when that test ran. So I just used a different name in my tests but don't know if that's an issue with test isolation?


helixbass:
Merged in #4586 (the compiler TypeError fix) so that it all targets `2` here

Added @GeoffreyBooth's example from #3906 as a test for something that still should be an error but would've thrown a TypeError previously

GeoffreyBooth:
These examples are truly perverse. There are some weird people out there.

Issue #3446, opened 2014-04-13
Lexer: Leading dot doesn’t close function passed as argument and not wrapped in parentheses


bwiklund:
On 1.7.1, this breaks:

``` coffeescript
angular.module 'foo'
.controller 'EmailLoginCtrl', ->
.controller 'EmailSignupCtrl', ->
```

While this is fine:

``` coffeescript
angular.module 'foo'
.controller 'EmailLoginCtrl', (->)
.controller 'EmailSignupCtrl', (->)
```

And so is this:

``` coffeescript
angular.module 'foo'
.controller 'EmailLoginCtrl', -> return
.controller 'EmailSignupCtrl', -> return
```

Is this the way it's supposed to work?

GeoffreyBooth:
You’re really trying to spread across multiple lines what is really a one-line call. When you do so, you should use `\`, and parentheses as necessary:

```coffee
angular.module('foo')\
.controller('EmailLoginCtrl', ->)\
.controller('EmailSignupCtrl', ->)
```

lydell:
I still think this is a bug.

GeoffreyBooth:
Fixed by #4588.

Issue #1043, opened 2011-01-12
doit:(a=->, b=->)-> 'multiple functions as default parameters'


chadams:
## multiple default values work

```
obj = ->
  doit:(a=1, b=2)->
    say()
```
## functions as default values work

```
obj = ->
  doit:(a=->)->
    say()
```
## multiple functions as default parameters throw (PARSE ERROR)

```
obj = ->
  doit:(a=->, b=2)->
    say()
```

GeoffreyBooth:
Closing this issue as it doesn’t appear to be affecting enough people to warrant attention. If someone wants to do further investigation into this and can devise a fix, a pull request is welcome.

GeoffreyBooth:
Fixed by #4588.

Issue #3845, opened 2015-02-10
Lexer: Chaining: Leading dot throws error when previous line ends with empty function


bioball:
This is not a huge deal, but not the behavior I was expecting. I'd like to be able to chain method calls with function callbacks together, even if the function bodies were empty. The use case is for cases like this:

``` coffeescript
doThing()
.then (@result) =>
.catch handleError
```

Where I'd expect this to compile into:

``` js
doThing().then((function(_this) {
  return function(account) {
    _this.account = account;
  };
})(this))["catch"](handleError);
```

Currently, this just throws an error: `Error on line 3: unexpected .`

edgebal:
@bioball you can use parentheses to encapsulate empty functions.

```
oThing()
.then ( (@result) => )
.catch handleError
```

bioball:
Yup, that's what I'm currently doing. It's not such much that I can't do it, it's just that this works:

```coffeescript
doThing()
.then (@result) => undefined
.catch handleError
```

While this doesn't

```coffeescript
doThing()
.then (@result) =>
.catch handleError
```

Adding parentheses is fine, I just expected functions with empty bodies to work the same way functions with bodies do.

epidemian:
> I just expected functions with empty bodies to work the same way functions with bodies do.

I would expect the same. Especially considering that:

```coffeescript
foo ->
bar ->
```

... compiles to:

```javascript
foo(function() {});
bar(function() {});
```

... and:

```coffeescript
obj
.foo ->
```

... compiles to:

```javascript
obj.foo(function() {});
```

Then:

```coffeescript
obj
.foo ->
.bar ->
```

... should compile to:

```javascript
obj.foo(function() {}).bar(function() {});
```

Unless i'm missing something, this seems like a bug.

nfour:
@epidemian 

You mean:

```coffee
obj
.foo ->
.bar ->
```
Should compile to

```coffee
_ref = obj
_ref.foo(function() {})
_ref.bar(function() {})
```

Panoplos:
I am having a similar issue with *bluebird* `Promise` code, as follows:

```coffeescript
doSomething argument
  .then (res) ->
    console.log res
  .catch (err) ->
    console.error err
```

this result in the following JavaScript:

```javascript
doSomething(argument).then(function(res) {
  console.log(res);
})["catch"](function(err) {
  console.error(err);
});
```

Which of course is seriously wrong...

nfour:
[...]

PR #4588, opened 2017-06-25
[CS2] Fix for comma after function glyph [#1043]

helixbass:
The one "unambiguous" thing I saw in #1043 was that a `,` immediately following `->` or `=>` can indicate an empty function body, since otherwise it seems to be a syntax error (ie a function body can't start with a comma). So this uses the rewriter to interpret `-> ,` and `=> ,` that way

@GeoffreyBooth I targeted `2`, I guess this could safely go on `master` since the previous meaning was a syntax error, but you seem to prefer only targeting `master` if really necessary?

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -441,7 +441,7 @@
           return this.tokens.splice((this.tag(i - 1) === ',' ? i - 1 : i), 0, outdent);
         };
         return this.scanTokens(function(token, i, tokens) {
-          var j, k, ref, ref1, tag;
+          var j, k, ref, ref1, ref2, tag;
           [tag] = token;
           if (tag === 'TERMINATOR') {
             if (this.tag(i + 1) === 'ELSE' && this.tag(i - 1) !== 'OUTDENT') {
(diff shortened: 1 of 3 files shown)


helixbass:
@GeoffreyBooth actually it was super easy to lump in a fix for #3446 (which is what led me to #1043) / #3845. Same thinking applies: treat `-> .` or `=> .` as an empty function body followed by chained whatever, and I believe here also the previous meaning would always be a syntax error. So now those chained calls with empty function args work eg:
```
a
.b ->
.c
```

So fixes #3446 and #3845, added tests with examples from those issues

GeoffreyBooth:
This all looks good to me. @lydell?

connec:
This does also allow `a -> .b`, which is a bit odd.

helixbass:
@connec ya I'd thought there was no way to distinguish between "inline" chaining and line-break chaining since the leading `.` suppresses line tokens (eg `INDENT`/`TERMINATOR`) but looked more closely and the `->` token will have `.newLine` set in the line-break case. So opened #4590 to distinguish between these so that eg `a -> .b` will still error

PR #4590, opened 2017-06-26
Refine #4588: disallow inline chained call after function glyph

helixbass:
Per @connec's comment in #4588, distinguish between "inline" and line-break chaining when interpreting `->` followed by `.` in the token stream as `(->).`

So `a -> .b` will error (added test) while
```
a ->
.b
```
is still interpreted as `a(->).b` via #4588


connec:
Nice, thanks for investigating :smile:

PR #4592, opened 2017-06-26
[CS2]: Fix #4591: multiple accesses after super

### LINKED ISSUE #4591: [CS2] super.x.y throws a compiler error
Starting to go through `nodes.coffee` for comprehension, noticed that in the `SimpleAssignable` grammar rules, `Value Accessor` assumes that it can call `.add()` on the `Value`, but `Super` (which is a `Value` in the grammar) doesn't have an `.add()` method. So things like `super.x.y` and `super['x'].y` throw compiler errors when trying to call the missing `.add()` method

helixbass:
Surprisingly, just wrapping a `new Value()` around the `Super` in the grammar didn't seem to break anything and made `super.x.y` and `super['x'].y` work

Just looking through references to `Super` in `nodes.coffee`, I didn't see anything that looked like it would need updating, but there could be gotchas I'm not aware of


test/classes.coffee
```diff
@@ -1808,3 +1808,19 @@ test 'Bound method of immediately instantiated class with expression base class
 
   {derivedBound} = a
   eq derivedBound(), 3
+
+test "#4591: super.x.y, super['x'].y", ->
```
GeoffreyBooth: Usually `super.foo` refers to a method named `foo`, as in `super.foo()`. We should add another test that tests `super.x.y()` and `super['x'].y()`

GeoffreyBooth: Also `super.x['y']` and `super.x['y']()`.

vendethiel: well, `super.foo` still has the implicit call there – it means `super(arguments...).foo`

connec: @vendethiel that changed in `2` to support ES-style `super` invocations. Compare [bare super], [super with parens], [super with access], and [super with both].

Deeper accesses against a `super` field could be useful if you're trying to `bind`/`call`/`apply` it, I guess.

```coffee
class A
  f: -> super.f.apply(this, arguments)
```

[bare super]: http://coffeescript.org/v2/#try:class%20A%0A%20%20f%3A%20-%3E%20super
[super with parens]: http://coffeescript.org/v2/#try:class%20A%0A%20%20f%3A%20-%3E%20super()
[super with access]: http://coffeescript.org/v2/#try:class%20A%0A%20%20f%3A%20-%3E%20super.f
[super with both]: http://coffeescript.org/v2/#try:class%20A%0A%20%20f%3A%20-%3E%20super.f()

vendethiel: ah, I didn't follow that. That's good to know

GeoffreyBooth: @helixbass do you mind adding these additional tests, unless there’s some reason they don’t make sense?

helixbass: @GeoffreyBooth added tests for the cases you outlined

@connec that's a more realistic use case than anything I thought of, I just noticed the hole in the grammar

GeoffreyBooth:
@helixbass if you don’t mind merging and resolving conflicts, I think this is ready. @lydell, @vendethiel or @connec, any notes?

vendethiel:
`super::` is not allowed – is that useful, and do we want to allow it?

vendethiel:
Also, that error is pretty surprising (it's not from this PR, this happens as well on `2`):

```
$ ./bin/coffee -bce "class then a: -> super"  
[stdin]:1:15: error: unexpected ->
class then a: -> super
```

vendethiel:
WRT #4601:
On 2 right now:

```
$ ./bin/coffee -bce 'super"b"'
[stdin]:1:6: error: unexpected string
super"b"
```

This PR hasn't been rebased yet, so the same error happens. Once rebased I think it should work, but we might want to add a test.

helixbass:
@GeoffreyBooth merged `2` and resolved conflicts

@vendethiel none of those seem directly relevant to this PR (or #4601). I don't know if `super"b"` (ie `super` as the "function" of a tagged template literal) actually makes sense or not but it wouldn't start working b/c of this PR or #4601, in `2` `super` (the token) is only currently allowed in the grammar as part of a `Super` (ie `super` with access/index) or an `Invocation` (ie `super(...)`) so you'd need an additional rule to allow it with a tagged template literal

GeoffreyBooth:
@connec or @vendethiel or @lydell, any other notes?

vendethiel:
Nope. If needed we can revisit the rest later.

PR #4604, opened 2017-07-05
[CS2] Support `await` in REPL without wrapper function

### LINKED ISSUE #4603: [CS2] Enhancement: Support for 'await' on REPL without having to wrap it inside a function
Hello good people of coffeescript: 

Currently, the way I can test how a library that heavily rely on promises works on the REPL is as follow: 
``` coffeescript
coffee> wikijs = require('wikijs')
coffee> batman = ""
coffee> wikijs().page('batman'). then (response)=> batman = response
Promise {
  <pending>,
  domain: 
   Domain {
     domain: null,
     _events: { error: [Function: debugDomainError] },
     _eventsCount: 1,
     _maxListeners: undefined,
     members: [] } }
coffee>  batman
' '  // too soon.. the promise hasn't resolved yet
coffee>  batman 
{
  links: [Function: links],
  categories: [Function: categories],
  coordinates: [Function: coordinates],
  info: [Function: n],
  backlinks: [Function: backlinks],
  rawImages: [Function: f],
  mainImage: [Function: mainImage]
 }
coffee>  mainImage = ''
coffee> wikijs().page('batman'). then (response)=> response.mainImage()). then ( img) => mainImage = img
Promise {
  <pending>,
  domain: 
   Domain {
     domain: null,
     _events: { error: [Function: debugDomainError] },
     _eventsCount: 1,
     _maxListeners: undefined,
     members: [] } }
coffee> mainImage
'https://upload.wikimedia.org/wikipedia/en/7/74/Batman_Detective_Comics_Vol_2_1.png'
```
I can also define a function, but doing that on a REPL is awkward. I would rather create a file and set up an environment that auto run my script on save instead. 

It would great if I could just take advantage of the new async/await features of coffeescript 2.0 and  do this instead: 
``` coffeescript
coffee> wikijs = require('wikijs')
coffee> batman = await wikijs().page('batman')
coffee> batman
{
  links: [Function: links],
  categories: [Function: categories],
  coordinates: [Function: coordinates],
  info: [Function: n],
  backlinks: [Function: backlinks],
  rawImages: [Function: f],
  mainImage: [Function: mainImage]
 }
coffee> await batman.mainImage()
'https://upload.wikimedia.org/wikipedia/en/7/74/Batman_Detective_Comics_Vol_2_1'
```
Which adjust perfectly with the philosophy of a REPL which is providing you with an environment to quickly and easily test and prototype stuff out.

connec:
I went ahead and made a fix for #4603 because I think it could be a really useful feature for playing with promise libraries.

I took care to ensure non-await evaluations will still happen fully synchronously (e.g. `process.nextTick -> console.log 'async'` works as before).

I didn't write any tests, because putting them in `test/repl.coffee` or `test/async.coffee` would involve quite a bit of copypasta. I did think of making the async test globally available (e.g. as `test.async` or `testAsync`) but that was a bigger change than I'd planned to make. If someone has an opinion on how to deal with it I'm happy to add some tests.

---

Rather than compiling top-level await expressions directly, which will
always throw a syntax error, the REPL wrapper will now wrap them in a
closure and handle the returned promise before calling back to the REPL.

Fixes #4603.


lydell:
How come you could implement this so easily, while nodejs/node#13209 is still open? Some kind of backwards compatibility thing? What am I missing?

connec:
Interesting... it seems a lot is to do with ambiguity with `await` as a variable, which isn't something CS needs to deal with as `await` is always parsed as an 'operator'. 

One thing I didn't think about that came up in the thread is what should happen if the user ctrl+c's an awaiting promise. The way this is implemented right now, if the promise hangs forever there is no observable issue (same as doing `new Promise ->` in the REPL), however if the promise eventually resolves (e.g. `await new Promise (r) -> setTimeout r, 5000`) then the REPL will eventually log the result (in this case `undefined`), even after ctrl+c is pressed.

screen shot 2017-07-05 at 19 50 45

That said, with some acrobatics it should be possible to intercept the `SIGINT` and prevent the REPL handler from being called to print the result (which won't prevent the computation from *running* of course, which is the nature of async in JS).

connec:
@AlvaroBernalG do you want to have a play with this branch and see if it works as you'd hoped?

AlvaroBernalG:
@connec I have been playing with your branch and the only thing that I found is that I am unable to assign the result of an await to a variable. Which useful in cases when you want to comfortably inspect the element returned. 

@princejwesley has a [gist](https://gist.github.com/princejwesley/a66d514d86ea174270210561c44b71ba) where he solve a similar problem by appending the assignment variable to 
the global object.

So if you have an await whose result is being assigned to a variable like so:

```coffeescript
coffee>myVariableResult = await wiki().page('batman')
```

It would get transpiled into:
```js
// `(async() => global.${myVariableResult}=${awaitExpression})()`

(async() => global.myVariableResult = await wiki().page('batman'))()

```

if there is no variable assignment then: 
```js
// `(async() => ${awaitExpression})()`

(async() => await wiki().page('batman'))()
```

lydell:
@AlvaroBernalG Good example. That's certainly unexpected behavior.

connec:
Good catch indeed, didn't think about assignment. I suppose I need to update the `isAwait` condition to also accept `Assign`'s whose `value` is an `Await`.

Just to be clear, the second example ("if there is no variable assignment") is working as expected?

AlvaroBernalG:
@connec I confirm that besides that everything is working as expected ( Even canceling the promise before it finish with ctrl +c .. which is awesome :) )

connec:
I found a much more robust approach that also supports `await` inside arbitrary expressions, e.g.

```coffee
console.log await Promise.resolve 1

# and assignment, of course :)
foo = await Promise.resolve 1
```

About the cancelling... that's actually a bit weird, and I don't think there's anything we can easily do about it. For example, running the following in the REPL will result in `foo` and `_` being set to the 'cancelled' value (`1`).

```coffee
foo = await new Promise (resolve) -> setTimeout (-> resolve 1), 5000
# ^C
foo = 2
# wait a few seconds...
foo
# 1
```

This is because it's impossible to actually cancel a native promise.

That said, in its current form it's useful for playing with promise libraries, and the same problem would exist in manual usage:

```coffee
foo = null
(new Promise (resolve) -> setTimeout (-> resolve 1), 5000).then (result) -> foo = result
foo = 2
# wait a few seconds
foo
# 1
```

GeoffreyBooth:
@AlvaroBernalG can you please confirm that this PR works as expected?

@lydell, any notes or is this ready to merge?

lydell:
I haven't tested it, but the code looks good.

AlvaroBernalG:
@connec @GeoffreyBooth Apologies for the delay. For some reason I missed the emails and warnings on Github.

I tested the branch and I can confirm that it is working as expected. 

@connec Thank you very much. Coffeescript REPL is now my default REPL.

Once again apologies for the delay.

GeoffreyBooth:
Great, thanks!

connec:
Glad it's working out for you @AlvaroBernalG :smile:

PR #4605, opened 2017-07-05
Restrict indented return arg to object

helixbass:
See discussion in #4599 - this restricts the newly introduced `return` with indented arg syntax to just allow an indented object. So eg this is allowed:
```
return
  a: 1
```
but this is not:
```
return
  a
```
Currently non-implicit objects will also work:
```
return
  {a: 1}
```
but it should be possible (though perhaps ugly) to disallow this if desired, [here's](https://github.com/jashkenas/coffeescript/pull/4599#issuecomment-313183609) one idea how

Fixes the `return` equivalent of #3199 eg this will now not split the implicit object:
```
if 1 then return
  a: 1
  b: 2
```

Also gets rid of some unintuitive cases that are currently allowed eg this will now not compile:
```
  return
    a for b in c
```
rather than compile to the unintuitive:
```
for (i = 0, len = c.length; i < len; i++) {
  b = c[i];
  return a;
}
```


test/formatting.coffee
```diff
@@ -298,3 +298,10 @@ test "#1275: allow indentation before closing brackets", ->
     a = 1
    )
   eq 1, a
+
+test "#3199: throw multiline implicit object", ->
```
vendethiel: throw -> return

helixbass: @vendethiel ya I was just being lazy about trying to minimize merge conflicts once this is eventually merged into `2` with #4599 but you're right, updated

Issue #85, opened 2010-01-12
function composition in coffescript as operator


francescoagati:
should be nice implements function composition as make in ruby with Proc class in this article:

http://media.fngtps.com/rubybanter/2007/006/

lass Object
  def op(c)
    lambda { |x| eval(c) }
  end
end

class Proc
  class << self
    def complement(f = nil, &b)
      b ||= f
      lambda { |_a| !b.call(_a) }
    end

```
def compose(*a)
  a.map! { |b| b.to_proc }

  if a.length == 2
    lambda { |*args| a[0].call(a[1].call(*args)) }
  elsif a.length > 2
    compose(a.shift, compose(*a))
  else
    lambda { nil }
  end
end

def conjoin(*a)
  lambda { |*args| a.all? { |f| f.call(*args) } }
end

def disjoin(*a)
  lambda { |*args| a.any? { |f| f.call(*args) } }
end
```

  end

  def complement
    self.class.complement(self)
  end

  def compose(_a)
    self.class.compose(_(a.reverse + [self]))
  end

  def conjoin(_a)
    self.class.conjoin(_([self] + a))
  end

  def disjoin(_a)
    self.class.disjoin(_([self] + a))
  end

  alias_method :c, :complement
  alias_method :*, :compose
  alias_method :&, :conjoin
  alias_method :|, :disjoin

  alias_method :[], :call
end

PR #4606, opened 2017-07-07
Spread syntax triple dots on either right or left

zdenko:
See [#85](https://github.com/coffeescript6/discuss/issues/85).

```coffeescript
{a, b, r...} = {a:1, b:2}
{a, b, ...r} = {a:1, b:2}

o = {obj1..., a:1, ...obj2}
a = [a, b, arr1..., c, d, ...arr2]

f args...
f ...args

({a, r...}) ->
({a, ...r}) ->

```


GeoffreyBooth:
Very clean. Looks good to me. @lydell? @connec?

lydell:
LGTM

src/rewriter.coffee
```diff
@@ -235,9 +235,10 @@ exports.Rewriter = class Rewriter
 
       # Recognize standard implicit calls like
       # f a, f() b, f? c, h[0] d etc.
+      # Added support for spread dots on the left side: f ...a
       if (tag in IMPLICIT_FUNC and token.spaced or
           tag is '?' and i > 0 and not tokens[i - 1].spaced) and
-         (nextTag in IMPLICIT_CALL or
+         (nextTag in [IMPLICIT_CALL..., '...'] or
```
connec: Bit of a nitpick but... this could be `nextTag in IMPLICIT_CALL or nextTag is '...'` and would probably be a bit cheaper and, I think, more readable (assuming we can't simply add `...` to `IMPLICIT_CALL`).

connec:
I'm still not entirely convinced it's a good idea :stuck_out_tongue: but apart from a small nitpick it LGTM.

What's our strategy going to be for documenting this?

GeoffreyBooth:
> What’s our strategy going to be for documenting this?

Tiny note in the changelog, and nothing else 😉

src/rewriter.coffee
GeoffreyBooth: Hmmm, looks like it’s much cheaper: https://jsperf.com/in-spread-array-vs-two-comparisons/1

Issue #4600, opened 2017-07-03
[CS2] Enhancement: Support object spread syntax in multiline objects without braces or commas


laurentpayot:
```coffeescript
foo =
  a: 1
  b: 2

bar =
  foo...
  c: 3
```
results in
```
error: unexpected ...
	foo...
	   ^^^
```
For now we have to use:
```coffeescript
bar = {
  foo...,
  c: 3
}
```
It's a bit ugly and it adds an exception to the nice CS way of object notation.

connec:
Copying https://github.com/jashkenas/coffeescript/pull/4493#issuecomment-312258561

---

That's probably related to [`looksObjectish`](https://github.com/jashkenas/coffeescript/blob/master/src/rewriter.coffee#L110) in the rewriter, which seems to only accept implicit object literals when there's a `:` involved (which is consistent with being unable to use shorthand properties in implicit objects.

Tbh we probably want to leave it like this, or we'll need to face down ambiguities in expressions like

```coffee
f k: v, rest...
```

Which, currently, will compile to two arguments (an object literal and a splat param) but would be ambiguous if we allowed splats to exist in implicit objects.

connec:
As an aside, commas are still optional, even with explicit braces:

```coffee
{
  foo...
  bar
  baz: 'qux'
}
```

GeoffreyBooth:
I ran into this just writing the documentation. I think it's a thing that will be very common for people to want to do. Could we find a way to allow spreads in multiline objects, while leaving function arguments unaffected (i.e. to avoid ambiguity problems)?

zdenko:
@GeoffreyBooth, @connec, @helixbass 
 I've started to play with this and made some progress.

```javascript
foo =
  bar...
  c: 3
```
compiles to
```javascript
foo = Object.assign({}, bar, {
    c: 3
  });
```

`f a:x, rest...` into `f({a:x}, ...rest)`

`f {a:x, rest...}` into 
```javascript
f(Object.assign({
    a: x
  }, rest))
```

And if I'm correct, there is also this case
```javascript
foo = 
  bar
  c: 3
```
which should compile to
```javascript
foo = {
   bar,
   c: 3
}
```

vendethiel:
> And if I'm correct, there is also this case
> ```
> foo = 
>   bar
>   c: 3
> ```
> which should compile to
> ```
> foo = {
>    bar,
>    c: 3
> }
> ```

[...]

PR #4607, opened 2017-07-08
[WIP][CS2] CSX spread attributes: <div {props…} />

zdenko:
This PR addresses the issue with CSX spread attributes (#4551).

Since spread in JSX is special notation different from the ES6 generic spread syntax, CS will simply compile `{props...}` to `{...props}`.

```coffeescript
# <div {props...}>hello</div>
<div {...props}>hello</div>

# <div {props1...} id="tagId" {props2...}>hello</div>
<div {...props1} id="tagId" {...props2}>hello</div>
```

There is also an "accidental" side effect and this:
```coffeescript
<div {props..., a: "x"}>hello</div>
<div {props1..., props2...}>hello</div>
```

will compile to

```javascript
<div {...props} a="x">hello</div>
<div {...props1} {...props2}>hello</div>
```


GeoffreyBooth:
I wonder if we should be doing this checking in the lexer . . . shouldn’t the check if a previous interpolation is a spread be happening in `nodes.coffee`?

Other than that it looks good to me. @lydell or @connec?

zdenko:
> shouldn’t the check if a previous interpolation is a spread be happening in nodes.coffee?

@GeoffreyBooth I don't think this is currently possible since object spread syntax without comma isn't valid syntax (#4600).

Lexer (originally) converts `<div id="bar" {foo...} />` into
`[CSX_TAG div] [CALL_START (] [{ {] [PROPERTY id] [: :] [STRING "bar"] [, ,] [( (] [{ {] [IDENTIFIER foo] [... ...] [} }] [) )] [, ,] [} }] [CALL_END )]`.

This PR converts the same code into
`[CSX_TAG div] [CALL_START (] [{ {] [PROPERTY id] [: :] [STRING "bar"] [, ,] [IDENTIFIER foo] [... ...] [, ,] [} }] [CALL_END )]`.

xixixao:
This is indeed what I had in mind in terms of what it would compile to. Nice! No opinion on the implementation.

GeoffreyBooth:
@zdenko Those tokens roll up into the `Call` node, where you have `compileCSX` that works with `attributes` and `content`. Is it not possible to look inside those arrays to check for duplicates there?

zdenko:
@GeoffreyBooth I've tried, but compilation fails at parsing step (before the Call).
Since @xixixao's lexer method `csxToken` is already transforming CSX tokens, I just added a simple check for spread attribute and alter the transformation.

But, I'm also working on the PR for #4600 and it might be possible to move this into the `compileCSX` if necessary.

GeoffreyBooth:
In general we try to do as much as possible in Nodes rather than in Lexer, because checks in Lexer are much more brittle and harder to maintain. If that’s not possible that’s fine.

Should we merge this in now or wait until your other PR is ready?

zdenko:
I would merge now. I'm not sure when the PR for #4600 will be ready.

GeoffreyBooth:
Okay. @lydell and @connec, any notes?

connec:
I guess syntactically it's a bit weird that `<div {a: 1}>` compiles but `<div {object}>` won't. Unfortunately my lexer-fu is weak, so I'm not sure if there's a clean way of blocking `a: 1` whilst still allowing splats?

zdenko:
I added some validation for CSX attributes. 
Valid CSX syntax:
```
<div id="value" name={someName} {props...} attr="#{attribute}" checked />
```
Since CSX attributes are transformed into `Obj` in the `lexer` (@xixixao might explain this better), the following CS syntax is also allowed and will produce the same result.
```
<div {id:"value"} {name:{someName}} {props...} {attr:"#{attribute}"} {checked} />
```

zdenko:
@GeoffreyBooth I think my work is done here. Do I have to add anything else?

GeoffreyBooth:
Assuming this now covers all the cases @connec mentioned, then hopefully this is ready. @connec, did you have any other notes?

GeoffreyBooth:
@connec, any notes on this?

src/nodes.coffee
```diff
@@ -1219,12 +1220,22 @@ exports.Obj = class Obj extends Base
...
+          else unless prop.bareLiteral?(IdentifierLiteral)
+            prop = new Assign prop, prop, 'object'
+        else
+          # Spread in CSX.
+          prop = if @csx then new Literal "{#{prop.compile(o)}}" else prop
+      # Check if CSX attribute is valid.
+      if @csx and prop instanceof Assign
+        prop.variable.error "Unexpected token" unless prop.variable.unwrap() instanceof PropertyName
+        propVal = prop.value.unwrap()
+        unless ((propVal instanceof Parens or propVal instanceof StringWithInterpolations) and propVal.body instanceof Block) or propVal instanceof StringLiteral or propVal instanceof Obj
```
connec: Could we reformat this line to something like:

```coffee
unless propVal instanceof StringLiteral or propVal instanceof Obj or
    ((propVal instanceof Parens or propVal instanceof StringWithInterpolations) and propVal.body instanceof Block)
  prop.value.error "..."
```

It might also benefit from a comment stating what is allowed (e.g. what is `(propVal instanceof Parens or propVal instanceof StringWithInterpolations) and propVal.body instanceof Block` ensuring?).

connec:
Thanks for the update @zdenko, that makes sense to me now!

Unfortunately, there are still a couple of 'regressions' in terms of what will compile now and the errors:

- Given:
    ```coffee
    <div {obj}></div>
    ```
    Used to throw:
    ```
    [stdin]:1:10: error: unexpected ,
    <div {obj}></div>
             ^
    ```
    Now compiles to:
    ```js
    <div obj></div>;
    ```
    This definitely shouldn't compile. The `{}` in the source implies that `obj` is a variable name, but it is instead compiled to a boolean flag.

- Given:
    ```coffee
    <div {"#{a}"}></div>
    ```
    Used to throw:
    ```
    [stdin]:1:13: error: unexpected ,
    <div {"#{a}"}></div>
                ^
    ```
    Now throws:
    ```
    SyntaxError: Unexpected token
    ```
    For some reason this error has lost location data.

- Given:
    ```coffee
    <div {props..., a: "x"}>hello</div>
    <div {props1..., props2...}>hello</div>
    ```
    As you note in the PR description, these now compile to:
    ```coffee
    <div {...props} a="x">hello</div>
    <div {...props1} {...props2}>hello</div>
    ```
    I'm not entirely comfortable with this for a couple of reasons:
    - From these examples you might infer that any object expression is valid, however `<div {a: 1}></div>` is invalid.
    - This leads to two ways of adding attributes to a CSX expression, should `<div {a..., b: "c"}>` or `<div {a...} b="c">` be preferred?
    I don't want to belabour this point too much though, if someone else thinks it's acceptable (@GeoffreyBooth / @lydell / @vendethiel / @xixixao).

microdou:
@zdenko actually added new functionality to the original jsx spreads, by allowing the following cases:
```coffeescript
<div {props..., a: "x"}>hello</div>
<div {props1..., props2...}>hello</div>
```
However, as @connec points out, the new syntax may be causing some problems, so it's debatable that whether these new syntax should be allowed in `coffeescript v2.0.0`.

 ```coffeescript
# If this is allowed
<div {props..., a: "x"}>hello</div>

# Should this be allowed?
<div {a: "x"}>hello</div>
```
JSX only allows `{...obj}` inside the bracket, which is not a real spread at all. I think this syntax is only introduced into JSX so that an object won't be treated as a Boolean flag.
```javascript
<div {...obj}>hello</div>
// transpiles to
React.createElement(
  "div",
  obj,
  "hello"
);
// obj is treated as an object

<div obj>hello</div>
// transpiles to
React.createElement(
  "div",
  { obj: true },
  "hello"
);
// obj is treated as a Boolean flag
```

Though ugly, the "jsx spread" `obj` can actually appear in a complex form and be correctly transpiled in babel.
```javascript
<div {...{...obj, a: "x"}}>hello</div>
// transpiles to
React.createElement(
  "div",
  _extends({}, obj, { a: "x" }),
  "hello"
);
// a new object {...obj, a: "x"} is created and gets "jsx spread"
// babel correctly recognize the two different "..." here
```

I would suggest that CSX follows JSX syntax strictly and safely, by only allowing `{obj...}` inside `< >`. Nothing else should appear side-by-side with`obj...` inside `{obj...}`. Neither `{obj..., a: "x"}` nor `{obj1..., obj2...}` is valid. However, a complex form of obj can be allowed, which follows just regular coffeescript syntax, such as `{{obj..., a: "x"}...}`, `{{obj1..., obj2...}...}` and `{{a: "x"}...}`. `{{obj1..., obj2...}...}` is of course ugly,  `{obj1...} {obj2...}` inside `< >` is a better form.

zdenko:
@connec, @microdou you're right. CSX syntax should follow JSX.
The main issue is that most of the CSX transformation happens inside `lexer` where attribute `checked` in `<div checked />` or `<div {checked} />` is transformed into `Obj` {checked} in both cases.
This means that most of the CSX syntax checking has to be placed inside the `lexer`.

zdenko:
BTW this new "functionality" (e.g. `<div {props..., a:1} ?>`) was just side effect caused by enabling spread synatx in the `lexer`. It wasn't intentional.

src/lexer.coffee
```diff
@@ -10,7 +10,7 @@
 # are read by jison in the `parser.lexer` function defined in coffeescript.coffee.
 
 {Rewriter, INVERSES} = require './rewriter'
-
+log = console.log
```
GeoffreyBooth: This shouldn’t be here.

src/lexer.coffee
```diff
@@ -109,6 +109,20 @@ exports.Lexer = class Lexer
   # though `is` means `===` otherwise.
   identifierToken: ->
     inCSXTag = @atCSXTag()
+    # Check CSX properties synatx.
```
GeoffreyBooth: “syntax”

src/nodes.coffee
```diff
@@ -1207,7 +1207,7 @@ exports.Obj = class Obj extends Base
       else
         ',\n'
       indent = if isCompact or prop instanceof Comment then '' else idt
-
+      log = console.log
```
GeoffreyBooth: Please remove.

GeoffreyBooth:
> This means that most of the CSX syntax checking has to be placed inside the lexer.

And is there no way we can shift more of this logic into `nodes.coffee`?

zdenko:
> And is there no way we can shift more of this logic into nodes.coffee?

@GeoffreyBooth I think it's possible, but it would also require rewriting existing CSX parts in the `lexer`.
I also tried to move all syntax checking stuff from `lexer` into `rewriter`'s `@enforceValidCSXAttributes`.

The current solution is far from optimal, but I think it's stable enough to merge into `2`.
However, I'll continue with refactoring and hopefully soon come with a better solution.

src/lexer.coffee
```diff
@@ -561,6 +575,7 @@ exports.Lexer = class Lexer
     i = @ends.length - 1
     i-- while @ends[i]?.tag is 'OUTDENT' or depth-- > 0 # Ignore indents.
     last = @ends[i]
+    # log @csxDepth, last?.tag
```
lydell: Please remove this one as well :)

zdenko: I think it's already removed.

xixixao:
Are there no tests on this PR? Results look good, how bad was moving the checking to the rewriter? The regexes are far from elegant/performant.

zdenko:
@xixixao there are no tests yet. I'll add them in a day or two. I wanted to come with an optimal solution first.
Checking is still in the `lexer` since I couldn't find a better option. I tried to reuse your logic where you set property `csxColon` for regular attributes token (`a={value}` => `a:({value})`) and then validate them in the `rewriter`, but I couldn't come up with an elegant way for marking invalid attributes. 
If they pass the `lexer` they will be converted to tokens, e.g. `{props} {props..., a:"b"}` => `{, props, props, ..., a, :, ", b, ", }`, and to marked them it would require some detail checking just to be able to throw an error in the next step.
So, I figured since I already have to check them in the `lexer` before the token is created, why not throw an error a.s.a.p.

And, yeah, I agree regarding the regex. I'm far from regex-fu mastery. I would appreciate if someone could take a look and advise.

connec:
Would it be worth stepping back from putting it all in the rewriter, and instead get the rewriter to emit some new token types, which we then parse (instead of overloading `Obj`)? Then we could do all the validation we like inside a `CSXTag(name, attributes)` kinda node.

xixixao:
@connec You meant "lexer" instead of "rewriter" I presume. The reason I didn't do this in the lexer is that you'd have to include the CSX-specific nodes everywhere where they need to be to in the syntax declaration to get the full possible syntax for tags (and CS always tried to keep the parser declaration simple, opting for magic in the rewriter mainly).

@zdenko You should be able to tag the tokens though in the lexer for the rewriter (to distinguish them).

zdenko:
@connec I'm working on something similar, e.g. to pass CSX attributes directly into the `nodes`.
@GeoffreyBooth should I create another PR for this improvement? 
I think current solution is good enough and can be merged in `2`. Besides, I don' know how long it will take to finish improvements.

GeoffreyBooth:
@zdenko Sure, a separate PR would be fine.

As for this PR, I reviewed some of the examples @connec and @microdou listed above, and this one still isn’t fixed:

```coffee
<div {"#{a}"}></div>
```

This should have a descriptive error, with location data, like our other errors.

@connec and @microdou, are there any other examples you can think of that this PR isn’t handling well? Whether or not they’re code that should compile or throw errors.

Also this PR needs tests. `<div {...{...obj, a: "x"}}>hello</div>` would be a good example to include.

microdou:
@GeoffreyBooth @zdenko  It passed all my examples! Though, as you already pointed out, error handling can be improved by providing description and location information.

@connec 's examples currently raises different errors, which should have been more consistent by simply pointing out that an object syntax is not allowed in `< >` except being wrapped by `{  ...}`.
```coffeescript
<div {obj}></div>
# error: Unexpected token, expected ...
# <div {obj}></div>
#         ^

<div {"#{a}": 1}></div>
# error: Unexpected token
# <div {"#{a}": 1}></div>
#       ^^^^^^

<div {"#{a}"}></div>
# SyntaxError: Unexpected token
```
I'm not familiar with how error should be reported, but is the following better by just pointing out entire `{   }`?
```coffeescript
<div {obj}></div>
# error: some error
# <div {obj}></div>
#      ^^^^^

<div {"#{a}": 1}></div>
# error: some error
# <div {"#{a}": 1}></div>
#      ^^^^^^^^^^^

<div {"#{a}"}></div>
# error: some error
# <div {"#{a}"}></div>
#      ^^^^^^^^
```

GeoffreyBooth:
Errors generally look like this:

```
[stdin]:1:5: error: octal literal '000' must be prefixed with '0o'
foo 000
    ^^^
```

The `1:5` is the location info (the error starts at row 1, column 5). I’m not too particular about exactly what gets underlined, as long as it’s clear to the user what the problem is.

microdou:
OK, here are two more examples that need attention:
1. a wrong compilation, which should have thrown an error
```coffeescript
<div ...obj></div>

# Wrongly compiles to
# <div {...obj}></div>;
```
2. error information needs improvement
```coffeescript
<div obj...></div>

# error: unexpected }
# <div obj...></div>
#            ^
```

zdenko:
OK. This was major rework.

Everything regarding the CSX attributes is now processed in the `nodes`.
The only change in the `lexer` is the transformation of the attributes which are now wrapped in the `Arr` instead of `Obj`. All other changes from this PR are reverted to the original code.
In this way, I was able to control and check each attribute in the `nodes`. Plus, IMHO, wrapping attributes in the `[]` is more logical.

All errors (I hope) are now caught in the `@compileCSX` where the list of attributes is processed.
The error description is now more verbose.
In the `Obj`, I added method `@compileCSXAttributes` where I moved processing CSX from the `@compileNode`.

@GeoffreyBooth I was able to finish this sooner, so I just committed to the same PR.

microdou:
@zdenko Great work!
Now errors are consistent.
```
error: Unexpected token. Allowed CSX attributes are: id="val", src={source} or {props...}
```
But there is actually one more attribute allowed in CSX: a single prop defaulting to true
```coffeescript
<div prop></div>
# equivalents to
<div prop={true}></div>
```
So the error message may be updated to
```
error: Unexpected token. Allowed CSX attributes are:
id="val", src={source}, {props...}, or a single prop defaulting to true
```

microdou:
Some side-effects are back:
```coffeescript
<div {obj..., a: "x"}></div>
# compiles to
<div {...obj} a="x"></div>;

<div {obj1..., obj2...}></div>
# compiles to
<div {...obj1} {...obj2}></div>;
```
I think they are OK to live with most of the time. Users need to avoid these bad jsx/csx syntax anyway.
However, if a user actually tries to use this "hack" for production, he/she may accidentally cause some confusion, in the following example:
```coffeescript
<div {a..., b}></div>
# compiles to
 <div {...a} b></div>;
```
It's debatable whether the `b` inside this object-like syntax should be treated as `b: b` or a single prop `b` defaulting to true.
So it's still best to throw error for these cases in my opinion.

zdenko:
>
> Some side-effects are back:
>
> <div {obj..., a: "x"}></div># compiles to<div {...obj} a="x"></div>
>
>
Hmm, I'll check this. This should throw an error. It was already working.
-- 
Sent from Gmail Mobile

microdou:
@zdenko Perfect!

microdou:
I just realized that `{...props}` is allowed.
```coffeescript
<div {...props}></div>
# compiles to
<div {...props}></div>;
```
I thought it was an side effect until I found that it appears to be universally allowed in coffeescript now.
```coffeescript
newObj = {...obj, a: "x"}
# compiles to
var newObj;
newObj = Object.assign({}, obj, {
  a: "x"
});
```
@GeoffreyBooth Is the **left-sided-spreads** officially supported by coffeescript now in addition to the legacy right-sided-spreads?

GeoffreyBooth:
> Is the left-sided-spreads officially supported by coffeescript now in addition to the legacy right-sided-spreads?

Yes, since #4606. Dots on the left are not the preferred style, so I don’t plan to add it to the docs, but it seemed like a minor enough convenience to provide for people copying and pasting ES code.

src/lexer.coffee
```diff
@@ -500,25 +503,30 @@ exports.Lexer = class Lexer
       [input, id, colon] = match
       origin = @token 'CSX_TAG', id, 1, id.length
       @token 'CALL_START', '('
-      @token '{', '{'
+      csxStart = @token '[', '['
```
connec: Unused assignment.

src/lexer.coffee
```diff
@@ -500,25 +503,30 @@ exports.Lexer = class Lexer
...
-        @token '}', '}', 0, 2
+        @token ']', ']', 0, 2
         @token 'CALL_END', ')', 0, 2
         @csxDepth--
         return 2
       else if firstChar is '{'
-        token = @token '(', '('
+        if prevChar is ':'
+          token = @token '(', '('
+          @csxObjAttribute = no
```
connec: Not sure what the right fix here is, but because this is limited to a single state, alternately nesting assignment and splats causes errors:

```
[stdin]:1:24: error: unexpected }
<A child={<B {...params} />} />
                       ^
```

I suppose that's what `@pair` is for, but that will currently throw if it doesn't match... It would be nice to avoid a new stack though...

zdenko: @connec the latest commit fixes this bug.

microdou:
@connec @zdenko Thanks for pointing out and the fix!
Yes, it's quite useful to use JSX expression as attributes for oneliners, for example in `react-bootstrap`:
```coffeescript
<ButtonToolbar>
  <OverlayTrigger overlay={<Tooltip>Hello</Tooltip>}>
    <Button>Click me!</Button>
  </OverlayTrigger>
</ButtonToolbar>
```
The nested JSX expression can be refactored out though:
```coffeescript
tooltip = <Tooltip>Hello</Tooltip>
<ButtonToolbar>
  <OverlayTrigger overlay={tooltip}>
    <Button>Click me!</Button>
  </OverlayTrigger>
</ButtonToolbar>
```
It may be unusual to have spreads in the nested expression, but it's awesome to have the edge case covered!

GeoffreyBooth:
So @microdou does the current code cover all the relevant cases, in your opinion? In other words, is it ready to merge in?

microdou:
@GeoffreyBooth I couldn't find any problem. @zdenko did such an awesome job!

GeoffreyBooth:
@connec, any more notes?

src/nodes.coffee
```diff
@@ -880,7 +880,19 @@ exports.Call = class Call extends Base
...
-    fragments.push attributes.compileToFragments(o, LEVEL_PAREN)...
+    if attributes.base instanceof Arr
+      for obj in attributes.base.objects
+        attr = obj.base
+        attrProps = attr?.properties or []
+        # Catch invalid CSX attributes: <div {a:"b", props} {props} "value" />
+        if not (attr instanceof Obj or attr instanceof IdentifierLiteral) or (attr instanceof Obj and not attr.generated and (attrProps.length > 1 or not (attrProps[0] instanceof Splat)))
+          obj.error """
+            Unexpected token. Allowed CSX attributes are: id="val", src={source}, {props...} or attribute.
+            Example: <div id="val" src={getTheSource()} {props...} checked>hello world</div>.
```
GeoffreyBooth: We don’t normally put examples in error messages. Please update the docs with this if the docs are insufficient.

test/csx.coffee
```diff
@@ -724,3 +692,11 @@ test 'unspaced less than after CSX works but is not encouraged', ->
 
       res = 2 < div;
     '''
+
+test 'CSX invalid attribute errors', ->
```
GeoffreyBooth: All our error-checking tests are in `error_messages.coffee`. Perhaps they shouldn’t be, but that’s a refactor for another PR. Please move these checks over there and follow the form in that file, where you confirm that the error returned is as expected (i.e. not just that an exception is thrown).

GeoffreyBooth:
My notes are fairly minor. This is much cleaner than before, thanks for reworking it @zdenko!

src/nodes.coffee
zdenko: I'll remove it.

test/csx.coffee
zdenko: Agree. I'll move and improve the tests later today.

connec:
I'm now seeing the following issue, but it's getting very esoteric now:

```coffee
<A {...{child: <B {...params} />}} />
# [stdin]:1:34: error: unexpected )
# <A {...{child: <B {...params} />}} />
#                                  ^
```

The same thing [compiles with babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=true&presets=react&targets=&browsers=&builtIns=false&debug=false&code_lz=DwQQBA3gdDEMYAsCWAbAJgLjMAQpGUADgIYBOxAtgM4C-YA9AHw11NA&experimental=false&loose=false&spec=false&playground=true&stage=0) and also works fine if you extract the child into a variable:

```coffee
child = <B {...params} />
<A {...{child}} />
```


I think the issue is still to do with that `csxObjAttribute` field - in the presence of arbitrary nested mixes of `{...*}` and `=*` we would need a stack to keep track of what the lexer is currently scanning in order to correctly match `}` or `)`.

connec:
Another failing example:

```coffee
<A {...{child: <B foo={bar} />}} />
```

The problem occurs because:

1. The lexer sees [`{` without `:`] and sets `csxObjAttribute = yes`.
2. The lexer sees [`{` with `:`] and sets `csxObjAttribute = no`.
3. The lexer sees [`}` with `not csxObjAttribute`] and emits a `)` to close the field.
3. The lexer sees [`}` with `not csxObjAttribute`] and emits a `)` to close the field - this is the incorrect step that needs to instead emit `}`.

The story with `<A {...{child: <B {...params} />}} />` is similar:

1. The lexer sees [`{` without `:`] and sets `csxObjAttribute = yes`.
2. The lexer sees [`{` without `:`] and sets `csxObjAttribute = yes`.
3. The lexer sees [`}` with `csxObjAttribute`] and emits a `}` to close the field, and sets `csxObjAttribute = no`.
3. The lexer sees [`}` with `not csxObjAttribute`] and emits a `)` to close the field - this is the incorrect step that needs to instead emit `}`.

[`{` without `:`]: https://github.com/GeoffreyBooth/coffeescript/blob/7de2eb9818def18072cc46b05f7828b37f870a70/src/lexer.coffee#L521
[`{` with `:`]: https://github.com/GeoffreyBooth/coffeescript/blob/7de2eb9818def18072cc46b05f7828b37f870a70/src/lexer.coffee#L518
[`}` with `not csxObjAttribute`]: https://github.com/GeoffreyBooth/coffeescript/blob/7de2eb9818def18072cc46b05f7828b37f870a70/src/lexer.coffee#L554
[`}` with `csxObjAttribute`]: https://github.com/GeoffreyBooth/coffeescript/blob/7de2eb9818def18072cc46b05f7828b37f870a70/src/lexer.coffee#L551

zdenko:
@connec good catch. I improved tracking of the splats inside the `CSX` tag.

GeoffreyBooth:
@connec Looks like the latest commit fixed both examples:

```jsx
// <A {...{child: <B {...params} />}} />
<A {...{
    child: <B {...params} />
  }} />;


// <A {...{child: <B foo={bar} />}} />
<A {...{
    child: <B foo={bar} />
  }} />;
```

And ditto with the dots on the right side.

GeoffreyBooth:
Great work @zdenko!

Issue #4610, opened 2017-07-11
Tests failing on node v4.x


MylesBorins:
Seems like you are using a RegExp feature that is not supported in v4.x

The failure is caused by the test found on line 290 of `test/regexps.coffee` 

```
failed 1 and passed 852 tests in 3.70 seconds

  #4248: Unicode code point escapes
  SyntaxError: Invalid flags supplied to RegExp constructor 'u'
    at new RegExp (native)
    at Function.<anonymous> (/Users/mborins/code/coffeescript/test/regexps.coffee:1:1)
    at global.test (/Users/mborins/code/coffeescript/Cakefile:1:1)
    at Object.<anonymous> (/Users/mborins/code/coffeescript/test/regexps.coffee:1:1)
    at Object.<anonymous> (/Users/mborins/code/coffeescript/test/regexps.coffee:1:1)
    at Module._compile (module.js:1:1)
    at Object.exports.run (/Users/mborins/code/coffeescript/lib/coffee-script/coffee-script.js:1:1)
    at runTests (/Users/mborins/code/coffeescript/Cakefile:1:1)
    at Object.action (/Users/mborins/code/coffeescript/Cakefile:1:1)
    at helpers.extend.invoke (/Users/mborins/code/coffeescript/lib/coffee-script/cake.js:1:1)
    at Object.exports.run (/Users/mborins/code/coffeescript/lib/coffee-script/cake.js:1:1)
    at Object.<anonymous> (/Users/mborins/code/coffeescript/bin/cake:1:1)
    at Module._compile (module.js:1:1)
    at Object.Module._extensions..js (module.js:1:1)
    at Module.load (module.js:1:1)
    at Function.Module._load (module.js:1:1)
    at Function.Module.runMain (module.js:1:1)
    at startup (node.js:1:1)
    at node.js:1:1
```

GeoffreyBooth:
We don’t aim to have the tests succeed in Node 4. See #4502.

If you’d like to submit a PR that disables this test in Node versions below the version that supports unicode regexes, I would merge it, but the tests are assumed to be running in current latest Node.

MylesBorins:
I've already disabled in our smoke testing suite... we will keep it disabled.

Node v4 is officially supported until April 2018... so it is possible other people will show up fairly confused.  I highly suggest updating the engines field in your package.json

GeoffreyBooth:
[...]

PR #4617, opened 2017-07-15
1.12.7

GeoffreyBooth:
Tiny release, but I guess we’re not expecting any other 1.x updates anytime soon so might as well get these fixes out there.


GeoffreyBooth:
@lydell One last quick fix: 8597477. I assume you’re okay with that? It gets the tests passing in Node 4.8.2, fixing #4610. If people are treating 1.x as our LTS release to go with Node’s oldest LTS, we probably should try not to break that (especially when it’s this easy).

GeoffreyBooth:
@helixbass do you mind please merging `master` into `2` and resolving conflicts? (Or submitting a PR for that)? I tried, but it’s a bit confusing since some of the bugs you fixed on `master` you also submitted PRs into `2` for. I’m not sure how to resolve the conflicts and get all the tests to pass.

PR #4618, opened 2017-07-19
[CS2] Object spread syntax in multiline objects without braces or commas

zdenko:
This is initial commit for object spread syntax in multiline objects without braces or commas (#4600).
Left and right position of spread dots are supported. 
It's still rough around the edges and needs refinements. The trickiest part was detecting the starting position of the spread and detecting if next token is spread. 

Examples:

```coffeescript
# bar =
#   foo...
#   c: 3
bar = Object.assign({}, foo, {
    c: 3
  });
  
# bar = a:1, foo...
bar = Object.assign({a:1}, 
    foo);

# bar = a:1, ...foo
bar = Object.assign({a:1}, 
    foo);
```

Function parameters compiles as usual.

```coffeescript
# f a:1, obj...
f({
    a: 1
  }, ...obj);

# f ...obj, a:1
f(...obj, {
   a: 1
 });

# f {a:1, obj...}
f(Object.assign({
    a: 1
  }, obj));
```

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -133,8 +133,96 @@
         return i + j + fuzz - 1;
       }
 
+      detectObjectSpreadStart(j) {
+        var expEnd, expStart, levels, ref, ref1, ref2, ref3, ref4, ref5, ref6, ref7, ref8, s, spreadStarters, startIdx, tag;
+        spreadStarters = ['IDENTIFIER', 'SUPER', '@', 'THIS'];
+        expStart = ['(', '{', 'CALL_START', 'PARAM_START'];
+        expEnd = [')', '}', 'CALL_END', 'PARAM_END'];
+        if (this.tag(j) === '...' && (ref = this.tag(j + 1), indexOf.call([...spreadStarters, '(', '{', 'CALL_START'], ref) >= 0)) {
+          return j;
+        }
+        startIdx = j - 1;
+        if (!(startIdx > -1)) {
+          return -1;
+        }
+        if (ref1 = this.tag(startIdx), indexOf.call(spreadStarters, ref1) >= 0) {
+          return startIdx;
+        }
+        if (this.tag(startIdx) === 'PROPERTY' || (ref2 = this.tag(startIdx), indexOf.call(expEnd, ref2) >= 0)) {
+          s = startIdx - 1;
+          levels = (ref3 = this.tag(startIdx), indexOf.call(expEnd, ref3) >= 0) ? 1 : 0;
+          while (tag = this.tag(s)) {
+            if ((tag === '[' || tag === 'INDEX_START' || tag === 'PARAM_START') && levels <= 0) {
+              return -1;
+            }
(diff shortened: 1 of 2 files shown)


vendethiel:
-1 for me, just because of the sheer amount of complexity introduced for this one case.

jashkenas:
Yeah, this is some extreme syntax.

GeoffreyBooth:
This looks like my first attempt at import/export, where I tried to do everything in the lexer. Why is all of this logic in the rewriter? Can't you take advantage of the grammar and nodes?

zdenko:
Well, I have to say, I agree with @vendethiel and @jashkenas. 
I started working on this out of fun, but at the end, it felt strange.
It's not consistent. Everywhere else with object destructuring, you have to use braces.
```
f({a, b, obj...}) ->
f {a:1, obj...}
{a, b, c...} = obj
```

And then, there is also a  question of property value shorthand `obj = {a, b, c} = {a:a, b:b, c:c}` which would require support for this syntax:
```
obj = 
  a
  b
  c
```

zdenko:
@GeoffreyBooth can you give me a hint how to deal with this in `grammar` or `nodes`?
Cause, I don't think we could skip the rewriter where in `@addImplicitBracesAndParens` this
```
obj = 
  a:1
  b:2
```
is compiled into
```
obj = {
  a:1,
  b:2
}
```

So, something like `obj = a:1, obj..., b:2`would be rewrited into `obj = {a:1}, obj..., {b:2}` before it reaches `nodes`.
I think `lexer` is better place, but then we also have to move other logic for the object from the rewriter.
It's a lot of complex work for a limited use case. And, like I said, the next step would be supporting property value shorthand, which is way harder, e.g. compile `obj = a, b, c` into `obj = {a, b, c}`.

helixbass:
Also -1, @zdenko there's a reason that objects which include shorthands eg `{a: 1, b}` require explicit braces - regardless of whether it could be grammatically unambiguous, it'll always be potentially weird/ambiguous to read and the same thinking applies with objects that include a `spread...`. So I think it should stay consistent with existing Coffeescript syntax (ie require braces) and for good reason

zdenko:
@helixbass like said before, I agree. This syntax is very limited and confusing. 
But, since I already started this PR (based on #4600), I wanted to finish and see where it ends.

GeoffreyBooth:
So do we have consensus that we should close this PR? And if people want object spread syntax in multiline objects, they just need to use braces?

Or is the objection just to the implementation, rather than the goal?

vendethiel:
I have objections to both

helixbass:
@GeoffreyBooth yes my objection is to the syntax, I think braces should be required

GeoffreyBooth:
Okay, closing for now. Thanks for the effort @zdenko.

laurentpayot:
Thanks @zdenko 👍 

> Everywhere else with object destructuring, you have to use braces.

We should add in the documentation that braces are required in multiline objects for any kind of destructuring. It is a simple rule so I have no problem with that.

PR #4625, opened 2017-07-28
[CS2] CLI: Propagate SIGINT and SIGTERM signals when node is forked

### LINKED ISSUE #4185: Killing coffee process does not kill child node process
See https://github.com/remy/nodemon/issues/195

When running `coffee --nodejs ...`, a `node` child  process is created.

Killing/restarting the `coffee` process does not kill the `node` process and can continue tying up resources that the new process needs to access. For example, if `node` was started with `--debug`, then it will keep using port 5858, preventing the newly spawned `node` process from opening that port.

Workaround is to abandon the `coffee` command and use:

```
node --debug ./node_modules/.bin/coffee path/to/script.coffee
```

However, it seems the `coffee` command is essentially useless when launching node and working with automated restart workflows.

Proposed solution: `coffee` should handle `SIGINT` and `SIGTERM` and kill any child processes it has started.

Gobie:
When running commands like `coffee --nodejs --max_old_space_size=128 index.coffee` the coffee process doesn't forward signals to node process, which was already discussed in #4185.

Problem is that tools that use flow SIGTERM, delay, SIGKILL leave zombie processes.

This fixes #4185.

### DIFF
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -462,6 +462,9 @@ forkNode = ->
     cwd:        process.cwd()
     env:        process.env
     stdio:      [0, 1, 2]
+  ['SIGINT', 'SIGTERM'].forEach (signal) ->
+    process.on signal, ->
+      p.kill signal
   p.on 'exit', (code) -> process.exit code
 
 # Print the `--help` usage message and exit. Deprecated switches are not
--- a/src/command.coffee
+++ b/src/command.coffee
@@ -462,9 +462,9 @@ forkNode = ->
     cwd:        process.cwd()
     env:        process.env
     stdio:      [0, 1, 2]
-  ['SIGINT', 'SIGTERM'].forEach (signal) ->
-    process.on signal, ->
-      p.kill signal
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
Please retarget this against `2`.

src/command.coffee
```diff
@@ -444,6 +444,9 @@ forkNode = ->
     cwd:        process.cwd()
     env:        process.env
     stdio:      [0, 1, 2]
+  ['SIGINT', 'SIGTERM'].forEach (signal) ->
```
GeoffreyBooth: Our prefer style is to use `for…in`:

```coffee
for signal in ['SIGINT', 'SIGTERM']
```

This compiles into a traditional `for` loop, without a new function or scope.

Gobie: I thought so looking at the surrounding code but in this case the new scope is necessary for the code to work properly.

Anyway I retargeted against `2` and used `for...in` with `do` for the scope.

GeoffreyBooth:
Thanks. @lydell or @connec?

Issue #4623, opened 2017-07-27
[CS2] Lexer: `do super` in constructor


Inve1951:
[code:](http://coffeescript.org/v2/#try:class%20a%0Aclass%20b%20extends%20a%0A%20%20constructor%3A%20-%3E%0A%20%20%20%20do%20super)
```coffee
class a
class b extends a
  constructor: ->
    do super
```
output:
```
[stdin]:4:12: error: unexpected indentation
    do super
           ^
```
expected output is same as for [`super()`](http://coffeescript.org/v2/#try:class%20a%0Aclass%20b%20extends%20a%0A%20%20constructor%3A%20-%3E%0A%20%20%20%20super())
```js
var a, b;

a = class a {};

b = class b extends a {
  constructor() {
    super();
  }

};
```

GeoffreyBooth:
Fixed via #4627.

PR #4627, opened 2017-07-30
[CS2] fix for “`do super` in constructor” bug

zdenko:
A small fix for the #4623 (`do super` in constructor)

```coffeescript
class b extends a
  constructor: ->
    do super
```

```javascript
b = class b extends a {
    constructor() {
      super();
    }
}
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -136,6 +136,11 @@ exports.Lexer = class Lexer
     if id is 'default' and @seenExport and @tag() in ['EXPORT', 'AS']
       @token 'DEFAULT', id
       return id.length
+    if id is 'do' and /^super(?:\s+?)?(?!\(\))/.exec @chunk[3...]
+      @token 'SUPER', 'super'
+      @token 'CALL_START', '('      
+      @token 'CALL_END', ')'
+      return 8
 
     prev = @prev()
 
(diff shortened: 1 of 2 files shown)


zdenko:
I improved RegExp and used `exec` so I can catch whitespace between `do` and `super`, and return correct length.

src/lexer.coffee
```diff
@@ -136,6 +136,12 @@ exports.Lexer = class Lexer
     if id is 'default' and @seenExport and @tag() in ['EXPORT', 'AS']
       @token 'DEFAULT', id
       return id.length
+    if id is 'do' and regExSuper = /^(\s*super)(?!\(\))/.exec @chunk[3...]
+      @token 'SUPER', 'super'
+      @token 'CALL_START', '('      
+      @token 'CALL_END', ')'
+      [input, sup] = regExSuper
+      return sup.length + 3
```
GeoffreyBooth: When would `sup.length` not be 5? And 3 is the length of `do` plus a space? What if someone types `do super` with two or more spaces between the words?

zdenko: Yes, 3 is the length of `do` plus a space. 
The `\s*` in `\s*super` was added to catch all spaces between `do` and `super`.
So, when `do` is detected, the regex will match zero or more spaces before `super`, starting from third characters of current match e.g. `do␣`.

If you change line 144 in `lexer` to `return 8` it will work for one space, but will fail when you add more spaces, e.g. `do␣␣␣␣super`.

GeoffreyBooth:
Verified that this works with any number of whitespace characters between `do` and `super`.

PR #4640, opened 2017-08-14
[CS2] Fix #4413: Generated variables in function parameters

### LINKED ISSUE #4413: [CS2] Nodes: Missing declarations for generated variables in parameter defaults
**Note:** this is for the `2` branch only.

```coffee
(a = foo() ? bar()) ->
```

```js
(function(a = (ref = foo()) != null ? ref : bar()) {});
```

Note that `ref` is used despite there being no `var ref` anywhere.

A simple patch to fix this is

```diff
diff --git a/src/nodes.coffee b/src/nodes.coffee
index 08b0b4e..65fc7cf 100644
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2005,10 +2005,11 @@ exports.Code = class Code extends Base
     modifiers.push '*'        if @isGenerator
 
     signature = [@makeCode '(']
+    paramO    = extend o, scope: o.scope.parent
     for param, i in params
       signature.push @makeCode ', ' if i
       signature.push @makeCode '...' if haveSplatParam and i is params.length - 1
-      signature.push param.compileToFragments(o)...
+      signature.push param.compileToFragments(paramO)...
     signature.push @makeCode ')'
 
     body = @body.compileWithDeclarations o unless @body.isEmpty()
```

I'm not sure though if such functions should also be compiled in a safety wrapper, though

```js
(function () {
  var ref
  return (function(a = (ref = foo()) != null ? ref : bar()) {});
})()
```

GeoffreyBooth:
See #4413. If a function parameter is an expression that triggers the creation of a generated variable (e.g. a variable named `ref` or `ref1` etc.), that variable needs to be declared in the parent scope, not the function scope.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -145,6 +145,10 @@
         return fragments;
       }
 
+      compileToFragmentsWithoutComments(o, lvl) {
+        return this.compileWithoutComments(o, lvl, 'compileToFragments');
+      }
+
       // Statements converted into expressions via closure-wrapping share a scope
       // object with their parent closure, to preserve the expected lexical scope.
       compileClosure(o) {
@@ -3671,7 +3675,7 @@
       // parameters after the splat, they are declared via expressions in the
       // function body.
       compileNode(o) {
-        var answer, body, boundMethodCheck, comment, condition, exprs, haveBodyParam, haveSplatParam, i, ifTrue, j, k, l, len1, len2, len3, m, methodScope, modifiers, name, param, paramNames, paramToAddToScope, params, paramsAfterSplat, ref, ref1, ref2, ref3, ref4, ref5, ref6, ref7, ref8, salvagedComments, signature, splatParamName, thisAssignments, wasEmpty;
+        var answer, body, boundMethodCheck, comment, condition, exprs, haveBodyParam, haveSplatParam, i, ifTrue, j, k, l, len1, len2, len3, m, methodScope, modifiers, name, param, paramNames, paramToAddToScope, params, paramsAfterSplat, ref, ref1, ref2, ref3, ref4, ref5, ref6, ref7, signature, splatParamName, thisAssignments, wasEmpty;
         if (this.ctor) {
           if (this.isAsync) {
             this.name.error('Class constructor may not be async');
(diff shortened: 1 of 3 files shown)


vendethiel:
Seems dangerous WRT recursive calls and the like.

GeoffreyBooth:
How so?

connec:
Would the code I proposed in #4413 not be a bit simpler (e.g., compile parameters with the enclosing scope)?

GeoffreyBooth:
> Would the code I proposed in #4413 not be a bit simpler (e.g., compile parameters with the enclosing scope)?

I tried that, but I got lots of failed tests. Also I don't think we want all function parameter variables declared in the parent scope, only generated ones. New variables due to destructuring, for example, shouldn't get declared in there parent scope.

GeoffreyBooth:
This might help illustrate. This code:

```coffee
'use strict'

foo = -> null
bar = -> 33

f = (a = foo() ? bar()) -> a
g = (a = foo() ? bar()) -> a + 1

console.log f()
console.log g()
```
compiles to:
```js
'use strict';
var bar, f, foo, g, ref, ref1;

foo = function() {
  return null;
};

bar = function() {
  return 33;
};

f = function(a = (ref = foo()) != null ? ref : bar()) {
  return a;
};

g = function(a = (ref1 = foo()) != null ? ref1 : bar()) {
  return a + 1;
};

console.log(f());

console.log(g());
```
and when executed prints:
```
33
34
```

connec:
Fair enough, looking back at it I see why it may cause failures.

This LGTM I think, though I'd be interested to know which cases @vendethiel was thinking of above.

vendethiel:
I need to do some testing but I'm on holidays with no way to reliably clone/test, I'm afraid.

vendethiel:
I was thinking of some very contrived case, e.g.

```coffeescript
./bin/coffee -bce 'f = (rec, a = (if rec then (f false) ? 0 else 0), ref) -> rec'
```

Where the `ref` would be poisoned by a call to the function itself in an argument default.

GeoffreyBooth:
@vendethiel Your example becomes:

```js
var f, ref1;

f = function(rec, a = (rec ? (ref1 = f(false)) != null ? ref1 : 0 : 0), ref) {
  return rec;
};
```

And when run with e.g. `f(1, 2)` it returns `1`. Does this look good to you?

vendethiel:
I think it's fine, yes.

GeoffreyBooth:
Thanks @vendethiel.

@lydell, any notes? @connec, does this look good to you?

Issue #1349, opened 2011-05-11
Invalid JS: Splatted statement causes if block inside array element


satyr:
```
$ coffee -bpe '[a if b...]'
var __slice = Array.prototype.slice;
__slice.call(  if (b) {
    a;
  });

$ coffee -e '[a if b...]'

.:3
  __slice.call(    if (b) {
                   ^^
SyntaxError: Unexpected token if
```

GeoffreyBooth:
Either a duplicate or else very closely related to #4260.

GeoffreyBooth:
Fixed by #4644.

PR #4643, opened 2017-08-16
[CS1] fix #4260 and #1349: splat error with soak properties or expressions

### LINKED ISSUE #4260: Invalid JS: `[a?.b…]` compiles to invalid JavaScript

zdenko:
This is a fix for #4260 and #1349 when invalid JS is compiled if `Splat` contains soak accessor  (`?.`), or `if` statement.

Before:
```javascript
// [a if b...]
slice.call(  if (b) {
    a;
  });

// [a?.b...]
slice.call(  if (typeof a !== "undefined" && a !== null) {
    a.b;
  });

// arr.push(error.data?.errors...)
arr.push.apply(arr, if ((ref = error.data) != null) {
  ref.errors;
});
```

After:
```javascript
// [a if b...]
slice.call((b ? a : void 0));

// [a?.b...]
slice.call((typeof a !== "undefined" && a !== null ? a.b : void 0));

// arr.push(error.data?.errors...)
arr.push.apply(arr, ((ref = error.data) != null ? ref.errors : void 0));
```

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1827,7 +1827,15 @@ exports.Splat = class Splat extends Base
     if not (@name instanceof Value) or
         (not (@name.base instanceof Parens) and @propHasSoak())
       @name = new Value new Parens @name
-    @name.compileToFragments o
+      # We need to replace `void 0` with `[]` in compiled fragments.
+      # Example: [a?.b...]
+      # slice.call((typeof a !== "undefined" && a !== null ? a.b : []));
+      fragments = @name.compileToFragments o
+      for fragment, ix in fragments
+        fragments[ix].code = "[]" if fragment.code == "void 0" and fragment.type == "If"
+      fragments
+    else
+      @name.compileToFragments o
 
   unwrap: -> @name
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1812,30 +1812,8 @@ exports.Splat = class Splat extends Base
   assigns: (name) ->
     @name.assigns name
 
-  propHasSoak: ->
-    return no unless @name.properties
-    isSoak = yes for prop in @name.properties when prop.soak
-    isSoak ? no
-
-  compileToFragments: (o) ->
-    # Check if @name is not an instance of `Value` or @name properties contains soak accessor, e.g. ?.b,
-    # and ensure correct compilation by wrapping the @name in `Parens`.
-    # Examples:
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
See #4644. I’m not sure this bug is severe enough that it’s worth fixing in 1.x. Since the current output throws runtime errors, this bug doesn’t cause any bugs in anyone’s code; it’s just that some syntax that should be allowable just isn’t, until #4644 gets merged in. Or to look at it another way, we add support for this syntax in CoffeeScript 2 😄

zdenko:
@GeoffreyBooth `void 0` is the problem. When the `slice` is called the correct output should be:

```javascript
// [a?.b...]
slice.call((typeof a !== "undefined" && a !== null ? a.b : []));
```

I'll take a look.

zdenko:
@GeoffreyBooth I think this should be fixed now.

src/nodes.coffee
```diff
@@ -1812,7 +1812,7 @@ exports.Splat = class Splat extends Base
   assigns: (name) ->
     @name.assigns name
 
-  compileToFragments: (o) ->
+  compileNode: (o) ->
     @name.compileToFragments o
```
GeoffreyBooth: Should this get `LEVEL_OP` too?

connec: In this case no, since in CS1 splats are not operators, but are instead handled by `Arr` and `Assign`.

PR #4645, opened 2017-08-16
Fix #4558: Stack trace line numbers for scripts that compile CoffeeScript

### LINKED ISSUE #4558: CLI: Line numbers shown as `NaN` in stack trace when compiling CoffeeScript inside CoffeeScript
When calling the following script lineNumbers.coffee:
```coffee
cs = require 'coffee-script'
cs.compile ""

throw new Error "Some Error"
```
using `coffee lineNumbers.coffee`
I get the following stacktrace:
```
Error: Some Error
    at Object.<anonymous> (F:\test\lineNumbers.coffee:NaN:NaN)
    at Object.<anonymous> (F:\test\lineNumbers.coffee:NaN:NaN)
    at Module._compile (module.js:NaN:NaN)
    at Object.exports.run (F:\test\node_modules\coffee-script\lib\coffee-script\coffee-script.js:NaN:NaN)
    at compileScript (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:NaN:NaN)
    at compilePath (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:NaN:NaN)
    at Object.exports.run (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:NaN:NaN)
    at Object.<anonymous> (C:\Users\Administrator\AppData\Roaming\npm\node_modules\coffee-script\bin\coffee:NaN:NaN)
    at Module._compile (module.js:NaN:NaN)
    at Object.Module._extensions..js (module.js:NaN:NaN)
    at Module.load (module.js:NaN:NaN)
    at tryModuleLoad (module.js:NaN:NaN)
    at Function.Module._load (module.js:NaN:NaN)
    at Module.runMain (module.js:NaN:NaN)
    at run (bootstrap_node.js:NaN:NaN)
    at startup (bootstrap_node.js:NaN:NaN)
    at bootstrap_node.js:NaN:NaN
```

When commenting the compile call out:
```coffee
cs = require 'coffee-script_1_12_1_works'
#cs.compile ""

throw new Error "Some Error"
```
I get a reasonable line numbers:
```
Error: Some Error
    at Object.<anonymous> (F:\test\lineNumbers.coffee:4:7)
    at Object.<anonymous> (F:\test\lineNumbers.coffee:1:1)
    at Module._compile (module.js:571:32)
    at Object.exports.run (F:\test\node_modules\coffee-script\lib\coffee-script\coffee-script.js:173:23)
    at compileScript (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:224:29)
    at compilePath (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:174:14)
    at Object.exports.run (F:\test\node_modules\coffee-script\lib\coffee-script\command.js:98:20)
    at Object.<anonymous> (C:\Users\Administrator\AppData\Roaming\npm\node_modules\coffee-script\bin\coffee:15:45)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:427:7)
    at startup (bootstrap_node.js:151:9)
    at bootstrap_node.js:542:3
```

This occurs using coffee-script 1.12.6. This behaviour is new, everything works correctly using coffee-script 1.12.1 and below. I also get proper line numbers using the newest coffeescript version when I execute the file using `node -r coffee-script/register lineNumbers.coffee`

GeoffreyBooth:
Fixes #4558. Also fixes the `cake release` command, which broke somewhere around when Node 8 came out.

Feedback welcome. So basically the issue in #4558 is that when you have a CoffeeScript file that itself has `CoffeeScript = require 'coffeescript'` and then something like `CoffeeScript.compile`, there’s basically a script within a script here and we need to preserve the source maps within the source maps, so that stack trace line numbers work correctly. Unfortunately the child scripts are all cached as `<anonymous>`, and the runtime always throws errors referencing the filename of the top-level script, so there’s some dark arts involved in picking which source map to use for patching the error stack trace. The method I’ve chosen isn’t foolproof, if you have multiple source maps that happen to have identical code or mappings, but it should work most of the time. I don’t see how to get a foolproof method without the runtime providing the full source of the “file” throwing the error, which it doesn’t (as far as I can tell). There’s more gory detail in the comments of the PR.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -341,11 +341,15 @@ task 'doc:source:watch', 'watch and continually rebuild the annotated source doc
 
 
 task 'release', 'build and test the CoffeeScript source, and build the documentation', ->
-  invoke 'build:full'
-  invoke 'build:browser:full'
-  invoke 'doc:site'
-  invoke 'doc:test'
-  invoke 'doc:source'
+  execSync '''
+    cake build:full
+    cake build:browser
+    cake test:browser
+    cake test:integrations
+    cake doc:site
+    cake doc:test
+    cake doc:source''', stdio: 'inherit'
+
 
 task 'bench', 'quick benchmark of compilation time', ->
   {Rewriter} = require './lib/coffeescript/rewriter'
--- a/docs/v2/index.html
+++ b/docs/v2/index.html
@@ -4682,7 +4682,7 @@ $(document).ready ->
     else
       initializeScrollspyFromHash window.location.hash
       # Initializing the code editors might’ve thrown off our vertical scroll position
-      document.getElementById(window.location.hash.slice(1)).scrollIntoView()
+      document.getElementById(window.location.hash.slice(1).replace(/try:.*/, '')).scrollIntoView()
 
 </script>
 
--- a/documentation/v2/docs.coffee
+++ b/documentation/v2/docs.coffee
@@ -125,4 +125,4 @@ $(document).ready ->
     else
       initializeScrollspyFromHash window.location.hash
       # Initializing the code editors might’ve thrown off our vertical scroll position
-      document.getElementById(window.location.hash.slice(1)).scrollIntoView()
+      document.getElementById(window.location.hash.slice(1).replace(/try:.*/, '')).scrollIntoView()
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -4,7 +4,8 @@
   // on Node.js/V8, or to run CoffeeScript directly in the browser. This module
   // contains the main entry functions for tokenizing, parsing, and compiling
   // source CoffeeScript into JavaScript.
-  var Lexer, SourceMap, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, sourceMaps, sources, withPrettyErrors;
+  var FILE_EXTENSIONS, Lexer, SourceMap, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, sourceMaps, sources, withPrettyErrors,
+    indexOf = [].indexOf;
 
   ({Lexer} = require('./lexer'));
 
@@ -21,7 +22,7 @@
   // The current CoffeeScript version number.
   exports.VERSION = packageJson.version;
 
-  exports.FILE_EXTENSIONS = ['.coffee', '.litcoffee', '.coffee.md'];
+  exports.FILE_EXTENSIONS = FILE_EXTENSIONS = ['.coffee', '.litcoffee', '.coffee.md'];
 
   // Expose helpers for testing.
   exports.helpers = helpers;
@@ -67,10 +68,10 @@
   // a stack trace. Assuming that most of the time, code isn’t throwing
   // exceptions, it’s probably more efficient to compile twice only when we
   // need a stack trace, rather than always generating a source map even when
-  // it’s not likely to be used. Save in form of `filename`: `(source)`
+  // it’s not likely to be used. Save in form of `filename`: [`(source)`]
   sources = {};
 
-  // Also save source maps if generated, in form of `filename`: `(source map)`.
+  // Also save source maps if generated, in form of `(source)`: [`(source map)`].
   sourceMaps = {};
 
   // Compile CoffeeScript code to JavaScript, using the Coffee/Jison compiler.
@@ -93,7 +94,10 @@
     generateSourceMap = options.sourceMap || options.inlineMap || (options.filename == null);
     filename = options.filename || '<anonymous>';
     checkShebangLine(filename, code);
-    sources[filename] = code;
+    if (sources[filename] == null) {
+      sources[filename] = [];
+    }
+    sources[filename].push(code);
     if (generateSourceMap) {
       map = new SourceMap;
     }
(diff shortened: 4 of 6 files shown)


jashkenas:
Yeesh this looks a little hacky. Can we not correctly associate the scripts with the maps by maintaining a correct lookup somehow?

GeoffreyBooth:
That’s kind of what’s happening now. The issue is that there are multiple maps per file when a file runs `CoffeeScript.compile` or `run` etc. For example if this is `test.coffee`:

```coffee
console.log 'Hello!'

CoffeeScript.run '''
    CoffeeScript = require './lib/coffeescript
    CoffeeScript.compile ''
    throw new Error 'Some Error'
  '''
```

This generates three source maps, all under `test.coffee`: the entire file, the string passed to `.run`, and the string passed to `.compile`. Hence the `sources` and `sourceMaps` objects now are in the form where the key is `test.coffee` and the value is an array of these three sources/source maps. The issue is that when an exception is thrown, the browser just tells us that it’s from `test.coffee`—there’s no way to know which one of these three the browser refers to.

There is a method called `frame.getFunction().toString()` that gives us the source of the function containing the exception, but it’s not always available. I thought about using it to get the source and searching through the `test.coffee` sources array to find the first one that had the source of the function in question, which still wouldn’t be foolproof if the “child scripts” had the same source code for that function, but I found `getFunction` too unreliably available to use that as a solution.

So instead of that, I search through the array of source maps to find the first one that _has_ a matching source map for the exception. This is potentially even less foolproof, as source maps could easily have overlapping matches (especially `1:1`) but it was the best option I could find. There doesn’t seem to be a rule that can be followed reliably like ”exceptions are always from the newest source/source map at the end of the array.”

If you can think of a better solution, please suggest it. I agree that a hacky, non-foolproof matching of source maps line/column pairs is far from ideal, but it’s better than returning `NaN` and it seems to be the best we can do as far as I can tell.

GeoffreyBooth:
So what do we think? @jashkenas or @lydell, is this okay to merge in?

If you think of a better solution for the line numbers problem than this one, we can always submit a new PR later.

lydell:
- Does this cause a risk that code that _doesn't_ have something like `CoffeeScript.compile` in it (which is the by far common case) get wrong numbers?
- According to #4558 the numbers used to be correct. Do we know why they became `NaN`? If they could be correct without hacks before, why can't they now?

I don't doubt that you've found the best solution that can be thought of for the moment, but it would be nice to clear any confusion up before mergin :)

lydell:
- Does this cause a risk that code that _doesn't_ have something like `CoffeeScript.compile` in it (which is the by far common case) get wrong numbers?
- According to #4558 the numbers used to be correct. Do we know why they became `NaN`? If they could be correct without hacks before, why can't they now?

I don't doubt that you've found the best solution that can be thought of for the moment, but it would be nice to clear any confusion up before mergin :)

GeoffreyBooth:
> Does this cause a risk that code that _doesn’t_ have something like `CoffeeScript.compile` in it (which is the by far common case) get wrong numbers?

I hope not! We do have tests for that, though.

> If they could be correct without hacks before, why can’t they now?

They weren’t always correct before. If the exception was thrown in the topmost script (i.e. not a string passed to `compile` or `run` etc.) then yes, in 1.12.1 there were line numbers and after #4428 there weren’t. But if a child or grandchild etc. script (a string passed to `compile` etc.) threw an exception, its exception line numbers would be incorrect. After this PR both should be correct.

> Do we know why they became `NaN`?

So there’s a pretty long, tortured history.

- In 1.12.1 and earlier, we patched `Error.prepareStackTrace` to return line numbers based on the CoffeeScript source, but the implementation was problematic. Specifically, if an exception was thrown, the compiler took the filename from the exception and used `fs` to reopen the file, recompiled it to get a source map, and then used that source map to revise the stack trace. This didn’t work at all in a browser environment, and had other issues. Hence . . .
- In 1.12.2 we removed our patched `Error.prepareStackTrace` entirely. Line numbers in stack traces corresponded to the original runtime line numbers, e.g. the JavaScript source line numbers.
- In 1.12.3 we restored our patched `Error.prepareStackTrace`, per #4428. This version added caching for sources and source maps, to avoid the file reopen on exception. However the caching had an edge case issue discussed above, in that multiple files with the same filename (or multiple files called `<anonymous>`) would overwrite each other’s cached sources/source maps, which led to the issue #4558 that this PR is meant to fix. When a source map couldn’t be found in a cache, or it didn’t match the stack trace, `NaN` was the result. (This is the simplified version. You can read through the diffs of #4428 and this PR for the details.)

lydell:
IMO, we're good to go then.

GeoffreyBooth:
If we want to replicate the 1.12.1 behavior but keep the caching, we could cache the first appearance of each filename but ignore any subsequent appearances. That would be equivalent to the file-open approach, but it would share the bug with that version of incorrect line numbers of child scripts. This bug is more important than it might seem, because lots of our own tests are compilations of child scripts.

jashkenas:
Just throwing out thoughts here -- but what if we stopped using the filename as the key to the source map lookup, and instead used something more robust: like the SHA of the source code to-be-compiled?

GeoffreyBooth:
I thought of that. Unfortunately the issue is that when the runtime throws the exception, all it supplies is the filename, not the source. We have only a filename to go on to find the source/source map.

jashkenas:
Ah, I see. In that case your fix looks perfect.

That said, it seems like with any engine providing source map support, there really ought to be a way to link the specific JS to the map directly.

GeoffreyBooth:
Yes. It seems pretty clear that the designers of the JavaScript stack trace never imagined that there might be multiple different sources with the same filename (or that someone might be rewriting their stack trace, for that matter).

PR #4649, opened 2017-08-16
Documentation: Define the functions used in the examples

bendrexl:
Added a couple basic functions to make the Array Comprehensions example code more self-explanatory. 

The "menu" example especially was hard to read.


documentation/examples/array_comprehensions.coffee
```diff
@@ -1,3 +1,7 @@
+# Methods
+eat = (food) -> return "#{food} eaten."
+menu = (i, dish) -> return "Menu Item #{i}: #{dish}"
```
shreeve: Why 'return' here instead of just:

```coffeescript
eat = (food) -> "#{food} eaten."
menu = (i, dish) -> "Menu Item #{i}: #{dish}" 
```

GeoffreyBooth:
Please target the `2` branch.

documentation/examples/array_comprehensions.coffee
bendrexl: Great point, there's no need - force of habit on my part

documentation/examples/array_comprehensions.coffee
```diff
@@ -1,3 +1,7 @@
+# Methods
+eat = (food) -> "#{food} eaten."
```
GeoffreyBooth: I feel like this line should be just above the `eat food` line, to keep it together.

documentation/examples/array_comprehensions.coffee
```diff
@@ -1,3 +1,7 @@
...
+menu = (i, dish) -> "Menu Item #{i}: #{dish}"
```
GeoffreyBooth: And likewise, I feel like this should be above the `menu i + 1` line.

bendrexl:
Makes sense. It looks like switching the target branch here in the UI is what caused those other "changes" to be committed. The branch history for this PR is already way too messy for a 2-line change, I'll just submit a fresh one based on branch '2'. 

Feel free to close this one.

PR #4658, opened 2017-08-23
[CS2] Fix #4209: --require for filenames that are invalid identifiers

### LINKED ISSUE #4209: CLI: --require options doesn't work for file with illegal chars for JS variable
The following snippet tells everything...

``` bash
$ echo 'console.log "hi"' > /tmp/hello-world.coffee
$ echo 'console.log "hi"' > /tmp/hello_world.coffee
$ coffee -r /tmp/hello_world.coffee -e 'console.log("eval")'
hi
eval
$ coffee -r /tmp/hello-world.coffee -e 'console.log("eval")'
ReferenceError: hello is not defined
```

GeoffreyBooth:
Fixes #4209. I hate that this adds 15K to the size of the compiler, but fortunately the increase is only in the Node version, not the browser version (which doesn’t include `command.coffee`).

Previously `--require hello-world.coffee` would fail because the compiler was prepending `hello-world = require('hello-world.coffee');`. Now the compiler checks that `hello-world` is a valid identifier before outputting it; if it’s invalid, the compiler just outputs `require('hello-world.coffee')` without the assignment.


lydell:
There are other places where we (in theory) need to check for valid identifiers as well, but, after much discussion, chose not to include that crazy regex.

Can't we output something like `global['hello-world'] = require('hello-world.coffee');` instead?

GeoffreyBooth:
D’uh. That’s a much better solution.

jashkenas:
Whew!

PR #4659, opened 2017-08-24
[CS2] Fix #4629: interpolations (whether in strings or CSX tags) with only comments

### LINKED ISSUE #4629: [CS2] Rewriter: Inline block comments in CSX not getting output
`JSX` has a comment syntax `{/*  */}`:
```javascript
<div> {/* A JSX comment */} </div>

<div>
{/* 
  Multiline
  JSX comment
*/}
</div>
```
Currently there is no corresponding comment syntax in `CSX` yet. However, a workaround does exist, by using ```{`/*  */`}```:
```coffeescript
# CSX
<div> {`/* A JSX comment */`} </div>
<div>
{`/* 
  Multiline
  JSX comment
*/`}
</div>

# converts to JSX
<div> {/* A JSX comment */} </div>;
<div>
{/* 
  Multiline
  JSX comment
*/}
</div>;
```
Should this workaround be documented, or implement a specific comment syntax for `CSX`?
It can be `{###  ###}` or `{#  }` or something else. There was some discussion on this topic in the _depreciated `CJSX`_ https://github.com/jsdf/coffee-react-transform/pull/41, and _`CJSX`_ actually implemented the syntax `{#  }`.

GeoffreyBooth:
Fixes #4629.

An optimization in interpolated strings (which CSX tags essentially are) ignored empty interpolations: `"a#{}b"` compiled to just `` `ab` ``, not `` `a${}b` ``. This meant that an interpolation that had nothing but comments, like `"a#{### comment ###}b"`, was also getting output as just `` `ab` ``, with the comment lost.

This PR preserves both the optimization and the comments. If an “empty” interpolation has no comments, the current behavior remains. However if it has comments but no other tokens, a zero-length `JS` token is created in the interpolation to hold the comments; and the optimization isn’t performed.

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -1105,7 +1105,14 @@
                 continue;
               }
               // There are comments (and nothing else) in this interpolation.
-              placeholderToken = this.makeToken('JS', '');
+              if (this.csxDepth === 0) {
+                // This is an interpolated string, not a CSX tag; and for whatever
+                // reason `` `a${/*test*/}b` `` is invalid JS. So compile to
+                // `` `a${/*test*/''}b` `` instead.
+                placeholderToken = this.makeToken('STRING', "''");
+              } else {
+                placeholderToken = this.makeToken('JS', '');
+              }
               // Use the same location data as the first parenthesis.
               placeholderToken[2] = value[0][2];
               for (k = 0, len1 = value.length; k < len1; k++) {
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -1089,7 +1089,7 @@
     // of `'NEOSTRING'`s are converted using `fn` and turned into strings using
     // `options` first.
     mergeInterpolationTokens(tokens, options, fn) {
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Huh. I had no idea `` `a${/*test*/}b` `` is invalid JS. How weird! I wonder why?

So what should we do for strings? Compile to `` `a${/*test*/''}b` ``?

lydell:
Empty interpolations are invalid in JS. I don't know why but that's how it is. I think your solution with the empty string looks good.

GeoffreyBooth:
Done!

connec:
We should probably make `"a#{###test###}b"` invalid CS.

lydell:
@connec You mean empty interpolations in general?

connec:
Yeh, I don't think it's useful to allow empty interpolations.

That said, I think I misunderstood the outcome of the PR - treating empty interpolations as `''` seems reasonable too.

GeoffreyBooth:
It would be a breaking change to no longer allow `"a#{###test###}b"`, not that that would be so terrible. But I think since the solution is so simple, we might as well improve on template literals a little 😄

PR #4661, opened 2017-08-24
[CS2] Fix #2870: Allow specifying output filename

### LINKED ISSUE #2870: CLI: Allow user to set output filename
Suppose we have a file named "master.coffee".

Right now, the CLI for CoffeeScript allows us to specify a specific folder where we want the compiled Javascript file created:

```
--output /folder/subfolder/outputHere
```

This will generate the following file:

```
/folder/subfolder/outputHere/master.js
```

But in many cases, we might want the name of the output file to be different than the name of the source file. I would like the --output flag to support this possibility. If the supplied path ends with a file rather than a folder, the Coffee compiler should use that name for the output file. Example:

```
--output /folder/somefolder/awesome.js
```

CoffeeScript is the only "cutting edge" tool on the market that doesn't have the capability to specify complete output file paths. I know it may seem like a small thing, but when building CoffeeScript into other apps or tools, it really is a frustrating limitation. It's even more of an issue now that CS supports source maps, so we can't just change the file names on the fly; they're hard-coded into the maps.

GeoffreyBooth:
Fixes #2870. If the input is a file, not a folder; and `--output` is a path to a file, like `dist/file.js` or `build.js`; save as the desired output path/filename.

Examples:
```bash
coffee -co dist/testing.js test.coffee # Saves as dist/testing.js
coffee -co testing.js test.coffee      # Saves as testing.js
coffee -co dist test.coffee            # Saves as dist/test.js (same as current)
coffee -co dist src                    # Saves as dist/* (same as current)
```


src/command.coffee
```diff
@@ -102,7 +102,17 @@ exports.run = ->
   process.argv = process.argv[0..1].concat literals
   process.argv[0] = 'coffee'
 
-  opts.output = path.resolve opts.output  if opts.output
+  if opts.output
+    outputBasename = path.basename opts.output
+    if '.' in outputBasename
```
lydell: What about `--output dir/.` and `--output dir/sub/..`? I guess those two are supposed to mean `--output dir/` and `--output dir/`? Also, there's `--output .` and `--output ..`.

Also, a question: Is it possible to add a trailing slash to _force_ a directory? `--output dir/script.js/`

GeoffreyBooth: I was afraid of testing this, since this part of the codebase has no tests.

All those cases are ones I didn’t consider. Let me cover them . . .

GeoffreyBooth: Okay, so now if `path.basename` (which is everything from the last separator, `/` or `\`, to the end of the string) returns only periods, I’m just treating the output argument as a path (folder). This conforms to the old behavior, though it doesn’t seem to make much sense.

If the output argument ends in a slash, I’m forcing it to be treated as a folder name, even if it contains periods. So `coffee -co scripts.js/ test.coffee` will save into `scripts.js/test.js`.

GeoffreyBooth:
I wonder if we should keep this unmerged until 2.0.1, or if we’ve merged in so many bugfixes that we should release a 2.0.0-beta5. I’m eager to release 2.0.0, but I don’t want it to have too much code that hasn’t been more widely tested.

src/command.coffee
```diff
@@ -102,7 +102,19 @@ exports.run = ->
   process.argv = process.argv[0..1].concat literals
   process.argv[0] = 'coffee'
 
-  opts.output = path.resolve opts.output  if opts.output
+  if opts.output
+    outputBasename = path.basename opts.output
+    if '.' in outputBasename and
+       not helpers.ends(opts.output, path.sep) and
+       not /^\.+$/.test(outputBasename) # Not `.` or `..`.
```
lydell: Three or more dots is a valid file name, though. We should only special case `.` and `..`.

GeoffreyBooth: You gotta be kidding me.

lydell:
I vote for a 2.0.0-beta5 btw.

GeoffreyBooth:
Yeah I’m thinking that too. Might as well merge this in then 😄

Issue #3789, opened 2015-01-12
Invalid JS: Nested throw expressions


rightfold:
The following CoffeeScript code:

```
throw throw throw a
```

compiles to the following JavaScript code:

```
throw throw throw a;;;
```

which is invalid syntax in JavaScript.

Instead it should either compile to the following JavaScript code:

```
throw (function() {
    throw (function() {
        throw a;
    })();
})();
```

or be rejected by the CoffeeScript compiler.

Now this code is of course incredibly silly, but the CoffeeScript compiler should under no circumstance both accept the input and generate code with invalid syntax.

jashkenas:
I don't think it should compile. That should be a syntax error.

GeoffreyBooth:
See also #3709.

GeoffreyBooth:
Fixed by #4664.

PR #4664, opened 2017-08-25
[CS2] Fix #3709, #3789: ‘throw’ an ‘if’, ‘for’, ‘switch’, ‘while’

### LINKED ISSUE #3709: Invalid JS: Compiler emits invalid JS when `throw` followed by `if` or `for`
This seemingly valid CS code:

``` coffeescript
throw if pedantic 
  new FatalError
else
  new ResumableError
```

Compiles to invalid JS:

``` javascript
throw if (pedantic) {
  new FatalError;
} else {
  new ResumableError;
};
```

I say seemingly valid because the same thing works for a function call:

``` coffeescript
raise = (e) -> throw e
raise if pedantic 
  new FatalError
else
  new ResumableError
```

``` javascript
var raise;

raise = function(e) {
  throw e;
};

raise(pedantic ? new FatalError : new ResumableError);
```

GeoffreyBooth:
Fixes #3709 and #3789. If the expression to be thrown is of certain types (`If`, `For`, `Switch`, `While` or a nested `Throw`) wrap it in a closure, so that it becomes valid JavaScript.

I’m sure there are probably other expression types that should get wrapped, so please suggest any you can think of and I’ll add them to the list. But these should be the most likely ones that people would think to put after `throw`, (other than another `throw`, which is so unlikely it feels silly to even cover the case). And now we have a way to handle all similar cases.

I tried to find a way to deduce whether the expression should get wrapped in a closure, by looking at `shouldCache` or `isStatement` or the like, but I didn’t find a helper that was consistent. Ultimately I don’t think there are too many node types that make sense to put after `throw`, so just listing every possibility should hopefully be good enough.

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3258,6 +3258,7 @@ exports.Throw = class Throw extends Base
     if @expression instanceof For or
        @expression instanceof If or
        @expression instanceof Switch or
+       @expression instanceof Throw or
        @expression instanceof While
       fragments = @expression.compileClosure o
     else
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3255,7 +3255,14 @@ exports.Throw = class Throw extends Base
   makeReturn: THIS
 
   compileNode: (o) ->
-    fragments = @expression.compileToFragments o
+    if @expression instanceof For or
+       @expression instanceof If or
+       @expression instanceof Switch or
+       @expression instanceof While
+      fragments = @expression.compileClosure o
+    else
+      fragments = @expression.compileToFragments o
+
     unshiftAfterComments fragments, @makeCode 'throw '
     fragments.unshift @makeCode @tab
     fragments.push @makeCode ';'
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3255,15 +3255,7 @@ exports.Throw = class Throw extends Base
   makeReturn: THIS
 
   compileNode: (o) ->
-    if @expression instanceof For or
-       @expression instanceof If or
-       @expression instanceof Switch or
-       @expression instanceof Throw or
-       @expression instanceof While
-      fragments = @expression.compileClosure o
-    else
-      fragments = @expression.compileToFragments o
-
+    fragments = @expression.compileToFragments o, LEVEL_LIST
     unshiftAfterComments fragments, @makeCode 'throw '
     fragments.unshift @makeCode @tab
     fragments.push @makeCode ';'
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -3255,7 +3255,15 @@ exports.Throw = class Throw extends Base
   makeReturn: THIS
 
   compileNode: (o) ->
-    fragments = @expression.compileToFragments o
```
connec: Is there any chance that changing this to `@expression.compileToFragments o, LEVEL_LIST` would work as well?

connec: Looks like the tests still pass with:

```diff
diff --git a/src/nodes.coffee b/src/nodes.coffee
index 677d4df0..f3b89902 100644
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3255,7 +3255,7 @@ exports.Throw = class Throw extends Base
   makeReturn: THIS
 
   compileNode: (o) ->
-    fragments = @expression.compileToFragments o
+    fragments = @expression.compileToFragments o, LEVEL_LIST
     unshiftAfterComments fragments, @makeCode 'throw '
     fragments.unshift @makeCode @tab
     fragments.push @makeCode ';'
```

vendethiel: That would also fiw `throw try ...`.

vendethiel: Er, fix*.

GeoffreyBooth: `@expression.compileToFragments o, LEVEL_LIST` is interesting. Care to explain what that means? I've never fully understood the second parameter there.

vendethiel: Since CoffeeScript generates a JS string directly and not a JS ast, we carry around a notion of what context the node is gonna be compiled in, so we can generate expr/statement, surround with parentheses, etc not to mess with the precedence/... (like compiling `{a:5}` toplevel)

connec: That's a good way of putting it, re: there being no JS AST. It's exactly designed as a way of dealing with the situation in this PR (and #4644) - namely that a given JS expression (e.g. `If`, `Obj`, `Op` etc.) that's a child of another expression needs to be compiled in a certain way in order to be valid JS (e.g. wrapped in parens, closures, etc.).

jashkenas:
Just to add some context -- this is what `isStatement` is supposed to be for!

Expressions are expressions, and their values can be used directly -- statements need to be closure-wrapped to be used as values.

There should be a way to use adjust the basic handling mechanism in `compileToFragments` to take care of this, instead of hardcoding a list of node types.

GeoffreyBooth:
Well `@expression.compileToFragments o, LEVEL_LIST` seems to do the job. All the new tests I recently added pass, and simple code like `throw new Error()` aren’t wrapped in parentheses or closures. Thanks for the tip @connec!

jashkenas:
Perfect!

Issue #4576, opened 2017-06-15
Lexer: Breaking change in multiple row function chaining


algesten:
#4535 introduced a breaking change in function chaining over multiple rows.

I bisected master and this worked in 1.12.5, but 51c06574a084388ca9537415b6907a8cb3bac013 broke it for 1.12.6.

Turns out I rely on it pretty heavily to avoid christmas trees.

```coffee
test "Function chaining on separate rows", (cb) ->
  func = -> Promise.resolve().then ->
      true
  .then cb

  func()
```

GeoffreyBooth:
Well we’re not simply going to revert that bugfix, so the question is whether your specific case can be accommodated. Can you please post some example code that broke as a result of that change?

algesten:
Our internal code is mostly coffeescript, and we've encountered this problem at least two places there.

Here's an external example https://github.com/algesten/ductile/blob/master/src/ductile.coffee#L31

algesten:
Is it fair to call that commit a bug fix? It's changing the interpretation.

Obviously stricken up the language is fine, but maybe not in a patch release given that 2.x is around the corner?

jashkenas:
Here's the minimal case that is now broken, with `unexpected .`:

```coffeescript
->
  a
.b
```

Note that with `.b` indented to the level of `a`, or anywhere further in, in compiles as an immediate access against `a`

```coffeescript
->
  a
  .b
```

It seems like the first case ought to compile? And close the function body and then call against it?

It's not great CoffeeScript style, but...

algesten:
So. Any chance this is getting fixed?

GeoffreyBooth:
You’re welcome to submit a bugfix. Or just use parentheses:

```coffee
test "Function chaining on separate rows", (cb) ->
  func = (-> Promise.resolve().then ->
      true
  ).then cb

  func()
```

algesten:
I mean the easiest would be to back out that breaking commit, but I guess there's a case there that you want to defend against. Where do you suggest I start trying to fix this?

GeoffreyBooth:
Look in lexer.coffee and rewriter.coffee for code related to chaining or unclosed lines.

helixbass:
@algesten @GeoffreyBooth I looked at this a bit

@jashkenas it doesn't look like your example has ever compiled:
```
->
  a
.b
```
[...]

PR #4665, opened 2017-08-28
[CS2] Allow chained access of a property of a function literal

GeoffreyBooth:
Implements the proposed syntax discussed in #4576:

```coffee
->
  a
.call(…)
```

This doesn’t “fix” #4576 per se, because the output desired by that issue’s OP was simply incorrect and was fixed in 1.12.6; but this allows an alternative way to chain after function literals without needing to indent (“fixing” #4576 in the sense that the issue was about avoiding Christmas trees while chaining).

@helixbass is this what you had in mind?


helixbass:
@GeoffreyBooth looks good. The description is not quite right, this will change the compilation of the example from #4576 from breaking (current behavior) or compiling to something unintuitive (pre-1.12.6 behavior) to compiling in a way that's more consistent with other chaining and with the chaining-after-top-level-function-literal example you give

GeoffreyBooth:
Fortunately the description doesn't need to be merged. :smile: @lydell, any notes?

lydell:
Well, if it makes

```coffee
Promise
.resolve()
.then ->
  yes
.then ok
```

compile with such a small change then it looks good to me.

GeoffreyBooth:
Hmm, @lydell your example compiles now. I guess I should change the second test.

The first test is definitely good,
```coffee
->
  eq @a, 3
.call a: 3
```
doesn’t currently compile. @helixbass is there another example you can think of that doesn’t compile now, that should after this?

lydell:
@GeoffreyBooth I copied my example from the tests of this PR!

GeoffreyBooth:
I just meant, that test is bad because it compiles fine _without_ this PR. So it’s not testing the change that this PR makes. It might be a good test to keep anyway, as a general test of chaining, but I feel like I should have at least one more test specific to this PR’s change.

GeoffreyBooth:
Thinking about this more, it occurs to me that the only properties or methods you could access off a function literal are the [ones attached to `Function` itself](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function#Properties)—`.call`, `.apply`, `.bind` etc. Which is useful functionality to add, but maybe we don’t need separate tests for more than just `.call`.

I think I’ll merge this in, and if anyone suggests additional tests I’ll add them in a future PR.

Issue #4552, opened 2017-05-16
[CS2] Support property initializers


xixixao:
We should be able to write the equivalent of
```js
class C {
  b = 3;
  c = () => 4;
}
```
Can it be
```
class C
  b: 3
  c: () => 4;
```
or does the `b` property have different semantics?

GeoffreyBooth:
Is this the same as class properties, that are only in stage 2? See https://github.com/jashkenas/coffeescript/issues/4497#issuecomment-293724531

connec:
Bit of a brain dump incoming 😅 

I did some digging into the current status of the public fields proposal. From what I can tell, [this proposal](https://github.com/erights/Orthogonal-Classes) to standardise orthogonal syntax for class properties is the bleeding edge for the standards work, however there doesn't seem to be any concrete end in sight for standardising class properties due to various critiques about the semantics and syntax, and the dependencies between it and other proposals (e.g. private fields).

In general, there seems to be a bit of stagnation in the standards around classes, and there's a whole other discussion about how tools such as babel make that work harder by providing early access to non-standard features like class properties, which people then conflate with standard JS features.

There's no clear way for us to move forward here imo. We either:

- Throw our lot in with babel and others and support their syntax. This will probably mean sacrificing executable class bodies, and would require us to update the syntax if the standard changes (if it's ever standard).

- Become early adopters of the orthogonal class syntax (which isn't even staged yet, afaict). This would probably allow us to continue using ECBs if we were happy to make `own` a reserved word. We'd have to figure out an alternative sigil from `#` for private fields. We'd also have to update if the standard changes (again, if it's ever standard).

- Figure out a syntax and semantics that makes sense for CoffeeScript. This is (clearly) my preference, but it conflicts with the current goal of CS2 of greater alignment between CS and ES. The benefits are that we could consider each change to the ES spec separately and decide whether it's something we want CS to support, and how it should support it. The downside is people don't get the whitespace sensitive ES they're hoping for, and instead have to learn a parallel syntax.

- Continue to hold out for a standard to solidify. This is another robust option, but will undoubtedly cause some adoption resistance and issues asking for support. At some point people will ask why if it's good enough for babel, it's not good enough for CS.

As a bit of an aside, I think the main reasons to care about this feature at all are:

- Improved static analysis. It's much easier to extract fields etc. from a class syntax than scanning a constructor. I think this is a legitimate issue, and more declarative syntax often enables a lot of useful tooling that would otherwise be impractical.

[...]

PR #4668, opened 2017-08-30
Fix #4464: backticked expressions in class body

### LINKED ISSUE #4464: Lexer: Backticks should allow adding code to class body
ES2015 spc allows method definitions to be prefixed by `get` or `set` in which case they are transformed to property getters and setters.

Of the following, only A generates the corresponding ES2015 output:

```coffeescript
class A
	x: -> 42

class B
	get x: -> 42

class C
	`get` x: -> 42

class D
	`get x`: -> 42

class E
	`get x() { return 42; }`
```

B makes a hoisted class body (as expected), C and D are syntax errors, and E, which would seem like it should be the ultimate escape valve, generates a hoisted class body as well.

It would be nice to have some kind of escape valve for modifying the definitions within the non-hoisted portion of ES2015 class body. It would be particularly nice if backticks could be made to serve this function, as they can help support other possible future syntax (e.g. decorators) as well.

GeoffreyBooth:
Fixes #4464, and allows backticks to provide a workaround for class properties (#4552).

This PR treats backticked expressions in a class body as a special case that doesn’t get hoisted; they just get output in the class body:

```coffee
class A
  `b: 3`
  c: -> 4
```
```js
var A;

A = class A {
  b: 3;

  c() {
    return 4;
  }

};
```

This allows backticks to be used for the not-yet-finalized feature of class properties, as shown above (#4552); or any other unsupported code people might want to put in a class body:

```coffee
class A
  `get x() { return 42; }`
```
```js
var A;

A = class A {
  get x() { return 42; };
};
```

The special case is triggered only when the entire expression is backticked. A line that mixes backticked and unbackticked code, assuming it doesn’t throw a compiler error, will be output as it is now (in the hoisted class body). And the backticked expressions are always output first, above the rest of the class body. @connec @xixixao

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1636,17 +1636,16 @@ exports.Class = class Class extends Base
     super()
 
   compileNode: (o) ->
-    @name          = @determineName()
-    executableBody = @walkBody()
+    @name = @determineName()
+    @walkBody()
 
     # Special handling to allow `class expr.A extends A` declarations
     parentName    = @parent.base.value if @parent instanceof Value and not @parent.hasProperties()
     @hasNameClash = @name? and @name is parentName
 
     node = @
-
-    if executableBody or @hasNameClash
-      node = new ExecutableClassBody node, executableBody
+    if @executableBody or @hasNameClash
+      node = new ExecutableClassBody node, @executableBody
     else if not @name? and o.level is LEVEL_TOP
       # Anonymous classes are only valid in expressions
       node = new Parens node
@@ -1666,7 +1665,7 @@ exports.Class = class Class extends Base
 
   compileClassDeclaration: (o) ->
     @ctor ?= @makeDefaultConstructor() if @externalCtor or @boundMethods.length
-    @ctor?.noReturn = true
+    @ctor?.noReturn = yes
 
     @proxyBoundMethods() if @boundMethods.length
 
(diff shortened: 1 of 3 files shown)


connec:
@GeoffreyBooth as I mentioned to you I found a much smaller change that accomplishes the same thing.

GeoffreyBooth:
A simpler solution would be great! That's why I wanted you to look at this :smile

Can you push your version on this branch, to update the PR?

connec:
@GeoffreyBooth I did already 😄  https://github.com/jashkenas/coffeescript/pull/4668/files

GeoffreyBooth:
I don't see your commits.

connec:
I had squashed the original one, but I've split it out now.

src/nodes.coffee
```diff
@@ -1751,11 +1751,10 @@ exports.Class = class Class extends Base
       new Block expressions
 
   # Add an expression to the class initializer
-  #
-  # NOTE Currently, only methods and static methods are valid in ES class initializers.
-  # When additional expressions become valid, this method should be updated to handle them.
```
GeoffreyBooth: Is this comment no longer valid?

connec: I wavered on it a little. I felt it was confusing since the check now also allows `PassthroughLiteral`s, and I didn't want to add that to the comment.

connec: I guess I could go the other way and flesh it out a little to indicate that it's the key method for determining whether something appears in the class initializer or executable body?

GeoffreyBooth: Yeah I think more is better. Maybe even explicitly mention class fields/class properties, that when they reach Stage 4 this is where we detect them. That way future PRs to add things like that look more like your version of this PR and less like mine 😄

connec:
It seems I broke the PR 😢

PR #4669, opened 2017-08-30
[CS2] Fix #3098: Suppressed newline should be unsuppressed by semicolon

### LINKED ISSUE #3098: Lexer: `a = ; 5`: Suppressed newline should be unsuppressed by semicolon

GeoffreyBooth:
Fixes #3098. Now an error is thrown if a semicolon follows an `=` or anything in the `UNFINISHED` group of tokens. No tests fail as a result of this PR, so presumably this isn’t a breaking change (aside from the case of `a = ; 5` which we want to throw an error).

```coffee
> ./bin/coffee -bpe 'a = ; 5'
[stdin]:1:5: error: unexpected ;
a = ; 5
    ^
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -509,7 +509,7 @@ exports.Lexer = class Lexer
         @token 'OUTDENT', moveOut, 0, outdentLength
         moveOut -= dent
     @outdebt -= moveOut if dent
-    @tokens.pop() while @value() is ';'
+    @suppressSemicolons()
 
     @token 'TERMINATOR', '\n', outdentLength, 0 unless @tag() is 'TERMINATOR' or noNewlines
     @indent = decreasedIndent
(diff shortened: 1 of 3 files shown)


src/lexer.coffee
```diff
@@ -662,6 +662,7 @@ exports.Lexer = class Lexer
       @exportSpecifierList = no
 
     if value is ';'
+      @error 'unexpected ;' if prev?[0] in ['=', UNFINISHED...]
```
vendethiel: do we not want `=` in `UNFINISHED`?

vendethiel: or maybe the other way around: what does adding `in UNFINISHED` fixes, given that `prev?[0] is '='` is the relevant fix here? Same in `suppressSemicolons`.

GeoffreyBooth: See https://github.com/jashkenas/coffeescript/issues/3098#issuecomment-26680094

`UNFINISHED` covers all the similar tokens like `+=` etc. I think @jashkenas in his comment was worried about the special cases around `=`, like `=(newline)(object literal)`.

GeoffreyBooth: Also, putting `=` into `UNFINISHED` causes tests to fail.

vendethiel: Ok. LGTM then.

PR #4672, opened 2017-08-30
[CS2] Refine #4666: add parens to chained do IIFE with params

helixbass:
@connec pointed out that #4666 didn't handle `do <IIFE>` with arguments

So this extends #4666 to handle arguments. Added tests where the arguments are on separate lines and contain nested `PARAM_END` token

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -757,9 +757,7 @@
             return 1;
           }
           doIndex = i;
-          this.detectEnd(i + 2, condition, action, {
-            log: true
-          });
+          this.detectEnd(i + 2, condition, action);
           return 2;
         });
       }
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -744,11 +744,12 @@
           return this.tag(i - 1) === 'OUTDENT';
         };
         action = function(token, i) {
-          if (token[0] !== '.') {
+          var ref;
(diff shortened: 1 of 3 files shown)


jashkenas:
And if the arguments have default values or are destructuring?

helixbass:
@jashkenas added a test including destructured param (it was already working) - it's using the rewriter's `detectEnd()`, which takes nesting into account, so I don't think there will be weird specific types of params that this won't handle

My only argument for keeping it this way for now is that the pre-#4666 behavior shouldn't go into 2.0.0. It's obvious that a grammar-based solution will be superior but unless someone wants to get it done pre-2.0.0 (which seems impending?), I think this is ok as a stopgap. I was tempted to go ahead and try and tackle it myself now but don't think I'm quite there yet

jashkenas:
Cool!

GeoffreyBooth:
@connec this look good to you? 

@helixbass I would say go ahead and try putting this into the grammar as a follow-up PR. Even if it slips into 2.0.1, that would be fine. Just because we release 2.0.0 doesn't mean development will stop.

GeoffreyBooth:
If you wanted to pass arguments into an IIFE, wouldn't you need to not use `do`?

```coffee
((a) ->
  {b: a}
)(1).b
```

helixbass:
@GeoffreyBooth you can do it via default-valued args to `do`, at least:
```
do (a=1) ->
  {b: a}
.b
```

Issue #4674, opened 2017-08-31
[CS2] Polyfills?


GeoffreyBooth:
(From @helixbass https://github.com/jashkenas/coffeescript/pull/4652#issuecomment-326139705):

Babel outputs polyfills for code _it_ generates, but not for code the CoffeeScript compiler outputs. In other words, sending Babel this:
```js
var obj2 = {...obj1};
```
will result in [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=G4QwTgBA9gRgVgJggXggbwHRdnAjAXwG4g&debug=false&evaluate=false&lineWrap=true&presets=stage-3%2Cenv&prettier=false&showSidebar=true&targets=&version=6.26.0):
```js
"use strict";

var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };

var obj2 = _extends({}, obj1);
```
But sending it the CoffeeScript compiler output of `obj2 = {obj1...}`:
```js
var obj2;

obj2 = Object.assign({}, obj1);
```
Babel leaves it unmodified. Babel’s [babel-preset-env](https://github.com/babel/babel-preset-env) converts _syntax_ for compatibility with browser presets, but it doesn’t add polyfills that your target browsers may need (such as for `Object.assign` in this case).

I for one was under the mistaken impression that Babel covered us in such cases. I’ve done some research and there doesn’t appear to be any tool out there to scan your code to identify which polyfills you’ll need. There’s [some discussion](https://github.com/babel/babel-preset-env/issues/294) that Babel should do this, but it seems a ways off. Here’s [a great article](https://philipwalton.com/articles/loading-polyfills-only-when-needed/#areas-for-improvement) that mentions the same idea. There’s also http://jscc.info/, but it appears to be several years out of date (it doesn’t know about `Object.assign`, for example).

[...]

PR #4675, opened 2017-08-31
[CS2] use _extends utility instead of Object.assign() for object spreads

helixbass:
See #4674 

This replaces the direct calls to `Object.assign()` currently generated by object spreads with a call to a new utility function `_extends` (directly copied from Babel's equivalent, defaults to using `Object.assign` when it's supported)

Added `eqJS()` test that `_extends` is being generated and used when compiling object spread


GeoffreyBooth:
We should have a test that covers if a user creates a function named `_extends` in the same scope as object destructuring.

GeoffreyBooth:
@helixbass I’ve added the test. Anyone else have any notes? @lydell or @connec?

helixbass:
@GeoffreyBooth thanks for adding the addtl test and docs

Issue #4676, opened 2017-08-31
[CS2] unexpected token (CLI)


coderofsalvation:
```
[sqz@localhost~/Downloads]  $ npm install coffeescript@next -g
/home/sqz/opt/nvm/versions/node/v6.9.1/bin/coffee -> /home/sqz/opt/nvm/versions/node/v6.9.1/lib/node_modules/coffeescript/bin/coffee
/home/sqz/opt/nvm/versions/node/v6.9.1/bin/cake -> /home/sqz/opt/nvm/versions/node/v6.9.1/lib/node_modules/coffeescript/bin/cake
/home/sqz/opt/nvm/versions/node/v6.9.1/lib
└── coffeescript@2.0.0-beta4 

[sqz@localhost~/Downloads]  $ coffee
/home/sqz/opt/nvm/versions/node/v6.9.1/lib/node_modules/coffeescript/lib/coffeescript/repl.js:29
    eval: async function(input, context, filename, cb) {
                ^^^^^^^^
SyntaxError: Unexpected token function
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.exports.run (/home/sqz/opt/nvm/versions/node/v6.9.1/lib/node_modules/coffeescript/lib/coffeescript/command.js:114:14)
    at Object.<anonymous> (/home/sqz/opt/nvm/versions/node/v6.9.1/lib/node_modules/coffeescript/bin/coffee:15:45)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3

[sqz@localhost~/Downloads]  $ uname -a   
Linux localhost 3.18.0-15976-g75b344ab7870 #1 SMP PREEMPT Mon Aug 28 11:13:29 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux
[...]

PR #4679, opened 2017-09-01
[CS2] Don’t require async/await support to run `coffee`

GeoffreyBooth:
Aside from #4604, the `coffee` command runs in Node 6 (though the tests don’t). See #4676. I feel like we might as well support Node 6 if we can, and it seems that all we need to do so is convert one `await` into a `Promise`.

I also added a test for the functionality added in #4604, plus a guard so that the tests (other than the async ones) can also run in Node 6.

I noticed that `cake test` is returning before the async tests have completed; see #4680.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -413,7 +413,7 @@ runTests = (CoffeeScript) ->
     try
       new Function('async () => {}')()
       yes
-    catch exception
+    catch
       no
   else
     [major, minor, build] = process.versions.node.split('.').map (version) ->
--- a/Cakefile
+++ b/Cakefile
@@ -409,6 +409,17 @@ runTests = (CoffeeScript) ->
         description: description if description?
         source: fn.toString() if fn.toString?
 
+  global.supportsAsync = if global.testingBrowser
+    try
+      new Function('async () => {}')()
+      yes
+    catch exception
+      no
+  else
+    [major, minor, build] = process.versions.node.split('.').map (version) ->
+      parseInt version, 10
+    major >= 8 or (major is 7 and minor >= 6)
+
   helpers.extend global, require './test/support/helpers'
 
   # When all the tests have run, collect and print errors.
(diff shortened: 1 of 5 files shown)


Cakefile
```diff
@@ -409,6 +409,17 @@ runTests = (CoffeeScript) ->
         description: description if description?
         source: fn.toString() if fn.toString?
 
+  global.supportsAsync = if global.testingBrowser
+    try
+      new Function('async () => {}')()
+      yes
+    catch exception
```
lydell: `exception` isn't used here so you could just as well remove it.

Cakefile
```diff
@@ -409,6 +409,17 @@ runTests = (CoffeeScript) ->
         description: description if description?
         source: fn.toString() if fn.toString?
 
+  global.supportsAsync = if global.testingBrowser
```
connec: Do we really need this split? Could we not just always feature test it?

connec:
Also, if we want to support node 6 should we update `engines` in `package.json`? https://github.com/jashkenas/coffeescript/blob/2/package.json#L14

Cakefile
GeoffreyBooth: Unfortunately not. See the comment at the top of `test/async.coffee`. Even with a feature test at the top of that file, and a `return` if it fails, a runtime that doesn't support async will throw an error just from trying to parse the file. Also there's an async test in `repl.coffee` now, so I want to feature test this once and cache the result rather than having it in two places.

GeoffreyBooth:
I thought about updating the `engines` key, but I think we should either leave it at 7.6 or up it to 8, because otherwise we'll get people complaining that we don't really support Node 6 because we say that async is supported but if they try to use async in Node 6 (like in the REPL) it'll fail; I don't know, what do you think? I guess you could make the same argument about JSX.

connec:
Yeh... this is a bit of a murky area. I'm not sure how many people really pay attention to the `engines` stuff anyway.

I feel like `package.json` should mandate a minimum version for `coffee` to be functional, and make sure people are aware that syntax such as `await` and `yield` in compiled (or REPL-interpreted) JS will depend on runtime support or further compilation (e.g. babel).

GeoffreyBooth:
I just fear another #4610. But I guess if the tests pass, and `coffee` works, we should say we support Node 6+.

Maybe it’s time to set up Travis CI to run builds/tests in Node 6 and 8 on every push.

PR #4680, opened 2017-09-01
[CS2] Fix async tests

GeoffreyBooth:
Not sure when this started happening, but our current test runs complete before waiting for the async tests to resolve, either passing or failing, so if an async test were to start failing we would never know. To remedy this, I check if a test function is an `instanceof Promise`, [as all async functions are](https://medium.com/@bluepnume/learn-about-promises-before-you-start-using-async-await-eb148164a9c8), and if it is, I save it to an `asyncTests` array. After the synchronous tests are all run, I do `Promise.all asyncTests` to wait for all the async tests to resolve. The consumers of the `runTests` function now expect a `Promise` to be returned; for non-async runtimes, I return an immediately-resolving `Promise` that resolves to whether or not all the tests passed. (I wanted to preserve the ability for the tests to run in Node 6 and non-async-capable runtimes, since it isn’t so much extra effort at least for now. Node 6 supports promises.)

To test this, break one of the tests (async and/or non-async) and make sure the failure appears when you run `cake test`. Do the same when running in Node 6.

(This branches off of #4679, so that it’s possible to run `cake test` in Node 6. The diff should shrink once that PR is merged in.)


Cakefile
```diff
@@ -437,21 +461,25 @@ runTests = (CoffeeScript) ->
       CoffeeScript.run code.toString(), {filename, literate}
     catch error
       failures.push {filename, error}
-  return !failures.length
+
+  Promise.all asyncTests
```
connec: I'd suggest trying to get a 'final' pass/fail from this method if possible, e.g. something like:

```coffee
Promise.all(asyncTests).then  ->
  Promise.reject() if failures.length
```

Then, when handling the results, we can just do:

```coffee
runTests CoffeeScript
  .catch -> process.exit 1
```

GeoffreyBooth: Good call!

Cakefile
```diff
@@ -396,18 +395,40 @@ runTests = (CoffeeScript) ->
...
+        result.then ->
+          passedTests++
+        .catch (err) ->
+          onFail description, fn, err
+      else
+        passedTests++
+    catch err
+      onFail description, fn, err
+
+  global.supportsAsync = if global.testingBrowser
```
connec: I think I made this comment in another PR - but could we not just have the `try` feature-test only? It seems neater to me to have just the one:

```coffee
global.supportsAsync = try
  new Function 'async () => {}'
  yes
catch
  no
```

connec: Otherwise this LGTM :+1:

Issue #4673, opened 2017-08-30
[CS2] Destructuring invalid LHS generates invalid JS


connec:
See http://coffeescript.org/v2/#try:%7B%20%7Bb...%7D...%20%7D%20%3D%20c

```coffee
{ {b...}... } = c
````

Compiles to:

```js
({} = c);
Object.assign({}, b) = objectWithoutKeys(c, []);
```

I suspect we're missing something when verifying that an object is a valid LHS. Specifically, when a splat is present it's name must be a valid LHS.

GeoffreyBooth:
@zdenko?

connec:
I ended up fixing this whilst playing with an alternative fix for #4651 - https://github.com/jashkenas/coffeescript/compare/2...connec:fix-4651-alt

It will need rebased once #4652 is resolved.

GeoffreyBooth:
@connec It’s actually a little worse than your initial comment. Look at the `var` declarations line:

```js
var Object.assign({}, b), b,
  objectWithoutKeys = function(o, ks) { var res = {}; for (var k in o) ([].indexOf.call(ks, k) < 0 && {}.hasOwnProperty.call(o, k)) && (res[k] = o[k]); return res; };

({} = c);
Object.assign({}, b) = objectWithoutKeys(c, []);
```

`var Object.assign({}, b), b`?

helixbass:
Should be fixed by #4683. See its description, I think @connec's example here actually should compile, as eg `({...{...b}} = c)` seems to be valid ES6 (works in Chrome console)

GeoffreyBooth:
Fixed by #4683.

Issue #4657, opened 2017-08-21
[CS2] Array destructuring bug


boris-petrov:
```coffeescript
x = ([a..., b]) ->
  console.log a, b

x [1, 2, 3, 4]
```

This produces the incorrect:

```javascript
var x,
  slice = [].slice;

x = function(arg) {
  var arg, i;
  a = 2 <= arg.length ? slice.call(arg, 0, i = arg.length - 1) : (i = 0, []), b = arg[i++];
  return console.log(a, b);
};

x([1, 2, 3, 4]);
```

As you can see, the variable `a` is not defined anywhere and this blows up.

Using CoffeeScript 2.0.0-beta4.

PS. Actually `b` is not defined either.

vendethiel:
Fixed by #4640. Not yet released.

boris-petrov:
@vendethiel - is this supposed to be fixed in beta5? I have a case where it doesn't work with it. Should I try to create a small reproduction or the fix for this is not merged in beta5?

vendethiel:
Please try to create a repro

helixbass:
@GeoffreyBooth this seems to have been introduced by #4478. Should be reopened

`([a..., b]) ->` illustrates the problem - `a` and `b` aren't declared in the compilation:
```
var slice = [].slice;

(function(arg) {
  var arg, i;
  a = 2 <= arg.length ? slice.call(arg, 0, i = arg.length - 1) : (i = 0, []), b = arg[i++];
});
```

helixbass:
Should be fixed by #4683

boris-petrov:
Thanks for the quick fixes. I'll wait for beta6 then, hopefully it comes out soon. :)

GeoffreyBooth:
Fixed by #4683.

PR #4683, opened 2017-09-03
[CS2] Fix destructuring bugs #4673 and #4657

helixbass:
@GeoffreyBooth in looking at #4673, it turned out to be related in a way to #4657 (which is a bug and should be open). So this should fix both

#4673 has a lot of other broken cases (really most non-identifier splat "variables" in an object spread destructure) eg `{a.b...} = c`, `{{d}...} = e`. Including corresponding broken cases for destructured object params eg `({@a...}) ->`. I think all of these (including @connec's example in #4673) should compile based on what ES6 allows and what types of "variables" are allowed in other types of destructurings. But all are currently emitting crazy declarations for the "variable" eg `var a.b`

So it seemed like the cleanest way to get them all compiling correctly would be to fall back to `Assign`'s understanding of how to compile assignments with more complex left-hand-sides (including when to declare a `var`). This meant getting rid of `setScopeVar()`, which was sort of the culprit for these crazy `var` declarations, and just allowing `Assign` to take responsibility for the declarations

It's possible @zdenko introduced `setScopeVar()` partly because of #4657, which is about missing declarations when compiling functions that can't target ES6 destructuring (eg a non-last splat in a destructured array param). So to get both of these to correctly declare destructured vars (even in the presence of outer vars of the same name) was mostly a question of making sure `@param` was set on the various generated `Assign`s

I then ran into top-level generated object destructurings no longer wanting to wrap themselves in parens, so I introduced a distinction between `@param = yes` and `@param = 'alwaysDeclare'` which I think corresponds to "this is actually a param" vs "this needs to always get declared even if there's an outer var of the same name". There may be a cleaner way to accomplish this and it's possible that not all of the `alwaysDeclare`s that I introduced are actually necessary to cover all of the cases

But it looks like it does cover the different cases, I added tests for some of the various more complex object spread "variables" both in destructured assignment and as params as well as a test for the example from #4657

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -3201,7 +3201,7 @@
             if (this.moduleDeclaration) {
               return o.scope.add(name.value, this.moduleDeclaration);
             } else if (this.param) {
-              return o.scope.add(name.value, 'var');
+              return o.scope.add(name.value, this.param === 'alwaysDeclare' ? 'var' : 'param');
             } else {
               return o.scope.find(name.value);
             }
@@ -3794,7 +3794,7 @@
               } else {
                 if ((param.value != null) && !param.assignedInBody) {
                   ref = new Assign(new Value(param.name), param.value, null, {
-                    param: 'alwaysDeclare'
+                    param: true
                   });
                 } else {
                   ref = param;
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2109,15 +2109,15 @@
 
       // Check if object contains splat.
       hasSplat() {
-        var j, len1, prop, ref1, splat;
+        var j, len1, prop, ref1;
         ref1 = this.properties;
         for (j = 0, len1 = ref1.length; j < len1; j++) {
           prop = ref1[j];
           if (prop instanceof Splat) {
-            splat = true;
+            return true;
           }
         }
-        return splat != null ? splat : false;
+        return false;
       }
 
       compileNode(o) {
@@ -2240,7 +2240,7 @@
       }
 
       // Object spread properties. https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md
-      // `obj2 = {a: 1, obj..., c: 3, d: 4}` → `obj2 = Object.assign({}, {a: 1}, obj, {c: 3, d: 4})`
+      // `obj2 = {a: 1, obj..., c: 3, d: 4}` → `obj2 = _extends({}, {a: 1}, obj, {c: 3, d: 4})`
       compileSpread(o) {
         var _extends, addSlice, j, len1, prop, propSlices, props, slices, splatSlice;
         props = this.properties;
@@ -3200,6 +3200,8 @@
             this.checkAssignability(o, name);
             if (this.moduleDeclaration) {
               return o.scope.add(name.value, this.moduleDeclaration);
+            } else if (this.param) {
+              return o.scope.add(name.value, 'var');
             } else {
               return o.scope.find(name.value);
             }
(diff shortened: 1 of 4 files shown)


GeoffreyBooth:
When I recompiled the compiler source as a result of this PR, I noticed lots of changes across every file. It turns out, as a result of this PR, function parameter default values were now getting redeclared inside functions:

```coffee
fn = (options = {}) ->
  options.foo = yes
```
```js
var fn;

fn = function(options = {}) {
  var options;
  return options.foo = true;
};
```

This doesn’t seem to have any practical effect—all the tests still pass—but it looks wrong.

GeoffreyBooth:
In general, though, this PR is an improvement. If there was one bit of general feedback from the object destructuring work, it was that those PRs didn’t use enough of “the machinery,” e.g. the classes and methods we already have written for determining things like setting scope and handling the left hand sides of assignments. My feeling was that a lot of the object destructuring code would ultimately be temporary, as hopefully object destructuring will reach Stage 4 soon and we can just compile to it rather than into `Object.assign`, so I wasn’t too concerned; but it doesn’t hurt to start that refactoring work sooner if we can.

helixbass:
@GeoffreyBooth fixed param redeclaration - it also had to do with the `@param = yes` vs `@param = 'alwaysDeclare'` distinction. Not sure why those changes didn't show up in `lib/` when I'd compiled - I usually run `git checkout lib && bin/cake build:except-parser` but seems like for whatever reason this time `bin/cake build:full` was required

Agreed that some of this stuff will hopefully be pretty temporary but this was a lot of cases where the compiler was emitting broken JS (even if the use cases for object-spread-destructuring into a non-simple variable are slightly obscure) so worth fixing

GeoffreyBooth:
That seemed to have done the trick. Looks good to me, but @connec should also review it.

GeoffreyBooth:
@jashkenas / @lydell / @connec / @helixbass I’m thinking of merging this in (once it gets approval from more people than just me) and releasing it as part of 2.0.0. So basically 2.0.0 would be beta5 plus this PR, assuming no more bugs get reported. So there’s some risk with this plan, in that 2.0.0 wouldn’t have had a group review like beta5 is getting; but I think that’s preferable to either a beta6 or holding this PR for 2.0.1. What do you think?

And #4668 can be the first PR merged into 2.0.1, as it’s not a bugfix.

helixbass:
@GeoffreyBooth yikes I don't want to be the guy who broke 2.0.0. But if @connec thinks it looks ok and you don't want to release another beta then I guess that makes sense. I do think it should get merged in pre-2.0.0 - it fixes a lot of cases (a la #4673) where the compiler is currently outputting invalid JS (not sure quite how obscure the broken cases are, basically any complex "variable" being used in an object spread destructuring) and #4657 does come up in real code (as evidenced by the naturally surfacing bug report)

connec:
The code looks fine to me, the removal of `setScopeVar` is definitely the right thing to do. Unfortunately I'm 'low-tech' (phone+kindle - not the best development setup) for the next week so I can't really play with it, though the tests look good.

One thing I can't quite grok from the description+diff is the need for `alwaysDeclare` - is it just to handle 'params' that we also definitely need a `var` for (as opposed to most params, which we expect to be 'declared' in the function signature)? I have a memory of something in `Scope` already that forces a declaration, is that no use?

GeoffreyBooth:
@connec Assuming @helixbass has a good answer to your question, do you think this would be okay to merge in? Or would you rather we wait until you can get back and test it properly?

helixbass:
@connec yup that's exactly what I was trying to accomplish with `alwaysDeclare`. If there's a way to do that more gracefully with `Scope` I'm not too familiar with it but happy to take a closer look

connec:
I think this looks fine then. I can't remember (or check) enough to know if there's a scope method that can do this.  I'd say we merge it and enhance later if we find something.

GeoffreyBooth:
@lydell or @jashkenas any thoughts?

PR #4689, opened 2017-09-08
Fix #4686: Comments inside JSX interpolation should force wrapping by braces

### LINKED ISSUE #4686: Bug: Nodes: Comments in JSX end up in text nodes
```cjsx
<div>
   {
      # a useful comment
      <div />
   }
</div>

<div>
  <div anAttr={
    # a useful comment
    "value"
  } />
</div>
```
emits
```jsx
<div>
   // a useful comment
   <div />
</div>;

<div>
  // a useful comment
  <div anAttr={"value"} />
</div>;
```

These line comments are not in curly braces, so JSX/babel treats them as text, not comments, and we see 
```
a useful comment
a useful comment
```
 on screen in the browser.

Coffeescript2 should instead be generating

```jsx
<div>
   {
      // a useful comment
      <div />
   }
</div>;

<div>
  <div anAttr={
    // a useful comment
    "value"
  } />
</div>;
```

GeoffreyBooth:
Fixes #4686, at least for the examples in that issue. @jaredp, can you please try this PR on some more complex code and let me know if it works for more complicated cases?

So CSX/JSX is really just a StringWithInterpolations where the `<`/`>` characters are treated as the string delimiters (i.e. like quotation marks). And just like with ES2015 template literals (strings delimited by backticks) braces are used to delimit interpolations; but where ES2015 uses `${` to denote the start of an interpolation, JSX uses just `{`. CoffeeScript’s version of JSX follows JSX’s lead.

One type of “interpolation,” as far as the compiler is concerned, is a child JSX tag:
```jsx
<div>
  <div />
</div>
```
You _could_ write this with braces:
```jsx
<div>
  {<div />}
</div>
```
And it would be equivalent, and Babel’s JSX transform would produce exactly the same output. The braces here are redundant, like an extra pair of parentheses, so you won’t find much handwritten JSX code looking like this. The CoffeeScript compiler has a special check to omit these braces when compiling a JSX tag inside another JSX tag.

That special check, however, wasn’t accounting for comments. If there are comments attached to that inner `div`, that would cause them to appear on the line above the element, those comments need to trigger braces to wrap both the comments and inner JSX tag:

```cjsx
<div>
  {
    # comment
    <div />
  }
</div>
```
```jsx
<div>
  {  // comment
<div />}
</div>;
```

Admittedly, the indentation of the output is terrible, but Babel doesn’t care. It treats this comment as a comment, not part of the HTML content. Ditto for attributes:

```cjsx
<div>
  <div anAttr={
    # comment
    "value"
  } />
</div>
```
```jsx
<div>
  {  // comment
<div anAttr={"value"} />}
</div>;
```

This PR is a one-line fix, so ~~how much harm could it do~~ hopefully it doesn’t affect too much. @xixixao, do you mind taking a look?


GeoffreyBooth:
@lydell or @connec?

lydell:
A quote from the first section on coffeescript.org:

> The compiled output is readable, pretty-printed

I don't think the indentation of the output is acceptable. But it is definitely not enough to hold up the 2.0 release. It can always be improved later.

I think this looks good to merge.

GeoffreyBooth:
There’s also #4692, I think fixing that one should cover fixing the indentation of the output.

@jaredp, did you get a chance to try this PR on your project?

jaredp:
lgtm

PR #4695, opened 2017-09-10
CoffeeScript 2 announcement

GeoffreyBooth:
See https://github.com/GeoffreyBooth/coffeescript/blob/announcing-2/documentation/sections/announcing_coffeescript_2.md

Feedback very much welcome. This feels too long, yet I also feel like I’m missing topics that should be covered. There are also no code samples in this, though I’m not sure where I’d add any (especially not if I want to avoid this getting really really long).

For the final product I’ll find a way to output this like the current docs page, and save it into `/docs/v2/announcing-coffeescript-2/index.html` so that it appears at http://coffeescript.org/announcing-coffeescript-2/ once we launch 2.0.0 and make the 2 docs the primary docs.

@jashkenas @lydell @connec


documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,82 @@
...
+- Classes
+- Async functions, a.k.a. `async`/`await`
+- Bound/arrow functions
+- Function default parameters
+- Function splat/rest parameters
+- Destructuring, for both arrays and objects
+- Object rest/spread properties
+- Interpolated strings/template literals (ES2015 backticked strings)
+- Tagged template literals
+- ES2015’s `for…of` is now available as CoffeeScript’s `for…from` (we already had a `for…of`)
```
lydell: Idea: Will it be easier for folks who aren’t super familiar with the names of every JS feature if we add a “code example” to each bullet point?

```
- Modules (`import`/`export`)
- Classes (`class Animal`)
- Async functions (`async`/`await`)
- Bound/arrow functions (`=>`)
- Function default parameters (`(delay = 0) ->`)
- Function splat/rest parameters (`(items...) ->`)
- Destructuring, for both arrays and objects (`[first, second] = items`/`{length} = items`)
- Object rest/spread properties (`{options..., force: true}`/`{force, otherOptions...} = options`)
- Interpolated strings/template literals (ES2015 backticked strings) (`"Hello, #{user}!"`)
- Tagged template literals (`html"<strong>coffee</strong>"`)
- ES2015’s `for…of` is now available as CoffeeScript’s `for…from` (we already had a `for…of`)
```

documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,82 @@
...
+
+Not all CoffeeScript features were adopted as is; most notably, [default values](http://coffeescript.org/v2/#breaking-changes-default-values) in ES2015 and CoffeeScript 2 are only applied when a variable is `undefined`, not `undefined` or `null` as in CoffeeScript 1; and [classes](http://coffeescript.org/v2/#breaking-changes-classes) have their own differences. See the [breaking changes](http://coffeescript.org/v2/#breaking-changes) for the fine details.
```
lydell: I found it a bit confusing who was adopting what here at first. Doing something like the following might help:

- `Not all CoffeeScript features were adopted as is` → `Not all CoffeeScript features were adopted by ES2015 100% the same way`
- `default values in ES2015 and CoffeeScript 2` → `default values in ES2015 (and also CoffeeScript 2)`

documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,82 @@
...
+
+In our experience, most breaking changes are edge cases that should affect very few people, like ES2015’s [lack of an `arguments` object inside arrow functions](http://coffeescript.org/v2/#breaking-change-fat-arrow). There seem to be two breaking changes that affect a significant number of projects:
+
+- In CoffeeScript 2, “bare” `super` (calling `super` without arguments) is now no longer allowed, and one must use `super()` or `super arguments...` instead.
+- References to `this`/`@` cannot occur before a call to `super`, per the ES spec.
+
+See the [full details](http://coffeescript.org/v2/#breaking-changes-super-extends). Either the CoffeeScript compiler or Babel will throw errors for either of these cases, so updating your code is a simple matter of fixing each instance as the compiler errors on it, until your code compiles successfully.
```
lydell: Perhaps remove the word “simple” in this paragraph?

documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,82 @@
...
+
+Besides supporting new ES2015+ features and outputting older CoffeeScript features in ES2015+ syntax, CoffeeScript 2 has added support for the following:
+
+- [JSX](http://coffeescript.org/v2/#jsx)
+- [Line comments](http://coffeescript.org/v2/#comments) are now output (in CoffeeScript 1 they were discarded)
+- Block comments are now allowed anywhere, enabling [static type annotations](http://coffeescript.org/v2/#type-annotations) using Flow’s comment-based syntax
+
+There are many smaller improvements as well, such as to the `coffee` command-line tool. You can read all the details in the [changelog](http://coffeescript.org/v2/#changelog) for the 2.0.0 betas.
+
+## What About…
```
lydell: Perhaps `“What About …?”`

documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,82 @@
...
+
+Fortunately, most features added by ECMA in recent years haven’t required any updates at all in CoffeeScript. New global objects, or methods on global objects, are just supported and output as is. Want to use `Object.assign` in CoffeeScript? Just type `Object.assign`. Ditto for `Array.forEach` and `Array.map` and `String.includes` and any number of excellent additions to the language that ECMA has added. You may need to add [polyfills](https://babeljs.io/docs/usage/polyfill/) for these methods if your target runtime(s) don’t support them, but that would be no different than if you used these methods in a plain JavaScript project.
```
lydell: This section about polyfills is also a candidate for shortening or removal if you want to make the whole thing shorter.

GeoffreyBooth:
I guess the question is, who is the audience for this article? I feel like it should probably be written for the developer who may have heard of ES6, but isn't clear that it's the same as ES2015; as in, people who don't follow these topics closely.

_Does_ it feel too long, or should I not worry about length?

lydell:
I don’t think it’s too long.

boris-petrov:
It's a 10-minute read for slow readers. It's perfect, no need to touch it. :)

jashkenas:
My thoughts:

"clean, elegant" is a little horn-tooting. How about just "clean".

All of the "ES2015+" and "ES2017" monikers, and the ECMA history lesson is confusing, and a little beside the point. Let's just call it "modern JavaScript", or "modern JS", and later just "JS", and axe the "Quick History Lesson".

`=>` was not adopted directly from CoffeeScript. It was in other languages earlier, and ES4 had it on the roadmap before CoffeeScript existed.

> "but simply put: let and const and named functions aren’t necessary for compatibility or interoperability with other libraries, and supporting them would add unwanted complexity to CoffeeScript"
It's not so much about "unwanted complexity", just that they don't fit in to CoffeeScript. We purposefully leave them out.

The Future Compatibility section feels a little overly Babel-specific. Don't Traceur and Bublé work too? And whatever transpiler becomes popular next year? Seems like this section could be shrunk a bit to just state our Stage 4 syntax support policy and mention the transpilers you can use for broader compatibility.

Finally, it would be nice to sneak in a small mention that CoffeeScript 1.X still does and will continue to exist. If you want to target lowest-common-denominator JS, which still exists under rocks and in dark corners (Illustrator, remote countries, embedded systems, grandparents), you still can.

GeoffreyBooth:
Agree with most of the above.

> Finally, it would be nice to sneak in a small mention that CoffeeScript 1.X still does and will continue to exist. If you want to target lowest-common-denominator JS, which still exists under rocks and in dark corners (Illustrator, remote countries, embedded systems, grandparents), you still can.

I would argue that if you want lowest-common-denominator JS, you should still use CS2 and just transpile your output. I don’t see any benefit to CS1 over CS2-plus-Babel, other than that compilation is faster. Though with CS1 and no transpilation, you need to be careful not to use modules or tagged template literals or generators, the “modern JS” features we added to CS1 without conversion down to ES3/ES5 compatibility. That’s quite a caveat to make sure to remember, though, so the docs should probably be conservative and recommend transpilation (i.e. a post-CoffeeScript compiler Babel step, or #4697) even for CS1, in which case CS1 offers nothing over CS2.

jashkenas:
Fair enough!

GeoffreyBooth:
@jashkenas I tried to revise per your notes. Not sure I got it quite right regarding `let`/`const`; maybe I’m better off just not giving a reason here and letting the docs speak for us, assuming you don’t have any notes to that section of the docs (?).

I hadn’t heard of Bublé, it looks cool. Traceur Compiler hasn’t been updated in a year, so I’m not sure how bright its future is, though it is supported by Google. One thing that I found interesting is that all three transpilers support JSX. This reinforces my feeling that we were right to add support for JSX into CoffeeScript.

jashkenas:
Re: `var/let/const` — probably better to omit the reason from this page ... because this version ain't quite right either.

While it's the case that everything in CoffeeScript is an expression, and we don't declare variables — that's not really the reason why we don't support `var/let/const`. We could have added different assignment operators for each, or allowed the standard prefix modifiers and just shifted them up to the declaration line at the top of the block.

GeoffreyBooth:
Okay, hopefully this should be it. Any last notes?

If this text is good I’ll convert this into an HTML file like our docs. I’ll just do it once semi-manually, I don’t think I want to create a new `cake` task for this that won’t ever get used again.

GeoffreyBooth:
Take a look: http://rawgit.com/GeoffreyBooth/coffeescript/announcing-2/docs/v2/announcing-coffeescript-2/

What do you all think?

documentation/sections/announcing_coffeescript_2.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Future Compatibility
+
+Back when CoffeeScript 1 was created, ES2015 JavaScript and transpilers like [Babel](http://babeljs.io/), [Bublé](https://buble.surge.sh/) or [Traceur Compiler](https://github.com/google/traceur-compiler) were several years away. The CoffeeScript compiler itself had to do what today’s transpilers do, converting modern features like destructuring and arrow functions into equivalent lowest-common-denominator JavaScript.
+
+But transpilers exist now, and they do their job well. With them around, there’s no need for the CoffeeScript compiler to duplicate this functionality. All the CoffeeScript compiler needs to worry about now is converting the CoffeeScript version of new syntax into the JS version of that syntax, e.g. `import fs from 'fs'` into `import fs from 'fs'`. This makes adding support for new JavaScript features much easier than before.
```
lydell: Is this a better example?

> e.g. `"Hello #{name}!"` into `` `Hello ${name}!` ``

I find the `import fs from 'fs'` example a little bit confusing since the CoffeeScript syntax is identical to the JavaScript syntax.

GeoffreyBooth: Yes, good call.

Issue #4696, opened 2017-09-10
Proposal: Offer Babel transpilation within the CoffeeScript compiler?


GeoffreyBooth:
So inside `Cakefile` we already have a function [`transpile`](https://github.com/jashkenas/coffeescript/blob/d172405244b7510f8cf4455ad4483accbac85d1d/Cakefile#L69) that takes JavaScript input (the output from the CoffeeScript compiler) and pipes it through Babel, to get a more compatible JavaScript output. We’re currently using `transpile` to generate the browser compiler `.js` file, and the code for the docs site.

What if we added a flag (shudder) that told the compiler to run its final output through `transpile` before returning or saving it? Then people could use the CoffeeScript 2 compiler without necessarily needing to wire together a build chain. We would also need to provide some way for the user to set Babel’s options, probably via [babelrc](https://babeljs.io/docs/usage/babelrc/).

Some concerns:

- We currently have no production dependencies, and we like it that way. But we could add Babel to the [`optionalDependencies` key in `package.json`](https://docs.npmjs.com/files/package.json#optionaldependencies), so that Babel wouldn’t get automatically installed; and if `compile` is run with this new flag, we could throw an error if Babel isn’t `require`-able.
- The current uses of `transpile` don’t bother with source maps, but we would want to make sure that source maps work with this. There’s [some great code in the Meteor CoffeeScript package](https://github.com/meteor/meteor/blob/devel/packages/non-core/coffeescript-compiler/coffeescript-compiler.js) that sends code through the CoffeeScript compiler, then through Babel, and merges the source maps from each compiler. We would probably want to adapt this.

The latter point is probably the strongest reason _to_ do this. I’m still waiting on some good examples from #4615, so I don’t know how good or bad the current situation is regarding source maps surviving double compilation (CoffeeScript then Babel); but if it’s not great, then maybe this _is_ something we should take on, as source map juggling is probably beyond the capabilities of most build chains or the developers assembling them.

lydell:
If all of these things are true ...

- Babel isn’t automatically installed
- `--transpile` (or `--babel` or whatever we end up) gives a really nice error message if Babel isn’t installed
- Complete Babel customization is available through `.babelrc` is available
- Source maps Just Work™

... I think this sounds like an awesome feature.

GeoffreyBooth:
[...]

PR #4697, opened 2017-09-11
Option to transpile with Babel

GeoffreyBooth:
See #4696. This PR adds a `--transpile` option that lets the user tell the CoffeeScript compiler to pipe its output through Babel, with source maps updated in the process.

I discovered that Babel has [an option called `inputSourceMap`](https://github.com/babel/babel/issues/827) that takes a v3 source map object (like that we output) and it _updates it_ as needed after Babel’s transpilation. So I can just use that, I think, rather than doing any fancy footwork myself [like the Meteor package is doing](https://github.com/meteor/meteor/commit/73ed3ff27a6ba686c06438ca1bb8e8d34d628bac). Babel also will apparently automatically update inline source maps it finds in the code it is sent as input, but it won’t do both (update the object and the comment). Since we always generate the object, whether we output it inline in a comment or saved as a `.map` file, I chose to just always use `inputSourceMap` and we can convert it to the inline comment like we already were.

The tricky part of this was finding the options to send to Babel. It has automatic functionality to look for a `.babelrc` file up the path of whatever file it’s given as input, but the CoffeeScript compiler will always be sending Babel a string of code, rather than a string file path. So we need to find the `.babelrc` file (or a `package.json` file with a `babel` key) ourselves, and parse it and pass the options to Babel. For the Node API this is much easier, as the user can just pass an object as the value of the `transpile` option (and must, since the Node API always takes a string as input).


GeoffreyBooth:
Okay @lydell I think this is ready for review. How to test the CLI (most common options):

```bash
➜  cd /tmp
➜  mkdir transpile-test
➜  cd transpile-test
➜  npm install https://github.com/GeoffreyBooth/coffeescript.git#transpile-with-babel
➜  npm install babel-core babel-preset-env
➜  echo "import path from 'path'\nconsole.log path.sep" > test.coffee
➜  echo '{ "presets": ["env"] }' > .babelrc
➜  coffee --transpile --bare --no-header --print ./test.coffee
'use strict';

var _path = require('path');

var _path2 = _interopRequireDefault(_path);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

console.log(_path2.default.sep);
➜  coffee --transpile ./test.coffee
/
➜  coffee --transpile --eval "import path from 'path'; console.log path.sep"
/
➜  coffee --transpile --bare --no-header --compile --map --inline-map ./test.coffee
➜  cat ./test.js.map
{
  "version": 3,
  "file": "test.js",
  "sourceRoot": "",
  "sources": [
    "test.coffee"
  ],
  "names": [],
  "mappings": ";;AAAA,AAAO,AAAP;;;;;;AACA,AAAO,QAAC,AAAR,IAAY,AAAI,eAAC,AAAjB",
  "sourcesContent": [
    "import path from 'path'\nconsole.log path.sep\n"
  ]
}
➜  ./cat test.js
'use strict';

var _path = require('path');

var _path2 = _interopRequireDefault(_path);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

console.log(_path2.default.sep);
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoidGVzdC5qcyIsInNvdXJjZVJvb3QiOiIiLCJzb3VyY2VzIjpbInRlc3QuY29mZmVlIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiI7O0FBQUEsQUFBTyxBQUFQOzs7Ozs7QUFDQSxBQUFPLFFBQUMsQUFBUixJQUFZLEFBQUksZUFBQyxBQUFqQiIsInNvdXJjZXNDb250ZW50IjpbImltcG9ydCBwYXRoIGZyb20gJ3BhdGgnXG5jb25zb2xlLmxvZyBwYXRoLnNlcFxuIl19
//# sourceURL=/private/tmp/test.coffee
//# sourceMappingURL=test.js.map
```

The key here is that I created a `.babelrc` file. You can also add a `babel` key to your `package.json`:
```bash
➜  rm .babelrc
➜  coffee --transpile ./test.coffee
To use the transpile option, there must be a .babelrc file
(or a package.json file with a "babel" key) in the path of the file
to be compiled, or in the path of the current working directory.
If you are compiling a string via the Node API, the transpile option
must be an object with the options to pass to Babel.
See http://coffeescript.org/#usage
➜  echo '{ "name": "test", "version": "0.0.1", "babel": { "presets": ["env"] } }' > package.json
➜  coffee --transpile ./test.coffee
/
```

I added tests for the various ways to use `transpile` from the Node API. Are there any additional tests that this needs?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["env"]
+}
```
lydell: When is this file used? Some cake task?

src/command.coffee
```diff
@@ -441,14 +442,77 @@ parseOptions = ->
...
+      else
+        cantFindOptions()
+
+      loop
+        try
+          opts.transpile = JSON.parse fs.readFileSync path.join(checkPath, '.babelrc'), 'utf-8'
+          break
+        catch
+          try
+            opts.transpile = JSON.parse(fs.readFileSync(path.join(checkPath, 'package.json'), 'utf-8')).babel
```
lydell: If we find a package.json _without_ the `babel` key, or with `"babel: null"`, then the search will stop and `opts.transpile` will be `undefined`/`null` and `opts.transpile.transpile` a few lines down will crash. I suggest we only stop the search if `packageJson.babel` is truthy.

src/command.coffee
```diff
@@ -48,7 +48,8 @@ SWITCHES = [
   ['-r', '--require [MODULE*]', 'require the given module before eval or REPL']
   ['-s', '--stdio',             'listen for and compile scripts over stdio']
   ['-l', '--literate',          'treat stdio as literate style coffeescript']
-  ['-t', '--tokens',            'print out the tokens that the lexer/rewriter produce']
+  ['-t', '--transpile',         'pipe generated JavaScript through Babel']
```
lydell: We need to add this to the docs as well, right? Or is docs stuff done some other time?

Do we need to document the "breaking change" that `-t` now means `--transpile` instead of `--tokens`?

.babelrc
GeoffreyBooth: It’s used only if you want to manually test `coffee --transpile`. So if you have the repo checked out and you want to run `coffee --transpile --eval "import path from 'path'; console.log path.sep"` and have it just work, without needing to create `.babelrc` first. I could just as easily put this into a `package.json` `babel` key, or leave it out. It’s not used by any `cake` tasks or any other part of the normal workflow. I just thought it would make development more convenient.

src/command.coffee
GeoffreyBooth: I’m working on the docs, I think I’ll add the docs changes to this PR once I finish getting your feedback (like if you thought I should use `--babel` or something).

I don’t know, do you think anyone cares that I’m reassigning `-t`? Is that really a breaking change? Perhaps it’s small enough that I mention it only in the changelog?

src/command.coffee
GeoffreyBooth: Okay, how’s the latest commit?

GeoffreyBooth:
Thanks for checking the source maps. I was struggling to figure out how to write a test around those, without adding `source-map` as a dependency (which I guess wouldn’t be so bad, as it would only be a dev dependency). Is the current “merged source map” test robust enough, do you think? I guess all it really needs to do is test that the source map isn’t the same as CoffeeScript’s normal one, in which case Babel did _something,_ and if that something wasn’t correct it’s a bug on Babel’s part. I’m going at least a little beyond that though.

.babelrc
lydell: I see, good idea!

src/command.coffee
lydell: Let’s just mention it in the changelog.

lydell:
I think the source map test is very clever and should be enough. Source maps are hard to test anyway. Not sure using the source-map package would help.

GeoffreyBooth:
Updated docs: http://rawgit.com/GeoffreyBooth/coffeescript/transpile-with-babel/docs/v2/

GeoffreyBooth:
Also I think we’re okay to go straight to 2.0.0 with this? Skipping one last beta.

lydell:
Yeah, I don’t think a beta is required for this feature.

GeoffreyBooth:
Do we want to provide a default configuration somehow? Like if they don’t specify a Babel config, it checks that `babel-preset-env` is installed and loads `{ "presets": ["env"] }`. Or is that too prescriptive?

I’m just looking at http://coffeescript.org/v2/#transpilation and thinking it looks like a lot of steps.

lydell:
That sounds too magic to me. Besides, there's just three steps:

```
npm install --save-dev babel-core babel-preset-env
echo '{ "presets": ["env"] }' > .babelrc
coffee --transpile ...
```

What you could to is reorganizing the transpilation in the docs. I'll come back with what I mean soon.

lydell:
My idea is to start the section in a TL;DR way so you can quickly find how to get something going, and then go into more details and explanations. For example, we could start the section something like this:

---

CoffeeScript 2 generates JavaScript that uses the latest, modern syntax. Your runtime might not support all of that syntax. If so, you need to _transpile_ the JavaScript. There are several tools that can do that. But to make things a little easier, CoffeeScript has built in support for the popular [Babel](http://babeljs.io/) transpiler. This is how you use it:

```bash
npm install --save-dev babel-core babel-preset-env
echo '{ "presets": ["env"] }' > .babelrc
coffee --transpile --compile --map some-file.coffee
```

In other words, if you have installed and configured Babel, the `--transpile` flag will tell `coffee` to use Babel on the generated JavaScript.

----

... and then glue the above into the rest of what you've written (which is great, btw).

GeoffreyBooth:
Yeah, I like this. It's the “Quickstart” approach, where you drop some terse commands or steps first and then follow with the same instructions with explanation. Transpilation is important enough, and difficult enough, to be worth the detail I think.

src/command.coffee
```diff
@@ -441,14 +442,79 @@ parseOptions = ->
...
+      require 'babel-core'
+    catch
+      console.error '''
+        To use --transpile, you must have Babel installed and configured.
+        See http://coffeescript.org/#usage
+      '''
+      process.exit 1
+
+    # We’re giving Babel only a string, not a filename or path to a file, so
+    # it doesn’t know where to search to find a `.babelrc` file or a `babel`
```
loganfsmyth: Just came across this PR. Could you clarify this? Why can't you pass Babel the filepath and let it resolve the config on its own?

In Babel 7 we're adding more ways to configure Babel, some of which are programmatic with an API managed by `babel-core` so doing stuff like this is going to get harder. Even for config files we're adding caching logic for performance, so by doing this all of that gets sidestepped which will hurt performance. I'd like to understand what limitations exist that made this necessary, so we can fix them if possible.

package.json
```diff
@@ -51,5 +51,8 @@
     "underscore": "~1.8.3",
     "webpack": "~3.5.6"
   },
+  "optionalDependencies": {
+    "babel-core": "^6"
```
loganfsmyth: This seems like a scary thing to add. Won't this mean that every person using CoffeeScript will install Babel automatically, even if they aren't using it? It also limits people to only using Babel 6.x for the duration of CoffeeScript 2.x's lifetime, but we're getting pretty close to releasing 7.x and don't plan to really support 6.x much once that has happened.

loganfsmyth: `npm` automatically installs optional dependencies unless you explicitly install with `--no-optional`, it just doesn't error if they fail to install for some reason, so this is essentially the same as having a full dependency on `babel-core` since I can't think of a time `babel-core` would fail to install

GeoffreyBooth:
@loganfsmyth This feature is only a week old, so if there’s a better way you’d recommend users get config info into Babel I can still change it.

The reason we went down this road, of the CoffeeScript compiler searching for a `.babelrc` file or `package.json` with a `babel` key, is that I didn’t want the compiler to write an intermediate file and then tell Babel to transpile that; I wanted to take the JS output by the CoffeeScript compiler and send it as a string to Babel, and only write to disk what Babel returned. Since Babel wasn’t opening a file, it had no point of reference from which to search for a `.babelrc` file or similar; so we needed to pass in the config. This meant us replicating Babel’s functionality of searching for the `.babelrc` file or `package.json` `babel` key. We deliberately copied how Babel expects to do it, so that the systems would be interchangeable (i.e. if someone wants to transpile via Gulp or Webpack etc. instead of using `--transpile`, they’re already good to go). Is Babel 7 changing anything about where it expects to find options, e.g. `.babelrc` file or `package.json` `babel` key?

The reason not to send Babel a file path is because the compiler might be compiling a whole folder (or folder tree)’s worth of files, and calling Babel for each compiled file, and I didn’t want Babel to have to go read the options file every time. It’s more efficient if the CoffeeScript compiler reads the options into memory once, and just passes them to Babel over and over for each file.

Assuming Babel 7 doesn’t have any API changes that will break this API, I’m happy to change the version constraint to `>=6` or whatever you’d recommend. And yes, I’m okay with Babel being automatically installed for most users. Most users will and should be transpiling CoffeeScript 2’s output. The only situation where you shouldn’t is if you’re running the output under Node 8 and you’re careful not to use modules or JSX. If you’re in that minority, or you want to use a different transpiler other than Babel (and you care if Babel appears in your `node_modules`) then yeah, they can do some research and install with `--no-optional`. But since most people will probably have Babel installed anyway, I don’t think it’s too much of a concern.

And one final note: this is really aimed as a convenience for users running code via the `coffee` command. That’s a larger community than I would’ve thought, as dev ops shell-scripting people apparently love using CoffeeScript as an alternative to Bash `.sh` files (the two syntaxes do look a lot alike) and others just like the simplicity of not having to deal with sophisticated build chains. But most serious users I would expect are using something like Webpack or Gulp or Browserity with Coffeeify, and so they’ll chain Babel with that rather than using this option. Or if they do use this option with one of those tools, it will be via the Node API where the Babel options are required to be passed in as an object, so the “where to find the file” issue isn’t a concern.

loganfsmyth:
> The reason we went down this road, of the CoffeeScript compiler searching for a .babelrc file or package.json with a babel key, is that I didn’t want the compiler to write an intermediate file and then tell Babel to transpile that; I wanted to take the JS output by the CoffeeScript compiler and send it as a string to Babel, and only write to disk what Babel returned.

You can set `.filename` in options object for `babel.transform` and it'll resolve the `.babelrc` automatically. I agree that temporary files would be awful :)

> Is Babel 7 changing anything about where it expects to find options, e.g. .babelrc file or package.json babel key?

Babel 7.x includes a `.babelrc.js` support for arbitrarily generating a config based on environmental variables, and exposes a programmatic caching API so that the config is only re-evaluated if the env vars change, and we're planning to encourage people to move to that rather than the current `env` block we have now that confuses everyone.

In both 6.x and 7.x, `.babelrc` files can also include `extends` blocks that reference other files. With this logic, you've loaded that and passed it programmatically, so those paths will resolve relative to the wrong location.

> The reason not to send Babel a file path is because the compiler might be compiling a whole folder (or folder tree)’s worth of files, and calling Babel for each compiled file, and I didn’t want Babel to have to go read the options file every time.

Babel includes caching for all of the config lookups because of this exact concern, and in 7.x it's even better. Babel's own compilation steps are implemented as one-off calls to `babel-core` and at least since the caching was added long ago, I don't think we've ever heard complaints about it. I'd be extremely surprised if it was even a reasonable fraction of the total time Babel spends doing the actual transformation of files.

> Assuming Babel 7 doesn’t have any API changes that will break this API, I’m happy to change the version constraint to >=6 or whatever you’d recommend.

The problem is that the user needs to know. None of Babel's core transforms are compatible across versions. The expectation is that users of Babel 7.x will use the `7.x` versions of Babel's plugins and presets. We've done our best not to _break_ APIs, so chances are things could work across versions fine, but all the plugins written for 7.x are probably going to have assertions that they are being processed with the right version of `babel-core`. If coffeescript included `>=7`, that means that when we release Babel 7.x, and people have established configs written with Babel 6.x in mind, it would break those users.

> And one final note: this is really aimed as a convenience for users running code via the coffee command.

I totally respect that goal, but I think if you want to achieve that, you'd have to fully bundle Babel, ditch resolution of config files entirely, and expose your own babel-specific config options from Coffeescript itself. The users need to know what version their plugins work on, and need to decide which version to use.

GeoffreyBooth:
So what do you recommend? I pass in to Babel’s `filename` the path and filename of the original source `.coffee` file, and it uses that to search for its options using its usual logic? And then I can remove all this code of the CoffeeScript compiler looking for the options. The instructions in the docs actually wouldn’t need to change, since the same rules would apply (?) for users to define how to configure Babel.

And this only applies to `--transpile` from the CLI. Running CoffeeScript from Node doesn’t do any file system searching, it just requires the options get passed in as an object, which is forwarded on to Babel.

loganfsmyth:
Yep! As far as I'm aware passing `filename` should work nicely.

GeoffreyBooth:
Okay, and it won’t matter that it’s the filename of the `.coffee` file? Does Babel use `filename` for anything other than options lookup, and I’m assuming it also uses it to populate the source map object?

And does any of this change in 7? Remember that I’m relying completely on the user to define their own options, so it’s on them to craft an appropriate `.babelrc` or `.babelrc.js` file that works for them and their installed version of Babel.

loganfsmyth:
> Okay, and it won’t matter that it’s the filename of the .coffee file?

`.filename` is used for config resolution, source maps, error messages, and it is exposed to users in plugins. None of that cares about file extensions though, so you should be fine. We've got enough random file extensions going through Babel already with `.es6`,`.jsx`, and `.mjs` and such. One more won't hurt :)

> And does any of this change in 7?

Nothing in the programmatic interface for Babel. If you wanted to limit things to 6.x until 7.x is out and official and tested, we export `.version` from `babel-core`, so you could add an assertion there for now and remove it later as a non-breaking change. That said, I don't expect anything breaking to change on the public interface of `babel-core`.

lydell:
@GeoffreyBooth

> And yes, I’m okay with Babel being automatically installed for most users

I thought we all agreed that we _don't_ want Babel to be automatically installed?

GeoffreyBooth:
> I thought we all agreed that we _don’t_ want Babel to be automatically installed?

Yes, that would be ideal. That’s what I thought `optionalDependencies` did, but clearly I was wrong. See https://github.com/jashkenas/coffeescript/issues/4703, let’s continue discussing this over there.

GeoffreyBooth:
One last question @loganfsmyth: Is the `filename` setting documented anywhere as the starting point for Babel to search for its options? In https://babeljs.io/docs/usage/api/#options all it says about `filename` is that it’s used for error reporting. I don’t want to rely on this setting and then someday Babel stops using it for compilations of strings, since perhaps someone else on the Babel team might not consider that a breaking change since the current behavior is undocumented.

loganfsmyth:
Nah it's definitely used for more than that, guess that's just our docs lacking detail. It'd break all kinds of things if that ever changed without a new major version, and that won't be changed in 7.x.

PR #4709, opened 2017-09-19
Add note about approximating 'let'

jez:
I found [this discussion][1] talking about `let` and `const` to be
helpful to understand the decisions why these keywords were not added to
CoffeeScript v2.

In particular, I think we can improve the documentation by linking to
these examples on the section about why they're unsupported. Basically,
it comes down to using `do ->` to approximate `let` and `const`.

I'm down to re-word anything here, and I also completely understand if
this is not something you want in the documentation.

[1]: https://github.com/coffeescript6/discuss/issues/1

### DIFF
--- a/documentation/sections/unsupported_let_const.md
+++ b/documentation/sections/unsupported_let_const.md
@@ -4,6 +4,30 @@ When CoffeeScript was designed, `var` was [intentionally omitted](https://github
 
 `let` and `const` add a useful ability to JavaScript in that you can use them to declare variables within a _block_ scope, for example within an `if` statement body or a `for` loop body, whereas `var` always declares variables in the scope of an entire function. When CoffeeScript 2 was designed, there was much discussion of whether this functionality was useful enough to outweigh the simplicity offered by never needing to consider variable declaration in CoffeeScript. In the end, it was decided that the simplicity was more valued. In CoffeeScript there remains only one type of variable.
 
+If you really think that what you need is behavior like `let`, you can use the `do ->` idiom, which creates and immediately runs a function:
+
+```coffee
+if foo then do ->
+  x = 42
+  console.log x       # => 42
+
+console.log x         # => ReferenceError
+```
+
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
The issue with this is that such a detailed discussion of workarounds for the lack of `let` makes it sound like there’s some deficiency in the language, like “you can’t get block-scoped variables, but you need that functionality, so here’s the closest approximation of it.” The part about using function arguments to shadow outer variables is especially hacky; that’s basically taking advantage of a loophole in the compiler.

In the discussions around the design of CoffeeScript 2, it was seriously considered (https://github.com/coffeescript6/discuss/issues/62, https://github.com/coffeescript6/discuss/issues/58) that we would make a breaking change to make all variables block scoped, instead of always function scoped the way they are now. Ultimately we decided not to go that way because it would be a significant breaking change, that would prevent a huge number of projects from being able to upgrade; but I think it’s still on the table for a potential CoffeeScript 3. Ultimately I wanted CoffeeScript 2 to be a bridge to get all projects into the modern JS era, and to not fragment the community the way Python is split between 2 and 3. And we may still make the same decision about scoping when 3 comes along, as it might still be too drastic of a breaking change. The other option discussed was an operator where you could define whether a variable should be function-scoped or block-scoped (https://github.com/coffeescript6/discuss/issues/58), but that was rejected as then you lose the simplicity of not needing to think about variable declaration (all the more valued now that JavaScript has three ways to declare variables, between `var` and `const` and `let`, not to mention globals or function arguments etc.).

I think ultimately if you’re needing to use `do ->` to create little pockets of scope, and especially if you’re using `do (outerVariable) ->` to shadow variables, you probably have functions that are way too long. That’s not a polite answer, which is why we don’t say that in the docs, but ultimately CoffeeScript is very much an opinionated language (this is one of its opinions) and the opinions make assumptions about how code should be written; see also the `get` and `set` keywords. Coffeelint looks like it’s getting updated for CS2, and there’s https://github.com/pwlmaciejewski/coffeelint-variable-scope that’s supposed to help catch errors involving variables across scopes, so maybe that’s another way to go.

PR #4712, opened 2017-09-20
Fix #4464: backticked expressions in class body

### LINKED ISSUE #4464: Lexer: Backticks should allow adding code to class body

connec:
Reopened from #4668...


GeoffreyBooth:
@connec Any idea why the build failed? Something to do with force-pushing?

@lydell do you have any notes on this?

lydell:
LGTM

Issue #4724, opened 2017-09-29
Nodes: Class body passthrough expressions get hoisted if anything in a class is hoisted


stevefan1999:
http://tinyurl.com/yb476vtn
Although it was still a legit prototype construct, this is very problematic for backtick embeded JS expression.
http://tinyurl.com/ya7xo5fz
The entire `qux = 10` was shifted outside of the class declaration.
I do believe this is a compiler bug.

Workaround: 
http://tinyurl.com/y7gzwfup for backtick property-based apply/or
http://tinyurl.com/y8xa58a6 for constructor apply

*PS: backtick property-based apply will still be boil down to constructor apply, tested with babel online repl: http://tinyurl.com/yd94hv4y

stevefan1999:
I thought MobX was cool with CS2 with relaxation on backtick JS expression generation, turns out I was wrong,

GeoffreyBooth:
There’s a lot going on in your post, so I’m not sure exactly what bug or bugs you’re referring to. But I did find at least one issue in your examples, that I’ve simplified to the [minimal case](http://coffeescript.org/#try:class%20Class%0A%20%20%60field%20%3D%203%60%0A%20%20hoisted%3A%204):

```coffee
class Class
  `field = 3`
  hoisted: 4
```
```js
var Class;

Class = (function() {
  class Class {
    ;

  };

  field = 3;

  Class.prototype.hoisted = 4;

  return Class;

})();
```

It appears that whenever anything in a class is hoisted, all passthrough expressions get improperly hoisted as well. If you change `hoisted: 4` above to `hoisted: -> 4` instead, the backticked block appears in the class body as desired.

connec:
I will look into this asap.

GeoffreyBooth:
@connec I noticed that I still had my original “backticked expressions in class body” branch (the one you refactored for #4668 / #4712) on my local repo. Out of curiosity, I checked out my old branch, where I left off before you started working in it, and I tried the minimal case above. To my surprise, my old branch compiles it correctly:

```js
var Class;

Class = (function() {
  class Class {
    field = 3;
  };

  Class.prototype.hoisted = 4;

  return Class;

})();
```

I pushed the branch up: https://github.com/GeoffreyBooth/coffeescript/tree/backticked-expressions-in-class-body and added a test for this case.

[...]

PR #4736, opened 2017-10-06
Fix #4706: Flow generics

### LINKED ISSUE #4706: Flow generics in comments
Flow supports [generic types](https://flow.org/en/docs/types/generics/) using the comment syntax in a couple key locations:

```js
/* @flow */

function identity/*::<T>*/(value /*: T */)/*: T */ {
  return value;
}

class Container/*::<T>*/ {
  method/*::<U>*/() {
    return true;
  }
}

let arr = [1, 2, 3];
let copy = arr.map(/*:: <T> */(item /*: T */) /*: T */ => item);
```

But the following roughly equivalent CoffeeScript code:

```coffee
# @flow

identity = ###::<T>### (value ###: T ###) ###: T ### ->
  value

class Container ###::<T> ###
  method: ###::<U> ### () -> true
  
arr = [1, 2, 3];
copy = arr.map(###:: <T> ###(item ###: T ###) ###: T ### => item)
```

Which compiles to:

```js
// @flow
var Container, arr, copy, identity, val;

identity = function(/*::<T>*/value/*: T */)/*: T */ {
  return value;
};

Container/*::<T> */ = class Container {
  method() {
    return true;
  }

};

arr = [1, 2, 3];

copy = arr.map((/*:: <T> */item/*: T */)/*: T */ => {
  return item;
});

```

If CoffeeScript put the comments in the correct location for Flow generics, it would cover the entire Flow grammar 1-to-1

```diff
  // @flow
  var Container, arr, copy, identity, val;

- identity = function(/*::<T>*/value/*: T */)/*: T */ {
+ identity = function/*::<T>*/(value/*: T */)/*: T */ {
    return value;
  };

- Container/*::<T> */ = class Container {
+ Container = class Container /*::<T> */ {
    method() {
      return true;
    }

  };

  arr = [1, 2, 3];

- copy = arr.map((/*:: <T> */item/*: T */)/*: T */ => {
+ copy = arr.map(/*:: <T> */(item/*: T */)/*: T */ => {
    return item;
  });
```

GeoffreyBooth:
Fixes #4706. Placement of comments around function parameter parentheses and around class names is improved; top-level identifiers with trailing comments are wrapped in parentheses, as required by Flow to distinguish them from JavaScript labels.

@thejameskyle do you mind reviewing this? And especially testing it on a few non-trivial cases that you may have in your code. I think I covered all the cases you mentioned in #4706, except for the case of identifiers in an assignment wrapped in parentheses: `other = (arr ###: any ###)`. I think `other = arr: any` is always a syntax error, never a JavaScript label, so does it need parentheses? But all the other cases are covered; see `test/comments.coffee` in this PR.

**EDIT**: Covered that last case.


thejameskyle:
This looks awesome :)

> I think `other = arr: any` is always a syntax error, never a JavaScript label, so does it need parentheses? 

I'm afraid Flow will only parse it as `other = (arr: any)`, the parenthesis are always required.

GeoffreyBooth:
Okay, while I ponder that do you mind testing the other cases?

GeoffreyBooth:
Would it be problematic if an identifier followed by a block comment: `str ###:: String ###` is always output wrapped in parentheses: `(str /*:: String */)`, no matter where it appears? That would simplify things, but it would mean extra parentheses in unexpected places like function parameter lists.

GeoffreyBooth:
Update: I figured out how to handle the `other = arr/*: any */` case. I think they’re all covered now.

stevefan1999:
The use of Flow with CS2 is truly horrific. Nobody wants to see `###` all around. Instead Facebook presumed developers to miss `/*` and `*/` magically...

thejameskyle:
@stevefan1999 Sorry all these people are creating free stuff for you and expecting nothing in return.

Sorry people like @GeoffreyBooth and myself spend thousands of hours writing code, answering issues, writing documentation, speaking at conferences. Spending hundreds of thousands of dollars of our time and creating projects worth millions of dollars for which you pay nothing because we think it's the right thing to do. I can't imagine the pain you must be going through.

I have an idea for you: Don't use it.

@GeoffreyBooth Awesome, it's late now, but I'll take a look at this soon and try a bunch of stuff locally and report back.

stevefan1999:
@thejameskyle I'm not trying to taunt @GeoffreyBooth but I would ditch Flow if I'm ever going to see a lot of `###`s, what I'd call this is the "ultimate pure anti-pattern".

On the other hand, I'm sorry if I offended both you @thejameskyle @GeoffreyBooth

src/nodes.coffee
```diff
@@ -502,8 +502,6 @@ exports.Block = class Block extends Base
         # We want to compile this and ignore the result.
         node.compileToFragments o
         continue
-
-      node = node.unwrapAll()
```
GeoffreyBooth: @connec Do you think there’s any concern about me removing this line? I removed it in order to not necessarily always optimize away redundant parentheses (which is what we’re dealing with in this case).

GeoffreyBooth:
@thejameskyle Thanks. I’m thinking it would be good to have a CoffeeScript file that included every part of the Flow comments-based syntax. Then I could test that the entire Flow syntax works by compiling that file to JS, then having Flow check it. This might become part of `cake test:integrations`.

But anyway that can be a separate pull request. If you can confirm that this PR works, I’ll merge it in, and if you have the time to generate such an example file, I’d love to include it in the project.

src/nodes.coffee
connec: @GeoffreyBooth why would we not want to optimize away redundant parentheses? Otherwise nothing jumps out at me!

GeoffreyBooth: Because of this particular case. When parentheses wrap an `IdentifierLiteral` that has a comment, we want those parentheses to be output because Flow requires them. See https://github.com/jashkenas/coffeescript/issues/4706#issuecomment-330533014.

I’m assuming that removing this optimization doesn’t _break_ anything else, though, right?

GeoffreyBooth:
@thejameskyle Do you think you’ll have time to check this soon?

GeoffreyBooth:
@thejameskyle or @connec, any notes on this? This is the last thing before we release 2.0.2.

thejameskyle:
Hey sorry, been out of commission the last two days. I was looking for the parser tests in Flow but couldn't find them. I'll check this out now.

thejameskyle:
```js
test "#4706: Flow comments before class methods", ->
  eqJS '''
  class Container
    ###::
    method: (number) => string;
    method: (string) => number;
    ###
    method: -> true
  ''', '''
  var Container;

  Container = class Container {
    /*::
    method: (number) => string;
    method: (string) => number;
    */
    method() {
      return true;
    }

  };'''

test "#4706: Flow comments for class method params", ->
  eqJS '''
  class Container
    method: (param ###: string ###) -> true
  ''', '''
  var Container;

  Container = class Container {
    method(param/*: string */) {
      return true;
    }

  };'''


test "#4706: Flow comments for class method returns", ->
  eqJS '''
  class Container
    method: () ###: string ### -> true
  ''', '''
  var Container;

  Container = class Container {
    method()/*: string */ {
      return true;
    }

  };'''
```

Here's some more tests cases for class methods

thejameskyle:
There's also the question for how to handle class properties:

```coffee
class Foo
  prop: 1;
```

Get's compiled to:

```js
class Foo {}
Foo.prototype.prop = 1;
```

But for that to work in Flow it needs a type annotation:

```js
class Foo {
  /*:: prop: number */
}
Foo.prototype.prop = 1;
```

The easiest way to do that would be to write it like this:

```coffee
class Foo
  ###:: prop: number ###
  prop: 1;
```

But the output is currently:

```js
class Foo {};

/*:: prop: number */
Foo.prototype.prop = 1;
```

thejameskyle:
Another test case:

```js
test "#4706: Flow comments for function spread", ->
  eqJS '''
  method = (...rest ###: Array<string> ###) =>
  ''', '''
  var method;

  method = (...rest/*: Array<string> */) => {};'''
```

GeoffreyBooth:
Class properties are still an experimental syntax, so they're only supported via backticks currently. #4724. So within that backticked block you could add the comment. 

So aside from that, I should add these tests you're proposing, which I suppose all pass? Or you're welcome to, I've given you access. 

And then the only remaining thing for Flow is variable declarations? I'd rather tackle that in its own PR. Is everything in this one okay?

GeoffreyBooth:
@thejameskyle I’ve added the tests you posted above. See my last comment for class properties, and I’ve opened #4747 to track fixing comments to support Flow for local variable declarations.

Anything else for this PR?

GeoffreyBooth:
@thejameskyle Please see also #4753. Once these two PRs are merged in I’ll cut a new release. Please try them on your project and let me know how it goes?

PR #4737, opened 2017-10-07
Fix #4724

### LINKED ISSUE #4724: Nodes: Class body passthrough expressions get hoisted if anything in a class is hoisted

connec:
The handling of hoisted nodes in class bodies was incorrect, as the node was being unwrapped *before* checking if it was hoisted, meaning nodes that should have been hoisted would be output normally.

This affected `PassthroughLiteral`s as they were wrapped in a `Value`.

Fixes #4724.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -701,6 +701,12 @@
         ref1 = this.expressions;
         for (index = j = 0, len1 = ref1.length; j < len1; index = ++j) {
           node = ref1[index];
+          if (node.hoisted) {
+            // This is a hoisted expression.
+            // We want to compile this and ignore the result.
+            node.compileToFragments(o);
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Excellent, thanks!

PR #4744, opened 2017-10-12
Fix import/export list bug with aliased keywords

### LINKED ISSUE #4742: Lexer: Importing some reserved keywords fails
I'm using Ember.js and since version 2.16 they have a new modules API. There are a bunch of cases where I have to import a CoffeeScript-reserved keyword: `Ember.computed.and` becomes `import { and } from '@ember/object/computed'` for example. There are also cases for `or`, `not`, `on`.

I saw a [pull-request](https://github.com/jashkenas/coffeescript/pull/4347) that is merged and should allow this but this doesn't seem to work in my case. CoffeeScript version 2.0.0. I'm trying the following: `import { and as computedAnd } from '@ember/object/computed'` but it complains about an `unexpected and`. Am I doing something wrong, is this a regression, or what?

GabrielRatener:
Fixes #4742

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -111,7 +111,7 @@
     // referenced as property names here, so you can still do `jQuery.is()` even
     // though `is` means `===` otherwise.
     identifierToken() {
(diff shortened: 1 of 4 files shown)


GeoffreyBooth:
Nice work! Thanks for taking this on.

PR #4750, opened 2017-10-16
Fix missing transpile options

Fudoshiki:
Fixes https://github.com/webpack/webpack/issues/5831


GeoffreyBooth:
I don't understand what you're doing here. `transpile` should not be `filename: filename` by default. We don't want a default, because we don't want to always transpile.

PR #4754, opened 2017-10-20
Fix #4752: Error on calling super with @params in a derived class constructor

### LINKED ISSUE #4752: Invalid JS: super in constructor shouldn’t accept @params
This code:
```coffeescript
class A
  constructor: (@name) ->
    alert @name, 'name'
    
class B extends A
  constructor: (@name) ->
    super @name
```
produce wrong code (See class b#super):
```javascript
var A, B, b;

A = class A {
  constructor(name) {
    this.name = name;
    alert(this.name, 'name');
  }

};

B = class B extends A {
  constructor(name) {
    super(this.name);
    this.name = name;
  }

};
```
where `this is not defined`
If I replace `@name` with `name`, like here:
```coffeescript
class A
  constructor: (@name) ->
    alert @name, 'name'
    
class B extends A
  constructor: (@name) ->
    super name
```
will produce:
```javascript
var A, B, b;

A = class A {
  constructor(name1) {
    this.name = name1;
    alert(this.name, 'name');
  }

};

B = class B extends A {
  constructor(name1) {
    super(name);
    this.name = name1;
  }

};
```

GeoffreyBooth:
Fixes #4752, in the sense that now we throw a compiler error rather than outputting invalid JavaScript. If we want to go further and create some placeholder variables to engage in sleight of hand to make the impossible possible, well, that can be a follow-up PR.

```coffee
class A
  constructor: (@name) ->
    alert @name, 'name'
    
class B extends A
  constructor: (@name) ->
    super @name
```
```
test.coffee:7:5: error: Can't call super with @params in derived class constructors
    super @name
    ^^^^^^^^^^^
```

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -4078,17 +4078,22 @@
         return seenSuper;
       }
 
-      // Find all super calls in the given context node
-      // Returns `true` if `iterator` is called
+      // Find all super calls in the given context node;
+      // returns `true` if `iterator` is called.
       eachSuperCall(context, iterator) {
         var seenSuper;
         seenSuper = false;
         context.traverseChildren(true, (child) => {
           if (child instanceof SuperCall) {
-            if (!child.variable.accessor && child.args.some(function(arg) {
-              return arg.this;
-            })) {
-              child.error("Can't call super with @params in derived class constructors");
+            // `super` in a constructor (the only `super` without an accessor)
+            // cannot be given an argument with a reference to `this`, as that would
+            // be referencing `this` before calling `super`.
+            if (!child.variable.accessor) {
+              Block.wrap(child.args).traverseChildren(true, (node) => {
+                if (node.this) {
+                  return node.error("Can't call super with @params in derived class constructors");
+                }
+              });
             }
             seenSuper = true;
             iterator(child);
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -3908,7 +3908,9 @@
           exprs.unshift(new Assign(new Value(new Arr([
             new Splat(new IdentifierLiteral(splatParamName)),
             ...((function() {
-              var k, len2, results;
+              var k,
+            len2,
+            results;
               results = [];
               for (k = 0, len2 = paramsAfterSplat.length; k < len2; k++) {
                 param = paramsAfterSplat[k];
@@ -4083,6 +4085,11 @@
         seenSuper = false;
         context.traverseChildren(true, (child) => {
           if (child instanceof SuperCall) {
+            if (!child.variable.accessor && child.args.some(function(arg) {
+              return arg.this;
+            })) {
+              child.error("Can't call super with @params in derived class constructors");
+            }
             seenSuper = true;
             iterator(child);
           } else if (child instanceof ThisLiteral && this.ctor === 'derived' && !seenSuper) {
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2783,15 +2783,19 @@ exports.Code = class Code extends Base
 
     seenSuper
 
-  # Find all super calls in the given context node
-  # Returns `true` if `iterator` is called
+  # Find all super calls in the given context node;
+  # returns `true` if `iterator` is called.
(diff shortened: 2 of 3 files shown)


GeoffreyBooth:
@connec does this look okay to you?

GeoffreyBooth:
Anyone familiar with the codebase available to review this? @vendethiel? @xixixao?

This one's only two lines plus a test.

vendethiel:
This won't catch code like `class extends A then constructor: (@a) -> super((=> @a)())`

GeoffreyBooth:
[It looks like `super(=> a)` is valid](http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEBC0FMAeAXBA7AJjAgtA3gFDTTAD26EKATgK7ApnUAUY1A5gJQHEnQS0ADghbNuAXgB80Nl14BfQvKA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=&targets=&version=6.26.0), though `class extends A then constructor: (@a) -> super((=> @a)())` shouldn’t be valid for the same reason that `class extends A then constructor: (@a) -> super(@a)` shouldn’t be. Yikes this is tricky . . . I guess to do it right we need to traverse down the tree of everything passed to `super` as arguments, and check them all for `this` references. Maybe it’s easier to just do the shenanigans that @connec was thinking about, although those will surely have their own difficult cases.

vendethiel:
> It looks like super(=> a) is valid

I meant specifically using `=> @a`. It doesn't error at compile time, but I can't seem to get the try it online to evaluate code.

GeoffreyBooth:
> I meant specifically using `=> @a`. It doesn’t error at compile time, but I can’t seem to get the try it online to evaluate code.

That’s the issue: we’re generating valid JavaScript, that just happens not to correspond to the user’s intent. [See](http://coffeescript.org/#try:class%20extends%20parent%0A%20%20constructor%3A%20(%40arg)%20-%3E%0A%20%20%20%20super(%3D%3E%20%40arg)):

```coffee
class extends parent
  constructor: (@arg) ->
    super(=> @arg)
```
```js
(class extends parent {
  constructor(arg) {
    super(() => {
      return this.arg;
    });
    this.arg = arg;
  }
});
```

[Babel allows this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEbQ3gXwFDNJGAjaBTAHgC44B2AJjHPMtNMAPbEQEBOArsAXcwBRjMDmASgTUa0CKwAOOHt2EBeAHwixY5jgKtmxaAQAWASwgA6PvwDcomokGXV-o6YHR5sAXegpEQA&debug=false&circleciRepo=&evaluate=true&lineWrap=true&presets=&targets=&version=6.26.0), but it’s not the user’s intent because `this.arg` isn’t defined when it’s used inside the `super` argument.

GeoffreyBooth:
Okay @vendethiel, I’ve found a way to search the tree of nodes that make up `super`‘s arguments, to find any references to `this` within them. This covers the case you pointed out above.

vendethiel:
This recurses a bit *too much*:

```coffee
class extends A then constructor: (@a) -> super(class then constructor: (@b) -> @b)
```

GeoffreyBooth:
I tried using `Base::contains`, which stops traversing at scope boundaries, but it failed your example. I ran into a similar issue with function parameters, in that our checks don't work too well when someone puts gobs of code in there. Ultimately I think we couldn't find a solution that was perfectly precise, that went deep enough but not too deep, and so we opted to err on the side of too deep. The thinking was that blocks of code in function parameters is an edge case, easily refactored, and it's better to throw an unnecessary error than potentially miss throwing one. 

I think the same could be said for this case. Would you agree?

vendethiel:
`Base::contains` (or rather `Base::traverseChildren`) doesn't differentiate "scope crossing" between whether it is `this`-aware. This might be doable, this might not be, this might not be worth it. It's just a decision we take, I'm fine with it whichever way we decide.

GeoffreyBooth:
Thanks. I guess let's leave this open another day to see if anyone else has an opinion. Any other notes?

PR #4759, opened 2017-10-27
[CS1] Add postinstall message for installs of ‘coffee-script’ (with hyphen)

### LINKED ISSUE #4714: [CS1] Add postinstall message telling people about module rename, possible upgrade
We should release a 1.12.8 that has an NPM `postinstall` hook that checks if the module being installed is `coffee-script` (with hyphen), and if so prints this message (feedback welcome) for local installs:

```
CoffeeScript has moved! Please update your package.json from "coffee-script" to "coffeescript".
Also make sure you specify version bounds that you know work for you, such as "^1.12.8" or "1.x".
A breaking change version 2 of CoffeeScript has been released, which you can learn about at
http://coffeescript.org.
```

The `1.12.8` above would be pulled from the version being installed.

This would be the message for global installs:

```
CoffeeScript has moved! The primary NPM module is now "coffeescript", not "coffee-script".
You can run `npm install --global coffeescript` to get the latest version, or run
`npm install --global coffeescript@1` to get the latest 1.x version. Learn more at
http://coffeescript.org.
```

The 1.x versions of CoffeeScript are mirrored on `coffeescript`, so there’s no need to use `coffee-script` anymore.

Again, feedback welcome. One thing to note is that we have _not_ definitively decided that 2.x releases won’t be added to `coffee-script` at some point in the future. We put the 2.x versions only on `coffeescript` in the first place to avoid breaking downstream tools that may not be bounding versions of `coffee-script`, and that might be the way things stay indefinitely; or we may decide at some point in the future that downstream tools have had enough time to adapt and make both module names full mirrors of each other. Let’s not debate that here, as putting 2.x releases on `coffee-script` isn’t something we plan to do anytime soon. I just don’t want the messages above to make claims that block us from revisiting this decision in the future.

GeoffreyBooth:
Fixes #4714. Once this is merged in, when the user does `npm install coffee-script` or `npm install` in a folder that has `coffee-script` in the `package.json` but not yet installed, this message will appear:

```
CoffeeScript has moved! Please update your package.json to replace "coffee-script" with "coffeescript".
Also, a new major release is available. Learn more at http://coffeescript.org.
```

In my testing at least, the message doesn’t appear when `coffee-script` is installed as a dependency of some other module. It also doesn’t appear when running `npm upgrade coffee-script`, which is unfortunate. So this should catch new installs, and new checkouts of existing projects, but it won’t alert users of existing projects who already have any version of `coffee-script` installed (I think; maybe we need to actually push to NPM, perhaps under some other temporary module name, to truly know for sure).

This needs some thorough testing. This will become 1.12.8, hopefully the last 1.x release. I really don’t want to push a 1.12.9 because I screwed this up.


GeoffreyBooth:
Would it be better to use `npm deprecate`? https://docs.npmjs.com/cli/deprecate. We could just deprecate all `coffee-script` versions, with a message to install `coffeescript` instead.

Pinging @jashkenas @lydell @connec @vendethiel and anyone else who has NPM experience.

jashkenas:
I think that this approach is better than a deprecation — if you deprecate, you'll be dealing with a bunch of folks opening tickets asking why. This is friendlier.

We should also change the README on the 1.X npm package, visible at npmjs.com/coffee-script, to make it clear what's up.

Finally, it would be nice to include a line in the message mentioning that the major new release targets evergreen JS, and (maybe) may have some incompatibilities...

GeoffreyBooth:
Okay, I’ve found a way to test this, thanks to [Verdaccio](https://github.com/verdaccio/verdaccio), a way to mock the NPM registry with a locally-hosted version.

```bash
npm install --global verdaccio
verdaccio

# In a new shell:
npm logout
npm set registry http://localhost:4873/
npm adduser --registry http://localhost:4873
# ... create some dummy credentials

cd ~/Sites/coffeescript
git checkout redirect

# Publish to Verdaccio local NPM registry:
npm publish --force --registry http://localhost:4873

cd /tmp
mkdir test
cd test
npm init
npm install coffee-script
# The locally-hosted coffee-script is installed, and the postinstall script runs
```

I see output like this:

image

It’s a good thing I tested this, as I discovered that `Cakefile` doesn’t get uploaded to NPM when we publish! So the postinstall script of `node ./bin/cake postinstall` tries to load a `Cakefile` that doesn’t exist, and the whole thing crashes embarrassingly. To fix this, I just put the entire postinstall script into `package.json` instead.

I also tested what would happen when installing a package that depends on `coffee-script`:

```bash
# ... continuing from above
npm rm coffee-script
npm install gulp-coffee
```

Unfortunately, I see the warning again. Do we care? I’m not sure that there’s a way to detect whether the current package is being installed directly by the user or indirectly as a dependency of some other module.

GeoffreyBooth:
@jashkenas I’ve updated the language per your comment:

image

And similar in the [README](https://github.com/GeoffreyBooth/coffeescript/blob/1e1fd9bd8f5ac22831161c994abf258024f79ac4/README.md). Once we publish `coffee-script` 1.12.8 with this README, I’ll update it again to basically be a clone of the current 2.x README for publishing `coffeescript` 1.12.8 (since the `coffeescript` module doesn’t need a message about CoffeeScript moving).

jashkenas:
@GeoffreyBooth Looks lovely!

GeoffreyBooth:
@jashkenas How do you feel about this message appearing even when `coffee-script` is installed indirectly, i.e. via something like `npm install gulp-coffee`?

jashkenas:
I think it's not ideal, but if there's no way to tell if we're being installed as a dependency, it's alright.

alan-agius4:
@GeoffreyBooth I think it is possible like so; (not tested)

```js
const path = require("path");
const rootPkg = path.join(process.cwd(), "package.json");
const pkg = JSON.parse(require(rootPkg)); // you should probably verify that it exists

if ((pkg.dependencies && pkg.dependencies["coffee-script"]) || pkg.devDependencies && pkg.devDependencies["coffee-script"]) {
 // oops wrong coffee-script
} 
```

PR #4764, opened 2017-10-29
Fix #4763: Comments at beginning or end of REPL input shouldn’t throw errors

### LINKED ISSUE #4763: CLI: REPL: comments would cause SyntaxError like "return __ = ;// comments" which breaks code exec.
coffee>.editor
```coffeescript
# comments
1+2
```
return __ = (, 1 + 2);// comments
_____________^
SyntaxError: Unexpected token ,

OR
coffee>.editor
```coffeescript
###
comments
###
1+2
```
return __ = (, 1 + 2);/*
____________ ^
SyntaxError: Unexpected token ,

GeoffreyBooth:
Fixes #4763. Now if you type `# comment` in the REPL, when it evaluates the line the REPL evaluates it to `undefined`.

Nitty-gritty explanation: Comments are “attached” to tokens, and added to the output JavaScript at the end of compilation when tokens are converted into the final JS code. When there’s no nearby token to attach to, for example if your entire code string was just `# comment` or `### first comment ### # second comment`, the rewriter creates a placeholder PassthroughLiteral token that’s a zero-length string—``` `` ```, essentially—to attach the tokens to. This placeholder is only ever needed at the beginning or end the tokens array. When such a placeholder is added, a newline `TERMINATOR` token is also added, if necessary, after or the placeholder (if it starts the stream) or before it (if it ends the stream). The REPL doesn’t care about comments, so when we see these placeholder tokens at the beginning or end of the `tokens` array, we remove them before we assign the REPL result to `__`.

### DIFF
--- a/lib/coffeescript/repl.js
+++ b/lib/coffeescript/repl.js
@@ -29,7 +29,7 @@
     })(),
     historyMaxInputSize: 10240,
     eval: function(input, context, filename, cb) {
-      var Assign, Block, Call, Code, Literal, Value, ast, err, isAsync, js, referencedVars, result, token, tokens;
+      var Assign, Block, Call, Code, Literal, Value, ast, err, isAsync, js, ref, ref1, referencedVars, result, token, tokens;
       // XXX: multiline hack.
       input = input.replace(/\uFF00/g, '\n');
       // Node's REPL sends the input ending with a newline and then wrapped in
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
@ven do you or anyone else mind reviewing this?

GeoffreyBooth:
@helixbass or anyone else, should this be more generalized than what I have here? This solution is rather surgical, but I wonder if it should be any broader than this specific case; and if so, what else it should cover.

helixbass:
@GeoffreyBooth I don't have enough understanding of the comments implementation (or how the REPL works) to be of much help. This feels like a (REPL-only) rewriter pass, which often are pretty surgical to just handle some specific cases. And based on your explanation it sounds like this fix covers the necessary cases?

GeoffreyBooth:
This is _very_ surgical, just comments at beginnings and ends of input. It doesn’t touch comments in the middle, for example.

But I guess this is probably good enough, and if someone reports other REPL issues with comments I can expand it then. No point in making it broader than it needs to be, until I know how broad it really needs to be.

PR #4783, opened 2017-11-12
Fix #4774: `export default` followed by an object should always work

### LINKED ISSUE #4774: Lexer: `export default` and child braces sometimes trigger “unexpected indentation”
[Minimal case](http://coffeescript.org/#try:export%20default%0A%20%20a%3A%20%7B%0A%20%20%20%20b%3A%201%0A%20%20%7D%0A%20%20c%3A%202%0A):

```coffee
export default
  a: {
    b: 1
  }
  c: 2
```
```coffee
[stdin]:5:1: error: unexpected indentation
  c: 2
^^
```
Interestingly, if `export default` is replaced with `return`, it compiles as expected.

If braces are added around the entire exported object (`export default {`...`}`), it also compiles; so that can be the workaround for anyone dealing with this right now. I stumbled across it because of trying to use an object spread in the object:

```coffee
export default
  a: {
    b...
  }
  c: 2
```
Which requires braces around the `b...`.

GeoffreyBooth:
Fix #4774: `export default` followed by an object should always work, even if the object contains braces (or put another way, even if one of the object’s values is an explicit object, defined with braces).

Fixes a bug introduced in #4532, where `export default` was given the ability to export an implicit object. In that PR, this was implemented in an incorrect way: `default` followed by a newline caused a suppression of that newline. For objects that contain _explicit_ objects (i.e. braces), this breaks, as the suppressed newline screws up the tracking of indents, which matters when we encounter `}` tokens. Instead we should handle it in the grammar the same way returning an implicit object is handled.

@connec you had some comments on the earlier PR. @helixbass you did some work related to this in #4605.


helixbass:
@GeoffreyBooth just got notified now (post-merge) - lgtm, by explicitly restricting it to an indented `Object` in the grammar you're also getting comparable benefits to #4605 (for `return`) where unintended (ie non-object) indented cases are disallowed eg:
```
export default
  foo
```

GeoffreyBooth:
Awesome, thanks.

GeoffreyBooth:
@helixbass if you have time, I'd love some review of the other recent open pull requests :smile:

PR #4791, opened 2017-11-18
Fix #4787: Destructuring of objects within arrays can generate invalid JavaScript

### LINKED ISSUE #4787: Destructuring of objects within arrays can generate invalid JavaScript
The combination of object and array destructuring seems not to work correctly when using ... in the array destructuring part. For example

```
[...,{forename,surname}] = arr
```

generates

```
var forename, surname;

{forename, surname} = arr[arr.length - 1];
```

but should generate

```
var forename, surname;

({forename, surname} = arr[arr.length - 1]);
```

In CoffeeScript 1.9.1 the same code compiles to

```
var forename, ref, surname;

(ref = arr[arr.length - 1], forename = ref.forename, surname = ref.surname);
```

which is valid JavaScript.

zdenko:
Fixes #4787 
Destructuring assignment with expansion and object destructuring  within array compiles correctly.

```coffeescript
[...,{a,b}] = arr
```

Before:
```javascript
var a, b;
{a, b} = arr[arr.length - 1];
```

After:
```javascript
var a, b;
({a, b} = arr[arr.length - 1]);
```

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2209,7 +2209,7 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or o.level is LEVEL_TOP and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
+    if o.level > LEVEL_LIST or (o.level is LEVEL_TOP or o.level is LEVEL_LIST) and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
       @wrapInParentheses answer
     else
       answer
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2209,7 +2209,9 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or (o.level is LEVEL_TOP or o.level is LEVEL_LIST) and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
+    # The assignment is wrapped in parentheses if 'o.level' has lower precedence than LEVEL_LIST (3)
+    # (i.e. LEVEL_COND (4), LEVEL_OP (5) or LEVEL_ACCESS (6)), or if we're destructuring object, e.g. {a,b} = obj.
+    if o.level > LEVEL_LIST or isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
       @wrapInParentheses answer
     else
       answer
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -3286,7 +3286,7 @@
         answer = compiledName.concat(this.makeCode(` ${this.context || '='} `), val);
         // Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
         // if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-        if (o.level > LEVEL_LIST || o.level === LEVEL_TOP && isValue && this.variable.base instanceof Obj && !this.nestedLhs && !(this.param === true)) {
+        if (o.level > LEVEL_LIST || (o.level === LEVEL_TOP || o.level === LEVEL_LIST) && isValue && this.variable.base instanceof Obj && !this.nestedLhs && !(this.param === true)) {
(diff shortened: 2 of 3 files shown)


GeoffreyBooth:
This is pretty surgical. @connec, any notes?

src/nodes.coffee
```diff
@@ -2209,7 +2209,7 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or o.level is LEVEL_TOP and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
+    if o.level > LEVEL_LIST or (o.level is LEVEL_TOP or o.level is LEVEL_LIST) and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
```
vendethiel: The first > could be >=?

GeoffreyBooth: That sounds like an improvement, yes.

Side note: Can someone who really understands `LEVEL_TOP`, `LEVEL_LIST` etc. write up an explanation on one of the wiki pages?

zdenko: @vendethiel, @GeoffreyBooth 
The condition has basically just left and right side, and could be written as

```coffeescript
if o.level > LEVEL_LIST or (o.level is LEVEL_TOP and isLevel and ...)
```

So, if we change first `>` to `>=`, the left side condition will be also true in cases like
` a = (b=1) -> `, where the level of assignment `b=5` is `LEVEL_LIST`, and will be compiled to  
`a = function((b=1)) {}`.

However, the condition could be simplified a bit and changed to
```coffeescript
if o.level > LEVEL_LIST or o.level <= LEVEL_LIST and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
```

or even to
```coffeescript
if o.level > LEVEL_LIST or isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
```

As far as I understood values of the levels, they're used to control precedence of the compiled code fragments.
From the `nodes.coffee`
```coffeescript
# Levels indicate a node's position in the AST. Useful for knowing if
# parens are necessary or superfluous.
LEVEL_TOP    = 1  # ...;
LEVEL_PAREN  = 2  # (...)
LEVEL_LIST   = 3  # [...]
LEVEL_COND   = 4  # ... ? x : y
LEVEL_OP     = 5  # !...
LEVEL_ACCESS = 6  # ...[0]
```

I guess, it can be confusing  sometimes since level with value 1 (`LEVEL_TOP`) has higher precedence then 2, i.e. `1 > 2` .

GeoffreyBooth: Okay, so let’s wrap this up. @zdenko and @vendethiel, what do you think the best check is?

And @zdenko you should probably add a comment explaining the conditional, similar to what you’ve written here.

vendethiel: I'd say `if o.level > LEVEL_LIST or isValue and @variable.base instanceof Obj and not @nestedLhs and not @param`

zdenko: Agree. It's shorter and IMHO, somehow less confusing.

src/nodes.coffee
```diff
@@ -2209,7 +2209,9 @@ exports.Assign = class Assign extends Base
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
     # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
     # if we’re destructuring without declaring, the destructuring assignment must be wrapped in parentheses.
-    if o.level > LEVEL_LIST or o.level is LEVEL_TOP and isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
+    # The assignment is wrapped in parentheses if 'o.level' has lower precedence than LEVEL_LIST (3)
+    # (i.e. LEVEL_COND (4), LEVEL_OP (5) or LEVEL_ACCESS (6)), or if we're destructuring object, e.g. {a,b} = obj.
+    if o.level > LEVEL_LIST or isValue and @variable.base instanceof Obj and not @nestedLhs and not (@param is yes)
```
connec: This looks good to me. I'd suggest it could be a bit more readable if the condition was rearranged slightly:

```js
if o.level > LEVEL_LIST or (isValue and @variable.isObject() and not @nestedLhs and not @param)
```

GeoffreyBooth: I appreciate the comment, but can you please explain why parentheses are required for those lower levels?

zdenko: @connec  `@variable.base instaceof Obj` is needed for cases like `{}.p = 0`. 
Changing `not (@param is yes)` to `not @param` cause two tests to fail, but I can't figure which ones exactly and why   ¯\\\_(ツ)\_/¯

@GeoffreyBooth I'm not sure exactly about lower levels. But, it seems that values for `LEVEL_COND` (e,g. `Existence`, `If`,...) and `LEVEL_ACCESS` (e.g. `Slice`, `In`) are already wrapped in parentheses elsewhere, before passing thru the `Assign`.
This leaves only `LEVEL_OP`, (e.g. `+`, `-`, ...) as in this case `a = b - c = 2` which has to be compiled to `a = b - (c = 2)`.
So, changing condition from `o.level > LEVEL_LIST` to `o.level is LEVEL_OP` works and all tests pass.   
On the other hand, wrapping expressions into another pair of parentheses doesn't hurt either.  
Perhaps someone else, more familiar with the internals, could shed some light on this.

PR #4792, opened 2017-11-19
Fix #4790: Double-check that we’re not creating a bound generator function

### LINKED ISSUE #4790: Invalid JS: Possible to write => generator when yield is under two comprehensions
With the latest compiler,

```coffee
=>
  for x in []
    for y in []
      yield tpl
```

compiles into `*() => ...`

GeoffreyBooth:
Fixes #4790. So that case was interesting:

```coffee
=>
  for x in []
    for y in []
      yield z
```

This compiled, even though it should’ve thrown an error about `yield` not being allowed in `=>` functions (ES2015 just doesn’t allow bound generator functions, for whatever reason). The reason our usual check for `yield`s inside `=>` functions missed this case was because one of the `for` loops here compiled into an immediately-invoked function expression (IIFE), which created a new function scope around the `yield`; and that new function was a normal one, not bound:

```js
(*() => {
  var i, len, ref, results, x, y;
  ref = [];
  results = [];
    x = ref[i];
    results.push((yield* (function*() {
      var j, len1, ref1, results1;
      ref1 = [];
      results1 = [];
      for (j = 0, len1 = ref1.length; j < len1; j++) {
        y = ref1[j];
        results1.push((yield z));
      }
      return results1;
    })()));
  }
  return results;
});
```

And all this happened because of all the contortions involved with returning the results of these `for` loops. Throw a `return` after the `yield` and the expected error message appears.

It looked hopelessly complicated to try to amend the code where the current check is (when compiling `yield`, we check that the function it’s inside isn’t bound) to try to know when to look up a level or two because its parent function might’ve been generated by the compiler. A simpler solution to this case was to add an extra error check at the source, the compilation of the function itself, so that’s what this PR does. Either we catch the bound generator-ness when compiling `yield` or when compiling `=>`.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -3729,7 +3729,7 @@
       // parameters after the splat, they are declared via expressions in the
       // function body.
       compileNode(o) {
-        var answer, body, boundMethodCheck, comment, condition, exprs, generatedVariables, haveBodyParam, haveSplatParam, i, ifTrue, j, k, l, len1, len2, len3, m, methodScope, modifiers, name, param, paramNames, paramToAddToScope, params, paramsAfterSplat, ref, ref1, ref2, ref3, ref4, ref5, ref6, ref7, ref8, scopeVariablesCount, signature, splatParamName, thisAssignments, wasEmpty;
+        var answer, body, boundMethodCheck, comment, condition, exprs, generatedVariables, haveBodyParam, haveSplatParam, i, ifTrue, j, k, l, len1, len2, len3, m, methodScope, modifiers, name, param, paramNames, paramToAddToScope, params, paramsAfterSplat, ref, ref1, ref2, ref3, ref4, ref5, ref6, ref7, ref8, scopeVariablesCount, signature, splatParamName, thisAssignments, wasEmpty, yieldNode;
         if (this.ctor) {
           if (this.isAsync) {
             this.name.error('Class constructor may not be async');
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
I think once this PR goes in we can release 2.0.3. Anyone else care to glance at it? @vendethiel? @connec? @zdenko? @helixbass?

zdenko:
Just from curiosity, I tried this approach in `Op:: compileContinuation `.   
It seems to be working but might miss some cases. I think your solution is better as it will catch all of them.
```coffeescript
@@ -3169,9 +3169,14 @@ exports.Op = class Op extends Base
   compileContinuation: (o) ->
     parts = []
     op = @operator
+    isRootBound = (exp) ->
+      return isRootBound(exp.expressions) if (exp?.expressions)
+      block = [].concat exp if exp
+      isGlyph = yes for b in block when b?.funcGlyph?.glyph is "=>"
+      isGlyph ? no
     unless o.scope.parent?
       @error "#{@operator} can only occur inside functions"
-    if o.scope.method?.bound and o.scope.method.isGenerator
+    if isRootBound(o.scope.root) or o.scope.method?.bound and o.scope.method.isGenerator
```

GeoffreyBooth:
Thanks @zdenko. So I guess this is good to merge?

Do you think we should keep the check in both places, or perhaps just keep the new one? I feel like it can’t hurt to check twice . . .

zdenko:
I didn't test my approach thoroughly, so I wouldn't include it. Besides, your PR solved the issue, seems more solid and can be merged IMHO.

PR #4795, opened 2017-11-22
2.0.3

### LINKED ISSUE #4775: Website still says it uses the text/coffeescript feature
...whereas it now has compiled minified JS. That's what you get for documenting technical details. :)

> In fact, the little bit of glue script that runs Try CoffeeScript, as well as the code examples and other interactive parts of this site, is implemented in just this way. View source and look at the bottom of the page to see the example.

GeoffreyBooth:
Fixes #4775, and recent bugfixes:

*   Bugfix for `export default` followed by an implicit object that contains an explicit object, for example `exportedMember: { obj... }`.
*   Bugfix for `key, val of obj` after an implicit object member, e.g. `foo: bar for key, val of obj`.
*   Bugfix for combining array and object destructuring, e.g. `[ ..., {a, b} ] = arr`.
*   Bugfix for an edge case where it was possible to create a bound (`=>`) generator function, which should throw an error as such functions aren’t allowed in ES2015.
*   Bugfix for source maps: `.map` files should always have the same base filename as the requested output filename. So `coffee --map --output foo.js test.coffee` should generate `foo.js` and `foo.js.map`.
*   Bugfix for incorrect source maps generated when using `--transpile` with `--map` for multiple input files.
*   Bugfix for comments at the beginning or end of input into the REPL (`coffee --interactive`).


GeoffreyBooth:
This look good everyone? @vendethiel @lydell @jashkenas @connec @helixbass @zdenko ?

Issue #4684, opened 2017-09-05
Proposal: Elision


danielbayley:
Following on from #3972 (and this [comment] in particular)…

Should we not align with ES6 and implement _elision_?

~~~ js
const [,,c] = [1,2,3];
console.log(c) // 3
~~~

~~~ coffee
[,,c] = [1,2,3]
console.log c # Error
~~~

[comment]: https://github.com/jashkenas/coffeescript/issues/3972#issuecomment-98233763

jashkenas:
Yes, as soon as we compile into ES6 destructuring, we should allow it.

(Distasteful as it may be)

GeoffreyBooth:
[We already are for arrays.](http://coffeescript.org/v2/#try:%5Ba%2Cb%2Cc%5D%20%3D%20%5B1%2C2%2C3%5D)

vendethiel:
@GeoffreyBooth It's about elision, so `[,,thing]`

GeoffreyBooth:
Yes. We don’t currently support elision (for anything) but as of 2.0.0-beta5 we output into ES syntax for array destructuring; and into ES2015 (fallback) syntax for object destructuring, until object destructuring reaches Stage 4 in ES2017 or ES2018.

So to support elision, we could support it for array destructuring now and output like the example above; but for object destructuring, if that even makes sense for elision, we would need to output into Stage 4-compatible syntax like how object destructuring is output now.

danielbayley:
> (Distasteful as it may be)

😄 I think this is one of those problems to which there is no real stand-out/elegant solution. If you just give the unused params a name that will throw linter errors when [as you ideally would be] checking for unused variables in general.

It does feel more elegant to me than other suggestions Iv'e seen… and of much more use than say, trailing commas in arrays.

Also, I wonder if this should be a thing with function parameters? (for consistency)
~~~ coffee
fn = (,,c) => console.log c
fn 1,2,3 # 3
~~~

connec:
Don't forget that 'expansions' are a thing in CS:

```coffee
[..., c] = [1, 2, 3]
((..., c) -> c) 1, 2, 3
```

jashkenas:
Trailing commas in arrays are a good example of why elisions are kinda stinky, syntax-wise.

Because if: `[,b,] = [1,2,3]` destructures a three-element array ... then `[1,2,].length` ought to equal `3`. But it doesn't.

GeoffreyBooth:
[...]

PR #4796, opened 2017-11-26
[Enhancement] Elision

zdenko:
This PR implements `elision` (#4684). There are probably some rough edges and errors, but so far the following code is properly compiled.

```coffeescript
arr = [1,2,3,4,5,6,7,8,9]

[,a,,b,,c,,,d] = arr  
# a = 2, b = 4, c = 6, d = 9

[,a,,,b...] = arr 
# a = 2, b = [5,6,7,8,9]

[,a,...,,b,,c] = arr 
# a = 2, b = 7, c = 9

[...,a,,,b,,,] = arr
# a = 4, b = 7

foo = ([,a,,b,,c,,,d]) -> [a,b,c,d]
[a,b,c,d] = foo arr
# [a,b,c,d] = [2,4,6,9]
```


test/arrays.coffee
```diff
@@ -26,6 +26,40 @@ test "incorrect indentation without commas", ->
...
+  foo = ([,a,,b,,c,,,d]) -> [a,b,c,d]
+  [a,b,c,d] = foo arr
+  arrayEq [a,b,c,d], [2,4,6,9]
+
+  [,a,,,b...] = arr
+  arrayEq [a,b], [2,[5,6,7,8,9]]
+  [,a,...,,b,,c] = arr
+  arrayEq [a,b,c], [2,7,9]
+  [...,a,,,b,,,] = arr
+  arrayEq [a,b], [4,7]
```
GeoffreyBooth: Shouldn't this be `[3,6]`?

src/grammar.coffee
```diff
@@ -533,6 +534,34 @@ grammar =
...
+    o 'ArgElisionList , Elisions INDENT ArgElisionList OptComma OUTDENT', -> $1.concat $3, $5
+    o 'ArgElisionList OptComma INDENT ArgElisionList , Elisions OUTDENT', -> $1.concat $4, $6
+  ]
+
+  ArgElision: [
+    o 'Arg',                  -> [$1]
+    o 'Elisions Arg',         -> [].concat $1, $2
+  ]
+
+  Elisions: [
```
GeoffreyBooth: This should be called `ElisionList` for consistency with `ArgList`.

src/grammar.coffee
```diff
@@ -533,6 +534,34 @@ grammar =
...
+    o 'Elision'
+    o 'Elisions Elision',     -> [].concat $1, $2
+  ]
+
+  Elision: [
+    o ',',                    -> new Elision
```
GeoffreyBooth: How does this play nice with `OptComma`? Is that why I think some of the tests are wrong, because they're discarding the last comma as optional?

test/arrays.coffee
```diff
@@ -26,6 +26,40 @@ test "incorrect indentation without commas", ->
   ok result[0][0] is 'a'
   ok result[1]['b'] is 'c'
 
+# Elisions
+test "array elisions", ->
+  eq [,1].length, 2
+  eq [,,1,2,,].length, 5
```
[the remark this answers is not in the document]
vendethiel: No, the very last comma is the optional end comma.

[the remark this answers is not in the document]
zdenko: `[a,].length is 1`
`[a,,].length is 2`
`[,a].length is 2`
`[,,a].length is 3`
In this case `[,,1,2,,,]` length is 5, because two `,,` at the beginning are count as two `Elisions` and the last `,` is the optional end comma (e.g. not `Elision`).

[Example](https://babeljs.io/repl/#?babili=false&browsers=%3E%203%25&build=&builtIns=false&code_lz=IYGwpgTgLgFA2gGgQRgQJiQXQHTgHYDmUAFgJQBQQA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=env&prettier=false&targets=&version=6.26.0) on Babel.

src/grammar.coffee
zdenko: Elision is `,` at the beginning of the array, e.g. `[,a]` (length is 2) or `,` preceded by `,`, e.g. `[a,,b]` (length is 3).
So, `OptComma` should not precede `Elision` in the `grammar`.

test/arrays.coffee
zdenko:
```javascript
// arr = [1,2,3,4,5,6,7,8,9]
// arr.length = 9
// [...,a,,,b,,,] = arr

var a, b, i;
i = arr.length - 6, a = arr[i++], i++, i++, b = arr[i++], i++, i++;
// i++ is used to skip 'Elisions'
```
`arr.length - 6` because `[a,,,b,,,].length is 6`
```javascript
i = arr.length - 6 = 9 - 6 = 3
a = arr[i++] = arr[3] = 4
// i = 4
i++, i++ 
// i = 6
b = arr[i++] = arr[6] = 7
// i = 7
i++, i++
// i = 9

```

zdenko:
@GeoffreyBooth 
> Assignment: arr = [0,,2]

This works, I'll add tests for the assignments.
```javascript
// arr = [0,,2]
var arr;
arr = [0, , 2];
```

> Function parameters: fn(first,,third)   

Did you mean `fn = (first, , third) ->`?
If I'm not mistaken, `elision` in function parameter list is not part of the ES specification.
Besides, the proposal #4684 was meant for the array `elisions`, which are part of the ES, and I wanted to finish that first.
If parameter list `elision` is required feature I'll create separate PR.

GeoffreyBooth:
I had forgotten that elision was already part of the ES spec. Since that’s the case, yes, please follow ES semantics for the trailing comma, and let’s not worry about elision in function parameters until ES supports it.

src/grammar.coffee
zdenko: Are you referring to the `Elisions`? They consist only of one or more `,` (`Elision`).
`ArgList` consists of  different elements: `Arg` (`Expression`, `Splat`, and `Expansion`) and ` ,`.

Or, perhaps you mean other elements: `ArgElision` or `ArgElisionList`?

GeoffreyBooth: I was referring to `Elisions`, but you make a good point.

jashkenas:
Looks good-ish, but can the grammar rules not be simplified further? The mixing of Elisions and OptCommas doesn't feel right, nor does the strange indentation combinations in the grammar rules.

Can we not just use the grammar to specify a Elision -> "," rule, like this PR has, and then change the existing grammar to read commas in places-where-elisions-are-allowed as Elisions?

If I'm not understanding the complexity of this change, feel free to ignore me.

zdenko:
At first, I tried to reuse existing grammar by placing `Elision` in allowed spaces, but  the existing `ArgList` rules are used for arrays and function parameter list. 
So, adding `Elision` there might break or cause errors for the function parameters.
This is the main reason I decided to create separate rules for arrays only (`ArgElisionLilst`). 

To avoid breaking existing array rules, where a comma is optional in some cases, and `Elision` requires preceding comma, I had to reuse some of the rules from `ArgList`. 
For example, `INDENT ArgList OptComma OUTDENT`  from the `ArgList` is repeated as `INDENT ArgElisionList OptComma OUTDENT` in `ArgElisionList`, but since multiple commas (elisions) are allowed at the array's end, additional rule `INDENT ArgElisionList , Elisions OUTDENT`is also required.

However, I do agree that rules are pretty messy (especially `OptComma` and `, Elision` combination).
So, I took another shot at it and tried to simplify it, by creating `OptElisions` rule which now covers both, `OptComma` and `Elisions`.

I think the result is now much better, as `ArgElisionList` is an almost identical copy of `ArgList`.

src/grammar.coffee
```diff
@@ -16,7 +16,7 @@
 
 # The only dependency is on the **Jison.Parser**.
 {Parser} = require 'jison'
-
+log = console.log
```
jashkenas: Was this left in here accidentally?

jashkenas:
Thanks for cleaning it up, that makes sense! LGTM.

src/grammar.coffee
zdenko: Yep, sorry about that. It's removed.

GeoffreyBooth:
Great work @zdenko, thanks!

danielbayley:
> Great work @zdenko, thanks!

This 👍

PR #4799, opened 2017-11-27
Fix #4798: Incorrect output for object rest destructuring inside array destructuring 

### LINKED ISSUE #4798: Nodes: Incorrect output for object rest destructuring inside array destructuring
I stumbled upon this while preparing tests for #4796.
` [{a, rest...}] = arr ` compiles to  ` [_extends({a}, rest)] = arr `, 
but instead it shoud be something like
```javascript
[ref] = arr; 
({a} = ref);
rest = objectWithoutKeys(ref, ['a']);
```
[Example](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBCA0FgpgZwC63GAvBA2gbxAFwCMsARgQEwC-AugNwBQeMEAdOwirRFuow0A&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=stage-3%2Cenv&prettier=false&targets=&version=6.26.0) on Babel.

zdenko:
Fixes #4798
```javascript
// [{a, rest...}] = arr
({a} = ref = arr[0]);
rest = objectWithoutKeys(ref, ['a']);

// [b,{a, rest...},c] = arr
b = arr[0], (({a} = ref = arr[1]), rest = objectWithoutKeys(ref, ['a'])), c = arr[2];
```

### DIFF
--- a/test/assignment.coffee
+++ b/test/assignment.coffee
@@ -185,6 +185,12 @@ test "#4787 destructuring of objects within arrays", ->
   eq b, arr[1].b
   deepEqual {a, b}, arr[1]
 
+test "#4798 destructuring of objects with splat within arrays", ->
+  arr = [1, {a:1, b:2}]
+  [...,{a, r...}] = arr
+  eq a, 1
+  deepEqual r, {b:2}
+
 test "destructuring assignment with splats", ->
   a = {}; b = {}; c = {}; d = {}; e = {}
   [x,y...,z] = [a,b,c,d,e]
--- a/test/assignment.coffee
+++ b/test/assignment.coffee
@@ -190,6 +190,11 @@ test "#4798 destructuring of objects with splat within arrays", ->
   [...,{a, r...}] = arr
   eq a, 1
   deepEqual r, {b:2}
+  [b, {q...}] = arr
+  eq b, 1
+  deepEqual q, arr[1]
+  eq q.b, r.b
+  eq q.a, a
 
 test "destructuring assignment with splats", ->
   a = {}; b = {}; c = {}; d = {}; e = {}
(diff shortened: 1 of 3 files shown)


test/assignment.coffee
```diff
@@ -185,6 +185,12 @@ test "#4787 destructuring of objects within arrays", ->
   eq b, arr[1].b
   deepEqual {a, b}, arr[1]
 
+test "#4798 destructuring of objects with splat within arrays", ->
+  arr = [1, {a:1, b:2}]
+  [...,{a, r...}] = arr
```
GeoffreyBooth: We should probably also have a test without an expansion.

GeoffreyBooth:
Other than that, it looks good to me. @connec, do you have any notes?

test/assignment.coffee
```diff
@@ -185,6 +185,17 @@ test "#4787 destructuring of objects within arrays", ->
...
+  eq a, 1
+  deepEqual r, {b:2}
+  [b, {q...}] = arr
```
GeoffreyBooth: This reuses `b` from the first test, it should get its own `test "#4798...` block.

zdenko: From the first assignment (`[...,{a, r...}] = arr`)  variables `a` and `r` are set.
The second assignment `[b, {q...}] = arr` sets variables `b` and `q`. 
I think `b` doesn't exist before the second assignment, so it's not reused.

zdenko:
@GeoffreyBooth do I need to add/change something to this PR?

GeoffreyBooth:
Nope, I'm satisfied if @lydell is satisfied.

GeoffreyBooth:
Sorry, I thought this was the other PR. @connec did you have any notes on this?

connec:
LGTM 👍

PR #4801, opened 2017-11-29
1.12.8

GeoffreyBooth:
This is it, the hopefully last 1.x release. 🤞 Very few changes from 1.12.7:

*   Added install message on `coffee-script` package encouraging people to use the `coffeescript` NPM package instead.
*   Backported bugfix for `export default` followed by an implicit object that contains an explicit object, for example `exportedMember: { obj... }`.
*   Backported bugfix for splats after existential operator soaks in arrays (`[a?.b...]`) or expressions.


cueedee:
Looks like the 1.12.8 release has not been published to npm yet.

GeoffreyBooth:
We won't be publishing it. You can use the `1` branch if you need those bug fixes. They're very minor. The point of 1.12.8 was the deprecation message.

cueedee:
Alright, thanks for clarifying.

Issue #4622, opened 2017-07-25
Proposal: Grammar: Computed property keys 


MegaUITeam:
Like this:
---
```javasript
w = 'prop'
q = {[w]:2}
```

There is no ambiguity, since after arrays `:` is never put.

I know, we can write like this:
```javascript
q = {"#{w}":2}
```
But it's not cool. In addition, we can not use link to Symbol, with the second use case
```Uncaught TypeError: Cannot convert a Symbol value to a string```

Besides
---
this is a correct javascript code, and compiles with Babel
```javascript
w = Symbol()
q = {[w]:2}
```

microdou:
It's ES6 **Computed Property Keys**, previously discussed in #4338.

GeoffreyBooth:
Closed via #4803.

PR #4803, opened 2017-11-29
[Enhancement] Computed property keys (#4622)

zdenko:
This PR implements computed property keys (#4622) syntax for object literals.
```coffeescript
i = 0
a =
  ['foo' + ++i]: i
  ['foo' + ++i]: i
  ['foo' + ++i]: i

# a = {foo1: 1, foo2: 2, foo3: 3}

param = 'size'
config =
  [param]: 12,
  ['mobile' + param.charAt(0).toUpperCase() + param.slice(1)]: 4

# conf = {size: 12, mobileSize: 4}

obj =
  [Symbol.iterator]: ->
    yield "hello"
    yield "world"

# [obj...] = ['hello', 'world']
```

Besides basic support for ES spec, it also enables use of property value shorthand syntax.
```coffeescript
key = 'a'
obj = { [key] }
# obj = {a: 'a'}

key = 5
obj = { [key] }
# obj = {5: 5}

foo = (k) -> "key_#{k}"
obj = { [foo(1)] }
# obj = { key_1: 'key_1' }
```

This is similar to "object keys with interpolations". The only difference is that property values will be the same type as the key.
```coffeescript
obj = { [5] } => { 5: 5 }
obj = { ['5'] } => { '5': '5' }

obj = { "#{5}" } => { '5': '5' }
obj = { "#{'5'}" } => { '5': '5' }

```

I temporarily disabled few "unexpected object keys" test in `test/error_messages.coffee`  until I sort out some issues.


src/nodes.coffee
```diff
@@ -765,6 +765,10 @@ exports.CSXTag = class CSXTag extends IdentifierLiteral
 exports.PropertyName = class PropertyName extends Literal
   isAssignable: YES
 
+exports.ComputedPropertyName = class ComputedPropertyName extends PropertyName
+  compileNode: (o) ->
+    [@makeCode "[" + @value.compile(o) + "]"]
```
GeoffreyBooth: These `+`s are casting whatever is output by `@value.compile(o)` as a string. Does `compile` always return a string?

I feel like this would be safer:
```coffee
  compileNode: (o) ->
    [@makeCode('['), @value.compile(o), @makeCode(']')]
```
Unless `@value.compile` should be `@value.compileToFragments` . . . ?

test/error_messages.coffee
```diff
@@ -866,7 +867,7 @@ test "unexpected object keys", ->
     {a: 1, [[]]: 2}
            ^
   '''
-
+###
```
GeoffreyBooth: Believe it or not, `{a: 1, [[]]: 2}` is valid JavaScript. Go ahead, try it!

So we should rewrite this test to check as appropriate and move it out of the `error_messages` file. And make `[[]]:` compilable if it isn’t already.

test/objects.coffee
```diff
@@ -582,6 +582,194 @@ test "#4324: Shorthand after interpolated key", ->
...
+  obj = {[a]: yes}
+  eq obj.a, yes
+  obj = {[a]}
+  eq obj.a, 'a'
+  obj = {[5]}
+  eq obj[5], 5
+  obj = {['5']}
+  eq obj['5'], '5'
+
+  # Commas in implicit object.
```
GeoffreyBooth: Please split up each section into a new `test`. Long tests are annoying, for if they fail there are lots of places to check to find where exactly the failure is. Ideally each `test` function would have only one validation (e.g. only one `eq`).

src/nodes.coffee
zdenko: Thanks. It's changed to   
`[@makeCode('['), @value.compileToFragments(o, LEVEL_LIST)..., @makeCode(']')]`

test/objects.coffee
zdenko: Agree. I copied the whole block from "object keys with interpolations".
I'll split them into smaller sections.

src/nodes.coffee
```diff
@@ -1501,6 +1505,15 @@ exports.Obj = class Obj extends Base
           [key, value] = prop.base.cache o
           key  = new PropertyName key.value if key instanceof IdentifierLiteral
           prop = new Assign key, value, 'object'
+        else if key instanceof Value and key.base instanceof ComputedPropertyName
+          # { [foo()] } => { [ref = foo()]: ref }
```
GeoffreyBooth: Do you mean:

```
# `{ [foo()] }` output as `{ [ref = foo()]: ref }`.
```
? Remember the comments are markdown and become the annotated source.

zdenko: Right. I'll keep that in mind.

GeoffreyBooth:
That’s all I’ve got. Any notes, @vendethiel or @lydell or @connec?

@zdenko thanks for tackling this, this is a good feature to add.

danielbayley:
Awesome work @zdenko! This one is nice to have, thanks.

Issue #4802, opened 2017-11-29
New React fragments syntax causes compilation error


papodaca:
Using the new fragment syntax supported by react [16.2](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) causes the following error:

```
ERROR in ./app/javascript/components/layout/index.coffee
Module build failed: Error: ./app/javascript/components/layout/index.coffee:10:4: error: missing / (unclosed regex)
  </>
   ^
    L9:   </>
```

Example input:
```coffeescript
import React from 'react'

Component = (props) =>
  <>
    <OtherComponent />
    <OtherComponent />
  </>

export default Component
```

Expected output:
```javascript
var Component;

import React from 'react';

Component = (props) => {
  return <>
    <OtherComponent />
    <OtherComponent />
  </>;
};

export default Component;
```

GeoffreyBooth:
Well, at least it’s a compile error at the moment, rather than compiling into something else. As a workaround for now, follow the [advice](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html#other-tools) of Facebook’s announcement:

> For other tools, please check with the corresponding documentation to check if there is support available. However, if you’re blocked by your tooling, you can always start with using the `<Fragment>` component and perform a codemod later to replace it with the shorthand syntax when the appropriate support is available.

```jsx
Component = (props) =>
  <Fragment>
    <OtherComponent />
    <OtherComponent />
  </Fragment>
```

GeoffreyBooth:
@clemmy I know we’re not as cool as Babel, Flow and TypeScript, but we’d appreciate being added to the list of tools that support JSX that you think about when making changes to it 😄

clemmy:
@GeoffreyBooth For sure 🙂. You can keep up to date on our plans/proposals for JSX [here](https://github.com/facebook/jsx/). If you can reference the PR/shoot a comment in there to notify us of the CoffeeScript progress, we'd be happy to add you into the list!

GeoffreyBooth:
Fixed by #4804.

PR #4804, opened 2017-11-29
[Enhancement] CSX fragments syntax (#4802)

zdenko:
React [fragment syntax](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) (#4802).
```coffeescript
Component = (props) =>
  <>
    <OtherComponent />
    <OtherComponent />
  </>
```
 Output:
```javascript
Component = (props) => {
  return <>
    <OtherComponent />
    <OtherComponent />
  </>;
};
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -597,7 +597,7 @@ exports.Lexer = class Lexer
         @mergeInterpolationTokens tokens, {delimiter: '"'}, (value, i) =>
           @formatString value, delimiter: '>'
         match = CSX_IDENTIFIER.exec(@chunk[end...]) or CSX_FRAGMENT_IDENTIFIER.exec(@chunk[end...])
-        if not match or match[0] isnt csxTag.name
+        if not match or match[1] isnt csxTag.name
           @error "expected corresponding CSX closing tag for #{csxTag.name}",
             csxTag.origin[2]
         afterTag = end + csxTag.name.length
@@ -1179,7 +1179,9 @@ CSX_IDENTIFIER = /// ^
 ///
 
 # Fragment: <></>
-CSX_FRAGMENT_IDENTIFIER = /// ^ ((?=>)) /// # Ends immediately with '>'.
+CSX_FRAGMENT_IDENTIFIER = /// ^
+  ()>
+  /// # Ends immediately with '>'.
 
 CSX_ATTRIBUTE = /// ^
   (?!\d)
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -556,7 +556,7 @@ exports.Lexer = class Lexer
     # Check the previous token to detect if attribute is spread.
     prevChar = if @tokens.length > 0 then @tokens[@tokens.length - 1][0] else ''
     if firstChar is '<'
-      match = CSX_IDENTIFIER.exec @chunk[1...]
+      match = CSX_IDENTIFIER.exec(@chunk[1...]) or CSX_FRAGMENT_IDENTIFIER.exec(@chunk[1...])
       return 0 unless match and (
         @csxDepth > 0 or
         # Not the right hand side of an unspaced comparison (i.e. `a<b`).
@@ -596,7 +596,7 @@ exports.Lexer = class Lexer
           @matchWithInterpolations INSIDE_CSX, '>', '</', CSX_INTERPOLATION
         @mergeInterpolationTokens tokens, {delimiter: '"'}, (value, i) =>
           @formatString value, delimiter: '>'
-        match = CSX_IDENTIFIER.exec @chunk[end...]
+        match = CSX_IDENTIFIER.exec(@chunk[end...]) or CSX_FRAGMENT_IDENTIFIER.exec(@chunk[end...])
         if not match or match[0] isnt csxTag.name
           @error "expected corresponding CSX closing tag for #{csxTag.name}",
             csxTag.origin[2]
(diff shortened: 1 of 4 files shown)


src/lexer.coffee
```diff
@@ -1178,6 +1178,12 @@ CSX_IDENTIFIER = /// ^
   ( (?: (?!\s)[\.\-$\w\x7f-\uffff] )+ ) # Like `IDENTIFIER`, but includes `-`s and `.`s.
 ///
 
+# Fragment: <></>
+CSX_FRAGMENT_IDENTIFIER = /// ^
+  (?![\d<]) # Must not start with `<`.
+  ( (?=>) ) # Ends immediately with '>.
```
vendethiel: why does this need to be a lookahead?

zdenko: For normal CSX tags RegEx will capture tag name, e,g, `div` from `<div>`.
The fragment tag is empty (`<>`) which means that tag name is '', and has to be compared to '' from closing tag `</>`.
I'm far from RexEg mastery and open to any better suggestion.

GeoffreyBooth: Aren’t you just trying to match exactly `<>` or `</>`? So

```js
/(<>)|(<\/>)/
```
https://regex101.com/r/68Imt8/2

test/csx.coffee
```diff
@@ -742,3 +742,27 @@ test '#4686: comments inside interpolations that also contain CSX attributes', -
...
+    </>
+  ''', '''
+    <>
+      Some text.
+      <h2>A heading</h2>
+      More text.
+      <h2>Another heading</h2>
+      Even more text.
+    </>;
+  '''
```
GeoffreyBooth: Can we also add a test for the example from #4802?

```coffee
Component = (props) =>
  <Fragment>
    <OtherComponent />
    <OtherComponent />
  </Fragment>
```

src/lexer.coffee
zdenko: Yes, something like that.  But, the way `CSX` processing works in the `lexer` is more complex and I'm trying to reuse existing code by adding RegEx for empty tag `<>`. This basically means I need to capture `` in `<>` which is then used for creating `CSX` token and finding the matching closing tag.
I'm working on RegEx improvement at the moment.

test/csx.coffee
zdenko: Done.

src/lexer.coffee
```diff
@@ -1178,6 +1178,12 @@ CSX_IDENTIFIER = /// ^
   ( (?: (?!\s)[\.\-$\w\x7f-\uffff] )+ ) # Like `IDENTIFIER`, but includes `-`s and `.`s.
 ///
 
+# Fragment: <></>
+CSX_FRAGMENT_IDENTIFIER = /// ^
+  (?![\d<])   # Must not start with number or `<`
+  (.{0}(?=>)) # Ends immediately with '>.
+///
```
GeoffreyBooth: I think this could become:

```coffee
CSX_FRAGMENT_IDENTIFIER = /^((?=>))/ # Ends immediately with `>`.
```

You’re really only looking for the next character to be `>`, and capturing a zero-length string so that the regex `exec` matches. The stuff about “must not start with a number or `<`“ refers to the identifier inside a normal CSX tag, e.g. the `div` in `<div>`. But by definition a fragment has no identifier, so we don’t need to validate one.

src/lexer.coffee
```diff
@@ -1179,10 +1179,7 @@ CSX_IDENTIFIER = /// ^
 ///
 
 # Fragment: <></>
-CSX_FRAGMENT_IDENTIFIER = /// ^
-  (?![\d<])   # Must not start with number or `<`
-  (.{0}(?=>)) # Ends immediately with '>.
-///
+CSX_FRAGMENT_IDENTIFIER = /// ^ ((?=>)) # Ends immediately with '>`. ///
```
GeoffreyBooth: You want `` `>` ``, not `` '>` ``. And don't you think the comment should go outside of the regex? It describes the entire regex.

GeoffreyBooth: Actually, it’s odd that this even compiles. I would think that the latter `///` would be considered part of the comment, leaving this an unclosed regex.

GeoffreyBooth:
I think this is okay. @vendethiel?

src/lexer.coffee
lydell: Looks like you found a bug! #4811

src/lexer.coffee
```diff
@@ -1178,6 +1178,9 @@ CSX_IDENTIFIER = /// ^
   ( (?: (?!\s)[\.\-$\w\x7f-\uffff] )+ ) # Like `IDENTIFIER`, but includes `-`s and `.`s.
 ///
 
+# Fragment: <></>
+CSX_FRAGMENT_IDENTIFIER = /// ^ ((?=>)) /// # Ends immediately with '>'.
```
vendethiel: why not just `/// ^ () > ///`?

zdenko: I've tried that one before but didn't work.  At first, I thought I got it wrong and moved on to the different pattern.
But, now after your suggestion, I took a closer look and actually found a bug in the `lexer` where start and end CSX tag where compared. The code compares group match of the start tag with the full match of the end tag.
It accidentally worked since they are the same, but it failed on fragment tag since the full match is `>` and group match is ''.

GeoffreyBooth:
@vendethiel Did you have any other notes?

Issue #4805, opened 2017-11-30
Error while installing Coffee-script@1.12.8 'invalid or unexpected token' 


eashi:
When I try to install like below I get the error.
I am using NPM 5.6.0 on a Windows 10 Machine .

> npm install -g coffee-script@1.12.8
> C:\Users\myuser\AppData\Roaming\npm\cake -> C:\Users\myuser\AppData\Roaming\npm\node_modules\coffee-script\bin\cake
> C:\Users\myuser\AppData\Roaming\npm\coffee -> C:\Users\myuser\AppData\Roaming\npm\node_modules\coffee-script\bin\coffee
> 
> > coffee-script@1.12.8 postinstall C:\Users\myuser\AppData\Roaming\npm\node_modules\coffee-script
> > node --eval 'if (require("./package.json").name === "coffee-script") { var red, yellow, cyan, reset; red = yellow = cyan = reset = ""; if (!process.env.NODE_DISABLE_COLORS) { red = "\x1b[31m"; yellow = "\x1b[33m"; cyan = "\x1b[36m"; reset = "\x1b[0m"; } console.warn(red + "CoffeeScript has moved!" + reset + " Please update references to " + yellow + "\"coffee-script\"" + reset + " to use " + yellow + "\"coffeescript\"" + reset + " (no hyphen) instead."); console.warn("Also, a new major version has been released under the " + yellow + "coffeescript" + reset + " name on NPM. This new release targets modern JavaScript, with minimal breaking changes. Learn more at " + cyan + "http://coffeescript.org" + reset + "."); console.warn(""); }'
> 
> [eval]:1
> 'if
> ^^^
> SyntaxError: Invalid or unexpected token
>     at createScript (vm.js:56:10)
>     at Object.runInThisContext (vm.js:97:10)
>     at Object.<anonymous> ([eval]-wrapper:6:22)
>     at Module._compile (module.js:570:32)
>     at evalScript (bootstrap_node.js:354:27)
>     at run (bootstrap_node.js:123:11)
>     at run (bootstrap_node.js:390:7)
>     at startup (bootstrap_node.js:122:9)
>     at bootstrap_node.js:505:3
> npm ERR! code ELIFECYCLE
> npm ERR! errno 1
[...]

PR #4807, opened 2017-11-30
fix: error while installing Coffee-script@1.12.8 'invalid or unexpect…

alan-agius4:
…ed token'

Closes: #4805 

@vendethiel, @lydell  & @GeoffreyBooth.

### DIFF
--- a/warning.js
+++ b/warning.js
@@ -13,8 +13,8 @@ function readJsonFile(filePath) {
 var pkgCoffeeData = readJsonFile('./package.json');
 
 // In npm when a package is installed the 'package.json' is amended and some new fields are added
-// one of these fields is '_where' which indicated from where the package was installed.
-// in case _where is not defined we cannpt determine from where it was installed, thus show the warning.
+// one of these fields is '_where' which indicates from where the package was installed.
+// in case _where is not defined we cannot determine from where it was installed, thus show the warning.
 if (pkgCoffeeData && pkgCoffeeData._where) {
   if (pkgCoffeeData._where.indexOf('node_modules') !== -1) {
     return;
--- a/warning.js
+++ b/warning.js
@@ -1,0 +1,54 @@
+var fs = require('fs'),
+  path = require('path');
+
+function readJsonFile(filePath) {
+  try {
+    var data = fs.readFileSync(filePath, 'utf-8');
+    return JSON.parse(data);
+  } catch (error) {
+    return null;
+  }
+}
+
+var pkgCoffeeData = readJsonFile('./package.json');
+
+// In npm when a package is installed the 'package.json' is amended and some new fields are added
+// one of these fields is '_where' which indicated from where the package was installed.
+// in case _where is not defined we cannpt determine from where it was installed, thus show the warning.
+if (pkgCoffeeData && pkgCoffeeData._where) {
+  if (pkgCoffeeData._where.indexOf('node_modules') !== -1) {
+    return;
+  }
+
+  var pkgData = readJsonFile(path.join(pkgCoffeeData._where, './package.json'));
+  if (!pkgData) {
+    return;
+  }
+
+  var hasOldCoffeeScript = (pkgData.dependencies && pkgData.dependencies['coffee-script'])
+    || (pkgData.devDependencies && pkgData.devDependencies['coffee-script']);
+
+  if (!hasOldCoffeeScript) {
+    return;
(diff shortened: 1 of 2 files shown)


Alevale:
This will actually show the message for anybody using coffeescript, no matter if using the "coffeescript" package or "coffee-script" so I would say it's not valid because what you want is to say that you need to change the name to "coffeescript" only if you are using "coffee-script" and not also if you are using it.

alan-agius4:
@Alevale, in package.json, the name is always `coffee-script`

GeoffreyBooth:
I tested this in Windows by pasting the `package.json` into a folder and running `npm run postinstall` and it seems to work. It’s unfortunate that we lose the quotes around `"coffee-script"`, as that was meant to imply that string in `package.json`, but I can live with that.

@alan-agius4 have you tested this in Windows?

GeoffreyBooth:
Oh and I’ll just leave out the `postinstall` entirely in the `coffeescript` side. The `if` was meant so that there would be one consistent codebase between the two, but clearly it’s not worth the headache.

alan-agius4:
Yeah my screenshot i posted is on windows

GeoffreyBooth:
Great, do you mind really and truly testing it via the steps in https://github.com/jashkenas/coffeescript/pull/4759#issuecomment-347093273? I just don’t want to push more than one fix to this 😄 And I don’t have a great Windows environment to test with.

alan-agius4:
Unfortunately i am at home now, and I only have mac available.

Basicly my test was the following;
npm pack
npm install 

Anyone willing to test it out your way?

GeoffreyBooth:
@alan-agius4 If there’s a way to have this only print on direct installs, we might as well add that ability at the same time as we fix this. I like your code in https://github.com/jashkenas/coffeescript/issues/4805#issuecomment-348259369, but it needs to be rewritten to run in Node 0.8+ (so no `const` etc.) and anything risky should go in a `try` block. We could put it into a `warning.js` and change `postinstall` to `node ./warning.js`. That should also fix the Windows quotes issue.

alan-agius4:
Sure thing, though i’ll only be available to do it tomorrow. Lets say in
14hrs.

GeoffreyBooth:
Yes, I don’t want to try again until Friday night Pacific time at the earliest.

alan-agius4:
@GeoffreyBooth I pushed the implementation that we discussed yesterday, that said I had to do some modifications to the original snippet to get it working.

warning.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+var pgkCoffeeData = readJsonFile('./package.json');
+
+if (!pgkCoffeeData || !pgkCoffeeData._where || pgkCoffeeData._where.indexOf('node_modules') !== -1) {
+  return;
+}
+
+var pgkData = readJsonFile(path.join(pgkCoffeeData._where, './package.json'));
```
GeoffreyBooth: Can you explain this?

warning.js
```diff
@@ -0,0 +1,49 @@
...
+if (!pgkData) {
+  return;
+}
+
+var hasOldCoffeScript = (pgkData.dependencies && pgkData.dependencies['coffee-script'])
+  || (pgkData.devDependencies && pgkData.devDependencies['coffee-script']);
```
GeoffreyBooth: Also there are some misspellings: `hasOldCoffeeScript`, `pkgData`.

warning.js
alan-agius4: When you install a library using `npm` a new property will be added in the `package.json` indicated from where the installation happened

alan-agius4: Maybe it makes sense to display the warning as well when _where is not defined cause we cannot determin from where it was installed

alan-agius4: In npm when a package is installed the `package.json` is amended and some new fields are added one of these fields is `_where` which indicated from where the package was installed.

In case `_where` is not defined we cannot determine from where it was installed, thus show the warning.

GeoffreyBooth:
Let's pause on this, per https://github.com/jashkenas/coffeescript/issues/4714#issuecomment-348394552. I think we might just use `npm deprecate` instead after all.

GeoffreyBooth:
Closing as @jashkenas will `npm deprecate` per https://github.com/jashkenas/coffeescript/issues/4714#issuecomment-348394552.

PR #4814, opened 2017-12-01
Fix #4811: Heregex comments cannot contain three slashes in a row

### LINKED ISSUE #4811: Heregex comments cannot contain three slashes in a row
```coffee
///
 # Comment with three slashes: ///
///
```
```
[stdin]:3:1: error: missing ///
///
^^^
```

[Try CoffeeScript link](http://coffeescript.org/#try:%2F%2F%2F%0A%20%23%20Comment%20with%20three%20slashes%3A%20%2F%2F%2F%0A%2F%2F%2F)

This is because the lexing of heregexes does not take comments into account.

zdenko:
Fixes #4811. Heregex comment will consume `///` in the same row.
```coffeescript
/// .* # comment ///

# [stdin]:1:1: error: missing ///
# /// .* # comment ///
# ^^^
```

```coffeescript
///
  a | # comment with ///
  b   # /// another comment
///

# // comment with ///
# // /// 'heregex' in comment will be consumed
# /a|b/;
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1254,7 +1254,7 @@ VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
 HEREGEX      = /// ^
   (?:
-      # Skip `\`, `/`, `#` and `\s` characters.
+      # Match any character, except those that need special handling below.
       [^\\/#\s]
       # Match `\` followed by any character.
     | \\[\s\S]
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1256,9 +1256,9 @@ HEREGEX      = /// ^
       (?:
         [^\\/#]   |
         \\[\s\S]  |
+        /(?!//)   |
         # The comment should consume everything until the end of the line, including the '///'
-        \#(?!\{)(?:[^\n\r]*[\n\r])  |
-        /(?!//)
+        \#(?!\{)(?:[^\n]*[\n])
       )*
 ///
 
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1216,9 +1216,9 @@ HERE_JSTOKEN = ///^ ```     ((?: [^`\\] | \\[\s\S] | `(?!``) )*) ``` ///
 STRING_START   = /^(?:'''|"""|'|")/
 
 STRING_SINGLE  = /// ^(?: [^\\']  | \\[\s\S]                      )* ///
-STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] |           \#(?!\{) )* ///
+STRING_DOUBLE  = /// ^(?: [^\\"\#] | \\[\s\S] |           \#(?!\{) )* ///
 HEREDOC_SINGLE = /// ^(?: [^\\']  | \\[\s\S] | '(?!'')            )* ///
-HEREDOC_DOUBLE = /// ^(?: [^\\"#] | \\[\s\S] | "(?!"") | \#(?!\{) )* ///
+HEREDOC_DOUBLE = /// ^(?: [^\\"\#] | \\[\s\S] | "(?!"") | \#(?!\{) )* ///
 
 INSIDE_CSX = /// ^(?:
     [^
@@ -1253,19 +1253,20 @@ REGEX_FLAGS  = /^\w*/
 VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
 HEREGEX      = /// ^
-      (?:
-        [^\\/#]   |
-        \\[\s\S]  |
-        /(?!//)   |
-        # The comment should consume everything until the end of the line, including the '///'
-        \#(?!\{)(?:[^\n]*[\n])
-      )*
+  (?:
+      [^\\/#]   |
+      \\[\s\S]  |
+      /(?!//)   |
+      \\\#      | # Escaped hash.
+      # The comment should consume everything until the end of the line, including the '///'.
+      \#(?!\{)[^\n]*\n
+  )*
 ///
 
 HEREGEX_OMIT = ///
     ((?:\\\\)+)     # Consume (and preserve) an even number of backslashes.
   | \\(\s)          # Preserve escaped whitespace.
-  | \s+(?:#.*)?     # Remove whitespace and comments.
+  | \s+(?:\#.*)?     # Remove whitespace and comments.
 ///g
 
 REGEX_ILLEGAL = /// ^ ( / | /{3}\s*) (\*) ///
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1254,12 +1254,16 @@ VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
 HEREGEX      = /// ^
   (?:
-    # Comments consume everything until the end of the line, including `///`.
-    (?:\s+|^)(?:#(?!\{)[^\n]*)? |
-    [^\\/#]                     |
-    \\[\s\S]                    |
-    /(?!//)                     |
-    \#(?!\{)
+      # Skip `\`, `/`, `#` and `\s` characters.
+      [^\\/#\s]
+      # Match `\` followed by any character.
+    | \\[\s\S]
+      # Match any `/` except `///`.
+    | /(?!//)
+      # Match `#` which is not part of interpolation, e.g. `#{}`.
+    | \#(?!\{)
+      # Comments consume everything until the end of the line, including `///`.
+    | \s+(?:#(?!\{)[^\n]*)?
   )*
 ///
 
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1216,9 +1216,9 @@ HERE_JSTOKEN = ///^ ```     ((?: [^`\\] | \\[\s\S] | `(?!``) )*) ``` ///
 STRING_START   = /^(?:'''|"""|'|")/
 
 STRING_SINGLE  = /// ^(?: [^\\']  | \\[\s\S]                      )* ///
-STRING_DOUBLE  = /// ^(?: [^\\"\#] | \\[\s\S] |           \#(?!\{) )* ///
+STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] |           \#(?!\{) )* ///
 HEREDOC_SINGLE = /// ^(?: [^\\']  | \\[\s\S] | '(?!'')            )* ///
-HEREDOC_DOUBLE = /// ^(?: [^\\"\#] | \\[\s\S] | "(?!"") | \#(?!\{) )* ///
+HEREDOC_DOUBLE = /// ^(?: [^\\"#] | \\[\s\S] | "(?!"") | \#(?!\{) )* ///
 
 INSIDE_CSX = /// ^(?:
     [^
@@ -1254,19 +1254,19 @@ VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
 HEREGEX      = /// ^
   (?:
-      [^\\/#]   |
-      \\[\s\S]  |
-      /(?!//)   |
-      \\\#      | # Escaped hash.
-      # The comment should consume everything until the end of the line, including the '///'.
-      \#(?!\{)[^\n]*\n
+    # Comments consume everything until the end of the line, including `///`.
+    (?:\s+|^)(?:#(?!\{)[^\n]*)? |
+    [^\\/#]                     |
+    \\[\s\S]                    |
+    /(?!//)                     |
+    \#(?!\{)
   )*
 ///
 
 HEREGEX_OMIT = ///
     ((?:\\\\)+)     # Consume (and preserve) an even number of backslashes.
   | \\(\s)          # Preserve escaped whitespace.
-  | \s+(?:\#.*)?     # Remove whitespace and comments.
+  | \s+(?:#.*)?     # Remove whitespace and comments.
 ///g
 
 REGEX_ILLEGAL = /// ^ ( / | /{3}\s*) (\*) ///
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1252,7 +1252,15 @@ REGEX = /// ^
 REGEX_FLAGS  = /^\w*/
 VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
-HEREGEX      = /// ^(?: [^\\/#] | \\[\s\S] | /(?!//) | \#(?!\{) )* ///
+HEREGEX      = /// ^
+      (?:
+        [^\\/#]   |
+        \\[\s\S]  |
+        # The comment should consume everything until the end of the line, including the '///'
+        \#(?!\{)(?:[^\n\r]*[\n\r])  |
+        /(?!//)
+      )*
+///
 
 HEREGEX_OMIT = ///
     ((?:\\\\)+)     # Consume (and preserve) an even number of backslashes.
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1263,7 +1263,7 @@ HEREGEX      = /// ^
       # Match `#` which is not part of interpolation, e.g. `#{}`.
     | \#(?!\{)
       # Comments consume everything until the end of the line, including `///`.
-    | \s+(?:#(?!\{)[^\n]*)?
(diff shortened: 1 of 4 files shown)


src/lexer.coffee
```diff
@@ -1252,7 +1252,15 @@ REGEX = /// ^
 REGEX_FLAGS  = /^\w*/
 VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
-HEREGEX      = /// ^(?: [^\\/#] | \\[\s\S] | /(?!//) | \#(?!\{) )* ///
+HEREGEX      = /// ^
+      (?:
+        [^\\/#]   |
+        \\[\s\S]  |
+        # The comment should consume everything until the end of the line, including the '///'
+        \#(?!\{)(?:[^\n\r]*[\n\r])  |
```
lydell: As far as I remember, the lexer starts by normalizing all newlines, so only checking for `\n` should be enough.

https://github.com/jashkenas/coffeescript/blob/04959162c64d7e4212ffcb90026aec5a5b059ee2/src/lexer.coffee#L94

src/lexer.coffee
```diff
@@ -1252,7 +1252,15 @@ REGEX = /// ^
...
+        /(?!//)
```
lydell: Any particular reason you changed the order, moving this case last?

zdenko: Actually no. It's a leftover from playing with the RegExp rules.

src/lexer.coffee
zdenko: :thumbsup:

src/lexer.coffee
```diff
@@ -1252,7 +1252,15 @@ REGEX = /// ^
...
 VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
-HEREGEX      = /// ^(?: [^\\/#] | \\[\s\S] | /(?!//) | \#(?!\{) )* ///
+HEREGEX      = /// ^
+      (?:
+        [^\\/#]   |
+        \\[\s\S]  |
+        /(?!//)   |
+        # The comment should consume everything until the end of the line, including the '///'
+        \#(?!\{)(?:[^\n]*[\n])
```
lydell: Sorry, one last little thing: The non-capturing group doesn’t do anything here, does it?

zdenko: I think it doesn't. But I found another bug which I missed before. 
Capturing comment is too aggressive, and following valid RegExp doesn't compile.
```coffeescript 
/// ^(?: [^\\"#] | \\[\s\S] |           \#(?)* ///
```

lydell: If it doesn’t do anything, I’d say remove it.

Did you find a new bug introduced by this PR or separately?

The regex you posted looks invalid to me: `(?)` is not a valid part of a regex. But if you are talking about the hash sign inside `[^\\"#]`, here is my analysis:

A hash sign _always_ means a comment inside a heregex (unless it is escaped by a backslash), regardless of _where_ it is. They are allowed in character classes too. So you need to put a backslash in front of the `#` in this case.

zdenko: Sorry, my mistake. This is the correct code. It's from `lexer` (not part of this PR).
```coffeescript
STRING_DOUBLE  =  /// ^(?: [^\\"#] | \\[\s\S] |           \#(?!\{) )* ///
```
I think this bug is "double". It works because of bug you opened #4811 (but it is still a bug).
With this PR the whole part after the hash inside bracets is now consumed.
`#] | \\[\s\S] |           \#(?!\{) )* ///`.
So, hash inside heregex will be treated as a comment unless it is escaped.
I'm fixing this.

src/lexer.coffee
```diff
@@ -1253,19 +1253,20 @@ REGEX_FLAGS  = /^\w*/
...
+      # The comment should consume everything until the end of the line, including the '///'.
+      \#(?!\{)[^\n]*\n
+  )*
 ///
 
 HEREGEX_OMIT = ///
     ((?:\\\\)+)     # Consume (and preserve) an even number of backslashes.
   | \\(\s)          # Preserve escaped whitespace.
-  | \s+(?:#.*)?     # Remove whitespace and comments.
+  | \s+(?:\#.*)?     # Remove whitespace and comments.
```
lydell: Ah, I had forgotten about this. Comments in heregex must be preceded by at least one whitespace. I think the correct fix is to revert this commit and adjust HEREGEX.

src/lexer.coffee
zdenko: @lydell I think this is it.
Only unescaped hash is treated as a comment. 
I also corrected some invalid heregex patterns in `lexer.coffee`.
```coffeescript
HEREGEX      = /// ^
  (?:
      [^\\/#]   |
      \\[\s\S]  |
      /(?!//)   |
      \\\#      | # Escaped hash.
      # The comment should consume everything until the end of the line, including the '///'.
      \#(?!\{)[^\n]*\n
  )*
///
```

src/lexer.coffee
lydell: Something like this:

```coffee
HEREGEX = /// ^
  (?:
    [^\\/#\s] |
    \\[\s\S]  |
    /(?!//)   |
    \#(?!\{)  |
    # Comments consume everything until the end of the line, including `///`.
    \s+(?:#(?!\{)[^\n]*[\n])? |
  )*
///
```

zdenko: So, hash preceded with the whitespace is a comment, e.g. no need to be escaped?

zdenko: I'm not sure about the whitespace. Is this valid heregex?
```coffeescript
/// ^(?: [^\\"\#] | \\[\s\S] |# (?!\{) ///
```

lydell: Yes, I’d say that’s a valid regex.

`[$#]` – literal hash sign, because no space before.
`start#end` – same thing here.
`start #end` – this is a comment, because there’s a space before
`start \#end` – literal hash sign, because the hash is escaped.

This one should work and be treated as a comment since the hash is preceded by a newline (which is a whitespace character):

```coffee
///
[a-z]
# comment
\s+
```

However, my proposed regex currently doesn’t handle this case (hash sign preceded by nothing):

```coffee
///# comment
[a-z]///
```

lydell: Something like this should handle the “comment at the very beginning” case:

```coffee
HEREGEX = /// ^
  (?:
    [^\\/#\s] |
    \\[\s\S]  |
    /(?!//)   |
    \#(?!\{)  |
    # Comments consume everything until the end of the line, including `///`.
    (?:\s+|^) (?: #(?!\{) [^\n]*[\n] )? |
  )*
///
```

lydell: Also, if you look at the line I commented on you can see that I’m lot making this up. `REGEX_OMIT` only strips comments if preceded by a space.

lydell: Hmm, since `REGEX_OMIT` requires a space before comments, I don’t think comments at the very beginning of a heregex has ever worked. So perhaps no need to worry about that. Or should we consider that a bug and fix that as well?

zdenko: It seems that order of the patterns matters. I guess it's some kind of "first come, first served" rule. The pattern for heregex comment has to be placed as the first option.
And the `[\n]` is also removed from the pattern.
```coffeescript
HEREGEX      = /// ^
  (?:
    # Comments consume everything until the end of the line, including `///`.
    (?:\s+|^)(?:#(?!\{)[^\n]*)? |
    [^\\/#]                     |
    \\[\s\S]                    |
    /(?!//)                     |
    \#(?!\{)
  )*
///
```

zdenko: I guess to cover the case when hash starts immediately after the `///`, we also have to adjust `REGEX_OMIT`.

zdenko: Correcting myself:
> The pattern for heregex comment has to be placed as the first option.

The position of heregex comment pattern can be anywhere as long as it is before this pattern `[^\\/#] ` which will match `\s` required by comment pattern.

> [\n] is also removed from the pattern.

Since we're matching `[^\n]*` (everything not newline) there is no need to specify `[\n]` explicitly at the end.

So, here is the improved (I hope) `HEREGEX`.
```coffeescript
HEREGEX      = /// ^
  (?:
      # Skip `\`, `/`, `#` and `\s` characters.
      [^\\/#\s]
      # Match `\` followed by any character.
    | \\[\s\S]
      # Match any `/` except `///`.
    | /(?!//)
      # Match `#` which is not part of interpolation, e.g. `#{}`.
    | \#(?!\{)
      # Comments consume everything until the end of the line, including `///`.
    | (?:\s+)(?:#(?!\{)[^\n]*)?
  )*
///
```

src/lexer.coffee
```diff
@@ -1252,7 +1252,20 @@ REGEX = /// ^
...
+      # Match `\` followed by any character.
+    | \\[\s\S]
+      # Match any `/` except `///`.
+    | /(?!//)
+      # Match `#` which is not part of interpolation, e.g. `#{}`.
+    | \#(?!\{)
+      # Comments consume everything until the end of the line, including `///`.
+    | \s+(?:#(?!\{)[^\n]*)?
+  )*
+///
```
GeoffreyBooth: _Thank you_ for commenting this. One style note: shouldn’t the `|` line continuers be at the end of the line, rather than the start of the next one? That’s how we have things elsewhere in the codebase, e.g. `if firstCondition and \n secondCondition`.

@zdenko and @lydell, why does a heregex require a space before a `#` comment? That seems to go against the general rules for comments.

zdenko: > shouldn’t the | line continuers be at the end of the line, rather than the start of the next one

I kinda followed the style from other heregex patterns in `lexer.coffee`   
Examples:
```coffeescript
HEREGEX_OMIT = ///
    ((?:\\\\)+)     # Consume (and preserve) an even number of backslashes.
  | \\(\s)          # Preserve escaped whitespace.
  | \s+(?:#.*)?     # Remove whitespace and comments.
///g

REGEX_INVALID_ESCAPE = ///
  ( (?:^|[^\\]) (?:\\\\)* )        # Make sure the escape isn’t escaped.
  \\ (
     ?: (0[0-7])                   # octal escape
      | (x(?![\da-fA-F]{2}).{0,2}) # hex escape
      | (u\{(?![\da-fA-F]{1,}\})[^}]*\}?) # unicode code point escape
      | (u(?!\{|[\da-fA-F]{4}).{0,4}) # unicode escape
  )
///
```

> why does a heregex require a space before a # comment?

`HEREGEX_OMIT`, which removes comments from the heregex, requires a space before `#`   
`\s+(?:#.*)?`.
We could add the rule for escaping `#` unless it's a comment, which will allow comments without the preceding space.
But, there are many valid heregex patterns without escaped `#`, even in the `lexer.coffee`:  
`STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] | \#(?!\{) )* ///`, e.g. first `#` in `[^\\"#]`
The same heregex can be changed to 
`STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] |#(?!\{) )* ///` and it would still be valid:

```coffeescript
STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] | \#(?!\{) )* ///
# STRING_DOUBLE = /^(?:[^\\"#]|\\[\s\S]|\#(?!\{))*/;

STRING_DOUBLE  = /// ^(?: [^\\"#] | \\[\s\S] |#(?!\{) )* ///
# STRING_DOUBLE = /^(?:[^\\"#]|\\[\s\S]|#(?!\{))*/;

```

So, these existing patterns wouldn't be valid anymore if we allow heregex comments without preceding space.

lydell: > @zdenko and @lydell, why does a heregex require a space before a # comment? That seems to go against the general rules for comments.

What @zdenko said. In other words: Because that’s the way it’s always been. Changing that now would be a breaking change. Even CoffeeScript’s own source code would need to be updated! And it’s easy to see the original thinking: Using literal spaces in regexes is uncommon. This means that you can copy-paste regex and paste into your heregex without having to think about escaping any `#`s.

src/lexer.coffee
```diff
@@ -1252,7 +1252,20 @@ REGEX = /// ^
 REGEX_FLAGS  = /^\w*/
 VALID_FLAGS  = /^(?!.*(.).*\1)[imguy]*$/
 
-HEREGEX      = /// ^(?: [^\\/#] | \\[\s\S] | /(?!//) | \#(?!\{) )* ///
+HEREGEX      = /// ^
+  (?:
+      # Skip `\`, `/`, `#` and `\s` characters.
```
lydell: “Match any character, except those that need special handling below.”?

test/error_messages.coffee
```diff
@@ -1803,3 +1803,12 @@ test "#3098: suppressed newline should be unsuppressed by semicolon", ->
     a = ; 5
         ^
   '''
+
+test "#4811, heregex comments cannot contain three slashes in a row", ->
```
lydell: “/// inside a heregex comment does not close the heregex”?

src/lexer.coffee
```diff
@@ -1252,7 +1252,20 @@ REGEX = /// ^
...
+      # Match any character, except those that need special handling below.
+      [^\\/#\s]
+      # Match `\` followed by any character.
+    | \\[\s\S]
+      # Match any `/` except `///`.
+    | /(?!//)
+      # Match `#` which is not part of interpolation, e.g. `#{}`.
+    | \#(?!\{)
+      # Comments consume everything until the end of the line, including `///`.
+    | \s+(?:#(?!\{)[^\n]*)?
```
lydell: It could be argued that HEREGEX and HEREGEX_OMIT should be consistent. The former uses `[^\n]*` while the latter uses `.*`. What do you think?

zdenko: Yes, It's more consistent. `[^\n]*` had it's purpose when when `[^\n]*\n` pattern was used. But since `\n` is already consumed by `[^\\/#\s]`, we can change pattern for comment to    
`\s+(?:#(?!\{).*)?`.

lydell:
Awesome work @zdenko! I’ll leave it up to to @GeoffreyBooth to merge.

Note that there’s a small risk that people have code like this:

```coffee
///[1-9] |
   0(?!0) # zero must not be followed by zero///
```

... which won’t compile. It _is_ a bugfix, but potentially breaking. Versioning is hard.

GeoffreyBooth:
> … which won’t compile. It _is_ a bugfix, but potentially breaking. Versioning is hard.

As I understand versioning, a breaking change isn’t anytime _anything_ changes, it’s any time the externally-facing API changes. In our case, it feels like the externally-facing API is vast, as it’s all the syntax of all the code anyone could write. But I would argue that it’s slightly smaller than that: it’s the syntax _defined in the documentation._ The docs clearly say that from `#` to the end of a line is a comment and ignored; the docs don’t say that `#` to the end of a line, except for `///` within a heregex, is ignored. So this is a conflict between the behavior of the compiler and the syntax as defined per the docs (which are really our spec, since we have no formal one). For such conflicts, the docs should prevail, unless the change would be so drastic as to affect lots of old code (as you’re arguing that fixing the requirement for a space before `#` in a heregex would be). We had a similar case when we fixed chaining after a function call between 1.12.6 and 1.12.7. People complained, but we pointed out how the prior behavior was incorrect per the docs, and the world moved on.

That was at least a tougher case as both variants compiled. This is much easier: something that currently compiles no longer will. Anyone with `#...///` in their code will suddenly get a compiler error. So there’s no risk here that this change does harm to any apps. It’ll just annoy anyone who has to fix their syntax.

I think the next 2.x release should be 2.1.0, as we’re adding new features such as elision. So that’s another clue to people that the release might have more changes contained in it.

zdenko:
>  It is a bugfix, but potentially breaking. Versioning is hard.

I was thinking the same. It would be nice if the error message is more descriptive and thus less ambiguous and more helpful.
So, instead of this   
```
[stdin]:1:1: error: missing ///
///[1-9] |
^^^
```
... we could show something like this (if possible)
```
[stdin]:2:10: error: '///' inside a heregex comment does not close the heregex
   0(?!0) # zero must not be followed by zero ///
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I guess this belongs to a separate PR now.

@lydell @GeoffreyBooth

lydell:
Nice, then we’ve thought this through before releasing. 👍 

I don’t think it’s worth doing anything about the error message at this point.

PR #4818, opened 2017-12-03
[Enhancement] Improve heregex comment error message

zdenko:
A small enhancement to provide a more descriptive error message when heregex comment ends with `///`.

```coffeescript
///[1-9] |
   0(?!0) # zero must not be followed by zero///
```

Before:
```
[stdin]:1:1: error: missing ///
///[1-9] |
^^^
```

After:
```
[stdin]:2:10: error: '///' inside a heregex comment does not close the heregex
   0(?!0) # zero must not be followed by zero ///
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -1005,7 +1005,7 @@
     // This method allows us to have strings within interpolations within strings,
     // ad infinitum.
     matchWithInterpolations(regex, delimiter, closingDelimiter, interpolators) {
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
I'm not sure this is necessary. I think it's already pretty obvious why it breaks.

lydell:
I agree with @GeoffreyBooth.

Issue #4517, opened 2017-04-18
Nodes: Optimization: Destructuring with non-final spread should still use rest syntax


isiahmeadows:
Example:

```coffee
tag = "<impossible>"

[open, contents..., close] = tag.split("")
```

```js
// Recommended output:
var close, contents, open, tag, pop = [].pop;

tag = "<impossible>";

[open, ...contents] = tag.split("");
close = pop.call(contents);
```

zdenko:
This definitely looks cleaner.
But, I think, using `array.pop` will work only until the number of destructured variables (without the splat) is the same as the length of the array.

Example:

```
# CS:
[a, rest…, b, c, d] = ["a", "b", "c", "d"]

# JS:
[a, …rest] = ["a", "b", "c", "d"]
d = pop.call(rest)
c = pop.call(rest)
b = pop.call(rest)

# Result:
a = "a"
rest = []
d = "d"
c = "c"
b = "b"
```

However, in this case the result will not be as expected:
```
# CS:
[a, rest…, b, c, d] = ["a", "b"]

# JS
…

# Result:
a = "a"
rest = []
d = "b"
c = undefined
b = undefined
```

Perhaps using something like `array.splice` would be a solution:

```
# CS: 
# splice = [].splice
# [a, rest…, b, c, d] = ["a", "b"]

# JS:
[a, …rest] = ["a", "b"]
[b, c, d] = (3 <= rest.length) ? splice.call(rest, rest.length - 3, rest.length) : splice.call(rest, 0);

# Result:
a = "a"
rest = []
b = "b"
c = undefined
d = undefined
```

isiahmeadows:
Maybe, this, then?

```js
# [a, rest…, b, c, d] = ["a", "b"]

[a, …rest] = ["a", "b", "c", "d"];
_i = rest.length - 3; if (i < 0) i = 0;
if (_i !== rest.length) b = rest[_i++];
if (_i !== rest.length) c = rest[_i++];
if (_i !== rest.length) d = rest[_i++];
rest.length = _i;
```

That should avoid the issue without resorting to a method call.

zdenko:
Method calls are not the problem. It's just that `array.pop` method removes the **last** element from an array.

Your last solution would also work. There are few points which need correction:
1. A small typo: `_i` instead `i`
```coffeescript
 _i = rest.length - 3; if (_i < 0) _i = 0; 
```
Which can be also shortened:
```coffeescript
[...]

PR #4824, opened 2017-12-09
[Enhancement] improve Elision output

zdenko:
This PR improves compiled code in `Assign::compileDestructuring`.

Current output:
```javascript
// CS: [a, , , {b, c...}, d] = arr

a = arr[0], 1, 2, (({b} = ref = arr[3]), c = objectWithoutKeys(ref, ['b'])), d = arr[4];
            ^  ^
```
`1` and `2` are values of the `idx` and can be skipped in cases when there are no `Splat` or `Expansion` in the destructured array.

Improved output:
```javascript
// CS: [a, , , {b, c...}, d] = arr

a = arr[0], (({b} = ref = arr[3]), c = objectWithoutKeys(ref, ['b'])), d = arr[4];
```

### DIFF
--- a/test/arrays.coffee
+++ b/test/arrays.coffee
@@ -74,8 +74,8 @@ test "array elisions destructuring with splats and expansions", ->
   arrayEq [a,b], [2,[5,6,7,8,9]]
   [,c,...,,d,,e] = arr
   arrayEq [c,d,e], [2,7,9]
-  [...,e,,,f,,,] = arr
-  arrayEq [e,f], [4,7]
+  [...,f,,,g,,,] = arr
+  arrayEq [f,g], [4,7]
 
 test "array elisions as function parameters", ->
   arr = [1,2,3,4,5,6,7,8,9]
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -3626,7 +3626,10 @@
               subpattern: true
             }).compileToFragments(o, LEVEL_LIST));
           } else {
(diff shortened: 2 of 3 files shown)


test/arrays.coffee
```diff
@@ -74,8 +74,8 @@ test "array elisions destructuring with splats and expansions", ->
   arrayEq [a,b], [2,[5,6,7,8,9]]
   [,c,...,,d,,e] = arr
   arrayEq [c,d,e], [2,7,9]
-  [...,e,,,f,,,] = arr
-  arrayEq [e,f], [4,7]
+  [...,f,,,g,,,] = arr
+  arrayEq [f,g], [4,7]
```
GeoffreyBooth: Why did this change?

zdenko: This test threw some errors, so I changed variables, as `e` was used in the line above, and I thought this was the reason (it wasn't).
I can change it back as it's not needed for this PR.

GeoffreyBooth: It's better not to reuse variables. If it passes either way, let's keep the non-reused version.

GeoffreyBooth:
This looks good to me. @vendethiel or anyone else have notes?

@zdenko we should include this in 2.1.0. Are you working on anything else we should wait for?

zdenko:
@GeoffreyBooth I'm working on PR for #4517, but I don't think it's something that needs to be included in 2.1.0.
It will take another day or two, and besides the changes in compiled outup it won't bring any new feature.
Also, the changes in the `nodes.coffee`, could spawn longer discusion and thus prolong the release.

PR #4825, opened 2017-12-11
[Change]: Destructuring with non-final spread should still use rest syntax (#4517)

zdenko:
This PR simplifies compiled output for destructuring assignments (#4517).

```javascript
 // [a, b, c..., d, e] = arr
var a, b, c, d, e, splice = [].splice;
[a, b, ...c] = arr, [d, e] = splice.call(c, -2);

 // [a, b, ..., d, e] = arr
var a, b, d, e, slice = [].slice;
[a, b] = arr, [d, e] = slice.call(arr, -2);

// [a, , ,b, c..., d, ,e, {x, r...}, f, , ,] = arr
var a, b, c, d   ...
[a, , , b, ...c] = arr, ref = splice.call(c, -7), 
d = ref[0], e = ref[2], (({x} = ref1 = ref[3]), 
r = objectWithoutKeys(ref1, ['x'])), f = ref[4];

// [a, , ,{x, r...}, b, ..., c, d, ,e ,f, , ,] = arr
var a, b, c, d   ...
a = arr[0], (({x} = ref = arr[3]), r = objectWithoutKeys(ref, ['x'])), 
b = arr[4], [c, d, , e, f, , , ] = slice.call(arr, -7);
```

Same examples as above compiled on `master`:
```javascript
 // [a, b, c..., d, e] = arr
var a, b, c, d, e, i, slice = [].slice;
a = arr[0], b = arr[1], 
c = 5 <= arr.length ? slice.call(arr, 2, i = arr.length - 2) : (i = 2, []), 
d = arr[i++], e = arr[i++];

 // [a, b, ..., d, e] = arr
var a, b, d, e, i;
a = arr[0], b = arr[1], i = arr.length - 2, d = arr[i++], e = arr[i++];

// [a, , ,b, c..., d, ,e, {x, r...}, f, , ,] = arr
var a, b, c, d   ...
a = arr[0], b = arr[3], 
c = 12 <= arr.length ? slice.call(arr, 4, i = arr.length - 7) : (i = 4, []), 
d = arr[i++], i++, e = arr[i++], 
(({x} = ref = arr[i++]), r = objectWithoutKeys(ref, ['x'])), 
f = arr[i++], i++, i++;

// [a, , ,{x, r...}, b, ..., c, d, ,e ,f, , ,] = arr
var a, b, c, d   ...
a = arr[0], (({x} = ref = arr[3]), 
r = objectWithoutKeys(ref, ['x'])), b = arr[4], 
i = arr.length - 7, c = arr[i++], d = arr[i++], 
i++, e = arr[i++], f = arr[i++], i++, i++;
```


src/nodes.coffee
```diff
@@ -2402,79 +2380,106 @@ exports.Assign = class Assign extends Base
...
-        defaultValue = undefined
+    # Helper which outputs `[].slice` code.
+    compSlice = (vvar, start, end = no) ->
+      endVal = if end then ", #{end}" else ""
+      new Value new Literal "#{utility 'slice', o}.call(#{vvar}, #{start}#{endVal})"
+
+    # Helper which outputs `[].splice` code.
+    compSplice = (vvar, start, end = no) ->
+      endVal = if end then ", #{end}" else ""
+      new Value new Literal "#{utility 'splice', o}.call(#{vvar}, #{start}#{endVal})"
```
GeoffreyBooth: We should avoid using `new Literal` when we can. Can you rewrite this to make each node individually, e.g. `new Call` etc.? This helps in generating a more explicit AST, and for things like comments attached to nodes.

GeoffreyBooth:
Thanks. Anyone else have any notes? @connec or @vendethiel ?

src/nodes.coffee
```diff
@@ -2402,79 +2380,108 @@ exports.Assign = class Assign extends Base
...
+    # Examples:
+    # a) `Splat`
+    #   CS: [a, b, c..., d, e] = arr
+    #   JS: [a, b, ...c] = arr, [d, e] = splice.call(c, -2)
+    # b) `Expansion`
+    #   CS: [a, b, ..., d, e] = arr
+    #   JS: [a, b] = arr, [d, e] = slice.call(arr, -2)
+    if splatsAndExpans.length
+      expIdx = splatsAndExpans[0]
+      leftObjs = objects.slice 0, expIdx + (if isSplat then 1 else 0)
```
vendethiel: Care to comment this one?

src/nodes.coffee
```diff
@@ -2351,47 +2351,25 @@ exports.Assign = class Assign extends Base
     if olen is 1 and obj instanceof Expansion
       obj.error 'Destructuring assignment has no target'
 
+    # Count all `Splats`: [a, b, c..., d, e]
+    splats = (i for obj, i in objects when obj instanceof Splat)
+    # Count all `Expansions`: [a, b, ..., c, d]
+    expans = (i for obj, i in objects when obj instanceof Expansion)
+    # Combine splats and expansions and sort them so we can show error if needed.
+    splatsAndExpans = [splats..., expans...].sort()
```
vendethiel: Why sort?

zdenko: In case there are more than one `Splat` and/or `Expansion` I use `sort` so I can show the first error.
So, presume this example: `[a, b..., ..., c..., d, e]`
FIrst I collect indexes of all splats (`splats = [1, 3]`), and then expansions (`expans = [2]`).
First splat (`b...`, index 1) is valid, and I want to show the next error (`...`, index 2).
`[splats..., expans...].sort()` will put them in order.

src/nodes.coffee
zdenko: I guess you mean this one `leftObjs = objects.slice 0, expIdx + (if isSplat then 1 else 0)`.
If `Splat`or `Expansion` is present, we can split the array in half and process each of them independently.
Since `Splat` is ES6 valid, we can iincluded it in the compiled code, hence   
`objects.slice 0, expIdx + 1`

```javascript
// [a, b, c..., d, e] = arr
[a, b, ...c] = arr, [d, e] = splice.call(c, -2)
```

`Expansion` is not included in the output, so `objects.slice 0, expIdx`

```javascript
// [a, b, c, ..., d, e] = arr
[a, b, c] = arr, [d, e] = slice.call(arr, -2);
```

GeoffreyBooth:
@connec Did you have any notes on this? I’d like to merge it in if it’s ready.

src/nodes.coffee
GeoffreyBooth: This could be:

```coffee
if splats.length isnt 0 and expans.length isnt 0
  # Combine and sort here, then throw error
```

No need to sort if we’re not throwing an error.

zdenko: Nice. I moved `sort()` after the condition.

GeoffreyBooth:
@vendethiel or @connec, is this okay to merge in?

src/nodes.coffee
```diff
@@ -2402,79 +2381,108 @@ exports.Assign = class Assign extends Base
...
-            assigns.push val.compileToFragments o, LEVEL_LIST
-        continue
-      else
-        if obj instanceof Splat or obj instanceof Expansion
-          obj.error "multiple splats/expansions are disallowed in an assignment"
-        defaultValue = undefined
+    slicer = (type) -> (vvar, start, end = no) ->
+      args = [new IdentifierLiteral(vvar), new NumberLiteral(start)]
+      args.push new NumberLiteral end if end
+      slice = new Value new Literal "#{utility type, o}.call"
```
GeoffreyBooth: Let’s avoid `new Literal`, it messes up the nodes that get output. You can achieve the same via something like `new Value (utility type, o), [new Access new PropertyName 'call']`

src/nodes.coffee
```diff
@@ -2402,79 +2381,108 @@ exports.Assign = class Assign extends Base
...
+    #   JS: [a, b, ...c] = arr, [d, e] = splice.call(c, -2)
+    # b) `Expansion`
+    #   CS: [a, b, ..., d, e] = arr
+    #   JS: [a, b] = arr, [d, e] = slice.call(arr, -2)
+    if splatsAndExpans.length
+      expIdx = splatsAndExpans[0]
+      leftObjs = objects.slice 0, expIdx + (if isSplat then 1 else 0)
+      rightObjs = objects.slice expIdx + 1
+      processObjects leftObjs, vvarText if leftObjs.length
+      if rightObjs.length
```
GeoffreyBooth: This is a bit of a pet peeve, but I don’t like `if a.length` as a way of really saying `if a.length isnt 0`. I prefer the latter, because it makes explicit that you’re checking that `length` is nonzero, not that a property name `length` exists. Relying on the fact that when `length` is 0, that 0 evaluates as falsy feels to me like it goes against the “always `===`” principle that is in vogue nowadays and that CoffeeScript enforces with `is` or `==` both equaling `===`.

GeoffreyBooth:
@zdenko The new output is quite an improvement in readability over current `master`, and greatly simplifies the hairiest part of the codebase. Thanks for taking this on!

Since this is a rather significant rewrite, it would be good to run this on some codebases to ensure that there aren’t any cases that aren’t caught by our tests. @connec, does your YAML library or any other library make use of splats or expansions, that could be tested with this branch?

src/nodes.coffee
zdenko: 👍

PR #4826, opened 2017-12-15
Use free variables for anonymous classes

### LINKED ISSUE #4822: Nodes: Nested anonymous class reuses generated variable
CoffeeScript 2.0.3 compiles this:
```
class
 a: class
   b: 1
```

into this:

```
(function() {
  var _Class;

  _Class = class {};

  _Class.prototype.a = (function() {
    _Class = class {};

    _Class.prototype.b = 1;

    return _Class;

  })();

  return _Class;

})();
```

Nested `_Class` should be defined as local variable with `var`.

connec:
Fixes #4822.


GeoffreyBooth:
@connec Can you resolve the conflicts with this, then I’ll merge it in and release a version tonight.

PR #4828, opened 2017-12-15
Set the context of executable class body wrappers

### LINKED ISSUE #4827: Nodes: executable class body wrappers have incorrect context
When defining a class with an executable body whose superclass is an `@`-value the generated JS is incorrect:

[Example:](http://coffeescript.org/#try:class%20%40A%0Aclass%20%40B%20extends%20%40A%0A%20%20%40property%20%3D%201)

> ```coffee
> class @A
> class @B extends @A
>   @property = 1
> ```

Becomes:

> ```js
> this.A = class A {};
>
> this.B = (function() {
>   class B extends this.A {};
>
>   B.property = 1;
>
>   return B;
>
> })();

Notice that `this` in `this.A` will be the global object (or null) as the wrapper function is not given a context.

connec:
Fixes #4827.

I did first try to use an arrow function by setting `bound = true` on the wrapper, but this messed with the handling of `@` *within* the class body, where it should refer to the class being defined.

We could opt to perform the binding only if the class `extends` an `@`-value but that would add a bit of extra complexity in the compiler for equivalent and, imo, no less readable output.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -543,7 +543,7 @@
 
     return Base;
 
-  })();
+  }).call(this);
 
   //### HoistTarget
 
@@ -997,7 +997,7 @@
 
     return Block;
 
-  })();
+  }).call(this);
 
   //### Literal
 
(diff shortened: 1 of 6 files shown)


GeoffreyBooth:
LGTM.

connec:
Any other notes? I would quite like to get this and #4826 into a (patch) release.

PR #4848, opened 2018-01-07
Fix #4843: bad output when assigning to @prop in destructuring assignment with defaults

### LINKED ISSUE #4843: Nodes: Bad output when assigning to @prop in destructuring assignment with defaults
illustrated [here](http://coffeescript.org/v2/#try:works%20%3D%20%22maybe%22%0A%0Aclass%20A%0A%20%20constructor%3A%20(%7B%40works%20%3D%20%22no%22%7D)%20-%3E%0A%20%20%20%20alert%20%22provided%3A%20%23%7Barguments%5B0%5D.works%7D%2C%20handled%3A%20%23%7B%40works%7D%22%0A%0Anew%20A%20works%3A%20%22yes%22).
input:
```coffee
works = "maybe"
class A
  constructor: ({@works = "no"}) ->
```
current output:
```js
var A, works;

works = "maybe";

A = class A {
  constructor({works1 = "no"}) {
    this.works = works1;
  }

};
```
expected output:
```js
var A, works;

works = "maybe";

A = class A {
  constructor({works: works1 = "no"}) {
    this.works = works1;
  }

};
```

zdenko:
Fixes #4843.

```coffeescript 
works = "maybe"
drinks = "beer"
class A
  constructor: ({@works = 'no', @drinks = 'wine'}) ->

a = new A {works: 'yes', drinks: 'coffee'}
```
Ouput

```javascript
var A, a, drinks, works;

works = maybe;
drinks = beer;

A = class A {
  constructor({works: works1 = 'no', drinks: drinks1 = 'wine'}) {
    this.works = works1;
    this.drinks = drinks1;
  }

};

a = new A({
  works: 'yes',
  drinks: 'coffee'
});
```

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2802,7 +2802,7 @@ exports.Code = class Code extends Base
     if @front or (o.level >= LEVEL_ACCESS) then @wrapInParentheses answer else answer
 
   eachParamName: (iterator) ->
-    param.eachName iterator for param, i in @params
+    param.eachName iterator for param in @params
 
   # Short-circuit `traverseChildren` method to prevent it from crossing scope
   # boundaries unless `crossScope` is `true`.
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2903,7 +2903,7 @@ exports.Param = class Param extends Base
   # `name` is the name of the parameter and `node` is the AST node corresponding
   # to that name.
   eachName: (iterator, name = @name) ->
-    atParam = (obj) => iterator "@#{obj.properties[0].name.value}", obj, @, nObj
+    atParam = (obj, originalObj = null) => iterator "@#{obj.properties[0].name.value}", obj, @, originalObj
     # * simple literals `foo`
     return iterator name.value, name, @ if name instanceof Literal
     # * at-params `@foo`
@@ -2932,7 +2932,7 @@ exports.Param = class Param extends Base
           @eachName iterator, obj.base
         # * at-params within destructured parameters `{@foo}`
         else if obj.this
-          atParam obj
+          atParam obj, nObj
         # * simple destructured parameters {foo}
         else iterator obj.base.value, obj.base, @
       else if obj instanceof Elision
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2597,25 +2597,24 @@ exports.Code = class Code extends Base
 
     # Check for duplicate parameters and separate `this` assignments.
     paramNames = []
-    @eachParamName (name, node, param) ->
+    @eachParamName (name, node, param, obj) ->
       node.error "multiple parameters named '#{name}'" if name in paramNames
       paramNames.push name
+
       if node.this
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -2904,12 +2903,14 @@ exports.Param = class Param extends Base
...
   eachName: (iterator, name = @name) ->
-    atParam = (obj) => iterator "@#{obj.properties[0].name.value}", obj, @
+    atParam = (obj) => iterator "@#{obj.properties[0].name.value}", obj, @, nObj
     # * simple literals `foo`
     return iterator name.value, name, @ if name instanceof Literal
     # * at-params `@foo`
     return atParam name if name instanceof Value
     for obj in name.objects ? []
+      # Save original obj.
+      nObj = obj
```
vendethiel: why?

zdenko: In case of object destructuring parameter with the default value, e.g. `({@works = "no", @drinks}) ->`,  when replacement function is called from `@renameParam`, the `parent` is `Assign` in case of `@works = "no"`, and `Obj` in case of `@drinks`.
Because of this `@works = "no"` is changed to `works1 = "no"` and `@drinks` to `drinks:drinks1`.

The best way, I could find, to solve this, was to pass  `Assign` (`{works:works1}`)  to  `@renameParam` instead of `Identifier` (`works1`).
But, because `Param::eachName` passes `obj.variable`  (`@works` in this case) to the iterator,   I had to find a way to detect if param has a default value set in order to create a correct replacement. Passing original `obj` seems like a way to go.
Another way would be passing index and use it to check the argument in the `Code::eachParamName` (e.g.  `param.name.parameters[i]`).

vendethiel: Ok, now I get it. I don't like that `atParam` magically references a variable from the for loop, tbh. Can you please make it explicit?

zdenko: You mean to pass it as an argument?
```coffeescript
atParam = (obj, originalObject = no) => iterator "@#{obj.properties[0].name.value}", obj, @, originalObject

....

else if obj.this
  atParam obj, nObj

```

vendethiel: Yes.

zdenko:
While working on this PR I noticed that object destructuring variables in parameters are always replaced, even if the variable isn't reserved in the scope.

```coffeescript
class A
  constructor: ({@works}) ->

###
Output
var A;

A = class A {
  constructor({
      works: works
    }) {
    this.works = works;
  }
};
###
```

I don't think this is required, so I've changed the logic and now variables are replaced only if needed.

```coffeescript
class A
  constructor: ({@works}) ->

###
A = class A {
  constructor({works}) {
    this.works = works;
  }
};
###

foo = 1
class B
  constructor: ({@foo}) ->

###
foo = 1;
B = class B {
  constructor({foo:foo1}) {
    this.foo = foo1;
  }
};
###

```

Inve1951:
gj, thx @zdenko :+1:

PR #4849, opened 2018-01-09
Fix #3441: parentheses wrapping expression throw invalid error 

### LINKED ISSUE #3441: Lexer: Parentheses wrapping expression throw invalid error in `then` clause

zdenko:
Fixes #3441.

Cases where `statement` is used inside parentheses don't compile:
```coffeescript
(a(); break) for a in obj
# error: cannot use a pure statement in an expression
# (a(); break) for a in obj
#       ^^^^^

(a(); break) while a
# error: cannot use a pure statement in an expression
# (a(); break) while a
#       ^^^^^

for a in obj then (a(); break)
# error: cannot use a pure statement in an expression
# for a in obj then (a(); break)
#                         ^^^^^

while a then (a(); break)
# error: cannot use a pure statement in an expression
# while a then (a(); break)
#                    ^^^^^
```

With this PR:
```coffeescript

# (a(); break) for a in obj
# for a in obj then (a(); break)
for (i = 0, len = obj.length; i < len; i++) {
  a = obj[i];
  a();
  break;
}

# (a(); break) while a
# while a then (a(); break)
while (a) {
  a();
  break;
}
```

When `Parens` block includes a `StatementLiteral` (e.g. `(b; break) for a in arr`), it won't compile since `Parens` (`(b; break)`) is compiled as `Value` and pure statement (`break`) can't be used in an expression.  

This PR checks `Parenthetical` rule in the grammar and returns `Block` instead of `Parens`  if `StatementLiteral` is detected. The `Block` node is passed further to the `Expression`.

### DIFF
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -334,7 +334,7 @@ grammar =
     o 'Assignable'
     o 'Literal',                                -> new Value $1
     o 'Parenthetical',                          ->
-        if $1.constructor.name is 'Parens'
+        if $1.hasParentheses()
           new Value $1
         else
           $1
@@ -606,14 +606,12 @@ grammar =
         # it won't compile since `Parens` (`(b; break)`) is compiled as `Value` and
         # pure statement (`break`) can't be used in an expression.
         # For this reasons, we return `Block`, which is passed further to `Value` and `Expression`.
-        stm = $2.contains (n) -> n.constructor.name is 'StatementLiteral'
-        if stm
+        if $2.hasStatementLiteral()
           LOC(2)(Block.wrap [$2])
         else
           new Parens $2
     o '( INDENT Body OUTDENT )',                ->
-        stm = $3.contains (n) -> n.constructor.name is 'StatementLiteral'
-        if stm
+        if $3.hasStatementLiteral()
           LOC(3)(Block.wrap [$3])
         else
           new Parens $3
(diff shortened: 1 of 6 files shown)


src/grammar.coffee
```diff
@@ -597,8 +601,22 @@ grammar =
...
   # the trick.
   Parenthetical: [
-    o '( Body )',                               -> new Parens $2
-    o '( INDENT Body OUTDENT )',                -> new Parens $3
+    o '( Body )',                               ->
+        # When `Parens` block includes a `StatementLiteral` (e.g. `(b; break) for a in arr`),
+        # it won't compile since `Parens` (`(b; break)`) is compiled as `Value` and
+        # pure statement (`break`) can't be used in an expression.
+        # For this reasons, we return `Block`, which is passed further to `Value` and `Expression`.
+        stm = $2.contains (n) -> n.constructor.name is 'StatementLiteral'
```
vendethiel: I'd rather we don't add such code to the grammar. We can have a node function for that.

vendethiel: Also we try not to use `constructor.name` for obfuscation reasons IIRC.

src/grammar.coffee
```diff
@@ -334,7 +334,7 @@ grammar =
     o 'Assignable'
     o 'Literal',                                -> new Value $1
     o 'Parenthetical',                          ->
-        if $1.constructor.name is 'Parens'
+        if $1.hasParentheses()
```
vendethiel: you can use `unwrap`

zdenko: Nice.

PR #4850, opened 2018-01-09
Fix #4422: operator (++/--) on braced expression

### LINKED ISSUE #4422: Lexer: Plusplus operator (++) on braced expression leads to parse error.

zdenko:
Fixes #4422 

This PR adds support for parenthetical expression with unary operators `++` and `--` 
`++(foo)` compiles to `++foo`,  and `(foo.bar)--` to `foo.bar--`


connec:
I'm not sure we necessarily want this to compile? It would be a useful convention if parens always allowed substitutions for other equivalent values, which isn't possible with `++` and `--`.

```coffee
x = 1
(x)--
(1)-- # Runtime error 0 - Invalid left-hand side expression
```

Also this will allow code to compile that perhaps shouldn't, including the example above and more absurd forms such as `(class)--` etc.

lydell:
I think I agree with @connec. I was surprised to see that _I_ added the "bug" label to that issue.

zdenko:
I agree that `Parenthetical` is too dangerous, but these are valid [expressions](https://babeljs.io/repl/#?babili=false&browsers=%3E%201%25&build=&builtIns=false&code_lz=G4QwTgBCELwQjAbgFCkge1hA3iAXAgL4oAUIAlANSUrVnmkUC0TKL9p6AdBdbZSW4VOPcizZNBo0gBcAFgEsAzqL7I685dOQlNK5q2Ts90oA&debug=true&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=stage-2%2Cstage-3&prettier=false&targets=Node-8.8&version=6.26.0) in JS.
```javascript
var a = 1;
var o = {a: 1};
(a)++;
++(a);
(a)--;
--(a);
(o.a)++;
++(o.a);
(o.a)--;
--(o.a);
(this.a)++;
++(this.a);
(this.a)--;
--(this.a);
```

zdenko:
Additional validation of the `Parens` expressions is added.
```coffeescript
(class)++
###
error: Invalid left-hand side expression
(class)++
^^^^^^^
###

--(1)
###
error: Invalid left-hand side expression
--(1)
  ^^^
###

++(a())
###
error: Invalid left-hand side expression
++(a())
  ^^^^^
###
```

vendethiel:
But now we're just adding code to support an alternative spelling of something that's less obvious...

zdenko:
So, actually this is not a bug, and there is no need to support `++(a)` syntax?

vendethiel:
Awaiting @lydell 's input.
Sorry if we do end up ruling this as Not A Bug, that'd be wasted time for you.

lydell:
I honestly don't remember why I labeled that issue as a bug in the first place. Sorry :(

Just because JavaScript supports this it doesn't have to mean that CoffeeScript should support it. I've never seen this used in the wild. But I may be missing something.

IMO it's not worth spending code on this. Sorry for wasting your time :(

GeoffreyBooth:
@lydell, as penance, I sentence you to reviewing the open pull requests.

zdenko:
Someone owes me a :beer:, or at least :coffee:  :smiley:
@GeoffreyBooth I guess you should also close #4422.

Issue #948, opened 2010-12-18
Loop variables are captured in closures; others are not


TrevorBurnham:
So, riddle me this:

```
for i in [1, 2, 3]
  j = i
  func1 = -> console.log "i is #{i} and j is #{j}"
  setTimeout func1, 0
```

Clearly, `i` and `j` should always be identical when `func1` is called. Yet the result is

```
i is 3 and j is 1
i is 3 and j is 2
i is 3 and j is 3
```

Looking at the compiled JavaScript, you can see why this is happening:

```
var func1, i, j, _fn, _i, _len, _ref;
_ref = [1, 2, 3];
_fn = function(i) {
  j = i;
  func1 = function() {
    return console.log("i is " + i + " and j is " + j);
  };
  return setTimeout(func1, 0);
};
for (_i = 0, _len = _ref.length; _i < _len; _i++) {
  i = _ref[_i];
  _fn(i);
}
```

The behavior is the same under 0.9.6, and 0.9.5; and it's the same whether you're in a `for...in` or a `for...of` loop. Under 0.9.4, you get

```
i is 1 and j is 1
i is 2 and j is 2
i is 3 and j is 3
```

which is a nice behavior if you were expecting `i` and `j` to be captured each time the `func1` was defined; another good behavior (the one I'd expect) would be to get

```
i is 3 and j is 3
```

three times, since the timeouts don't get invoked until the code has finished executing, by which time `i` and `j` should both be 3.

I'd be happy with either of those behaviors, but the current approach feels seriously inconsistent. With 1.0 set to be released in a week, what's the right solution?

PR #4854, opened 2018-01-12
Fix #1403: Variables inconsistently masked inside of `do`

### LINKED ISSUE #1403: Scope: Variables inconsistently masked inside of `do`
``` coffeescript
# The variable `masked` is redeclared inside of the anonymous
# function created by `do`.
for dummy in [0...1]
  masked = 10
  do -> alert masked = masked + 1
alert 'a line needs to be here to trigger the issue'
```

``` coffeescript
# The variable `notMasked` is *not* redeclared. The only
# difference is that `+=` is used instead of simple
# assignment.
for dummy in [0...1]
  notMasked = 10
  do -> alert notMasked += 1
alert 'a line needs to be here to trigger the issue'
```

Here is the JavaScript generated by the first (broken?) example:

``` javascript
var dummy, masked, _fn;
_fn = function() {
  var masked;
  return alert(masked = masked + 1);
};
for (dummy = 0; dummy < 1; dummy++) {
  masked = 10;
  _fn();
}
alert('a line needs to be here to trigger the issue');
```

Here is the JavaScript generated by the second (working?) example:

``` javascript
var dummy, notMasked, _fn;
_fn = function() {
  return alert(notMasked += 1);
};
for (dummy = 0; dummy < 1; dummy++) {
  notMasked = 10;
  _fn();
}
alert('a line needs to be here to trigger the issue');
```

I believe I've reduced the issue to its simplest form, though I could be wrong.

The following prevents the masking from occurring:
1. Changing the `for` loop to an `if` or a `while` loop
2. Removing the line immediately following the `for` loop
3. As demonstrated, changing the simple assignment to a `+=`
4. Assigning the `masked` variable before the loop

zdenko:
Fixes #1403.
It seems that `For::pluckDirectCall` isn't needed.

Examples from #1403 
```coffeescript
for dummy in [0...1]
  masked = 10
  do -> alert masked = masked + 1
alert 'a line needs to be here to trigger the issue'
```
```javascript
for (dummy = i = 0; i < 1; dummy = ++i) {
  masked = 10;
  (function() {
    return alert(masked = masked + 1);
  })();
}
alert('a line needs to be here to trigger the issue');
```
---
```coffeescript
for dummy in [0...1]
  notMasked = 10
  do -> alert notMasked += 1
alert 'a line needs to be here to trigger the issue'
```
```javascript
for (dummy = j = 0; j < 1; dummy = ++j) {
  notMasked = 10;
  (function() {
    return alert(notMasked += 1);
  })();
}
alert('a line needs to be here to trigger the issue');
```

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -5288,7 +5288,7 @@
       // comprehensions. Some of the generated code can be shared in common, and
       // some cannot.
       compileNode(o) {
-        var body, bodyFragments, compare, compareDown, declare, declareDown, defPart, defPartFragments, down, forPartFragments, fragments, guardPart, idt1, increment, index, ivar, kvar, kvarAssign, last, lvar, name, namePart, ref, ref1, resultPart, returnResult, rvar, scope, source, step, stepNum, stepVar, svar, varPart;
+        var body, bodyFragments, compare, compareDown, declare, declareDown, defPart, down, forPartFragments, fragments, guardPart, idt1, increment, index, ivar, kvar, kvarAssign, last, lvar, name, namePart, ref, ref1, resultPart, returnResult, rvar, scope, source, step, stepNum, stepVar, svar, varPart;
         body = Block.wrap([this.body]);
         ref1 = body.expressions, last = ref1[ref1.length - 1];
         if ((last != null ? last.jumps() : void 0) instanceof Return) {
@@ -5397,7 +5397,6 @@
         if (this.pattern) {
           body.expressions.unshift(new Assign(this.name, this.from ? new IdentifierLiteral(kvar) : new Literal(`${svar}[${kvar}]`)));
         }
-        defPartFragments = [].concat(this.makeCode(defPart), this.pluckDirectCall(o, body));
         if (namePart) {
           varPart = `\n${idt1}${namePart};`;
         }
@@ -5415,10 +5414,7 @@
         if (bodyFragments && bodyFragments.length > 0) {
           bodyFragments = [].concat(this.makeCode('\n'), bodyFragments, this.makeCode('\n'));
         }
-        fragments = [];
-        if ((defPartFragments != null) && fragmentsToText(defPartFragments) !== '') {
-          fragments = fragments.concat(defPartFragments);
-        }
+        fragments = [this.makeCode(defPart)];
         if (resultPart) {
           fragments.push(this.makeCode(resultPart));
         }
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
Before we jettison `pluckDirectCall` entirely, can someone explain why it was added in the first place? From my Git diving I see that @satyr added it in 4e39e2eec41bf207c2b8717c2360ae060a7d7e03; @satyr, do you mind sharing what was on your mind back in 2010? The commit message was “made dynamic calls at the last of `for` expression be defined outside”.

I did some testing, and on `master` `pluckDirectCall` is currently called by the following tests, all in `comprehensions.coffee`:

- Ensure that the closure wrapper preserves local variables.
- Even a convoluted one.
- Scoped loop pattern matching.
- Issue #948. Capturing loop variables.
- Comprehensions over function literals.

I pulled those tests into their [own file](https://gist.github.com/GeoffreyBooth/c0e0d539c13ae181ee719627151d3402) and compiled them into JS using both `master` and this branch. [Here’s the diff.](https://www.diffchecker.com/sHxumxSy)

A minimal case seems to be this:

```coffee
for i in [1..3] then do -> console.log i
```

On `master`:

```js
var fn, i, j;

fn = function() {
  return console.log(i);
};
for (i = j = 1; j <= 3; i = ++j) {
  fn();
}
```

Per this PR:

```coffee
var i, j;

for (i = j = 1; j <= 3; i = ++j) {
  (function() {
    return console.log(i);
  })();
}
```

Both versions print `1`, `2`, `3`, which might not be obvious at first; the reference to `i` is consistent no matter where the `do` function is put. But the two outputs illustrate the difference `pluckDirectCall` makes, which is to shift a `do`-wrapped function outside of the `for` loop, presumably for scope reasons. Sounds a lot like “made dynamic calls at the last of `for` expression be defined outside.” So . . . why? What is gained by creating a temporary variable `fn` and assigning this `do` function to it, and calling `fn` inside the `for` loop rather than leaving the function itself there?

4e39e2eec41bf207c2b8717c2360ae060a7d7e03 has this test added, in obsolete CoffeeScript syntax:

```coffee
# If the last expression is dynamic call,
# define it outside and pass loop variables to it.
fs = for i from 0 to 2 when i > 0
  void
  do ->
    {callee} = arguments
    -> [i, callee]
[one, two] = (f() for f in fs)
eq one[0], 1
eq two[0], 2
eq one[1], two[1]
```
Even converting this into modern syntax:
```coffee
fs = for i in [0..2] when i > 0
  do ->
    {callee} = arguments
    -> [i, callee]
[one, two] = (f() for f in fs)
eq one[0], 1
eq two[0], 2
eq one[1], two[1]
```
none of these tests pass, assuming I converted it correctly. Perhaps `pluckDirectCall` is a relic of those times, when scope and comprehensions behaved differently?

vendethiel:
@GeoffreyBooth my memory is a bit fuzzy because it's been a long time... But AFAIK not defining the function inside the loop is an optimization (see coco's wiki page "Additions").
As to why the test passes: I don't remember everything, but I think at some point `do` auto-captured loop variables(?). I think that's why the test passed.

vendethiel:
Also, paging @satyr about some reasoning from 2010 is probably not going to work...

GeoffreyBooth:
Ah, an optimization makes sense. Well, it's not really worth keeping if it's buggy, which it appears to be; and it's an optimization the user could do themselves. If they want the function defined before the loop and referenced in the loop, they could write their code as such.

So this looks good to me. @vendethiel, any notes? And have we lost @satyr?

vendethiel:
LGTM. As for @satyr, I think their last commit was a few years back, so we'd have to wait for a while.

connec:
This *may* lead to non-trivial performance degradations for code using the old `for ... then do ->` construct:

```coffee
for a in [1, 2, 3] then do ->
  setTimeout (-> console.log a), 1000
```

PR #4860, opened 2018-01-18
Fix #3306: trailing comma in a function call in the last line throws a syntax error

### LINKED ISSUE #3306: Lexer: Trailing comma in function call sometimes is a syntax error and sometimes gets ignored
If a source file ends with a function call that has an extra trailing comma:

``` coffeescript
foo bar,
```

The compiler, fortunately, raises a compiler error:

```
[stdin]:1:9: error: unexpected end of input
foo bar,
        ^
```

But if the invalid function call is inside a block:

``` coffeescript
if baz
  foo bar,
```

The compiler doesn't complain about the comma:

``` javascript
if (baz) {
  foo(bar);
}
```

Same thing if some statements follow that block:

``` coffeescript
if baz
  foo bar,
qux
```

``` javascript
if (baz) {
  foo(bar);
}

qux;
```

And, if a statement follows the traling-comma function call at the same indentation level, the compiler does this:

``` coffeescript
foo bar,
if baz
  qux quxx
```

``` javascript
foo(bar, baz ? qux(quxx) : void 0);
```

Now, this last case might be an intended behaviour; though it definitely falls in the what-the-fuck-is-happening category for me. But i'm pretty sure the others are bugs.

Sorry if this is a duplicated issue. I couldn't find any previous discussion on this.

zdenko:
Fixes #3306.
Trailing comma in a function call in the last line throws syntax error because the implicit call isn't closed in `Rewriter::addImplicitBracesAndParens`.

```coffeescript
foo bar,
```

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -510,7 +510,7 @@
             while (inImplicit()) {
               [stackTag, stackIdx, {sameLine, startsLine}] = stackTop();
               // Close implicit calls when reached end of argument list
-              if (inImplicitCall() && prevTag !== ',') {
+              if (inImplicitCall() && prevTag !== ',' || (prevTag === ',' && tag === 'TERMINATOR' && (nextTag == null))) {
                 endImplicitCall();
               // Close implicit objects such as:
               // return a: 1, b: 2 unless true
--- a/src/rewriter.coffee
+++ b/src/rewriter.coffee
@@ -347,7 +347,8 @@ exports.Rewriter = class Rewriter
         while inImplicit()
           [stackTag, stackIdx, {sameLine, startsLine}] = stackTop()
           # Close implicit calls when reached end of argument list
-          if inImplicitCall() and prevTag isnt ','
+          if inImplicitCall() and prevTag isnt ',' or
+              (prevTag is ',' and tag is 'TERMINATOR' and not nextTag?)
             endImplicitCall()
           # Close implicit objects such as:
           # return a: 1, b: 2 unless true
--- a/test/compilation.coffee
+++ b/test/compilation.coffee
@@ -170,3 +170,10 @@ test "using transpile from the Node API requires an object", ->
 test "transpile option applies to imported .coffee files", ->
   return if global.testingBrowser
   doesNotThrow -> transpile 'run', "import { getSep } from './test/importing/transpile_import'\ngetSep()"
+
+test "#3306: trailing comma in a function call in the last line", ->
+  eqJS '''
+  foo bar,
+  ''', '''
+  foo(bar);
+  '''


GeoffreyBooth:
Can we add a test?

GeoffreyBooth:
Sorry, I just reread #3306. Why are we allowing this syntax?
```coffee
foo bar,
```
That feels to me like it should throw a syntax error. Admittedly #3306 is somewhat ambiguous, but I read it the opposite way: not that the above should compile, but that the other oddball examples currently do compile but shouldn’t.

CoffeeScript already has too many things that cause line continuations (like ending a line in `and`); we shouldn’t be adding more. I read #3306 as meaning that several of those other cases should cause syntax errors, even if they currently don’t.

zdenko:
So, the last comma is not allowed in function invocation?
In that case, we have to change the grammar rules
from

```coffeescript
Arguments: [
    o 'CALL_START CALL_END',                    -> []
    o 'CALL_START ArgList OptComma CALL_END',   -> $2
  ]
```
to

```coffeescript
Arguments: [
    o 'CALL_START CALL_END',           -> []
    o 'CALL_START ArgList CALL_END',   -> $2
  ]
```

Won't this change break existing code?

lydell:
Good point, @zdenko. I think this should be allowed:

```coffee
foo(
  1,
  2,
)
```

But `foo 1, 2,` looks really odd. Tricky.

zdenko:
@lydell `Rewriter::addImplicitBracesAndParens` wraps arguments in cases like `foo 1, 2,`.
I think this will create a lot of confusion if `foo(1, 2,)` compiles, but `foo 1, 2,` throws an error.
Besides, this 
```coffeescript
foo a, b,  
c = 2
```
currently compiles to `foo(a, b,  c = 2)`, so changing the rules for detecting the end of the implicit call will be tricky, and will cause even more broken code.

lydell:
@zdenko Good, I think you have nailed it. We have to support that comma, no matter how ugly it is. I guess #3306 is trying to point out some inconsistency that we should fix. Or perhaps this PR already fixes it?

GeoffreyBooth:
I think explicit parentheses make it acceptable. This is valid CoffeeScript and JavaScript:
```js
console.log(1,2,)
```
But currently `console.log 1,2,` in CoffeeScript throws a syntax error. This feels correct to me.

Parentheses can make it unambiguous that the user isn’t finished calling the function. As in:
```coffee
console.log(1,2,
x = 3
)
```
can be nothing other than `console.log(1, 2, x = 3)`; it’s unambiguous. But take away those parentheses and you have ambiguity.

GeoffreyBooth:
> I guess #3306 is trying to point out some inconsistency that we should fix. Or perhaps this PR already fixes it?

#3306 points out an inconsistency, yes, in that sometimes a function call is continued onto a following line and sometimes it isn’t. So the resolution is to either get stricter (continue onto the following line in fewer cases) or looser (continue onto the following line in more cases). This PR goes the “looser” route. #3306 doesn’t seem to pick a side as to whether the desired resolution is stricter or looser; I assumed the desired fix would have been to become more strict, that the compiler currently allows some bad syntax that it really shouldn’t, because when I read the examples that the compiler currently allows they look pretty awful and like things that should throw errors.

But if you think those currently-allowed cases are common and/or legitimate, that might appear in non-crazy code, I guess we’ll have to go looser—or just leave things as they are, and accept some logical inconsistency as a lesser evil than allowing even more bad syntax to be permissible.

lydell:
I don’t care very much which direction this goes as long as we’re careful not to break backwards compatibility unnecessarily.

GeoffreyBooth:
Taking the #3306 cases one by one:

``` coffeescript
foo bar,
```
```
[stdin]:1:9: error: unexpected end of input
foo bar,
        ^
```
The user who opened #3306 (and I) expect this to throw an error. So far so good. (This is the case that this PR makes permissible.)

``` coffeescript
if baz
  foo bar,
```
``` javascript
if (baz) {
  foo(bar);
}
```

This case the user and I consider to be a bug. This should throw a syntax error like the first case. I think if we made such a change, it wouldn’t really affect anyone; anyone who has a comma here most likely has more code that follows on the next line. Which brings us to:

``` coffeescript
if baz
  foo bar,
qux
```

``` javascript
if (baz) {
  foo(bar);
}

qux;
```
The user and I also both expect this to throw an error. The dedent, in particular, is very odd; even if we wanted to allow multiline implicit calls, I would think that each line should be at either the same level or indented. But let’s say for the sake of argument that that `qux` were indented, either to the same level as `foo` or one indent further in. _That_ would be a syntax that I think is reasonable and that probably appears in lots of code and that we need to continue to support. But this isn’t a case of a trailing comma; there’s no comma after `qux`, for example. This is a comma that triggers an implicit line continuation, similar to an `if` statement that ends in `and` in order to continue onto the next line. We could still cause a trailing comma that doesn’t cause a line continuation—such as a comma after `qux` here—to throw an error.

Finally, the user is befuddled by this one:
``` coffeescript
foo bar,
if baz
  qux quxx
```
``` javascript
foo(bar, baz ? qux(quxx) : void 0);
```
Which is really just the consequence of allowing a trailing comma to trigger a line continuation. Awfully surprising, though, it is.

But as I just wrote, I accept the need to allow a comma to force an implicit line continuation, because there’s too much code out there using it for that purpose for us to change that now. So the question is, do we also need to allow a _trailing_ comma, where there’s nothing on the next line to continue to?

It appears that we do, because this is allowed:

```coffee
obj =
  a: 1,
  b: 2,
```
I guess if we allow a trailing comma here, we need to allow it in `foo bar,` too. So maybe we should accept this PR after all. It’s just frustrating to make an already too-permissible compiler even more permissible, for a questionable syntax.

zdenko:
Some info about the trailing comma: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas
It seems that trailing comma is allowed in function calls and parameters, which means `foo bar,` is a valid code.
However, these are  still weird and confusing:

```coffeescript
foo a, b,

c

# compiles to
# foo(a, b, c)
```

```coffeescript
foo a, b,
if c then d

# or

foo a, b,

if c then d

# compiles to
# foo(a, b, c ? d : void 0);
```

I would say that `Rewriter` should be reworked to close implicit calls on trailing comma followed by two successive `TERMINATOR` tags.

GeoffreyBooth:
Anyone else have an opinion? @jashkenas @vendethiel @xixixao @connec

vendethiel:
I really don't like the dedent case. Otherwise lgtm

PR #4861, opened 2018-01-19
Fix #3909: Backslash to break line is inconsistent in 'for ... of' and 'for ... in' loops

### LINKED ISSUE #3909: Lexer: Backslash to break line is inconsistent in object comprehension
The following compiles correctly into JavaScript if I paste it into the "Try CoffeeScript" box on [the CoffeeScript homepage](http://coffeescript.org/):

```
x = ( key for own key of obj )
```

(Is that called "object comprehension?"  I don't know.)  This is a very small and uninteresting example, but I boiled it down to this case from the more interesting example in which I encountered it.

But we get inconsistent variations if I try to break the line at various points.  First, let me list those that compile as expected:

```
x \
  = ( key for own key of obj )

x = \
  ( key for own key of obj )

x = ( \
  key for own key of obj )

x = ( key \
  for own key of obj )

x = ( key for own key of \
  obj )

x = ( key for own key of obj \
  )
```

Now, those that don't compile.  All say "Error on line 2: unexpected of."  And yet, it's the exact same statement, just with breaks at different points.

```
x = ( key for \
  own key of obj )

x = ( key for own \
  key of obj )

x = ( key for own key \
  of obj )
```

This is my first issue for this project, so please forgive me if this idiosyncrasy is known or exists more broadly.  I did a brief search of the issues before posting, but certainly may have missed it.

zdenko:
Fixes #3909.
A syntax error is thrown when the backslash is used in `for of`  and `for in` loops, and `in` or `of` tags aren't in the same line with `for` tag.
The issue is in `Lexer::lineToken` where `for` tag is closed (`@seenFor = no`). This PR corrects this by checking the previous `tag`, e.g. `tag`  before the `TERMINATOR`.

```coffeescript
x = ( key for \
  own key of obj )

x = ( key for own \
  key of obj )

x = ( key for own key \
  of obj )

x = ( key for \
  key of obj )

x = ( key for \
  key of obj )

x = ( key for key \
  of obj )

x = ( key for \
  key in obj )

x = ( key for \
  key in obj )

x = ( key for key \
  in obj )
```

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -587,12 +587,16 @@
     // Keeps track of the level of indentation, because a single outdent token
     // can close multiple indents, so we need to know how far in we happen to be.
     lineToken(chunk = this.chunk) {
-      var diff, indent, match, minLiteralLength, newIndentLiteral, noNewlines, size;
+      var backslash, diff, indent, match, minLiteralLength, newIndentLiteral, noNewlines, prev, size;
       if (!(match = MULTI_DENT.exec(chunk))) {
         return 0;
       }
       indent = match[0];
-      this.seenFor = false;
+      prev = this.prev();
+      backslash = (prev != null) && prev[0] === '\\';
+      if (!(backslash && this.seenFor)) {
+        this.seenFor = false;
+      }
       if (!this.importSpecifierList) {
         this.seenImport = false;
       }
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Not that it should be fixed in the same PR, but what about #3909?

PR #4862, opened 2018-01-20
Fix #3933: Missing `then` doesn’t trigger nearby error

### LINKED ISSUE #3933: Lexer: Missing `then` doesn’t trigger nearby error

zdenko:
Fixes #3933.
`Rewriter::addImplicitBracesAndParens` recognizes code `for a in b do ->` or `for a in b ->` as implicit call. When `then` is missing, `b do ->` and `b ->` are compiled to `b(do ->)` and `b(->)`.

```coffeescript
for a in b do ->
  foo()
  bar()
baz()
```

After `Rewriter`:
```coffeescript
for a in b(do ->
  foo()
  bar())
baz()

# or in a single line

for a in b(do -> foo(); bar())
baz()
```

Other cases:

```coffeescript
for a in b do ->

for a in b ->

while a do ->

while a ->

switch
  when a do ->

switch
  when a ->
```

This PR fixes this by preventing implicit call and enables error to be positioned in correct place.

```coffeescript
for a in b do ->
  foo()

###
error: unexpected do
for a in b do ->
           ^^
###
```

### DIFF
--- a/src/rewriter.coffee
+++ b/src/rewriter.coffee
@@ -263,6 +263,20 @@ exports.Rewriter = class Rewriter
             stack.pop()
         start = stack.pop()
 
+      inControlFlow = =>
+        seenFor = @findTagsBackwards(i, ['FOR']) and @findTagsBackwards(i, ['FORIN', 'FOROF', 'FORFROM'])
+        controlFlow = seenFor or @findTagsBackwards i, ['WHILE', 'UNTIL', 'LOOP', 'LEADING_WHEN']
+        return no unless controlFlow
+        isFunc = no
+        tagCurrentLine = token[2].first_line
+        @detectEnd i,
+          (token, i) -> token[0] in LINEBREAKS
+          (token, i) ->
+            [prevTag, ,{first_line}] = tokens[i - 1] || []
+            isFunc = tagCurrentLine is first_line and prevTag in ['->', '=>']
+          returnOnNegativeLevel: yes
+        isFunc
+
       # Recognize standard implicit calls like
       # f a, f() b, f? c, h[0] d etc.
       # Added support for spread dots on the left side: f ...a
@@ -271,7 +285,8 @@ exports.Rewriter = class Rewriter
          (nextTag in IMPLICIT_CALL or
          (nextTag is '...' and @tag(i + 2) in IMPLICIT_CALL and not @findTagsBackwards(i, ['INDEX_START', '['])) or
           nextTag in IMPLICIT_UNSPACED_CALL and
-          not nextToken.spaced and not nextToken.newLine)
+          not nextToken.spaced and not nextToken.newLine) and
+          not inControlFlow()
         tag = token[0] = 'FUNC_EXIST' if tag is '?'
         startImplicitCall i + 1
         return forward(2)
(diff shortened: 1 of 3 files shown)


src/rewriter.coffee
```diff
@@ -263,6 +263,20 @@ exports.Rewriter = class Rewriter
             stack.pop()
         start = stack.pop()
 
+      inControlFlow = =>
+        seenFor = @findTagsBackwards(i, ['FOR']) and @findTagsBackwards(i, ['FORIN', 'FOROF', 'FORFROM'])
+        controlFlow = seenFor or @findTagsBackwards i, ['WHILE', 'UNTIL', 'LOOP', 'LEADING_WHEN']
+        return no unless controlFlow
+        isFunc = no
+        tagCurrentLine = token[2].first_line
+        @detectEnd i,
```
GeoffreyBooth: Speaking of line continuations, shouldn’t there be a `\` here? I don’t like relying on implicit line continuation.

zdenko: You mean for these cases
```coffeescript
for a in b \
-> c

for a in b \
do -> c

for \
a \
in \
b \
do \
-> \
```

vendethiel: we've accepted `,` since basically 2010. It's set to stay now :).

zdenko: Oh, that one.

GeoffreyBooth: Fine. 😞

GeoffreyBooth:
@zdenko I know I already merged this, but I just discovered an issue. I can open an issue if you think it warrants it. See dcd6de9a773d834426dbff472349f040bce2d652; after this PR was merged in (I’m pretty sure it was this one, but it might’ve been an earlier one) all `else if`s started to be compiled like `else\nif`:

```coffee
if no
  1
else if yes
  2
```
```js
if (false) {
  1;
} else {
  if (true) {
    2;
  }
}
```

The `else if` [should be output on the same line](http://coffeescript.org/#try:if%20no%0A%091%0Aelse%20if%20yes%0A%092%0A).

zdenko:
I'm working on it.

PR #4863, opened 2018-01-21
Fix #4437: variable scope in chained calls

### LINKED ISSUE #4437: Lexer: Chaining: Variable scope varies depending on whether its calling function starts with a leading dot
I have a question about the scoping of "myVar" here:
```coffeescript
# CoffeeScript:
foo myVar = 1
.bar -> myVar = 2

# Generates this Javascript:
var myVar;
foo(myVar = 1).bar(function() {
  var myVar;
  return myVar = 2;
});
```

In the example above, myVar gets scoped in two places, but if we remove the dot (.), it's only scoped in the enclosing scope:

```coffeescript
# CoffeeScript:
foo myVar = 1
bar -> myVar = 2

# Generates this Javascript:
var myVar;
foo(myVar = 1);
bar(function() {
  return myVar = 2;
});
```

My understanding of CoffeeScript scoping is: if the variable is assigned in the enclosing scope lexically-before it is used in a function sub-scope, the sub-scope will use that binding. The first case above breaks this rule. Is this a bug? Or am I not understanding CoffeeScript scoping fully?

zdenko:
Fixes #4437.

```coffeescript
foo(x = 3).bar(-> x = 4)
```
```js
var x;
foo(x = 3).bar(function() {
  return x = 4;
});
```
___
```coffeescript
foo(-> x = 3).bar(x = 4)
```
```js
var x;
foo(function() {
  return x = 3;
}).bar(x = 4);
```

### DIFF
--- a/test/function_invocation.coffee
+++ b/test/function_invocation.coffee
@@ -862,23 +862,23 @@ test "#4836: functions named get or set can be used without parentheses when att
   eq 7, @set @a
 
 test "#4473: variable scope in chained calls", ->
-  A = class A
-    foo: (a) ->
+  obj =
+    foo: -> this
+    bar: (a) ->
       a()
       this
-    bar: (a) -> this
 
-  a = (a) -> new A
-  b = (b) ->
-    b()
-    new A
-  c = (c) -> new A
+  obj.foo(a = 1).bar(-> a = 2)
+  eq a, 2
 
-  a(x = 1).foo( -> x = 2)
-  eq x, 2
+  obj.bar(-> b = 2).foo(b = 1)
+  eq b, 1
 
-  b(-> y = 2).bar(y = 1)
-  eq y, 1
+  obj.foo(c = 1).bar(-> c = 2).foo(c = 3)
+  eq c, 3
 
-  c(z = 1).foo(-> z = 2).bar(z = 3)
-  eq z, 3
+  obj.foo([d, e] = [1, 2]).bar(-> d = 4)
+  eq d, 4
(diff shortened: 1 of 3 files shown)


test/function_invocation.coffee
```diff
@@ -860,3 +860,25 @@ test "#4836: functions named get or set can be used without parentheses when att
   eq 13, @set 10
   eq 6, @get @a
   eq 7, @set @a
+
+test "#4473: variable scope in chained calls", ->
+  A = class A
+    foo: (a) ->
+      a()
+      this
+    bar: (a) -> this
```
GeoffreyBooth: Is this class perhaps more complicated than we need? What about this?

```coffee
obj =
  method: -> this

obj.method(x = 1).method(-> x = 2)
```

test/function_invocation.coffee
```diff
@@ -860,3 +860,25 @@ test "#4836: functions named get or set can be used without parentheses when att
...
+  c = (c) -> new A
+
+  a(x = 1).foo( -> x = 2)
+  eq x, 2
+
+  b(-> y = 2).bar(y = 1)
+  eq y, 1
+
+  c(z = 1).foo(-> z = 2).bar(z = 3)
+  eq z, 3
```
GeoffreyBooth: We should add tests for destructuring assignment too:

```coffee
obj.method([x, y] = [1, 2]).method(-> x = 2)
obj.method({x} = {x: 3}).method(-> x = 2)
```

PR #4867, opened 2018-01-28
Fix #4852, @get @set implicit object

### LINKED ISSUE #4852: @set vs this.set vs this?.set / obj?.set bug/inconsistency
```
> coffee -v
CoffeeScript version 2.1.1

> coffee
coffee> @set = -> 'ok'
[Function]
coffee> @set()
'ok'
coffee> @set 3
'ok'

coffee> @set moo: 3
[stdin]:1:2: error: 'set' cannot be used as a keyword, or as a function call without parentheses
@set moo: 3
 ^^^

coffee> this.set moo: 3
'ok'

coffee> this?.set moo: 3
[stdin]:1:7: error: 'set' cannot be used as a keyword, or as a function call without parentheses
this?.set moo: 3


coffee> boo = {}
{}
coffee> boo.set = -> 'ok'
[Function]
coffee> boo.set moo: 3
'ok'
coffee> boo?.set moo: 3
[stdin]:1:6: error: 'set' cannot be used as a keyword, or as a function call without parentheses
boo?.set moo: 3

coffee> @set( moo: 3 )
'ok'
coffee> this?.set( moo: 3 )
'ok'
coffee> boo?.set( moo: 3 )
'ok'
```

Seems like another edge case similar to https://github.com/jashkenas/coffeescript/issues/4836.

GeoffreyBooth:
This fixes all the edge cases noted in #4852.


src/lexer.coffee
```diff
@@ -192,13 +192,17 @@ exports.Lexer = class Lexer
     # what CoffeeScript would normally interpret as calls to functions named
     # `get` or `set`, i.e. `get({foo: function () {}})`.
     else if tag is 'PROPERTY' and prev
-      if prev.spaced and prev[0] in CALLABLE and /^[gs]et$/.test(prev[1]) and @tokens[@tokens.length - 2][0] isnt '.'
-        @error "'#{prev[1]}' cannot be used as a keyword, or as a function call without parentheses", prev[2]
+      if prev.spaced and prev[0] in CALLABLE and /^[gs]et$/.test(prev[1]) and
+         @tokens[@tokens.length - 2][0] not in ['.', '?.', '@']
```
vendethiel: We're relying on `@tokens[@tokens.length - 2]` existing here / `- 3` in the else, but it seems it's fine.

GeoffreyBooth: This is a good catch. If you file begins with `get foo: ->`, the compiler crashes. Updated. The compiler won’t throw the error if your file begins with `get` or `set`, but it will compile it as a function call; and it could only have been intended as a function call, as getters and setters need to be inside objects or classes.

PR #4869, opened 2018-01-29
Fix #4868: Incorrect ‘Can’t call super with @params’ error

### LINKED ISSUE #4868: Nodes: Incorrect ‘Can’t call super with @params’ error
```coffee
class A
  constructor: (fn) ->
class B extends A
  constructor: -> super -> @x
```

does not compile:

```
[stdin]:4:28: error: Can't call super with @params in derived class constructors
  constructor: -> super -> @x
                           ^^
```

`@x` is not `this` of the `class B`, but of the function `fn`. The code should compile.

zdenko:
Fixes #4868. 
The function body, passed as a parameter, to `super` is excluded from the arguments check in the `Code::eachSuperCall`. 

```coffeescript
class B extends A
  constructor: -> super -> @x

###
B = class B extends A {
  constructor() {
    super(function() {
      return this.x;
    });
  }
};
###

class extends A then constructor: (@a) -> super(class then constructor: (@b) -> @b)

###
(class extends A {
  constructor(a) {
    super(class {
      constructor(b) {
        this.b = b;
        this.b;
      }

    });
    this.a = a;
  }
});
###
```

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2856,13 +2856,21 @@ exports.Code = class Code extends Base
   eachSuperCall: (context, iterator) ->
     seenSuper = no
 
+    childArgs = (args) ->
+      cArgs = []
+      for arg in args
+        continue if arg instanceof Class or arg instanceof Code
+        cArgs.push arg
+      cArgs
+
     context.traverseChildren yes, (child) =>
       if child instanceof SuperCall
         # `super` in a constructor (the only `super` without an accessor)
         # cannot be given an argument with a reference to `this`, as that would
         # be referencing `this` before calling `super`.
         unless child.variable.accessor
-          Block.wrap(child.args).traverseChildren yes, (node) =>
+          cArgs = childArgs child.args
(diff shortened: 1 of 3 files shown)


vendethiel:
Can you please add a few tests?

Also make sure that this still errors: `class A then constructor: -> super ((a) -> a.x)(@)`, `class A then constructor: super => @a`, etc.

src/nodes.coffee
```diff
@@ -2856,13 +2856,21 @@ exports.Code = class Code extends Base
   eachSuperCall: (context, iterator) ->
     seenSuper = no
 
+    childArgs = (args) ->
```
vendethiel: we'd usually write this

```coffeescript
childArgs = (args) ->
  for arg in args when arg not instanceof Class and arg not instanceof Code
    arg
# or (I think filter is ES5)
childArgs = (args) ->
  args.filter (arg) -> arg not instanceof Class and arg not instanceof Code
```

GeoffreyBooth:
This looks great, aside from needing tests.

PR #4872, opened 2018-01-31
Fix #4871: `else if` no longer output together

### LINKED ISSUE #4871: Rewriter: `if else` no longer output together
See https://github.com/jashkenas/coffeescript/pull/4862#issuecomment-361824222. After #4862 (I think), `if else` gets output across two lines:

```coffee
if no
  1
else if yes
  2
```
```js
if (false) {
  1;
} else {
  if (true) {
    2;
  }
}
```

The `else if` [should be output on the same line](http://coffeescript.org/#try:if%20no%0A%091%0Aelse%20if%20yes%0A%092%0A). See dcd6de9a773d834426dbff472349f040bce2d652. @zdenko

zdenko:
Fixes #4871.

```coffeescript
if no
  1
else if yes
  2
```

```javascript
if (false) {
  1;
} else if (true) {
  2;
}
```

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -881,7 +881,7 @@
             tokens.splice(i + 1, 0, indent, outdent);
             return 1;
           }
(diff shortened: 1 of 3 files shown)


vendethiel:
Always feels good when fixes only remove code :-). :+1:

zdenko:
Yet another proof that less is more :smile:

PR #4855, opened 2018-01-12
Fix #4105: Duplicate variable `i` defined when evaluating `range`

zdenko:
Fixes #4105 

```coffeescript
a = 3
foo [0..a]
```
```javascript
var a;
a = 3;
foo((function() {
  var results = [];
  for (var i = 0; 0 <= a ? i <= a : i >= a; 0 <= a ? i++ : i--){ results.push(i); }
  return results;
}).apply(this));
```
---
```coffeescript
b = [1..a]
```
```javascript
var b;
b = (function() {
  var results = [];
  for (var i = 1; 1 <= a ? i <= a : i >= a; 1 <= a ? i++ : i--){ results.push(i); }
  return results;
}).apply(this);
```

PR #4873, opened 2018-02-01
2.2.0

GeoffreyBooth:
This release fixes *all* currently open bugs, dating as far back as 2014, 2012 and 2011. Heartfelt thanks and congratulations to @zdenko!

*   **Potential breaking change:** An inline `if` or `switch` statement with an ambiguous `else`, such as `if no then if yes then alert 1 else alert 2`, now compiles where the `else` always corresponds to the closest open `then`. Previously the behavior of an ambiguous `else` was unpredictable. If your code has any `if … then` or `switch … then` statements with multiple `then`s (and one or more `else`s) the compiled output might be different now, unless you had resolved ambiguity via parentheses. We made this change because the previous behavior was inconsistent and basically a bug: depending on what grammar was where, for example if there was an inline function or something that implied a block, the `else` might bind to an earlier `then` rather than a later `then`. Now an `else` essentially closes a block opened by a `then`, similar to closing an open parenthesis. #4838
*   When a required `then` is missing, the error more accurately points out the location of the mistake. #4862, #4872
*   An error is thrown when the `coffee` command is run in an environment that doesn’t support some ES2015 JavaScript features that the CoffeeScript compiler itself requires. This can happen if CoffeeScript is installed in Node older than version 6. #4839
*   Destructuring with a non-final splat/spread, e.g. `[open, contents..., close] = tag.split('')` is now output using ES2015 rest syntax. #4825
*   Functions named `get` or `set` can be used without parentheses in more cases, including when attached to `this` or `@` or `?.`; or when the first argument is an implicit object, e.g. `@set key: 'val'`. #4837, #4867
*   Statements such as `break` can now be used inside parentheses, e.g. `(doSomething(); break) while condition` or `(pick(key); break) for key of obj`. #4849
*   Bugfix for assigning to a property attached to `this`/`@` in destructuring, e.g. `({@prop = yes, @otherProp = no}) ->`. #4848
*   Bugfix for incorrect errors being thrown about calling `super` with a parameter attached to `this` when said parameter is in a lower scope, e.g. `class Child extends Parent then constructor: -> super(-> @prop)`. #4869
*   Bugfix to prevent a possible infinite loop when a `for` loop is given a variable to step by, e.g. `for x in [1..3] by step` (as opposed to `by 0.5` or some other primitive numeric value). #4853
*   Bugfix to no longer declare iterator variables twice when evaluating a range, e.g. `end = 3; fn [0..end]`. #4855
*   Bugfix for incorrect scope of variables in chained calls, e.g. `start(x = 3).then(-> x = 4)`. #4863
*   Bugfix for incorrect scope of variables in a function passed to `do`, e.g. `for [1..3] then masked = 10; do -> alert masked`. #4854
*   Bugfix to no longer throw a syntax error for a trailing comma in a function call, e.g. `fn arg1, arg2,`. #4860
*   Bugfix for an expression in a property access, e.g. `a[!b in c..]`. #4851
*   Bugfix to allow a line continuation backslash (`\`) at any point in a `for` line. #4861

http://rawgit.com/GeoffreyBooth/coffeescript/2.2.0/docs/v2/#changelog
@jashkenas @lydell @connec


jashkenas:
I speak from the bottom of my heart, when I say: Holy crap!

Three cheers for @zdenko!

GeoffreyBooth:
This is also our 1,000th closed pull request! 🎆

PR #4879, opened 2018-02-03
Fix #4878: Compile error when using destructuring with a splat or expansion in an array

### LINKED ISSUE #4878: Nodes: Compile error when using destructuring with a splat or expansion in an array
```coffee
foo = (list) ->
  [ first, ..., last ] = list
```

results in `TypeError: Cannot read property 'indexOf' of undefined`

That’s the simplest case I could come up with. It’s pretty finicky though, if you reference the values after the destructuring statement it compiles:

```coffee
foo = (list) ->
  [ first, ..., last ] = list
  [ first, last ]
```

However other more complex variations also fail, for instance having that statement inside conditional assignment:

```coffee
foo = (list) ->
  ret =
    if list.length
      [ first, ..., last ] = list
      [ first, last ]
    else
      []
```

removing the `ret = ` part of the last example compiles fine

Not a huge deal, it can be replaced with code that explicitly uses `list.slice -1`, it’s just we had some code that was compiling previously and stopped working with a minor update somewhere along the way

zdenko:
Fixes #4878.

### DIFF
--- a/test/assignment.coffee
+++ b/test/assignment.coffee
@@ -947,64 +947,41 @@ test "#4673: complex destructured object spread variables", ->
   eq g, 1
 
 test "#4878: Compile error when using destructuring with a splat or expansion in an array", ->
-  eqJS '''
-    foo = (list) ->
-      [first, ..., last] = list
-    ''', '''
-    var foo,
-      slice = [].slice;
+  arr = ['a', 'b', 'c', 'd']
 
-    foo = function(list) {
-      var first, last;
-      return [first] = list, [last] = slice.call(list, -1);
-    };
-    '''
+  f1 = (list) ->
+    [first, ..., last] = list
 
-  eqJS '''
-    foo = (list) ->
-      [first..., last] = list
-    ''', '''
-    var foo,
-      splice = [].splice;
+  f2 = (list) ->
+    [first..., last] = list
 
-    foo = function(list) {
-      var first, last;
-      return [...first] = list, [last] = splice.call(first, -1);
-    };
-    '''
+  f3 = (list) ->
+    ([first, ...] = list); first
 
-  eqJS '''
-    foo = (list) ->
-      ret =
-        if value.length
-          [ first, ..., last ] = value
-          [ first, last ]
-        else
-          []
-    ''', '''
-    var foo,
-      slice = [].slice;
+  f4 = (list) ->
+    ([first, ...rest] = list); rest
 
-    foo = function(list) {
-      var first, last, ret;
-      return ret = value.length ? (([first] = value, [last] = slice.call(value, -1)), [first, last]) : [];
-    };
-    '''
+  arrayEq f1(arr), ['d']
+  arrayEq f2(arr), ['d']
+  arrayEq f3(arr), 'a'
+  arrayEq f4(arr), ['b', 'c', 'd']
 
-  eqJS '''
-    foo = (list) ->
-      ret =
-        if value.length
-          [ first..., last ] = value
-          [ first, last ]
-        else
-          []
-    ''', '''
-    var foo,
-      splice = [].splice;
+  foo = (list) ->
+    ret =
+      if list.length
+        [first, ..., last] = list
+        [first, last]
+      else
+        []
 
-    foo = function(list) {
-      var first, last, ret;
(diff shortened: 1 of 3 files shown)


test/assignment.coffee
```diff
@@ -945,3 +945,66 @@ test "#4673: complex destructured object spread variables", ->
 
   {{g}...} = g: 1
   eq g, 1
+
+test "#4878: Compile error when using destructuring with a splat or expansion in an array", ->
+  eqJS '''
+    foo = (list) ->
+      [first, ..., last] = list
+    ''', '''
```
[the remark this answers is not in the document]
zdenko: This example would compile on `master`.
This PR fixes the issue where destructuring is the last line.
```coffeescript
foo = (list) ->
  [ first, ..., last ] = list
```
But, anyway, you're right. I'll change the tests.

test/assignment.coffee
```diff
@@ -947,64 +947,41 @@ test "#4673: complex destructured object spread variables", ->
...
+  f2 = (list) ->
+    [first..., last] = list
+
+  f3 = (list) ->
+    ([first, ...] = list); first
+
+  f4 = (list) ->
+    ([first, ...rest] = list); rest
+
+  arrayEq f1(arr), ['d']
```
GeoffreyBooth: In 2.1.1, `f1(arr)` returns `[ 'a', 'b', 'c', 'd' ]`

test/assignment.coffee
```diff
@@ -947,64 +947,41 @@ test "#4673: complex destructured object spread variables", ->
...
+  arrayEq f2(arr), ['d']
+  arrayEq f3(arr), 'a'
+  arrayEq f4(arr), ['b', 'c', 'd']
+
+  foo = (list) ->
+    ret =
+      if list.length
```
GeoffreyBooth: I don’t like using `list.length` being nonzero as a proxy for truthiness. If you’re evaluating that `list.length isnt 0`, you should be explicit about it. I think it’s a lot more readable than just `list.length` or other hacks like `!!`.

test/assignment.coffee
zdenko: In 2.1.1 this
```coffeescript
f1 = (list) ->
 [first, ..., last] = list
```

compiles into 

```javascript
f1 = function(list) {
  var first, last;
  return first = list[0], last = list[list.length - 1], list;
};
```

So, the returned value is `list`.

In 2.2.0 it's compiled to

```javascript
f1 = function(list) {
  var first, last;
  return [first] = list, [last] = slice.call(list, -1);
};
```

The returned value is `[last]'.

test/assignment.coffee
zdenko: Right. I copied the example from #4878.

test/assignment.coffee
GeoffreyBooth: Yes, but `[first, ..., last] = list` is shorthand for `return [first, ..., last] = list`, which looks to me like it should return the full array. As in, returning just `[last]` is wrong.

```coffee
do -> [a, b, c] = [1, 2, 3]
# returns [1, 2, 3]
```

zdenko: Yup. Sorry, my mistake. Not enogh coffe in the system :sleepy:

GeoffreyBooth: > Not enough coffee in the system

Isn’t that why you’re here? 😝

zdenko:
@GeoffreyBooth the last commit fixes the returned value, but I just found another issue, which I'll fix later today.
Sorry, I was too quick with the push.

GeoffreyBooth:
No worries.

The tests we have at least prove that this syntax compiles; it would be nice to also somehow prove that `first` and `last` are the values we expect them to be.

zdenko:
@GeoffreyBooth the "issue" I noticed before is a false alarm.
When I checked this compiled output,
```coffeescript
foo = (list) ->
  ret =
    if list.length > 0
      [first, ..., last] = list
      [first, last]
    else
      []

###
foo = function(list) {
  var first, last, ret;
  return ret = list.length > 0 ? (([first] = list, [last] = slice.call(list, -1), list), [first, last]) : [];
};
###
```

I thought that this part `(([first] = list, [last] = slice.call(list, -1), list)` shouldn't include `list` at the end.
Although it's not a bug, since the returned value is correct, it would be nicer if it could be omitted from such expressions. 
It's probably a task for some other PR.

PR #4880, opened 2018-02-03
Regular expressions `s` (dotAll) flag

GeoffreyBooth:
This adds support for ES2018’s [`s` (dotAll) flag](https://github.com/tc39/proposal-regexp-dotall-flag).

The syntax is only supported in Node 9+, so I added Node 9 to the list of versions we test in CI.

Merging this in would mean that we would need to expand the “Compatibility” section of the docs that currently say:

> With the exception of modules and JSX, all the modern JavaScript features that CoffeeScript supports can run natively in Node 7.6+.

We would probably need to put a table there, listing the features required by certain Node versions (i.e. async functions: 7.6+, `dotAll` flag: 9+, etc.).


aminland:
Should we wait for this to be stage 4 first? otherwise you have to use the plugin `transform-modern-regexp`

GeoffreyBooth:
This feature is at Stage 4: https://github.com/tc39/proposal-regexp-dotall-flag

GeoffreyBooth:
@jashkenas @lydell @zdenko @connec I updated the docs to start listing features that CoffeeScript supports that are newer than Node 7.6. See [this diff](https://github.com/jashkenas/coffeescript/pull/4880/commits/aefe43bc52e84802b1c3fe94fbd47b20168c7e29). Any feedback on this language?

Note also that Node’s new, native support for modules needs to be addressed. It was painful just writing the instructions for how to use `.mjs` files now; we should make this better. Please see #5018.

jashkenas:
Looks good! I would, personally, omit the "Why doesn't CoffeeScript simply always transpile?" paragraph.

GeoffreyBooth:
> I would, personally, omit the “Why doesn’t CoffeeScript simply always transpile?” paragraph.

Yeah, I think that made more sense back when we were introducing CoffeeScript 2, but it isn’t needed anymore.

PR #4881, opened 2018-02-03
Fix #4877: Exponentiation operators

### LINKED ISSUE #4877: Lexer, Nodes: Use ES syntax for exponentiation ** and **= operators
The [exponentiation operator](https://github.com/rwaldron/exponentiation-operator), e.g. `x ** y`, [was adopted](https://github.com/tc39/proposals/blob/master/finished-proposals.md) as part of ES2016 and is supported by [Node 7.5+](http://node.green/#ES2016-features-exponentiation------operator). We currently compile this into `Math.pow(x, y)`, but we can now output it as `x ** y`. Likewise for our `**=` operator (which is the same as `a = a * a`).

GeoffreyBooth:
Fixes #4877: Output ES2018 versions of `**` and `**=` operators.

You can’t do odd things like `!2 ** 2` or `~1 ** 5` in JavaScript, so I removed those tests. (Wrap the left side in parentheses and it’ll work.) I suppose technically this is a breaking change, but the test is really just testing order of operations and the examples seem unlikely to exist in production code.

These operators are only supported in Node 7.5+.


GeoffreyBooth:
I’ve updated this to account for #5003. I think this should be ready now. @zdenko do you have any notes?

I’d like to resolve #5014 and any other open issues first in a 2.2.4 patch release, then merge in all the enhancements that are ready and release them in a 2.3.0. So @zdenko if you think this PR looks good, please just add a comment saying so, and we can merge it into `master` once 2.2.4 is done.

zdenko:
LGTM :thumbsup:

aminland:
Should we wait for this to be stage 4 first? otherwise you have to use the plugin `transform-exponentiation-operator`

GeoffreyBooth:
This feature is Stage 4: https://github.com/tc39/proposals/blob/master/finished-proposals.md

GeoffreyBooth:
This feature doesn’t require a rewrite of the Compatibility section, so it’s safe to merge in as is.

PR #4883, opened 2018-02-03
Fix #4882: Range not declaring var for the "i"

### LINKED ISSUE #4882: Nodes: 2.2.0: Range not declaring var for the “i”
CoffeeScript code:

```coffeescript
      [0..100].forEach (idx) ->
        wrapped idx, (err, out) ->
```

Resulting JavaScript:

```javascript
      return function () {
        var results = [];
        for (i = 0; i <= 100; i++) {
          results.push(i);
        }
        return results;
      }.apply(this).forEach(function (idx) {
        return wrapped(idx, function (err, out) {
```

Note the missing `var i`.

This worked still in 2.1.1. Causes our browser tests to fail with:

> Strict mode forbids implicit creation of global property 'i'

Refs noflo/noflo#588

Potentially caused by #4825 ?

zdenko:
Fixes #4882.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2043,7 +2043,7 @@
         idx = del(o, 'index');
         idxName = del(o, 'name');
         namedIndex = idxName && idxName !== idx;
-        varPart = `${idx} = ${this.fromC}`;
+        varPart = known && !namedIndex ? `var ${idx} = ${this.fromC}` : `${idx} = ${this.fromC}`;
         if (this.toC !== this.toVar) {
           varPart += `, ${this.toC}`;
         }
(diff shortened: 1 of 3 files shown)


vendethiel:
We've long hidden the index from the user. Did we change that in the previous PR?

GeoffreyBooth:
Let’s please add some tests. I would think you could copy the example from the bug report and check if `global.i` is defined.

zdenko:
@vendethiel the previous PR (#4855) changed variable declaration of the index, e.g. `i`:
```coffeescript
# before
i = o.scope.freeVariable 'i', single: true

#after
i = o.scope.freeVariable 'i', single: true, reserve: no
```

The former declared `i` in the global scope.

The other thing, which I didn't notice before, but #4882 has found, is the different output of the `for` statement between `range`  with number literals (`[0..50]`) and variables (`[0..a]`) in `Range::compileArray`.

PR #4891, opened 2018-02-08
Fix #4889: for...range loop condition

### LINKED ISSUE #4889: `for` loop unexpected behavior
Hi,

I am trying to run this simple script:

```coffee
N = 1

for i in [0..N]
  console.log 'i = ' + i
  for j in [(i+1)..N]
    console.log 'j = ' + j
```

And I get:

```bash
i = 0
j = 1
i = 1
j = 2
j = 1
```

If I add `by 1` to the second loop:

```coffee
N = 1

for i in [0..N]
  console.log 'i = ' + i
  for j in [(i+1)..N] by 1
    console.log 'j = ' + j
```

... then I get: 

```
i = 0
j = 1
i = 1
j = 2
```

Is this behavior adequate? Why is `j` equal to `2`?

zdenko:
Fixes #4889.
It seems that #4853, besides preventing an infinite loop, also caused a breaking change in the existing code.
The condition which checks the upper and lower `range` bounds, in order to prevent the infinite loop,  always allows at least one iteration of the `for` loop.
Example: 
```coffeescript
a = 2
b = 1
s = 1
r = (x for x in [a..b] by s)
```
The value of `r`, prior to  #4853 was `[]`, but now it's `[2]`.

To me `[]` is a bug as I would expect the result to be `[2]`.   
Given that `r = (x for x in [1..2] by 1000)` is `[1]`, `r = (x for x in [1..2] by -1000)` should give the same result.

Anyway, to correct the breaking change caused by #4853, I prepared this PR which reverts the output.
And, I guess whether the result of `for x in [1..2] by -1` should be `[]`or `[1]`  belongs to a separate discussion and PR.

### DIFF
--- a/test/control_flow.coffee
+++ b/test/control_flow.coffee
@@ -1287,3 +1287,14 @@ test "#4871: `else if` no longer output together ", ->
      2;
    }
    '''
+
+test "#4889: `for` loop unexpected behavior", ->
+  n = 1
+  result = []
+  for i in [0..n]
+    result.push i
+    for j in [(i+1)..n] by 1
+      result.push j
+
+  arrayEq result, [0,1,1]
+
--- a/test/control_flow.coffee
+++ b/test/control_flow.coffee
@@ -1288,16 +1288,6 @@ test "#4871: `else if` no longer output together ", ->
    }
    '''
 
-test "#4889: `for` loop unexpected behavior", ->
-  n = 1
-  result = []
-  for i in [0..n]
-    result.push i
-    for j in [(i+1)..n] by 1
-      result.push j
-
-  arrayEq result, [0,1,1]
-
 test "#4898: Lexer: backslash line continuation is inconsistent", ->
   if ( \
       false \
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2031,7 +2031,7 @@
       // When compiled normally, the range returns the contents of the *for loop*
       // needed to iterate over the values in the range. Used by comprehensions.
       compileNode(o) {
-        var cond, condPart, from, gt, idx, idxName, known, lowerBound, lt, namedIndex, stepCond, stepPart, to, upperBound, varPart;
+        var cond, condPart, from, gt, idx, idxName, known, lowerBound, lt, namedIndex, ref1, ref2, stepCond, stepNotZero, stepPart, to, upperBound, varPart;
         if (!this.fromVar) {
           this.compileVariables(o);
         }
@@ -2054,12 +2054,11 @@
         // Generate the condition.
         [from, to] = [this.fromNum, this.toNum];
         // Always check if the `step` isn't zero to avoid the infinite loop.
-        stepCond = this.stepNum ? `${this.stepNum} !== 0` : `${this.stepVar} !== 0`;
-        condPart = known ? this.step == null ? from <= to ? `${lt} ${to}` : `${gt} ${        // from < to
-to}` : (lowerBound = `${from} <= ${idx} && ${lt} ${// from > to
-to}`, upperBound = `${from} >= ${idx} && ${gt} ${to}`, from <= to ? `${stepCond} && ${lowerBound}` : `${stepCond} && ${// from < to
-upperBound}`) : (lowerBound = `${this.fromVar} <= ${idx} && ${lt} ${// from > to
-this.toVar}`, upperBound = `${this.fromVar} >= ${idx} && ${gt} ${this.toVar}`, `${stepCond} && (${this.fromVar} <= ${this.toVar} ? ${lowerBound} : ${upperBound})`);
+        stepNotZero = `${(ref1 = this.stepNum) != null ? ref1 : this.stepVar} !== 0`;
+        stepCond = `${(ref2 = this.stepNum) != null ? ref2 : this.stepVar} > 0`;
+        lowerBound = `${lt} ${(known ? to : this.toVar)}`;
+        upperBound = `${gt} ${(known ? to : this.toVar)}`;
+        condPart = this.step != null ? `${stepNotZero} && (${stepCond} ? ${lowerBound} : ${upperBound})` : known ? `${(from <= to ? lt : gt)} ${to}` : `(${this.fromVar} <= ${this.toVar} ? ${lowerBound} : ${upperBound})`;
         cond = this.stepVar ? `${this.stepVar} > 0` : `${this.fromVar} <= ${this.toVar}`;
         // Generate the step.
         stepPart = this.stepVar ? `${idx} += ${this.stepVar}` : known ? namedIndex ? from <= to ? `++${idx}` : `--${idx}` : from <= to ? `${idx}++` : `${idx}--` : namedIndex ? `${cond} ? ++${idx} : --${idx}` : `${cond} ? ${idx}++ : ${idx}--`;
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1365,23 +1365,18 @@ exports.Range = class Range extends Base
     # Generate the condition.
     [from, to] = [@fromNum, @toNum]
     # Always check if the `step` isn't zero to avoid the infinite loop.
-    stepCond = if @stepNum then "#{@stepNum} !== 0" else "#{@stepVar} !== 0"
+    stepNotZero = "#{ @stepNum ? @stepVar } !== 0"
+    stepCond = "#{ @stepNum ? @stepVar } > 0"
+    lowerBound = "#{lt} #{ if known then to else @toVar }"
+    upperBound = "#{gt} #{ if known then to else @toVar }"
     condPart =
-      if known
-        unless @step?
-          if from <= to then "#{lt} #{to}" else "#{gt} #{to}"
-        else
-          # from < to
-          lowerBound = "#{from} <= #{idx} && #{lt} #{to}"
-          # from > to
-          upperBound = "#{from} >= #{idx} && #{gt} #{to}"
-          if from <= to then "#{stepCond} && #{lowerBound}" else "#{stepCond} && #{upperBound}"
-      else
-        # from < to
-        lowerBound = "#{@fromVar} <= #{idx} && #{lt} #{@toVar}"
-        # from > to
-        upperBound = "#{@fromVar} >= #{idx} && #{gt} #{@toVar}"
-        "#{stepCond} && (#{@fromVar} <= #{@toVar} ? #{lowerBound} : #{upperBound})"
+       if @step?
+         "#{stepNotZero} && (#{stepCond} ? #{lowerBound} : #{upperBound})"
+       else
+         if known
+           "#{ if from <= to then lt else gt } #{to}"
+         else
+          "(#{@fromVar} <= #{@toVar} ? #{lowerBound} : #{upperBound})"
 
     cond = if @stepVar then "#{@stepVar} > 0" else "#{@fromVar} <= #{@toVar}"
 
--- a/test/ranges.coffee
+++ b/test/ranges.coffee
@@ -130,43 +130,43 @@ test "#2047: Infinite loop possible when `for` loop with `range` uses variables"
 
   testData = [
     [1, 5, 1, [1..5]]
-    [1, 5, -1, [1]]
+    [1, 5, -1, []]
     [1, 5, up, [1..5]]
-    [1, 5, down, [1]]
+    [1, 5, down, []]
 
     [a, 5, 1, [1..5]]
-    [a, 5, -1, [1]]
+    [a, 5, -1, []]
     [a, 5, up, [1..5]]
-    [a, 5, down, [1]]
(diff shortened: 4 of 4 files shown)


vendethiel:
Ok, that code looks *much better*. Nicely done.

GeoffreyBooth:
Let’s add a test that matches #4889. Instead of logging the values, just push them into an array and then compare the array with what you’d expect it to be.

GeoffreyBooth:
Going back to the example in #4889, modified slightly so that it could become a test:

```coffee
results = []
n = 1

for i in [0..n]
  results.push 'i = ' + i
  for j in [(i+1)..n]
    results.push 'j = ' + j
```
This produces `[ 'i = 0', 'j = 1', 'i = 1', 'j = 2', 'j = 1' ]` in this branch, 2.1.1, 2.0.3, and 1.12.7. All well and good.

Now take the same as above, but add `by 1` to the second `for`:

```coffee
results = []
n = 1

for i in [0..n]
  results.push 'i = ' + i
  for j in [(i+1)..n] by 1
    results.push 'j = ' + j
```
Then we get:
* `[ 'i = 0', 'j = 1', 'i = 1', 'j = 2', 'j = 1' ]` in this branch
* `[ 'i = 0', 'j = 1', 'i = 1', 'j = 2' ]` in 2.1.1
* `[ 'i = 0', 'j = 1', 'i = 1' ]` in 2.0.3 and 1.12.7

The issue raised in #4889 was that `by 1` shouldn’t make a difference. And that certainly seems reasonable, it shouldn’t. So assuming the output of the first example is correct, this PR fixes the bug that `by 1` should be a noop.

It’s a separate question, which I think @zdenko is alluding at in his last comment, as to whether the output from the first example _is_ correct. But I think we should treat that as a separate bug. Could we get one PR that makes the existence or nonexistence of `by 1` irrelevant, and a separate issue/PR to discuss what the “correct” output should be in cases like the first example?

zdenko:
@GeoffreyBooth  my comment about the correctness of the output was regarding the second example, e.g. `for - range - by`.

Besides fixing infinite loop, #4853 also caused a breaking change of the `for - range - by` loop compilation, but I don't think this change is a bug.
And, I think that the result ([ 'i = 0', 'j = 1', 'i = 1' ]) user from #4889 expects is wrong.

Options:   
a) we merge this PR which will revert the output of `for - range - by` loop, and then create another PR to discuss changes in the output.
b) discard this PR and keep changes from the #4853 (with minor improvements in a separate PR)

I vote for b).

**Why the different result?**

The results from `for x in [a..b]` and `for x in [a..b] by 1]` are different because  `for - range` and `for - range - by` loops compiles differently.

1. `for x in [a..b]` 

No (significant) changes before and after #4853:

```coffeescript
for (x = a; a <= b ? x <= b : x >= b; a <= b ? ++x; --x)
```

The value of increment/decrement step is defined by the relation of `a` and `b`.

2. `for x in [a..b] by s`

Before #4853:

```coffeescript
for (x = a; s > 0 ? x <= b : x >= b; x += s)
```

The loop condition basically says if the step is larger then `0` it also means that the end index  ('b') of the range is larger then the start index (`a`).

After #4853:

```coffeescript
for (x = a; a <= b ? a <= x <= b : a >= x >= b; x += s)
```

The loop condition is changed: `x` must be between `a` and `b`

The original intention of this change was to prevent possible infinite loops. Back then I wasn't paying additional attention to the output (although I should), as all existing tests have passed.
But, using the value of 'by' to define loop condition seems wrong to me. The value of 'by` is an explicit increment/decrement step set by the user.

Here is the table that shows the difference between the `for - range - by` from the second example:

|  i  | j   | range | condition<br>before | condition<br>after| result<br>before | result<br>current   
| :---: | :---: | :---: | :----: | :-----: | :-----:   | :----:  
| 0   |  1  | [1..1] by 1 | j <= 1 |  1 <= j <= 1 | [i = 0, j = 1] | [i = 0, j = 1]
| 1   |  2  | [2..1] by 1|  **j <= 1** |  2 >= j >= 1 | **[i = 0, j = 1, i = 1]** | [i = 0, j = 1, i = 1, j = 2]
| 1   |  3  | [2..1] by 1 |   |  **2 >= j >= 1** |   | **[i = 0, j = 1, i = 1, j = 2]**

GeoffreyBooth:
I don’t have time right this minute to dig into this, but do you mind opening a PR that’s limited to making `by 1` and the lack of `by` equivalent? I think we can all agree that _that_ is a bug that should be fixed, and we should get that merged in.

As for changing/correcting the interpretation of `by`, I think I’ll probably agree with you, but I need to work through some examples first. This is complicated enough that I think it would benefit from its own PR for focus/discussion. That could maybe be this PR, if you want to split out the `by` stuff to another one.

zdenko:
Agree. I'll prepare a couple of tests for this one and create another PR for changing the interpretation of `by`.

GeoffreyBooth:
@zdenko If you don’t mind, let’s wrap up this work (both fixing the `by 1` inconsistency and any change to how ranges are handled, as a separate PR) so that I can get another bugfix patch release out. Once that’s out, we can merge in some of the new features/changed output PRs, which should trigger a new minor release (2.3.0).

zdenko:
@GeoffreyBooth I added a simple test, a code from the example.
Will you merge this into master, and I'll start another PR to deal with `range by` inconsistency.

test/control_flow.coffee
```diff
@@ -1288,6 +1288,16 @@ test "#4871: `else if` no longer output together ", ->
    }
    '''
 
+test "#4889: `for` loop unexpected behavior", ->
+  n = 1
+  result = []
+  for i in [0..n]
+    result.push i
+    for j in [(i+1)..n] by 1
```
GeoffreyBooth: Let’s either have both `for` loops say `by 1` or have neither do so.

If it makes a difference, in the other PR maybe we can copy this test with the other case (like one test with no `by`, and the other test with two `by 1`s).

zdenko: Without `by 1` the result will be `[ 0, 1, 1, 2, 1 ]`. I used the same code as the user in #4889 to show the final (expected) result. Adding `by 1 ` to the first loop will make no difference, but it's not actually the point of this PR (or #4889).    
The second loop range is `[1..1]` in the first iteration and `[2..1]` in the second, and this is the cause of different results if `by 1` is used or not.   
The purpose of this PR was to revert the compilation of `for - range - by` so the result would be the same as before #4853.

GeoffreyBooth: I don’t think we need a PR that produces the result the user expects, if we think that that result is wrong. I thought the plan was to have two PRs:
- one so that `by 1` and a lack of `by` are always equivalent
- one that fixes the range output to be what you think it should be

zdenko: Oh, sorry, my bad.   
In that case, if you agree,  I suggest just one PR which will cover both issues, as changed range output is prerequisite for correct `by 1 ` and w/o `by` output.

GeoffreyBooth:
Thinking out loud, starting with my example above:
```coffee
results = []
n = 1

for i in [0..n]
  results.push 'i = ' + i
  for j in [(i+1)..n]
    results.push 'j = ' + j
```

This could be rewritten as:
```coffee
results = []
for i in [0..1]
  results.push 'i = ' + i
  for j in [(i+1)..1]
    results.push 'j = ' + j
```

Now is where things get tricky. So the first range is `0..1`, and so the first `for` loops twice, once where `i` is 0 and once where `i` is 1 (remember that `..` is inclusive).

So there are _two_ second `for` loops created, once for each `i`. Those two ranges are `1..1` and `2..1`. There’s no `by` here, so what are we to do for `2..1`? Nothing at all, since a lack of `by` implies `by 1` but we can’t count up here? Or do we see that the start is greater than the end, and imply `by -1` instead?

The docs actually have an answer to this question. The example for ranges is this:
```coffee
countdown = (num for num in [10..1])
```
Which returns `[10,9,8,7,6,5,4,3,2,1]`, both now and in 1.12.7. So clearly the expectation is that we check whether we should count up or down, and choose `by 1` or `by -1` as appropriate.

So to unpack the loops:

```coffee
results = []
i = 0
results.push 'i = ' + i
for j in [1..1]
  results.push 'j = ' + j
i = 1
results.push 'i = ' + i
for j in [2..1]
  results.push 'j = ' + j
```

And finally:

```coffee
results = []
i = 0
results.push 'i = ' + i
j = 1
results.push 'j = ' + j
i = 1
results.push 'i = ' + i
j = 2
results.push 'j = ' + j
j = 1
results.push 'j = ' + j
```

And so we get `[ 'i = 0', 'j = 1', 'i = 1', 'j = 2', 'j = 1' ]`, which my “by hand” calculations determines to be what I would assume is the “correct” answer.

So I guess what I was assuming in earlier posts was wrong: a lack of `by` doesn’t always mean that there’s an implied `by 1`, but rather that there’s an implied `by 1` or `by -1`, depending on if the start of the range is greater or less than the end of the range. (If the start and end are equal, it doesn’t matter what the `by` value is, because the range returns just that one value.)

But what are we supposed to do with a range of `[2..1]` and `by 1`? You can’t count _up_ from 2 to 1. In 1.12.7, `[10..1] by 1` is an empty array. That feels correct to me. In 2.2.1, it’s `[10]`—as in, it does one loop first, then increments up and _then_ checks that it’s still within range. I can see the logic in this interpretation too.

Since both interpretations of `[10..1] by 1` are reasonable, I think we should stay with the 1.x handling unless there’s a persuasive case to be made for why it’s wrong. Otherwise the new output is just an arbitrary breaking change. Surely it must be possible to avoid infinite loops while preserving the empty-array output of `[10..1] by 1` that we got in 1.x.

zdenko:
> Surely it must be possible to avoid infinite loops while preserving the empty-array output of [10..1] by 1 that we got in 1.x.

@GeoffreyBooth this PR does exactly this. All tests from #4853 are included, I just corrected the empty array result in cases like `[2..1] by 1`.


> Since both interpretations of [10..1] by 1 are reasonable, I think we should stay with the 1.x handling 

I think the opposite. Here is simplified version of how `for x in [a..b] by s` is compiled in `nodes`:

```coffeescript
for (x = a; s > 0 ? x <= b : x >= b; x += s) 
```

The part `s > 0` feels wrong to me. Step `s` in explicit increment/decrement set by the user and shouldn't be used in the condition part. 
Let's see what happens if we use numbers:
```coffeescript
for x in [2..1] by 1
# for (x = 2; 1 > 0 ? x <= 1 : x >= 1; x += 1) 
```
The result is an empty array because `x <= 1` (e.g. `2 <= 1`) fails. It seems logic at first as you can't go up from `2`to `1` by `1`.

---

```coffeescript
for x in [2..2] by 1
# for (x = 2; 1 > 0 ? x <= 2 : x >= 2; x += 1) 
```
The result in this case will be `[2]`. So, here we say go up from `2` to `2` by `1` and we'll get `[2]`

---

```coffeescript
for x in [2..1] by -100
# for (x = 2; -100 > 0 ? x <= 1 : x >= 1; x += -100) 
```
This reads as go down from `2` to `1` by `-100`. The result is `[2]`.

---

It's probably just me, but I read `for x in [a..b]` as "go from `a` to `b`" and `for x in [a..b] by s` as "go from `a` to `b` with step of `s`".
So, the only difference is the step. The former will be implicitly set by CS, the latter by the user.
If  `by` is used as the condition, then it actually means "if `s` is positive go up to `b`, otherwise go down to `b`". The `a` is disregarded in this case, and IMHO it shouldn't be.
The loop starts with `a` and proceeds to `b` by given step:
```coffeescript
for x in [2..1] by 1    # = [2]
for x in [2..1] by -100 # = [2]
for x in [2..2] by 1    # = [2]
for x in [2..1] by 0    # = [2]
```

I know this is a breaking change, but I'm totally OK if the current output stays.

GeoffreyBooth:
I don't mean that we should keep the output of 1.x, of comparing the step against zero. I just mean that `[10..1] by 1` should still return an empty array, by however method makes sense.

zdenko:
Yes, I get that. I wanted to explain why the result shouldn't be an empty array. 
But, let's wrap this up. This PR provides the correct result (e.g. empty array) and prevents an infinite loop.
I say we merge this.

GeoffreyBooth:
@zdenko I’ve added a few more tests. Do they look correct to you?

Assuming they’re good, I’m okay with merging this in. Are there still additional changes you want to make to how ranges/`by` operate, that you’ll be opening a new PR for?

zdenko:
Tests are correct. I'll wait with the PR for `range-by` for now. It would be a breaking change to the existing code.  So, maybe in CS3.

GeoffreyBooth:
Thanks. Maybe open a new issue explaining it, with a little less focus on what the JavaScript output would be and a little more emphasis on what the result (empty array, array with certain values, etc.) would be. If you can explain why the current output is _wrong,_ and others agree, it’s something we can change in 2.x; if the current output is reasonable but not what a more thoughtful approach would be, it would be good to explain what the better result would be and why, and why we should change to it.

PR #4893, opened 2018-02-10
Fix #4875: Asynchronous iterators

### LINKED ISSUE #4875: Proposal: Support Asynchronous Iterators
[Asynchronous iterators](https://github.com/tc39/proposal-async-iteration) are increasingly well-supported. Should CoffeeScript 2 provide first-class support for them?

This would require supporting an additional `for` syntax corresponding to [`for await … of` in JavaScript](https://github.com/tc39/proposal-async-iteration#the-async-iteration-statement-for-await-of). Perhaps CoffeeScript can add `for await … from`?

The problem here is analogous to the problem that inspired `for … from`—namely, you cannot use `for` loops naturally with async iterators otherwise.

In addition, it would require allowing functions that contain both `await` and `yield` and generating `async function*` declarations, aka [async generator functions](https://github.com/tc39/proposal-async-iteration#async-generator-functions).

zdenko:
This PR adds support for [asynchronous iterators](https://github.com/tc39/proposal-async-iteration) (#4875)
It's still WIP and needs tests.

```coffeescript
readLines = (path) ->
  file = await fileOpen path
  try
    while not file.EOF
      yield await file.readLine()
  finally
    await file.close()

foo = ->
  res = []
  for await x from readLines
    res.push x
  res
```
compiles to

```javascript
var foo, readLines;
readLines = async function*(path) {
  var file, results;
  file = (await fileOpen(path));
  try {
    results = [];
    while (!file.EOF) {
      results.push((yield (await file.readLine())));
    }
    return results;
  } finally {
    await file.close();
  }
};

foo = async function() {
  var res, x;
  res = [];
  for await (x of readLines) {
    res.push(x);
  }
  return res;
};

```


src/grammar.coffee
```diff
@@ -669,16 +669,25 @@ grammar =
   ForBody: [
     o 'FOR Range',                              -> source: (LOC(2) new Value($2))
     o 'FOR Range BY Expression',                -> source: (LOC(2) new Value($2)), step: $4
-    o 'ForStart ForSource',                     -> $2.own = $1.own; $2.ownTag = $1.ownTag; $2.name = $1[0]; $2.index = $1[1]; $2
+    o 'ForStart ForSource',                     ->
+        $2[prop] = $1[prop] for prop in ['await', 'awaitTag', 'own', 'ownTag']
```
vendethiel: please move this kind of code to the AST noes

GeoffreyBooth:
Marking this as WIP because it lacks tests. You could convert the old error messages ones to be the basis of some new async iterators tests. Thanks for tackling this!

src/grammar.coffee
zdenko: OK. That code was already there, I just added 'await' parts.
But, I agree. It belongs to the `nodes`.

dyoder:
This is looking good…thank you @zdenko! And to @vendethiel for the review.

dyoder:
I wonder…creating an array of results here is potentially dangerous since async iterators will often, perhaps even typically, be used with an “infinite” collection, ex: events. Of course, _not_ doing that would be unexpected. Perhaps it's something we can document?

dyoder:
I guess that's no different than sync generators, though.

rdeforest:
I made a test for this and it failed 'cake test' because it needed --harmony. When I ran ```cake build:watch:harmony``` without my test, a different test ("classes with value'd constructors") failed with the error "TypeError: Class constructors may only return object or undefined".

In order that my work not go to waste I'll include it in my post here:

```coffee
    succeedLater = (val, ms) ->
      new Promise (resolve, reject) ->
        setTimeout (-> resolve val), ms

    asyncIterator = (iterations, operation) ->
      for i in [1 .. iterations]
        yield await operation()

    test "success loop finishes", ->
      a = do ->
        for await result from acsyncIterator 5, -> succeedLater 'ok', 10
          result

      eq a.length, 5
      eq -1, a.findIndex (x) -> x isnt 'ok'
```

I'll work on the error case next. :)

GeoffreyBooth:
I don’t know if this is already on this branch by now, but earlier this was tagged WIP because we need to add something to `Cakefile` similar to the feature detection for async function support, to test for async iterator support; and the async iterator tests would be in their own file that only gets loaded when supported. It’s a pattern we’re going to need to increasingly follow as we add support for more newer features.

rdeforest:
I see this in Cakefile:

```coffee
  unless global.supportsAsync # Except for async tests, if async isn’t supported.
    files = files.filter (filename) -> filename isnt 'async.coffee'
```

This seems like it wouldn't scale. Is there already work underway on generalizing this kind of thing? I didn't see a tracking issue for it. I'd love to take the project on if nobody else is doing it already.

GeoffreyBooth:
@rdeforest That would be great. Maybe do it as a separate PR?

The codebase assumes support for all features supported by Node 6, as that’s our minimum supported version. So there aren’t _that_ many features added to ES since then that need special casing. Async iterators, exponentiation operator (see #4881) and rewriting the modules tests to really run rather than do string comparisons. I’m not sure there are too many more; and in a year or two when there might be, we’ll probably bump up our minimum Node version to 8 or so.

rdeforest:
I'll take a stab at it and submit a new PR and in the interest of 'one change at a time' I'll base it on master and not including the tests I posted above. I'm currently unemployed and procrastinating so I should have something today or tomorrow.

GeoffreyBooth:
P.S. Take a look at the `1` branch `Cakefile`: https://github.com/jashkenas/coffeescript/blob/1/Cakefile#L403-L422

There were more special cases in the 1.x days, when we were supporting Node 0.8+.

rdeforest:
I opened #4996 to move this topic fork to.

GeoffreyBooth:
@zdenko I think we have the testing feature detection worked out, so the next step for this PR is to create a `tests/async_iterators.coffee` and put the async iterator tests in there. They should be normal tests, not string comparisons.

src/grammar.coffee
```diff
@@ -660,26 +660,31 @@ grammar =
...
   # or postfix, with a single expression.
   For: [
-    o 'Statement  ForBody',                     -> new For $1, $2
-    o 'Expression ForBody',                     -> new For $1, $2
-    o 'ForBody    Block',                       -> new For $2, $1
-    o 'ForLineBody  Block',                     -> new For $2, $1
+    o 'Statement    ForBody',  -> $2.addBody $1
+    o 'Expression   ForBody',  -> $2.addBody $1
+    o 'ForBody      Block',    -> $1.addBody $2
+    o 'ForLineBody  Block',    -> $1.addBody $2
```
GeoffreyBooth: Can you please explain all these grammar changes and why they’re necessary (or an improvement)?

src/nodes.coffee
```diff
@@ -2610,8 +2610,8 @@ exports.Code = class Code extends Base
         @isGenerator = yes
       if (node instanceof Op and node.isAwait()) or node instanceof AwaitReturn
         @isAsync = yes
-      if @isGenerator and @isAsync
-        node.error "function can't contain both yield and await"
+      if (node instanceof For and node.isAwait())
```
GeoffreyBooth: This feels a lot less readable than `if @isGenerator and @isAsync`. Why was this changed?

src/grammar.coffee
zdenko: @vendethiel suggested moving some code from `grammar` into `nodes`, e.g.

```coffeescript
 o 'ForStart ForSource',  ->
        $2[prop] = $1[prop] for prop in ['await', 'awaitTag', 'own', 'ownTag']
```

to which I agreed, and then reworked `For` in `nodes`.
It's just a cleanup.

GeoffreyBooth: Ah, okay. Maybe next time that should be its own PR, so it’s easier to track the changes that relate to the cleanup as opposed to adding support for async iterators, but I have no objection.

src/nodes.coffee
zdenko: Becuase I'm checking if this is async iterator `for await x from y`. 
`if @isGenerator and @isAsync` was used to check if both condition above returned true and throw an error. Since functions can now contain `yield` and `await`  this condition isn't required.
Besides, I think we can't use `@isAsync` here, e.g. `if node instanceof For and @isAwait` here, as `@isAwait` will be false.

GeoffreyBooth:
Okay, anyone else have any notes? @vendethiel?

PR #4990, opened 2018-02-25
Fixes 'get' and 'set' keyword example in docs

zdenko:
Fixes error in the example for `get` and `set` keyword syntax in docs.

```coffeescript
screen =
  width: 1200
  ratio: 16/9

Object.defineProperty screen, 'height',
  get: ->
    this.width / this.ratio
  set: (val) ->
    this.width = val * this.ratio

# this.height = 675

this.height = 900

# this.width = 1600
```


GeoffreyBooth:
Ah. Yes. Thanks 😄

Issue #4991, opened 2018-02-25
Syntactic sugar proposal: Sub-structuring


aminland:
Note that some prior discussion on this was started in #4919, before that thread turned into a discussion on something unrelated. 

Right now, destructuring is only supported as a means to assign properties to vars in top level scope.
The proposal is for a new syntax to support assigning properties from one object into another object (i.e. building / extending objects).

The proposed sugar would behave as follows (js output shown):
```coffeescript
# rhs substructuring (basically _.pick)
y = data{a, b, c}    
# >>    y = ({a:_a, b:_b, c:_c} = data, {a:_a, b:_b, c:_c})
# lhs substructuring
x{a, b, c} = data    
# >>    ({a: x.a, b: x.b, c: x.c} = data);
# or for easier implementation, reuse rhs output:
# >>    x = _extends(x, ({_a, _b, _c} = data, {a:_a, b:_b, c:_c}))
```

I think this would make a lot of things simpler in coffeescript, as the syntax is much more concise. The important thing to note here is that there must _not_ be a whitespace between the source object and the `{` symbol, allowing easy differentiation between this syntax and a function call.

Thoughts?

aminland:
Just following up on the thought process... If we fully implement sub-structuring, following should also work:
```coffeescript
user{title, name:full_name} = data 
# >>    ({title: user.title, name: user.full_name} = data);
user{addr:{street:[street1, street2], city, state, zip} } = data    
# >>    ({addr:{street:[user.street1, user.street2], city:user.city, state:user.state, zip:user.zip} } = data);
book = data{title, authors:[{name:author},{name:coauthor=null}]}
# >>     book = ({title:_title, authors:[{name:_name1},{name:_name2=null}]} = data, {title:_title, author:_name1, coauthor:_name2});
```

We can optionally support rest/spread syntax, with a way to specify unwanted props:
```coffeescript
user{name, birthdate, -password, ...extra_info} = data 
# >>    ({name: user.name, password:_blackhole, ...user.extra_info} = data);
record = data{id, -unwanted, ...properties}
# >>    record = ({id:_id, unwanted:_blackhole, ..._properties} = data, {id:_id, properties: _properties});
```

Sub-structuring doesn't have to used in an assignment:
```coffeescript
console.log data{id, -unwanted, ...properties}
# >>    console.log({id:_id, unwanted:_blackhole, ..._properties} = data, {id:_id, properties: _properties});
```

[...]

PR #4995, opened 2018-02-28
WIP: Sub-structuring

zdenko:
This PR adds features proposed in #4991: assigning properties from one object into another object (i.e. building / extending objects).

```coffeescript
data =
  title: "Mr"
  name: "Coffee"
  address: "Sesame Street"

view = data{title, name:full_name}
# view = { title: "Mr", full_name: "Coffee" }

view = data{title, name:full_name, rest...}
###
view = {
  title: "Mr", 
  full_name: "Coffee",
  rest: {
    address: "Sesame Street"
  }
}
###

console.log  data{title, name:full_name}
#  { title: "Mr", full_name: "Coffee" }
```

---

```coffeescript
data =
  title: "Mr"
  name: "Coffee"
  address: "Sesame Street"

view{title, name:full_name} = data
###
error: the variable 'view' has not been declared before
view{title, name:full_name} = data
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
###
```

---

```coffeescript
data =
  title: "Mr"
  name: "Coffee"
  address: "Sesame Street"

view = address: "Elm Street"
view{title, name:full_name} = data
###
view = { address: 'Elm Street', title: 'Mr', full_name: 'Coffee' }
###

```

---

```coffeescript
data =
  title: "Mr"
  name: "Coffee"
  address: "Sesame Street"
  zip: 10001

view = city: "London"
view{title, name, -zip, rest...} = data
# view = {city: "London", title: "Mr", name: "Coffee", rest: {address: "Sesame Street"}}
```

---

```coffeescript
data =
    addr:
      street: ["Elm", 12]
      city: "London"
      state: "England"
      zip: "GH 201 X"
      floor: 30
      appartment: "XJ-1"
    phone:
      home: "999-555-111"
      work: "888-333-555"

  user = name: "CS"
  user{
    addr: {
      street: [street1, street2],
      city, state, -zip, -floor, ext...
    },
    phone: { home:phone_home }
  } = data

###
user = {
  name: "CS",
  street1: "Elm",
  street2: 12,
  city: "London",
  state: "England",
  ext: {
    appartment: "XJ-1"
  },
  phone_home: "999-555-111"
}
###


user = data{
      addr: {
        street: [street1, street2],
        -zip, -floor, ext...
      },
      phone: { home:phone_home }
    }

###
user = {
  street1: "Elm",
  street2: 12,
  ext: {
     city: "London",
     state: "England",
     appartment: "XJ-1"
  },
  phone_home: "999-555-111"
}
###
```


vendethiel:
-1 for "Excluded properties". I don't think we should introduce new syntax, instead we should use e.g. `{x: []}` or `{x: {}}` or `{x: null}` or `{x:,}` or...

src/nodes.coffee
```diff
@@ -760,6 +760,45 @@ exports.IdentifierLiteral = class IdentifierLiteral extends Literal
...
+      else if prop.base not instanceof ExProperty
+        ret.push prop.base
+    ret
+
+  finalProps: (props = []) ->
+    ret = []
+    getProps = (props) => (p for p in @finalProps @objProperties props)
+    for prop in props
+      if prop instanceof Obj
+        ret = [ret..., (getProps prop.properties)...]
```
vendethiel: we usually write calls `getProps(prop.properties)`

src/nodes.coffee
```diff
@@ -760,6 +760,45 @@ exports.IdentifierLiteral = class IdentifierLiteral extends Literal
...
+        ret.push prop.value.base
+      else if prop instanceof Splat
+        ret.push prop.name
+      else if prop.base not instanceof ExProperty
+        ret.push prop.base
+    ret
+
+  finalProps: (props = []) ->
+    ret = []
+    getProps = (props) => (p for p in @finalProps @objProperties props)
```
vendethiel: why the for/in here? it's not here L792

zdenko: Right. I forgot to remove it. It's a leftover from the previous version.

aminland:
Nice work @zdenko !

Soo I'm just looking at what babel ends up doing when you destructure with `{a:{}} = {a:2}`:

```
function _objectDestructuringEmpty(obj) { if (obj == null) throw new TypeError("Cannot destructure undefined"); }
var _a = { a: 2 };
_objectDestructuringEmpty(_a.a);
_a;
```

So basically it will throw an error if the key doesn't exist, which imho is a poor design choice by the tc39 folks. The only use that it provides is basically a shorthand for validation. 

The reason I like the `-key` syntax is that it is concise way to say you don't care about certain properties (not that you want errors thrown). It is also, in the vain of everything else coffeescript, semantically clear. Together with the rest of the PR, it would allow full elimination of _.pick and _.omit, which are two of the most used lodash functions...

That's not to say we should prevent folks from doing the {a:{}} syntax, but at least this way people can have an option to be deliberate about throwing runtime errors. Note that regular destructuring does not cause babel to check that something is defined (i.e. `{a,b}=c` is a straightforward conversion, with no possible exceptions stemming from key choice). Thus, the js behavior is not really consistent, and it would behoove us to right some JS wrongs ;)

zdenko:
@aminland `-key` syntax is not removed from this PR. It compiles into `key:{}`, which is a valid ES syntax for dropping keys. Besides, once #4884 is merged, it will work without a change. Otherwise, we would have to keep parts of the current polyfill from the `Access` in order to remove keys. 
Also, the `-key` syntax is mostly useful with rest destructuring ` {a, -b, r...}`.  I see no point compiling `{a, -b}` into `{a}`. In that case, we would have to be consistent and also compile `{a, b{}}` into `{a}`.  And how to deal with this case `{-a}`?
Now, let's presume that `o = {b:1}` and user writes `{a:{}, b} = o`. The type error will enable him to correct the mistake, whose cause can be somewhere else, e.g. result from the function from some module.  I think we should allow the user to shoot himself in the foot.
If my intention is to remove some keys, I have to be sure these keys exist in the first place or use a third-party library or custom function to safely remove unwanted keys from the object.

GeoffreyBooth:
Sorry to join this discussion so late, I’m just starting to wrap my head around this. To take your initial example:

```coffee
data =
  title: "Mr"
  name: "Coffee"
  address: "Sesame Street"

view = data{title, name:full_name}
# view = { title: "Mr", full_name: "Coffee" }
```
So the `view =` assignment line is basically a shorthand for:
```coffee
view = Object.assign {}, {title: data.title, full_name: data.name}
```

Which gives me two questions:
1. Obviously the “sub-structuring” syntax is shorter, but is this enough of an improvement to be worth adding? Why?
2. Why is it `view = data{name: full_name}` and not `view = data{full_name: name}`? The latter feels more natural to me, as when you’re assigning to a variable in general the keys for the object in the new variable go on the left, e.g. `view = {full_name: "Coffee"}`. See also the `Object.assign` example.

zdenko:
> Why is it view = data{name: full_name} and not view = data{full_name: name}

This syntax is the same as assigning to new variable names in object destructuring, i.e. the new name is on the right side.

```coffeescript
o = {p: 42, q: yes}
{p: foo, q: bar} = o
# foo = 42, bar = yes
```

GeoffreyBooth:
Closing for now per https://github.com/jashkenas/coffeescript/issues/4991#issuecomment-386505347

Issue #4996, opened 2018-03-01
Proposal: generalize exempting of tests requiring advanced V8 features


rdeforest:
Per discussion on https://github.com/jashkenas/coffeescript/pull/4893, tests which require switches be passed to node/coffee have their special handling hacked into Cakefile.

My proposal (and I'm already working on a PR) is to move special-case test files to their own directories under test/, accompanied by a config.json which contains the particulars of what's special about those tests.

(I figured it made more sense to open this than to continue off-topic discussion on the aforementioned PR.)

rdeforest:
The first problem I'm seeing is that in the current architecture the scanning of test/ happens after re-invocation of bin/cake with (or without) --harmony. I will probably have to make some relatively invasive changes to implement this suggestion.

GeoffreyBooth:
I don’t think you need to worry about whether `--harmony` is invoked or not. Just use feature detection for each set of tests we might test, and if the runtime supports it (whether because it’s a new enough version or because `--harmony` is used) then we run those tests.

I also don’t like the idea of mixing a JSON file into the codebase. I know its function would be primarily to store data, hence it’s good to use a “data” file type, but editing that file would look very different from editing the rest of the codebase. Could we just use a regular `.coffee` file that simply exports an object?

rdeforest:
I think you just cleared up all my confusion and I think I can keep the PR small. I should have something in an hour or so (fingers crossed).

GeoffreyBooth:
Fixed via #5003.

PR #4997, opened 2018-03-02
simplify skipping unsupported tests (#4996)

rdeforest:
Per discussion on #4893, managing exclusion of tests when newer features are required is currently non-obvious. The goal of this PR is to make maintenance of such exceptions easy and surprise-free.

The existing method was a hard-wired exception for the file 'async.coffee'. When I added async_iterators for #4893 I was surprised because I didn't know about the ```build:watch:harmony``` build target nor the hardwired exception in runTests. Now both the feature test and filename match are handled on the same lines near the top of Cakefile.

I included two currently unused pattern/feature pairs to demonstrate the intent and value of this approach. Both will likely be used for #4877 and #4893.

I didn't write a test for this change since it's a change to the test system.

After posting this I'm going to figure out how to remove the extra unrelated empty commits, but I wanted to get this out there before the weekend to follow through on my previous commitment.


rdeforest:
Bah, forgot to push my changes to shorten the diff. Fixing that right away...

rdeforest:
I remembered how to use ```git merge --squash```.  Now I'm going to create a new PR with the clean branch and resolve this in favor of that...

rdeforest:
This PR obsoleted by #4998.

PR #5000, opened 2018-03-03
Fix: destructuring assignment with an empty array in object

zdenko:
While testing sub-structuring (#4995) I noticed wrong output of destructuring with an empty array in the object.

Current branch:

```coffeescript
{a:[], b} = obj

###
var b;
obj.undefined, b = obj[1];
###
```

Since `@variable.isAssignable()` in `Assign` returns false because of an empty array,  the `::compileDestructuring` is invoked. 
The syntax is valid and can be output directly.

This PR:

```coffeescript
{a:[], b} = obj

###
var b;
({
  a: [],
  b
} = c);
###
```

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2179,7 +2179,7 @@
       }
 
       isAssignable() {
-        var j, len1, message, prop, ref1;
+        var j, len1, message, prop, ref1, ref2;
         ref1 = this.properties;
         for (j = 0, len1 = ref1.length; j < len1; j++) {
           prop = ref1[j];
@@ -2188,7 +2188,7 @@
           if (message) {
             prop.error(message);
           }
-          if (prop instanceof Assign && prop.context === 'object') {
+          if (prop instanceof Assign && prop.context === 'object' && !(((ref2 = prop.value) != null ? ref2.base : void 0) instanceof Arr)) {
             prop = prop.value;
           }
           if (!prop.isAssignable()) {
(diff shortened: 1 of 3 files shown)


aminland:
This would be a breaking change since babel runs `{a: []}` through `function _toArray(arr) { return Array.isArray(arr) ? arr : Array.from(arr); }`.

As an example `Array.from(null)` and `Array.from(undefined)` throws a type error, so anyone who was using it to drop keys will be screwed.

zdenko:
I think my [comment](https://github.com/jashkenas/coffeescript/pull/4995#issuecomment-371619067) from #4995 applies here as well. 
I see no difference between `{a: []} = b` or `{a:{}} = b`. It is upon user to take care of destructuring variables.
And, I don't think anyone will be screwed, because currently, if an empty array is used in the destructuring`, you won't get correct [result](http://coffeescript.org/#try:o%20%3D%20%7Ba%3A%5B1%5D%2C%20b%3A2%7D%0A%7Ba%3A%5B%5D%2C%20b%7D%20%3D%20o%0A%0Aconsole.log(b)) even if the key exists in the object.

GeoffreyBooth:
I don’t think we should care what Babel does. If we output valid JS, that’s all we need to worry about.

This is correct, isn’t it?

```js
obj = {a: [1]}
result = ({a: []} = obj)
```
In Chrome console, `result` equals `{a: [1]}`.

So @zdenko and @aminland is this safe to merge in?

zdenko:
@GeoffreyBooth go for it.

PR #5006, opened 2018-03-06
Access object members with '.' and literal

zdenko:
As proposed in #1334 and #3828, this PR adds options to access object and array members by using dot notation with literals (`string`, `number`).

```coffeescript
obj =
    "a-b-c": 1
    a:
      "b-c": 11
    "a-b":
      2: 22
    42: 3
    "f-o-o": -> "foo"
    "b-a":
      "-r": -> "bar"
      5: -> "b52"
    4:
      2: 42
    5:
      "2-0": "five-two-zero"

obj."a-b-c"
obj.42
obj."f-o-o"()
obj.'b-a'."-r"()
obj.4.2
obj.5."""2-0"""

obj."a-b".2 + .5  # = 22.5

# String interpolation is supported.
obj."#{"a" + "-"}b#{"-"}c"

# Arrays
arr = [
  [1, 2, 3]
  [4, 5, 6]
  [7, 8, 9]
]

arr.0.2
arr."#{1 + 1}"."#{6 * 3 / 9}"
```


aminland:
This is starting to get messy... Just throwing my two cents in: I'm not a fan of this syntax :(

vendethiel:
Just write that.

aminland:
Also I see it getting messy depending on variable names: `temp2.2` vs `temp2 .2`
Also `temp.2.2` is unclear whether it means `temp[2][2]` or `temp[2.2]`

vendethiel:
We went over this previously.

GeoffreyBooth:
What about the issues raised in #3828 regarding floats without leading zeros, e.g. `.5`? We should at least have tests that those didn’t break, e.g.
```coffee
test "#3828: Floats with leading periods still work", ->
  eq (0.5).toFixed(1), (.5).toFixed(1)
```

connec:
For backwards compatibility we should also check things like `f .2` and `f .2.toFixed()` as @aminland mentioned above.

test/numbers.coffee
```diff
@@ -86,3 +86,10 @@ test "Infinity", ->
 
 test "NaN", ->
   ok isNaN 1/NaN
+
+test "#3828: Floats with leading periods still work", ->
+  eq (0.5).toFixed(1), (.5).toFixed(1)
+
+  foo = (n) -> n
+  eq 0.2, foo .2
```
GeoffreyBooth: This really _shouldn’t_ work . . . in JavaScript, there’s a good chance that `0.2 === 0.2` fails, because one of them might be treated as `0.20000000001` or so.

vendethiel: No, that’s not true.

zdenko: I think this one can fail sometimes `.2 === 2/10 === .1 * 2`.

connec: The literals would be fine - it would be a different story if it was testing...

```coffee
foo = (n) -> n + 0.2
eq 0.6, foo .4
```

...because `0.2 + 0.4 isnt 0.6` (it's `0.6000000000000001`).

GeoffreyBooth:
@zdenko @helixbass @vendethiel @connec @aminland Any more notes on this, or is this okay to merge in?

@helixbass Does this interfere with any of the refactoring you’re doing? Obviously this will introduce some merge conflicts with your string refactoring, but aside from fixing merge conflicts, does allowing this syntax expansion interfere with anything you’re working on regarding getting us closer to an AST and Prettier support

Per semver this would bump us into 2.4.0, unless we want to merge it in and wait for another feature or two to be ready to release as part of 2.4.0.

src/lexer.coffee
```diff
@@ -242,6 +242,22 @@ exports.Lexer = class Lexer
     number = match[0]
     lexedLength = number.length
 
+    # Property number notation, e.g. a.3
+    prev = @prev()
+    if prev and not prev.spaced
+      if prev[0] in [CALLABLE..., '?.', '::', '?::']
+        s = 0
```
helixbass: @zdenko in general I'd prefer more readable variable names. Makes it much easier to decipher what's going on. Eg here `s` -> `offset`, `nProp` -> `remaining` or `numChunk`, `x` -> `index`

src/lexer.coffee
```diff
@@ -242,6 +242,22 @@ exports.Lexer = class Lexer
...
+        nProp = number
+        if nProp[0] is "."
+          @token '.', '.', s++, 1
+          nProp = nProp[1..]
+
+        for num, x in nProp.split "."
+          @token '.', '.', s++, 1 if (x+1) % 2 is 0
```
helixbass: @zdenko since `NUMBER` will match at most one `.`, the `(x+1) % 2 is 0` logic is overly generic and therefore hard to follow. Should work just to say `if x is 1`

src/lexer.coffee
```diff
@@ -287,6 +303,11 @@ exports.Lexer = class Lexer
     {tokens, index: end} = @matchWithInterpolations regex, quote
     $ = tokens.length - 1
 
+    if prev and not prev.spaced and prev[0] in ['.', '?.', '::', '?::']
+      stringProp = prev[0]
```
helixbass: @zdenko naming maybe `stringProp` -> `precedingAccessor`. Or since it's effectively just being checked for its presence, you could name it `precededByAccessor` and not even bother setting it to `no` in an `else`, eg
```
precededByAccessor = yes if prev and not prev.spaced and prev[0] in ['.', '?.', '::', '?::']
```

src/lexer.coffee
```diff
@@ -830,7 +851,12 @@ exports.Lexer = class Lexer
   # `options` first.
   mergeInterpolationTokens: (tokens, options, fn) ->
     if tokens.length > 1
-      lparen = @token 'STRING_START', '(', 0, 0
+      sTag =
```
helixbass: @zdenko naming maybe `sTag` -> `delimitingTags`

test/arrays.coffee
```diff
@@ -287,3 +287,21 @@ test "for-from comprehensions over Array", ->
 
   array2 = (a + b for [a, b] from [[10, 20], [30, 40], [50, 60]] when a + b >= 70)
   ok array2.join(' ') is '70 110'
+
```
helixbass: @zdenko a couple other ideas for tests:
- check that eg `f .5` -> `f(.5)` (as suggested by @connec)
- some longer accessor chains eg `a.1.2.3.4.5`
- some tests with `@` eg `@1`, `@1.'a'` (note that eg `@'b'` compiles into a tagged template expression so there's a slight inconsistency)

src/grammar.coffee
```diff
@@ -154,8 +154,19 @@ grammar =
     o 'CSX_TAG',                                -> new CSXTag $1
   ]
 
+  # PropertyNumber is `NumberLiteral` `Accessor`, e.g. obj.42
```
helixbass: @zdenko this explanation is a bit confusing/misleading, when I read
```
`NumberLiteral` `Accessor`
```
I think it means that `PropertyNumber` is the grammar sequence `NumberLiteral Accessor`

Maybe say eg
```
PropertyNumber is a `NumberLiteral` as an access, e.g. obj.42
```

aminland:
@GeoffreyBooth how is it that your point from https://github.com/jashkenas/coffeescript/issues/4991#issuecomment-386505347 doesn't apply here as well?

GeoffreyBooth:
@aminland I had thought that this wasn’t and couldn’t be valid JavaScript syntax, and so there isn’t a risk of “getting ahead” of the ECMA committee on something they may someday adopt. But . . . _could_ this someday be valid JavaScript syntax? Now I’m not so sure.

Though you could counterargue that even if it does someday become allowed in JavaScript, would the syntax ever represent anything other than this? In which case it wouldn’t matter.

But yes, let’s discuss. This feature isn’t such a groundbreaking improvement that it’s worth risking a potential breaking change.

GeoffreyBooth:
I came across this, on the forum where new proposals to JavaScript get discussed:
https://esdiscuss.org/topic/extended-dot-notation-pick-notation-proposal

So the ES community is looking at the space after the period in dot notation and considering what else could be put there. That said, would they ever consider an integer there to be anything other than an index reference (what we're considering here)? Perhaps a champion for this proposal should ask on that forum.

GeoffreyBooth:
So I just ran the tests on #5215 and one of them is failing on `1._23`, which parses as:

```
Block
  Value NumberLiteral: 1
    Access PropertyName: _23
```

So `1._23` shouldn’t parse, which is the issue in that branch. But aside from that, this error is making me think that maybe we should close this PR after all, since we wouldn’t want to add this new syntax and then get into trouble later if/when JavaScript adds anything else along the lines of new numeric separators or some way to signal true decimals as opposed to floats, etc. Any thoughts?

I’m sorry for all the back-and-forth @zdenko, you put in great effort with this PR 😄

Inve1951:
`1._23` parses correctly. It's CS sugar like `42.toString()` which has been around for ages but only rarely is used. I didn't think of it in #5215 and will change the test to explicitely allow for this (not as a number literal tho).

GeoffreyBooth:
And `42.toString()` compiles into `42..toString()` which . . . works. Somehow. I don’t quite follow how.

helixbass:
I'd also vote to close this PR, I don't think this syntax is very nice or adds much and I agree with the thoughts that it is hard to read/parse and potentially setting us up for conflicts with future syntax

> And `42.toString()` compiles into `42..toString()` which . . . works. Somehow. I don’t quite follow how.

@GeoffreyBooth I don't 100% understand this but I know JS is effectively parsing it like `(42.).toString()`

GeoffreyBooth:
So besides `BigInt` and numeric separators, there seem to be other [number-related proposals](https://neil188.github.io/07-11-2018/JavaScript-Numbers-proposals/) in various stages of progress. I think the prudent thing to do would be to close this, to avoid setting ourselves up for trouble in the event of a syntax clash with one of the future JavaScript enhancements related to numbers. I’m open to being persuaded otherwise, and we can also always reopen some time later once the proposals pipeline has slimmed down a bit and we see where things land with regards to numbers in JavaScript.

PR #5007, opened 2018-03-06
Feature test directories (alternative to PR 5003)

rdeforest:
This is an alternative to #5003 for making some tests not run (or even compile) when the runtime doesn't support a feature they rely on.

This change removes the need to edit Cakefile whenever tests are added for which rely on a feature not previously relied on. The tests need only be placed in a directory containing an index.coffee which tests for the feature they require.

The following pending PRs would benefit from this:
- #4893 
- #4877 
- #4880 
- #4884


GeoffreyBooth:
P.S. I don’t mean my closing this to be the final say on the matter. I’m happy to reopen if the other maintainers feel that this is the better of the two PRs. Just trying to de-clutter.

PR #5012, opened 2018-03-13
Make `node --harmony bin/cake test` pass on Node 9

rdeforest:
Make classMaker() explicitly return an object

Without this change, `node --harmony bin/cake test` the test "classes with value'd constructors" fails because classMaker returns a number.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - 6
   - 8
+  - 9
 
 cache:
   directories:
@@ -14,5 +15,6 @@ script:
   - node ./bin/cake build:full
   - node ./bin/cake build:browser
   - node ./bin/cake test
+  - node --harmony ./bin/cake test
   - node ./bin/cake test:browser
   - node ./bin/cake test:integrations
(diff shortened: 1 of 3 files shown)


rdeforest:
Should I also add

  - node --harmony ./bin/cake test

to appveyor.yml?

GeoffreyBooth:
We should add Node 9 for both AppVeyor and Travis. And harmony as one of the test commands we run. If harmony tests fail in Node < 9, then we should have harmony only run for Node 9 (because why would someone run an old version of Node in harmony mode?).

GeoffreyBooth:
P.S. See https://github.com/jashkenas/coffeescript/pull/4880/files

rdeforest:
Seeing my Cakefile patch in use in #4880 was the best birthday present I got yesterday. :)

Back on topic, harmony tests are currently passing on v6 and v8 for me. I agree in principle that there's little to no customer value in making bleeding-edge features tests run and/or pass on non-bleeding-edge runtimes, but until they actually fail I propose that they could serve as a kind of canary-in-a-coal-mine for other surprises. They might reveal obscure changes in older runtimes, dependencies or invalid assumptions in our feature tests.

PR #5016, opened 2018-03-15
Optimize 'for-range-by' loop when 'by' is literal number

### LINKED ISSUE #5015: Regression in generated output when using a fixed range
Older versions of CoffeeScript (I'm using version 2.1.1), given this input:

### Input Code
```coffee
a = []

for x in [0...a.length] by 1
	alert(x)
```

generate an optimized output for the "for" statement

### Expected Behavior
```js
  for (x = i = 0, ref = a.length; i < ref; x = i += 1) {
```

### Current Behavior
Newer versions of CoffeeScript (Sorry, I don't know when this new behaviour was introduced), generate some useless checks
```js
for (x = i = 0, ref = a.length; 1 !== 0 && (1 > 0 ? i < ref : i > ref); x = i += 1) {
```
The problem is that  1!==0 and 1 > 0 are checked at each loop iteration, and are completely useless, as they can be (and were) optimized at compile time.

### Possible Solution
Triage where this optimization has been cancelled out, and put it back in (not sure if possible, given the refactoring of the compiler), or implement again the optimization when the part after the "by" keyword is a constant.

zdenko:
Fixes #5015. Simplify and optimize `for-range-by` loop when `step` is a literal number.

```coffeescript
for x in [1..a] by 1
# for (x = i = 1, ref = a; i <= ref; x = i += 1)

for x in [a..1] by 1
# for (x = i = ref = a; i <= 1; x = i += 1)

for x in [a..b] by 1
# for (x = i = ref = a, ref1 = b; i <= ref1; x = i += 1)

for x in [1..2] by 1
# for (x = i = 1; i <= 2; x = i += 1)

for x in [1..a] by -1
# for (x = i = 1, ref = a; i >= ref; x = i += -1)

for x in [a..1] by -1
# for (x = i = ref = a; i >= 1; x = i += -1)

for x in [a..b] by -1
# for (x = i = ref = a, ref1 = b; i >= ref1; x = i += -1)

for x in [2..1] by -1
# for (x = i = 2; i >= 1; x = i += -1)
```

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1371,7 +1371,7 @@ exports.Range = class Range extends Base
     upperBound = "#{gt} #{ if known then to else @toVar }"
     condPart =
       if @step?
-        if @stepNum? and @stepNum != 0
+        if @stepNum? and @stepNum isnt 0
           if @stepNum > 0 then "#{lowerBound}" else "#{upperBound}"
         else
           "#{stepNotZero} && (#{stepCond} ? #{lowerBound} : #{upperBound})"
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1370,12 +1370,15 @@ exports.Range = class Range extends Base
     lowerBound = "#{lt} #{ if known then to else @toVar }"
     upperBound = "#{gt} #{ if known then to else @toVar }"
     condPart =
-       if @step?
-         "#{stepNotZero} && (#{stepCond} ? #{lowerBound} : #{upperBound})"
-       else
-         if known
-           "#{ if from <= to then lt else gt } #{to}"
-         else
+      if @step?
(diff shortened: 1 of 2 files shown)


src/nodes.coffee
```diff
@@ -1370,12 +1370,15 @@ exports.Range = class Range extends Base
...
     upperBound = "#{gt} #{ if known then to else @toVar }"
     condPart =
-       if @step?
-         "#{stepNotZero} && (#{stepCond} ? #{lowerBound} : #{upperBound})"
-       else
-         if known
-           "#{ if from <= to then lt else gt } #{to}"
-         else
+      if @step?
+        if @stepNum? and @stepNum != 0
```
GeoffreyBooth: `isnt`

PR #5025, opened 2018-03-29
2.2.4

GeoffreyBooth:
Hopefully the last bugfix release before we release 2.3.0 with the enhancement PRs. In 2.2.4:

*   When the `by` value in a `for` loop is a literal number, e.g. `for x in [2..1] by -1`, fewer checks are necessary to determine if the loop is in range. #5016, fix for #5015.
*   Bugfix for regression in 2.2.0 where a statement inside parentheses, e.g. `(fn(); break) while condition`, was compiling. Pure statements like `break` or `return` cannot turn a parenthesized block into an expression, and should throw an error. #5014, fix for #5013.


zdenko:
@GeoffreyBooth this LGTM.

PR #5043, opened 2018-04-27
2.3.0

GeoffreyBooth:
**This release adds support for all the new features and syntaxes in ES2018 that weren’t already possible in CoffeeScript.**

*   Asynchronous iterators are now supported. You can now `yield` an `await` call, e.g. `do -> until file.EOF then yield await file.readLine()`. #4893
*   Object splats/destructuring, a.k.a. object rest/spread syntax, has been standardized as part of ES2018 and therefore this release removes the polyfill that had previously been supporting this syntax. Code like `{a, b, rest...} = obj` now outputs more or less just like it appears, rather than being converted into an `Object.assign` call. Note that there are [some subtle differences](https://developers.google.com/web/updates/2017/06/object-rest-spread) between the `Object.assign` polyfill and the native implementation. #4884
*   The exponentiation operator, `**`, and exponentiation assignment operator `**=` are new to JavaScript in ES2018. Now code like `a ** 3` is output as it appears, rather than being converted into `Math.pow(a, 3)` as it was before. #4881
*   The `s` (dotAll) flag is now supported in regular expressions. #4880 

http://rawgit.com/GeoffreyBooth/coffeescript/2.3.0/docs/v2/


GeoffreyBooth:
I just tested this branch with my biggest app and it seemed to work fine, so hopefully this is good to go.

PR #5046, opened 2018-04-30
Fix #5034: Adjacent JSX elements must be wrapped in an enclosing tag

### LINKED ISSUE #5034: Bug: Last JSX tag returned, when compiler should throw error
Because of the parenthesis BOTH <Row> tags should be included in the return statement.  In error only the last one is:

CoffeeScript:
```coffee
		JSXFunction = -> (
				<Row>ABC</Row>
				<Row>DEF</Row>
			)
```

Errant JavaScript:
```js
// Generated by CoffeeScript 2.2.4
  JSXFunction = function() {
        <Row>ABC</Row>;
        return <Row>DEF</Row>;
      };
```

Correct JavaScript:
```js
  JSXFunction = function() {
        return <Row>ABC</Row>
            <Row>DEF</Row>;
      };
```

GeoffreyBooth:
```coffee
render = ->
  <Row>a</Row>
  <Row>b</Row>
```
```log
[stdin]:3:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
  <Row>b</Row>
   ^^^^^^^^^^^
```
Not sure why the initial `<` isn’t part of the error highlight, but that’s a separate bug. @xixixao

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -671,13 +671,22 @@
       // A Block node does not return its entire body, rather it
       // ensures that the final expression is returned.
       makeReturn(res) {
-        var expr, len;
+        var csxCheckIndex, expr, j, len, ref1;
         len = this.expressions.length;
         while (len--) {
           expr = this.expressions[len];
           this.expressions[len] = expr.makeReturn(res);
           if (expr instanceof Return && !expr.expression) {
             this.expressions.splice(len, 1);
+          }
+          // We also need to check that we’re not returning a CSX tag if there’s an
+          // adjacent one at the same level; JSX doesn’t allow that.
+          if (expr.unwrapAll().csx) {
+            for (csxCheckIndex = j = ref1 = len; (ref1 <= 0 ? j <= 0 : j >= 0); csxCheckIndex = ref1 <= 0 ? ++j : --j) {
+              if (this.expressions[csxCheckIndex].unwrapAll().csx) {
+                expr.error('Adjacent JSX elements must be wrapped in an enclosing tag');
+              }
+            }
           }
           break;
         }
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -482,6 +482,12 @@ exports.Block = class Block extends Base
       expr = @expressions[len]
       @expressions[len] = expr.makeReturn res
       @expressions.splice(len, 1) if expr instanceof Return and not expr.expression
+      # We also need to check that we’re not returning a CSX tag if there’s an
+      # adjacent one at the same level; JSX doesn’t allow that.
+      if expr.unwrapAll().csx
+        for csxCheckIndex in [len..0]
+          if @expressions[csxCheckIndex].unwrapAll().csx
+            expr.error 'Adjacent JSX elements must be wrapped in an enclosing tag'
       break
     this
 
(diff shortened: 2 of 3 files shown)


vendethiel:
It seems to check a bit too many cases, though:

```coffeescript
$ echo -e "->\n  <a></a>\n  foo\n  <b></b>" | ./bin/coffee -bcs
[stdin]:4:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
  <b></b>
   ^^^^^^
```

This is probably OK (why would you put a random toplevel element), but means we check every child of a block to check if it's a JSX element, if the returned value is a JSX element, which might be a bit wasteful.

GeoffreyBooth:
It only checks siblings if the value to be returned is a JSX element. So yes, it’s a little wasteful for whenever you’re returning a JSX element, but I don’t know how to avoid checking every element if siblings are disallowed.

But I think the bigger problem is that [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IAoCUFkD4IG8AoCCAHhB1IHoLiyAjSqxwgX0KA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2):
```jsx
var render = () => {
  <a></a>
  <b></b>
}
```
throws the `Adjacent JSX elements must be wrapped in an enclosing tag` error in Babel, whereas [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IAoCUFkD4IG8AoCCAHhB1IHoKBuYsgI0qufoF9Cg&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2):
```jsx
var render = () => {
  <a></a>;
  <b></b>;
}
```
does not. I guess it’s because the first example’s lack of semicolons makes it equivalent to [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IAoCUFkD4IG8AoCCAHhB1IHoLSAjSqhwgX0KA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2):
```jsx
var render = () => {
  <a></a><b></b>
}
```
which throws the error.

So what to do? It would appear that this PR has made our error checking too strict, since [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IAoCUFkD4IG8AoCCAHhB1IHoKBuYqaAFwFcxYyAjSq7-gX0JA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2):
```js
var render = () => {
  <a></a>;
  return <b></b>;
}
```
should be allowable JSX, at least if Babel is to be trusted.

vendethiel:
Well, we can loop from 1 to length, and if both `nodes[i]...csx && nodes[i - 1]...csx`, we have found adjacent tags.

PR #5048, opened 2018-05-02
Normalize :: syntax

helixbass:
@GeoffreyBooth a couple Prettier-related changes to try and make generation of `::` constructs safer:
- Currently, `.`, `?.` and `::` are `LINE_CONTINUER`s but `?::` is not. So add it so that line-breaking before any kind of `Accessor` (other than a "computed" one, ie an `Index` eg `['a']`) should be safe
- Currently, there is "standalone" `::` (ie just referencing the `.prototype` object) as well as `::` followed by a `Property`. `?::` has no standalone version in the grammar. So add it, since currently in Prettier I'm auto-transforming *any* `.prototype` access into `::`, so if the original source were eg `a?.prototype` it'd generate `a?::`, which currently won't compile. If it ends up making sense to only generate `::` when explicit `::` was in the original source, that should just be a question of eg adding a `prototypeShorthand: yes` flag to the `Access` in the grammar, but there's no harm in allowing standalone `?::` regardless


helixbass:
@connec looks like there were existing checks that `?::` behaves as expected in `test/operators.coffee`, so i expanded that test to test that `?::` behaves the same way when continuing a line

PR #5049, opened 2018-05-02
Fix #5046: Adjacent JSX

zdenko:
Fixes the issue in previous PR (#5046).
In `Block::makeReturn`, the last two expressions are checked before the main loop starts.
Instead of `unwrapAll()`, I used `unwrap()` to avoid false errors with `foo <a></a>` and when explicit `return` is used.

```coffeescript
<a></a>
foo <b></b>
###
<a></a>;
foo(<b></b>);
###
```
---
```coffeescript
render = ->
  <a></a>
  return <b></b>
###
render = function() {
  <a></a>;
  return <b></b>;
};
###
```


GeoffreyBooth:
What about this case?

```coffee
render = -> (
  <Row>ABC</Row>
  <Row>DEF</Row>
)
```
The parentheses get removed as redundant, and then `<Row>DEF</Row>` gets returned. That’s clearly not the user’s intent, and Babel throws an error on what I would think is the [equivalent](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IDMCusDGAXAlgPawAUAlBAN4BQEU0uGYsEJtdEAPAEqEDuAPgCCAIQDCnAPS9BAbnZ0e_AQBEAogDEpMgfLplqAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=react&prettier=false&targets=&version=6.26.0&envVersion=) of what the user had in mind:

```jsx
var render = function() {
  return (
    <Row>ABC</Row>;
    <Row>DEF</Row>;
  )
}
```

Curiously though, this is the error:
```error
repl: Unexpected token, expected , (3:18)
  1 | var render = () => {
  2 |   return (
> 3 |     <Row>ABC</Row>;
    |                   ^
  4 |     <Row>DEF</Row>;
```

GeoffreyBooth:
Also, this:
```coffee
render = ->
  <Row>ABC</Row>
  <Row>DEF</Row>
```
_should_ compile, as it would become [this](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IDMCusDGAXAlgPawAUAlBAN4BQEEAPAEqEDuAfAIIBCAwvQPTN2AblpRouDGFgMhbACIBRAGIC5ogL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=react&prettier=false&targets=&version=6.26.0&envVersion=):
```jsx
var render = function() {
  <Row>ABC</Row>;
  return <Row>DEF</Row>;
}
```
which is valid JSX.

zdenko:
User from #5034 wants this
```coffeescript
JSXFunction = -> (
  <Row>ABC</Row>
  <Row>DEF</Row>
)
```

to compile into this

```javascript
JSXFunction = function() {
  return <Row>ABC</Row>
     <Row>DEF</Row>;
};
```

But, this is an error. Your PR solved this, but was to greedy and was throwing and error even if `return` was explicit, e.g.

```coffeescript
render = -> (
  <Row>ABC</Row>
  return <Row>DEF</Row>
)
```

My PR only corrects greediness.

zdenko:
Besides, this

```coffeescript
render = ->
  <Row>ABC</Row>
  <Row>DEF</Row>
```

already [compiles](http://coffeescript.org/#try:render%20%3D%20-%3E%0A%20%20%3CRow%3EABC%3C%2FRow%3E%0A%20%20%3CRow%3EDEF%3C%2FRow%3E) on the `master`

```javascript 
render = function() {
  <Row>ABC</Row>;
  return <Row>DEF</Row>;
};
```

GeoffreyBooth:
The non-parentheses version compiles today in 2.3.0, but my PR made it break. It _should_ compile, because it’s valid JSX.

zdenko:
OK, So only JSX in parentheses should throw an error?

```coffeescript
render = -> (
  <Row>ABC</Row>
  <Row>DEF</Row>
)
```

Since CS [removes](http://coffeescript.org/#try:foo%20%3D%20-%3E%20(%0A%20%20%20a%0A%20%20%20b%0A)) parentheses in any case, JSX will get special treatment.

And, what about this case?

```coffeescript
render = -> (
  a += 1
  <Row>ABC</Row>
  <Row>DEF</Row>
)
```

I guess we could inspect expressions in the parentheses and only if JSX is present we would throw an error.

GeoffreyBooth:
I’m not sure what to do. I just want to make sure we at least don’t throw errors on _valid_ JSX.

zdenko:
The only way I see is checking parentheses.
And, we can treat any expression inside the same way as [Babel](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IDMCusDGAXAlgPawAUAlBAN4BQEEokAHgNy1TS4ZiwQkRt1GEANSoAjAIgAeEAD4pAJUIB3WQEEAQgGEpAeiWq2ASEUrZAEQCiAMT0H5uuWzLUAvkA&debug=true&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=Node-9.8&version=6.26.0&envVersion=1.6.2) does.

zdenko:
The last commit fixes cases discussed above.

Valid CSX:
```coffeescript
render = ->
  <Row>ABC</Row>
  <Row>DEF</Row>

###
render = function() {
  <Row>ABC</Row>;
  return <Row>DEF</Row>;
};
###
```
---
```coffeescript
<a></a>
foo <b></b>

###
<a></a>;
foo(<b></b>);
###
```

Compiler throws an error when adjacent CSX in parentheses is not wrapped in an enclosing tag.

```coffeescript
render = -> (
  <Row>ABC</Row>
  <Row>DEF</Row>
)

###
error: Adjacent JSX elements must be wrapped in an enclosing tag
  <Row>DEF</Row>
   ^^^^^^^^^^^^^
###
```
---

Only last two expressions are checked.

```coffeescript
render = -> (
  a = 2
  <Row>ABC</Row>
  <Row>DEF</Row>
)

###
error: Adjacent JSX elements must be wrapped in an enclosing tag
  <Row>DEF</Row>
   ^^^^^^^^^^^^^
###
```

This will compile.

```coffeescript
render = -> (
  <Row>ABC</Row>
  a = 2
  <Row>DEF</Row>
)

###
render = function() {
  var a;
  <Row>ABC</Row>;
  a = 2;
  return <Row>DEF</Row>;
};
###
```

vendethiel:
I think even the non-parenthesized version should break. It will never be what the user meant, Afaik. The way CoffeeScript works makes this a bit ambiguous maybe... I think we had a similar thing with

```
->
  a: 1
  b: 2
```

zdenko:
>  It will never be what the user mean

Agree. Even the user from #5034 expected something else.
The question is how to deal with mixed expressions:

```coffeescript
render = ->
   x = 1
   y = 2
  <Row>ABC</Row>
  <Row>DEF</Row>
```
 
I think in such cases we should return the last expression, e.g., no error.

GeoffreyBooth:
So what else needs to be done to wrap up this PR?

zdenko:
As per @vendethiel's suggestion, I would treat 

```coffeescript
render = ->
  <Row>ABC</Row>
  <Row>DEF</Row>
``` 

in the same way as 

```coffeescript
render = -> (
  <Row>ABC</Row>
  <Row>DEF</Row>
)
```

However, this case would return the last expressions
```coffeescript
render = ->
   x = 1
   y = 2
   <Row>ABC</Row>
   <Row>DEF</Row>

###
render = function() {
  var x, y;
  x = 1;
  y = 2;
  <Row>ABC</Row>;
  return <Row>DEF</Row>;
};
###
```

GeoffreyBooth:
Isn’t that the current behavior before we started changing things?

zdenko:
@GeoffreyBooth it was, but then your PR changed it and in this PR you [commented](https://github.com/jashkenas/coffeescript/pull/5049#issuecomment-386039511) that non-parenthesized JSX should compile.

GeoffreyBooth:
Right, so what's the difference between this PR and reverting mine?

vendethiel:
> I think in such cases we should return the last expression, e.g., no error.

Why? I don't think that makes sense.

Inve1951:
@vendethiel It's valid coffee/js/jsx code and coffee implicitely returns the last statement of a function body. JSX tags are still only js objects. `<Row>ABC</Row>` is same as `React.createElement(Row, null, "ABC")`.

I don't see the need for special treatment either but it's true that it's unlikely that the user ment for the code to turn out that way.
I see 2 solutions:
* Ignore it and have ESlint deal with it.
* Implicitely wrap the JSX tags into [Fragments](https://reactjs.org/docs/fragments.html).

zdenko:
@GeoffreyBooth 

```coffeescript
->
  <a></a>
  foo
  <b></b>
```

Your PR:

```
[stdin]:5:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
  <b></b>
   ^^^^^^
```

This PR:

```javascript
(function() {
  <a></a>;
  foo;
  return <b></b>;
});
```

GeoffreyBooth:
@zdenko I get that, what I mean is, what's the difference between current 2.3.0 (a.ka. reverting my PR to the prior state of `master`) and merging this PR in?

vendethiel:
> @vendethiel It's valid coffee/js/jsx code and coffee implicitely returns the last statement of a function body. JSX tags are still only js objects. <Row>ABC</Row> is same as React.createElement(Row, null, "ABC").

In general I'd agree, but here we're just copying an error that's reported by babel; because the error is lost when compiling.

zdenko:
I get an error on [Babel](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IDMCusDGAXAlgPawAUAlBAN4BQEU0uGYsEJdtdEAPAEqEDuAPgCCAIQDCXAPR8hAbmoBIXgMEARAKIAxabMELOZBQF8gA&debug=true&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=Node-9.8&version=6.26.0&envVersion=1.6.2) for 
```javascript
var render = function() {
  return (
    <Row>ABC</Row>;
    <Row>DEF</Row>;
  )
}
```

This [works](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=G4QwTgBGCmB2Am1IF4IDMCusDGAXAlgPawAUAlBAN4BQEU0uGYsEJdtdEAPAEqEDuAPgCCAIQDCXAPR8hAGmoBIXgMEARAKIAxabMEcIZANzUAvkA&debug=true&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=react%2Cenv&prettier=false&targets=Node-9.8&version=6.26.0&envVersion=1.6.2)
```javascript
var render = function() {
  return (
    <Row>ABC</Row>,
    <Row>DEF</Row>
  )
}
```

If we revert your PR and merge this one, we'll (currently) get an error only if JSX in enclosed in parentheses.

```coffeescript
-> (
  <a></a>
  <b></b>
)
###
[stdin]:4:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
  <b></b>
   ^^^^^^
###
```

Non-parenthesized JSX will compile:

```coffeescript
->
  <a></a>
  <b></b>
###
(function() {
  <a></a>;
  return <b></b>;
});
###
```

GeoffreyBooth:
Okay, so regarding what _would_ trigger an error in this PR:

```coffee
-> (
  <a></a>
  <b></b>
)
```
That’s equivalent to [this JS](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=BQMwrgdgxgLglgewsAlAAgN4Cg1oE4CmMYeEawOuaAPAIYB81A9AwNyW7UBGjTP7uFFgC-KIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=true&presets=react&prettier=false&targets=&version=6.26.0&envVersion=):
```jsx
(function() {
  return (
    <a></a>;
    <b></b>;
  )
})
```
Which throws this error in Babel:
```
repl: Unexpected token, expected , (3:11)
  1 | (function() {
  2 |   return (
> 3 |     <a></a>;
    |            ^
```
Which isn’t really equivalent to “Adjacent JSX elements must be wrapped in an enclosing tag.” Do we care?

vendethiel:
It’s not equivalent because of our everything-is-an-expression semantics. I think we do care.

GeoffreyBooth:
Maybe “do we care” is a bit harsh 😄 I guess the question is what we consider
```coffee
(
  <a />
  <b />
)
```
to be equivalent to, since it _is_ an expression.

Some more ways of putting it, in JS, with the errors from Babel:
```jsx
var someJsx = (
  <a />
  <b />
);
```
```
repl: Adjacent JSX elements must be wrapped in an enclosing tag (3:2)
  1 | var someJsx = (
  2 |   <a />
> 3 |   <b />
    |   ^
```
and
```jsx
var someJsx = (
  <a />;
  <b />;
);
```
```
repl: Unexpected token, expected , (2:7)
  1 | var someJsx = (
> 2 |   <a />;
    |        ^
```
and finally,
```jsx
var someJsx = (
  <a />,
  <b />
);
```
compiles successfully, into
```js
var someJsx = (React.createElement("a", null), React.createElement("b", null));
```
So in CoffeeScript, each newline triggers a semicolon; and there doesn’t seem to be an equivalent for the comma operator or a way to trigger its output. ([This is worth reading.](https://stackoverflow.com/a/10284572/223225))

I think Babel’s `expected ,` error, asking for the comma operator, doesn’t make sense for CoffeeScript to output since we don’t have that operator (I think). And I think Babel is only asking for that because asking for a semicolon is pointless here, since the semicolon version throws a different error. Really though I think Babel asking for a comma is misguided, since I don’t see the sense in using that operator here. All it means in this case, to use the last example, is that `someJsx` gets the value of `React.createElement("b", null)`, which doesn’t seem terribly useful. I think the `Adjacent JSX elements must be wrapped` error is actually more informative, as it’s closer to the source of the problem. @vendethiel what do you think?

test/error_messages.coffee
```diff
@@ -1664,25 +1664,26 @@ test 'CSX error: invalid attributes', ->
 
 test '#5034: CSX error: Adjacent JSX elements must be wrapped in an enclosing tag', ->
   assertErrorFormat '''
-    render = ->
+    render = -> (
```
vendethiel: I think we should test that it errors out both with and without parens

vendethiel: I talked about this a bit earlier.
You could say that, but you could also say that

```coffeescript
->
  <a />
  <b />
```
really is
```coffeescript
->
  return (
    <a />
    <b />
  )
```
just like
```coffeescript
->
  a: 1
  b: 2
```
is
```coffeescript
->
  return {
    a: 1
    b: 2
  }
```
instead of
```coffeescript
->
  {a: 1}
  return {b: 2}
```

GeoffreyBooth: I don’t think you could say that parentheses are implied. JSX objects on subsequent lines aren’t assumed to be part of a unified construct the way an object is. They don’t compile into a single entity, for one thing, even with parentheses:

```coffee
jsx = (
  <a />
  <b />
)
```
```jsx
jsx = (<a />, <b />);
```

That’s not the way objects behave. The separate keys of an object are unified by virtue of being all part of the same block, as defined by indentation. Adjacent JSX tags are just that, adjacent JSX tags, unless they’re enclosed by a parent tag or fragment.

vendethiel: OK. I also want to point out that this:

> I agree it’s likely a mistake on the user’s part, but catching “most of the time this is wrong” errors is the job of a linter, not the compiler.

Could very well apply to the original warning. Babel could compile it to two createElement doing nothing.

GeoffreyBooth: > Could very well apply to the original warning. Babel could compile it to two createElement doing nothing.

It could, but the JSX spec tells it not to. I think that's the distinction.

test/error_messages.coffee
```diff
@@ -1664,25 +1664,26 @@ test 'CSX error: invalid attributes', ->
...
     render = -> (
+      a = "foo"
       <Row>a</Row>
       <Row>b</Row>
     )
   ''', '''
-    [stdin]:3:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
+    [stdin]:4:4: error: Adjacent JSX elements must be wrapped in an enclosing tag
       <Row>b</Row>
        ^^^^^^^^^^^
```
GeoffreyBooth: @zdenko We should also add a test for the non-error case:

```coffee
->
  <a />
  <b />
```

That would be a good place to put a comment with a link to this PR, and an explanation for why this shouldn’t error.

GeoffreyBooth:
@zdenko @vendethiel I added the last test that I had requested. I’m eager to wrap this up. Any more notes?

zdenko:
@GeoffreyBooth LGTM

PR #5050, opened 2018-05-02
WIP: Output strings with newlines as template literals

zdenko:
Fixes (partially) #5019.
Currently, only block strings with interpolation are covered as other cases break code in many places.
This works:
```coffeescript
s = """
  
  #{a}
  
      c

"""

###
s = `
${a}

    c
`;
###
```
---
```coffeescript
s = """\n
  \n
  #{a}
  
      c
\n

  \nd

"""

###
s = `\n
\n
${a}

    c
\n

\nd
`;
###
```

I think strings without interpolation won't work. For example, this code from the `Cakefile`

```coffeescript
execSync '''
  cake build:full
  cake build:browser
  cake test:browser
  cake test:integrations
  cake doc:site
  cake doc:test
  cake doc:source'''
```

isn't working if compiled into

```javascript
execSync(`
cake build:full
cake build:browser
cake test:browser
cake test:integrations
cake doc:site
cake doc:test
cake doc:source`)
```

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -341,6 +341,12 @@
         index: end
       } = this.matchWithInterpolations(regex, quote));
       $ = tokens.length - 1;
+      // nlRegEx = """(\n|\r|\u2028|\u2029)"""
+      // str = @chunk[quote.length..end-4].replace ///^#{nlRegEx}///, ""
+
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Any reason to keep this PR open? It seems like #5057 replaces it and isn’t WIP, so we should close this one in favor of that one?

PR #5053, opened 2018-05-04
Allow implicit call with class with no body

### LINKED ISSUE #5052: Bug: implicit call of class with no body fails
Expected behavior:
Should be able to do an implicit call of a class without a body, as you can do an implicit call of a class with a body

eg this fails to compile (the rewriter generates an implicit opening `(` but not the corresponding `)`):
```
f class A
```

Will open a fixing PR momentarily

helixbass:
Fixes #5052 

Tweaks `addImplicitBracesAndParens()` to correctly close an inline implicit call which has an argument  that's a class with no body

So now eg `f class A` compiles

The issue was that it was generating the implicit opening `(` but failing to generate the closing `)` at the end of the line since when it saw `class` it pushed a `CONTROL` token that expected to see an `INDENT`-`OUTDENT` pair before unblocking the closing of the implicit call

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -542,6 +542,8 @@
               // the continuation of an object.
               } else if (inImplicitObject() && tag === 'TERMINATOR' && prevTag !== ',' && !(startsLine && this.looksObjectish(i + 1))) {
                 endImplicitObject();
+              } else if (inImplicitControl() && tokens[stackTop()[1]][0] === 'CLASS' && tag === 'TERMINATOR') {
+                stack.pop();
               } else {
                 break;
               }
--- a/src/rewriter.coffee
+++ b/src/rewriter.coffee
@@ -379,6 +379,8 @@ exports.Rewriter = class Rewriter
           else if inImplicitObject() and tag is 'TERMINATOR' and prevTag isnt ',' and
                   not (startsLine and @looksObjectish(i + 1))
             endImplicitObject()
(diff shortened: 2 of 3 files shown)


Inve1951:
this syntax for a function call used in the test is currently a hack
```coffee
f
  b: 1
  class A
  a
```

helixbass:
@Inve1951 what do you mean exactly? I included that test case to check that in a multiline implicit call (with an implicit object first argument) the new code didn't trigger a premature ending of the implicit call

helixbass:
@Inve1951 yup the fact that an indented implicit object as the first arg can start an implicit call is a special case. That's why I thought it was worth testing since it's a different way for an implicit call and an "inline" class to interact

GeoffreyBooth:
@Inve1951 or anyone else, is this okay to merge in?

Inve1951:
I'm good with it. My only concern is that the test relies on the implicit object starting a call, which I'm not sure about whether that's a supported feature or not. But I'm certainly in favor to have that as a language feature. In fact I'd expect the same from the broken cases I demonstrated in my previous comment.

vendethiel:
> My only concern is that the test relies on the implicit object starting a call, which I'm not sure about whether that's a supported feature or not.

It's very much a special case.  Your "broken cases" do not compile on purpose.

GeoffreyBooth:
@vendethiel are there any issues with this PR, or is it okay to merge in?

vendethiel:
I'm very bad at rewriter stuff, it the tests pass then I'd say go for it.

PR #5056, opened 2018-05-06
fix JSX expression indentation bug

### LINKED ISSUE #5055: Bug: reasonable JSX indentation fails
Even though you can use `if` in Coffeescript instead of the canonical-for-standard-JSX `&&`, this is still something people might reasonably want to be able to write, and it currently fails:
```
<div>
  {someCondition &&
    <span />
  }
</div>
```

Will open a fixing PR momentarily

helixbass:
Fixes #5055 

So now eg this compiles:
```
<div>
  {someCondition &&
    <span />
  }
</div>
```

Basically it was choking on the closing empty `INDENT`/`OUTDENT` pair at the end of the parsed tokens inside the nested `{...}` expression

So in `matchWithInterpolations()`, similar to the existing tweaking of `nested` to remove a leading `TERMINATOR`, added a tweaking to remove a closing `INDENT`/`OUTDENT` pair

### DIFF
--- a/test/csx.coffee
+++ b/test/csx.coffee
@@ -799,3 +799,27 @@ test '#5055: JSX expression indentation bug', ->
       {someCondition && <span />}
     </div>;
   '''
+
+  eqJS '''
+    <div>
+      {someString +
+          "abc"
+      }
+    </div>
+  ''', '''
+    <div>
+      {someString + "abc"}
+    </div>;
+  '''
+
+  eqJS '''
+    <div>
+      {a ?
+      <span />
+      }
+    </div>
+  ''', '''
+    <div>
+      {typeof a !== "undefined" && a !== null ? a : <span />}
+    </div>;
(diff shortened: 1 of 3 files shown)


test/csx.coffee
```diff
@@ -786,3 +786,16 @@ test 'JSX fragments: fragment with component nodes', ->
...
+    <div>
+      {someCondition &&
+        <span />
+      }
+    </div>
+  ''', '''
+    <div>
+      {someCondition && <span />}
+    </div>;
+  '''
```
GeoffreyBooth: How about a test that confirms that this works with a few other types of line continuators, besides just `&&`? And what if the next line is at the same indentation, and what if it’s normal code rather than a CSX tag?

helixbass: @GeoffreyBooth sure, added more tests with those variants

Issue #5036, opened 2018-04-18
Unexpected token <


iamMonzurul:
Choose one: is this a bug report or feature request?

### Input Code

```coffee
reactDom.render  <ViewContainer opts... />, eleRef
```

### Context

I am trying to render ViewContainer component but while I try to compile it, I am getting unexpected "<".

### Environment

* CoffeeScript version: 2.2.4
* Node.js version: 8

vendethiel:
This is the error message I get:

```
Unexpected token. Allowed CSX attributes are: id="val", src={source}, {props...} or attribute.
reactDom.render  <ViewContainer opts... />, eleRef
```

iamMonzurul:
Could you please tell me what would be the correct way to write that line? @vendethiel

vendethiel:
The syntax, it seems for both JSX *and* CoffeeScript, is `<ViewContainer {opts...} />`, as the error message points out.

iamMonzurul:
`<ViewContainer {opts...} />, `

In that case I am getting "expected end of line" error in webstorm.

vendethiel:
We do not support the CoffeeScript WebStorm plugin, you'll have to take that issue to JetBrains. Seems like it works fine on CoffeeScript's end.

iamMonzurul:
Well, I understand you guys don't support webstorm plugin but if you can go to http://js2.coffee and put this line `reactDom.render  <ViewContainer {props...} />, eleRef` you will see it's not a problem with webstorm.

Let me know if you have any explanation on that. I would really appreciate that.

vendethiel:
js2coffee hasn't been updated (except for deps) since mid 2016. So it doesn't support CSX.
You'll have to take this with the js2coffee people.
See https://github.com/js2coffee/js2coffee/issues/369.

GeoffreyBooth:
If you’re looking for an online CoffeeScript testing environment, the CoffeeScript webpage has one called Try CoffeeScript that is always up to date:

http://coffeescript.org/#try:reactDom.render%20%3CViewContainer%20%7Bprops…%7D%20%2F%3E%2C%20eleRef

PR #5057, opened 2018-05-07
Preserve original string literal

helixbass:
@GeoffreyBooth @zdenko I started on being able to preserve original string formatting into the nodes tree (eg for passing to Prettier). Haven't handled preserving heregex formatting yet but this should currently be preserving relevant string/heredoc info and all existing tests are passing (had to modify some since now a `STRING` token doesn't include the quotes etc)

The basic approach I took to getting the extra data through the grammar was similar to how we expose `generated` to the grammar for object braces: wrap a token-value string (`token[1]`) in a `new String()` object, at which point you can set arbitrary additional properties that will be visible on the token that the grammar sees (eg `$1`)

So far haven't tried to emit (via AST)/generate (via Prettier) the original quote eg `'''`/`"""` (or get it to shift the heredoc indentation level along with the surrounding indentation level, which should be fun), but I have made it so that for a `StringLiteral` that contains a newline, it generates a `TemplateLiteral` AST with the original string value which Prettier then reprints with original spacing/linebreaks intact (on my `prettier` branch/`prettier-plugin-coffeescript` repo)


src/helpers.coffee
```diff
@@ -233,3 +233,28 @@ exports.nameWhitespaceCharacter = (string) ->
     when '\r' then 'carriage return'
     when '\t' then 'tab'
     else string
+
+exports.dump = (...args, obj) ->
+  util = require 'util'
+  console.log ...args, util.inspect obj, no, null
```
GeoffreyBooth: We put the dots on the right in the CoffeeScript codebase; but besides, I don’t think we want debugging helpers in the codebase.

src/grammar.coffee
```diff
@@ -166,8 +166,8 @@ grammar =
   ]
 
   String: [
-    o 'STRING',                                 -> new StringLiteral $1
-    o 'STRING_START Body STRING_END',           -> new StringWithInterpolations $2
+    o 'STRING',                                 -> new StringLiteral $1, quote: $1.quote, initialChunk: $1.initialChunk, finalChunk: $1.finalChunk, indent: $1.indent, double: $1.double
```
GeoffreyBooth: So this is the method for passing extra data through the parser? Clever, I can get behind this.

This particular object is ambiguous though, it should have braces.

src/lexer.coffee
```diff
@@ -870,8 +885,14 @@ exports.Lexer = class Lexer
...
           if i is 2 and firstEmptyStringIndex?
             @tokens.splice firstEmptyStringIndex, 2 # Remove empty string and the plus.
+          markInitialChunk token if i is 0
+          markFinalChunk token if i is $
+          markIndent token
+          attachQuote token
+          attachDouble token
           token[0] = 'STRING'
-          token[1] = @makeDelimitedLiteral converted, options
+          # token[1] = preserveTagged(token) @makeDelimitedLiteral converted, options
```
GeoffreyBooth: Leftover?

src/nodes.coffee
```diff
@@ -747,13 +747,48 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
...
+  constructor: (@originalValue, {@quote, @initialChunk, @finalChunk, @indent, @double} = {}) ->
+    super ''
+    @fromSourceString = @quote?
+    @quote ?= '"'
+    @originalValue = unwrapStringObject @originalValue
+    @formatValue()
+
+  formatValue: ->
+    heredoc = @quote.length is 3
+    @value = do =>
```
GeoffreyBooth: Why the wrapper of `do =>`?

src/nodes.coffee
```diff
@@ -747,13 +747,48 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
...
+      makeDelimitedLiteral val, {
+        delimiter: @quote.charAt 0
+        @double
+      }
+
   compileNode: (o) ->
-    res = if @csx then [@makeCode @unquote(yes, yes)] else super()
+    return [@makeCode @unquote(yes, yes)] if @csx
+    # @formatValue()
+    super o
```
GeoffreyBooth: Cleanup?

src/nodes.coffee
```diff
@@ -3430,8 +3465,9 @@ exports.Parens = class Parens extends Base
 #### StringWithInterpolations
 
 exports.StringWithInterpolations = class StringWithInterpolations extends Base
-  constructor: (@body) ->
+  constructor: (@body, {@quote} = {}) ->
     super()
+    # dump {@quote}
```
GeoffreyBooth: Cleanup

src/nodes.coffee
```diff
@@ -3484,6 +3520,7 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
     fragments.push @makeCode '`' unless @csx
     for element in elements
       if element instanceof StringLiteral
+        # element.formatValue()
```
GeoffreyBooth: Cleanup

src/nodes.coffee
```diff
@@ -3894,3 +3934,6 @@ unfoldSoak = (o, parent, name) ->
   parent[name] = ifn.body
   ifn.body = new Value parent
   ifn
+
+unwrapStringObject = (str) ->
+  str.valueOf?() ? str
```
GeoffreyBooth: Shouldn’t this be implemented as `unwrap` on the `StringLiteral` class?

src/rewriter.coffee
```diff
@@ -41,6 +41,7 @@ exports.Rewriter = class Rewriter
   # like this. The order of these passes matters—indentation must be
   # corrected before implicit parentheses can be wrapped around blocks of code.
   rewrite: (@tokens) ->
+    # dump {@tokens}
```
GeoffreyBooth: Cleanup

test/location.coffee
```diff
@@ -123,7 +123,7 @@ test 'Verify locations in string interpolation (in "string", multiple interpolat
   eq c[2].last_column, 2
 
 test 'Verify locations in string interpolation (in "string", multiple interpolation and starting with line breaks)', ->
-  [a, b, c] = getMatchingTokens '"\n#{a}\nb\n#{c}"', 'a', '" b "', 'c'
+  [a, b, c] = getMatchingTokens '"\n#{a}\nb\n#{c}"', 'a', '\nb\n', 'c'
```
GeoffreyBooth: `"\nb\n"` needs to still compile as `" b "`. To do otherwise would be a breaking change.

test/location.coffee
```diff
@@ -436,6 +436,9 @@ test "#3822: Simple string/regex start/end should include delimiters", ->
   eq regexToken[2].last_line, 0
   eq regexToken[2].last_column, 6
 
+strVal = (str) ->
+  str.valueOf?() ? str
```
GeoffreyBooth: Cleanup

src/nodes.coffee
helixbass: @GeoffreyBooth I'm a fan of this pattern as a way of mini-encapsulating multi-step logic for calculating a single value - by nesting it inside `do =>`, you know that everything inside is geared towards calculating `@value`

You could also extract the value calculation into a named method, but here the logic feels like it belongs inline inside `formatValue()`. Happy to refactor to just assign to `@value` at the end if this pattern feels weird

src/nodes.coffee
helixbass: @GeoffreyBooth I'm not sure this should be conflated with our concept of `unwrap()`, this is just ensuring that we normalize `String` objects to primitive strings (since I think I ran into Prettier not treating a `String` object the same) - updated the name of the helper to `normalizeStringObject()`

test/location.coffee
helixbass: @GeoffreyBooth yup so the difference is at the token level, since the original formatting is now preserved (rather than already transformed by the lexer before creating the tokens). Since these tests are checking against `CoffeeScript.tokens()` directly, they had to be updated to look for different `token[1]` values

Then it's in `StringLiteral::formatValue()` that the originally-formatted string `@originalValue` (which also doesn't have surrounding quotes included) gets transformed and quoted (eg here into `" b "`) and assigned to `@value`

So it still *compiles* to `" b "`, we have lots of tests for various string types that would be failing if that weren't the case

test/location.coffee
helixbass: @GeoffreyBooth this one is actually getting used (see below), since `eq()` was failing when comparing a primitive string to a `String()` object

It is equivalent to the `unwrapStringObject()` (now `normalizeStringObject()`) helper though, I don't know if there's an existing way to share a helper between `src/` and tests?

src/helpers.coffee
```diff
@@ -233,3 +233,28 @@ exports.nameWhitespaceCharacter = (string) ->
     when '\r' then 'carriage return'
     when '\t' then 'tab'
     else string
+
+exports.dump = (args..., obj) ->
+  util = require 'util'
+  console.log args..., util.inspect obj, no, null
```
helixbass: > We put the dots on the right in the CoffeeScript codebase; but besides, I don’t think we want debugging helpers in the codebase.

@GeoffreyBooth ok updated dots - can already picture having a linting config in the codebase with rules like this...

As far as including this helper, I've been using it constantly on my `prettier` branch while developing, but obviously it's not needed in production. The argument I'd see in favor of including it would be to reference it in the docs for hacking on the compiler ("a convenient way to inspect eg a node object is to use `helpers.dump()`"). But it's a simple enough util to recreate manually when needed and tbh I don't know what's standard practice here

src/grammar.coffee
```diff
@@ -166,8 +166,8 @@ grammar =
   ]
 
   String: [
-    o 'STRING',                                 -> new StringLiteral $1
-    o 'STRING_START Body STRING_END',           -> new StringWithInterpolations $2
+    o 'STRING',                                 -> new StringLiteral $1, {quote: $1.quote, initialChunk: $1.initialChunk, finalChunk: $1.finalChunk, indent: $1.indent, double: $1.double}
```
helixbass: > So this is the method for passing extra data through the parser? Clever, I can get behind this.
> This particular object is ambiguous though, it should have braces.

@GeoffreyBooth yup credit to whoever figured out how to expose `.generated` to the grammar, that's where I got the pattern of wrapping in a `new String()`

Ok added explicit braces - for the sake of my thinking about Prettier formatting, can you try and explain why you consider this trailing object ambiguous while the one on the next line (`quote: $1.quote`) is ok?

GeoffreyBooth: The next line isn’t great either 😉 Technically the next line isn’t ambiguous because it doesn’t have any commas: it’s an object with just one key. The line I objected to was an inline object with multiple key-value pairs separated by commas, and it’s ambiguous (to a human reader, at least) whether those commas refer to the object or to arguments of the `new StringLiteral` constructor/function call. But I’m not a big fan of implicit inline objects in function calls, there’s just so much already going on with a function call that braces are generally helpful here.

src/helpers.coffee
GeoffreyBooth: So far we haven’t been including debugging helpers, I think probably because everyone has their own opinion as to how they’d want their `dump` function to behave and that just leads to chaos. There is one exception, in that there’s a built-in debugging logger at the top of `rewriter.coffee`, and I confess I had a hand in that one; it used to be just a comment (“uncomment the next line to print the token stream”) and I added the environment variables so that all one had to do to enable the debugging output was set the environment variable rather than uncomment the line, because I was afraid of accidentally committing the uncommented line back into the codebase.

So I can’t really talk 😄but personally I’d prefer to hold the line at the rewriter debugger, and avoid putting a totally generic “print any object” debugging helper in the codebase, especially since it gets included all over the place and it’s an invitation for people to bikeshed and reimplement it according to their own preferences. But my preference doesn’t need to be the final say here, if others feel strongly otherwise.

GeoffreyBooth:
I was just about to suggest you unwrap the `String` objects in `grammar.coffee` like `generated` is, but you beat me to it 😄

Please take a look at #5063 and hopefully approve merging it in, it has some overlap with what you’re working on here. I’m fine with changing the signature of `makeToken` to take an object, but I’d like to merge #5036 in first and then you can resolve the conflicts as you merge that into this branch.

helixbass:
@GeoffreyBooth ok sure will look at #5063 in a few

Ya I took your suggestions on #5059 and started applying them on this branch (wasn't going to make sense to update both this and that one in parallel, so started here and then rebased #5059 onto this branch), I think we're landing on a much cleaner approach - using `"#{str}"` instead of `normalizeStringObject()` is a lot lighter weight, good call

So for now #5059 will show you the diff of this branch as well as its own - I guess I (or you since you have permissions) could push this branch to `origin` and then retarget #5059 against it for the time being to keep the diff there to just the relevant changes?

GeoffreyBooth:
I think it would be easiest to just do things in sequence:
- Merge #5063 into `master`
- Merge `master` into this branch, resolve conflicts
- Finish up this branch/PR, merge it into `master`
- Merge `master` into #5059, resolve conflicts
- Finish up #5059, merge it into `master`

There aren’t any breaking changes in any of the three branches, so the changes should be irrelevant to 2.3.1 or any upcoming releases. It’s all just under-the-hood refactoring.

Cakefile
```diff
@@ -50,22 +50,34 @@ run = (args, callback) ->
...
   files = fs.readdirSync 'src'
-  files = ('src/' + file for file in files when file.match(/\.(lit)?coffee$/))
+  files = ('src/' + file for file in files when file.match(/\.(lit)?coffee$/) and if match then file.match(match) else true)
   run ['-c', '-o', 'lib/coffeescript'].concat(files), callback
 
 build = (callback) ->
   buildParser()
   buildExceptParser callback
 
+buildWithUpdatedGrammar = (callback) ->
```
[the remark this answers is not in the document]
GeoffreyBooth: I solved the problem through a Gulp task that watches my files as they change, and runs the relevant `build` command based on which ones have changed (e.g. including the parser or not):
https://github.com/GeoffreyBooth/coffeescript-gulp/blob/master/gulpfile.coffee

It’s not a great solution, though, and I would be open to a solution that’s built into CoffeeScript/the `Cakefile` itself. This should probably be its own PR, though.

src/lexer.coffee
```diff
@@ -873,8 +858,12 @@ exports.Lexer = class Lexer
           # empty string.
           if i is 2 and firstEmptyStringIndex?
             @tokens.splice firstEmptyStringIndex, 2 # Remove empty string and the plus.
+          addTokenData token, initialChunk: yes if i is 0
+          addTokenData token, finalChunk: yes   if i is $
+          addTokenData token, {indent, quote, double}
+          addTokenData token, {heregex} if heregex
           token[0] = 'STRING'
-          token[1] = @makeDelimitedLiteral converted, options
+          token[1] = '"' + converted + '"'
```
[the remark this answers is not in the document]
GeoffreyBooth: Tell me again why we're removing the quotes from the string value? Why not just pass them through and deal with them in the node class?

src/lexer.coffee
```diff
@@ -416,27 +399,26 @@ exports.Lexer = class Lexer
...
         @error "invalid regular expression flags #{flags}", offset: index, length: flags.length
       when regex or tokens.length is 1
-        if body
-          body = @formatRegex body, { flags, delimiter: '/' }
-        else
-          body = @formatHeregex tokens[0][1], { flags }
-        @token 'REGEX', "#{@makeDelimitedLiteral body, delimiter: '/'}#{flags}", 0, end, origin
+        delimiter = if body then '/' else '///'
+        body ?= tokens[0][1]
+        @validateUnicodeCodePointEscapes body, {delimiter}
```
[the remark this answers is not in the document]
GeoffreyBooth: All nodes should have location data, there are helpers to add it if missing. One of our goals is to shift as much logic as possible out if the lexer into the node classes, so I'd rather not call this function from the lexer if we can avoid it.

Cakefile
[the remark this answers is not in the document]
helixbass: @GeoffreyBooth ok removed for now. Ya I can take a look at your Gulpfile and try and think through the cases a little more, but I'm pretty sure the simplest solution is to ensure that the state of the `lib/coffeescript/` files that it's using to compile things stays the same (except `grammar.js`) until it's done compiling everything, then replace `lib/coffeescript`. I'd even argue that whenever you run `bin/cake build` (ie build both parser and others) it should work this way

But I'm curious what others have run into/how they typically build when they're working on the grammar? I actually didn't know that `bin/cake build` existed, I would typically run `git checkout lib && bin/cake build:except-parser && bin/cake build:parser`. And then if I run into compilation weirdness, I try switching the order or invoking one more than once. Or I've resorted to trying to "massage" the transition eg if I'm making an incompatible change to the signature of a node constructor, I'll temporarily make the node constructor able to accept either signature, recompile everything, commit (so the updated grammar will be used next compilation), and then make the incompatible change

But that's a real pain, which is what prompted this - I believe all that pain just stems from the fact that we're trying to compile things using a partially updated set of files in `lib/coffeescript/` while we're in the middle of a full compilation, when what we want is to compile everything against the internally-consistent current state of `lib/coffeescript/` (with the exception of `grammar.js`, which needs to be updated before the parser is built but that doesn't cause problems)

src/lexer.coffee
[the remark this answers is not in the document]
helixbass: @GeoffreyBooth the issue as I understand it is that if you look at how the grammar/node construction works, `LOC()` (ie `addDataToNode()`) gets called on the newly constructed node, and that's what adds the location data to the node from the tokens. But that means that at the time of node construction, it doesn't have location data yet. So throwing an error from code called by `constructor()` was missing location data

I'd gone partially down the road of adding an `onLocationDataInitialized()` callback that `addDataToNode()` would call if present (and there was no previously existing location data) and then trying to wire this up to use that, but felt like this deserves more thought and shouldn't get lumped in with this PR

Eg the need for some kind of callback flow would be avoided if we could synchronously include location data (ie pass as an argument to node constructors), but that seems to imply a huge overhaul of how `LOC()` etc work

So given that it was quite clean (didn't require modifying `replaceUnicodeCodePointEscapes()` at all) to just trigger the errors from `lexer` but shift the transformation to `nodes` in this case, I felt like this was the right solution for the scope of this PR

src/lexer.coffee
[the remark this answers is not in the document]
helixbass: @GeoffreyBooth ya this deserves further explanation

So previously we were always including quotes around `STRING` tokens since we were fully transforming/formatting them in the lexer before constructing the token. Specifically `makeDelimitedLiteral()` is responsible for adding the quotes. Then some would get output like that, but some (eg string chunks of a template literal) would then get their (artificialy generated) quotes re-stripped

So once I moved transformation/`makeDelimitedLiteral()` to the `nodes` side, it was sort of the default thing to do to pass them bare. But then I ran into the fact that in many places in the lexer, it just checks a string against `@value()`, so now string chunks that happened to match operators could cause false positives. So then I re-added quotes around `STRING` tokens to avoid this, but have the grammar strip them before passing to `new StringLiteral()` as I think it's weird to include fake quotes that weren't necessarily in the original source (eg if the `STRING` came from an interpolated string or a heregex) in the value that `StringLiteral` sees. You could have `StringLiteral::constructor()` strip them itself, but I'd rather contain it as a token artifact that the grammar is aware of (I also considered stripping them out in the token data cleanup pass at the end of the rewriter)

The state of `StringLiteral`s being quoted is still unresolved imo, eg `StringWithInterpolations::compileNode()` manually replaces the `value` of its `StringLiteral` children with the result of calling their `unquote()` method. But for now preserved the existing `@value` interface of `StringLiteral`s (which always includes quotes)

src/nodes.coffee
GeoffreyBooth: `formatValue` is really just an extension of the constructor, isn’t it? You’d never do something like `stringLiteralNode = new StringLiteral(args); stringLiteralNode.formatValue()`. Since that’s the case, it feels like all this logic belongs in the constructor.

I tend to avoid things like `do =>` because it’s been ingrained into me that creating a new function comes with a performance cost. Maybe that’s not the case realistically anymore, so I’m not sure what to do here. I know that for best performance, we would want to define the function outside of the class itself (or as a separate method in the class) so that it only gets _defined_ once, and executed lots and lots of times for every `StringLiteral` class that gets instantiated; `do =>` both defines _and_ executes it for each new instantiation. But again, I’m not sure how big a hit this is, if it’s worth optimizing for; it’s probably more important to keep logic that deserves to be part of the constructor actually in the constructor, rather than outside of the class entirely.

src/nodes.coffee
```diff
@@ -752,18 +752,75 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
...
+
+  formatValue: (val) ->
+    heregex = @delimiter is '///'
+    endDelimiterIndex = val.lastIndexOf '/'
+    @flags = val[endDelimiterIndex + 1..]
+    @value = do =>
+      val = @originalValue = val[1...endDelimiterIndex]
+      val = val.replace HEREGEX_OMIT, '$1$2' if heregex
+      val = replaceUnicodeCodePointEscapes val, {@flags}
+      "#{makeDelimitedLiteral val, delimiter: '/'}#{@flags}"
```
GeoffreyBooth: Ditto here. `formatValue` _is_ the entire constructor.

GeoffreyBooth:
Does this branch fix #5019?

GeoffreyBooth:
Looks good to me. Any notes @zdenko or @vendethiel ?

helixbass:
@GeoffreyBooth closing since this branch was also merged into `ast` with the merge of #5079 (which was based on this branch)

PR #5059, opened 2018-05-07
Preserve aliased operators

helixbass:
@GeoffreyBooth @zdenko another WIP similar to #5057: this one preserves the original operator eg `on`/`yes`/`true` or `is`/`==`

Uses the same technique of stashing `.original` on a `new String()` wrapper around the token value. While I followed the same pattern in the node constructor of immediately calling `normalizeStringObject()` to unwrap to a primitive string, I ran into more cases within `lexer.coffee` itself where it wanted to do direct string comparisons against the token value so I had to call `normalizeStringObject()` in other places where it could be wanting to compare to a wrapped `String()`

I've wired this one up fully with Prettier/AST and am able to see the original operator in reformatted source

The only other thing I was thinking of including in this PR was preserving `@` vs `this`? And I guess `::` vs `.prototype`


helixbass:
A couple other things that ultimately will need to be preserved:
- whether the `...` of a splat was prefix or postfix
- whether a function with no params used `() ->` or just `->`
- whether a call used parens or not

As well as things like whether the postfix form of `if`/`while`/`for` was used - I've started doing this "inline" on my `prettier` branch but can extract into separate PR if it makes sense to, maybe I can group all the ones (like this) that don't require token data-stashing but just require passing a flag to the node constructor in certain grammar rule variants into their own PR

src/helpers.coffee
```diff
@@ -233,3 +233,6 @@ exports.nameWhitespaceCharacter = (string) ->
     when '\r' then 'carriage return'
     when '\t' then 'tab'
     else string
+
+exports.normalizeStringObject = (str) ->
+  str?.valueOf?() ? str
```
GeoffreyBooth: I think we can do without this function. Everywhere you’re currently calling it, you could do
```coffee
"#{str}"
```
instead. Proof:
```coffee
primitiveString = 'foo'
stringObject = new String 'foo'
primitiveString is stringObject # false
primitiveString is "#{stringObject}" # true
```

helixbass:
@GeoffreyBooth merged `ast` and updated to use `$1.toString()` pattern in grammar

src/grammar.coffee
```diff
@@ -836,7 +836,7 @@ grammar =
   ]
 
   Operation: [
-    o 'UNARY Expression',                       -> new Op $1 , $2
+    o 'UNARY Expression',                       -> new Op $1.toString(), $2, null, null, originalOperator: $1.original
```
GeoffreyBooth: Rather than passing `null` to skip arguments, we should pass `undefined`, as a general rule. Since only `undefined` allows function parameter default values to be set.

src/nodes.coffee
```diff
@@ -3204,6 +3206,9 @@ exports.Op = class Op extends Base
     @operator in ['<', '>', '>=', '<=', '===', '!==']
 
   invert: ->
+    if @isIn()
```
GeoffreyBooth: Until I read farther down, I thought this meant if is inverted.

helixbass: Ya this is a tricky name - updated to `@isInExpression()` but even that could be misinterpreted. `@isAnInExpression()`, though less ambiguous, looks weird to me

src/grammar.coffee
helixbass: updated to `undefined` (in all the updated rules)

src/nodes.coffee
GeoffreyBooth: Well doesn’t `in` refer to the operator? And this is in the `Op` class, so shouldn’t we call this `isInOperator`?

helixbass: @GeoffreyBooth sure updated to `isInOperator()`

PR #5067, opened 2018-05-14
Docs headings and changelog refactor

GeoffreyBooth:
This PR is meant to make #5058 easier.
https://rawgit.com/GeoffreyBooth/coffeescript/docs-headings/docs/v2/

- The changelog entries are now split into separate files, and each one is in its own `<section>` tag like the rest of the docs’ sections; and they use `h3` tags like the other sub-sections do. The HTML structure should be consistent now between sections like Breaking Changes and its children and Changelog and its children.
- I made the dates in the changelog headings formatted in the user’s locale by default. So for US English users they appear the same as before, but a U.K. user should see e.g. “13 May 2018” and a French user “13 mai 2018”. @zdenko, if you have any devices with non-US locale settings, please let me know if it works for you 😄The formatting falls back to ISO, e.g. “2018-05-13”.
- I added separate sidebar links for Introduction and Overview, so that the scrolling to those sections works as expected. This also simplifies the structure, as there aren’t any sections with two `htmlFor` calls (other than for children).

I considered making a `sectionFor` helper, to slim down `body.html`, but ultimately figured it would be more hassle than it’s worth because of the complexity of how to handle children. Hopefully this isn’t too hard to parse as is.


zdenko:
LGTM

PR #5070, opened 2018-05-21
Chained do IIFE: use grammar, not rewriter

helixbass:
@GeoffreyBooth in playing with whether [`INDENT_SUPPRESSOR`](https://github.com/jashkenas/coffeescript/pull/5068) could be used for chaining, I came across #4666/#4672, where the consensus was that a grammar-based solution for handling chained `do` IIFE eg:
```
do ->
  a
.b
```
would be better than the rewriter-based one added in those PRs

At this point I was comfortable tackling a grammar-based solution, so this PR removes that rewriter pass and instead adds tagging in the lexer to distinguish `DO` (not followed by IIFE) from `DO_IIFE` and grammar rules for both (and assigns higher precedence to `DO_IIFE`) - this should be more or less what @jashkenas was [pushing for](https://github.com/jashkenas/coffeescript/pull/4666#issuecomment-325776253)


GeoffreyBooth:
@helixbass Do you want to resolve the conflicts so we can wrap this up?

GeoffreyBooth:
This looks good to me. @zdenko or @vendethiel or anyone else, any notes?

PR #5096, opened 2018-08-15
Fix (possible) typo in expressions example (docs).

Zyndoras:
As this example is about functions automatically returning their final value, you probably want this expression to be an actual function.

New output:
coffee

### DIFF
--- a/documentation/examples/expressions.coffee
+++ b/documentation/examples/expressions.coffee
@@ -6,4 +6,4 @@ grade = (student) ->
   else
     "C"
 
-eldest = if 24 > 21 then "Liz" else "Ike"
+eldest = -> if 24 > 21 then "Liz" else "Ike"


GeoffreyBooth:
I’m not sure about this. The first paragraph might mention only functions, but the section overall talks about how almost everything in the language is an expression. The `eldest` line is an example of how a ternary assignment is also an expression.

jashkenas:
It's actually more important than that — in JavaScript, `if/else` is a statement. You can't pass a complex `if/else` as a value. In CoffeeScript, you can. That's what the example is showing.

GeoffreyBooth:
Maybe the example should be a little more complex, then? Because the JavaScript it maps to is almost identical:

```coffee
eldest = if 24 > 21 then "Liz" else "Ike"
```
```js
eldest = 24 > 21 ? "Liz" : "Ike";
```

and therefore it doesn’t look much like “a complex`if`/`else`.”

GeoffreyBooth:
The `eldest` example looks like an expression to me _in JavaScript._ Perhaps an example where it’s clearly a statement in JavaScript would get the point across better.

Don’t ask me for such an example though 😄

Inve1951:
maybe [#try](https://coffeescript.org/v2/#try:eldest%20%3D%20switch%20student.age%0A%20%20when%2024%20then%20%22yes%22%0A%20%20when%2023%20then%20%22almost%22%0A%20%20else%20%22no%22):
```coffee
eldest = switch student.age
  when 24 then "yes"
  when 23 then "almost"
  else "no"
```
why does that compile to a function btw? seems unneccessary.

vendethiel:
Easier compilation really

PR #5098, opened 2018-08-22
Include range in location data

helixbass:
@GeoffreyBooth here's a first PR carved off my `prettier` branch:

We'll need to include `range`s (the raw, `0`-based character indexes in the file) for Babel-style node location data (as expected by eg Prettier). So this PR updates our location data objects to include `range: [<start-index>, <end-index>]`

The "end range" (ie `range[1]`) here is *inclusive*, while the one expected by Babel is *exclusive* (eg here a single-character identifier `x` might have `range: [1, 1]` while Babel would expect `[1, 2]`). I believe I did it this way to be compatible with Jison's ranges. We'll have fixed places where our location data objects get converted to Babel-style location data, so that's where we'll convert inclusive -> exclusive end range

I haven't exposed the ranges on the generated ASTs yet - the next PR will probably extend the generic AST node generation logic and that's where it'll make sense to include that

Since it's not exposed here, it's not the easiest to test - I'd recommend debug-logging the root node object from eg `Block::compileRoot()` and inspecting the dumped node objects' location data, which should include `range`s (corresponding to their `first_line`/`first_column`/`last_line`/`last_column`). There'll actually be a lot of location-data-related stuff in these AST PRs, so will for sure be including related tests as we go


GeoffreyBooth:
Why have two styles of ranges at all? If we weren’t using ranges previously, why not output only the exclusive ranges that Babel and Prettier expect?

src/helpers.coffee
```diff
@@ -107,6 +107,10 @@ buildLocationData = (first, last) ->
     first_column: first.first_column
     last_line: last.last_line
     last_column: last.last_column
+    range: [
+      first.range[0]
+      last.range[1]
```
GeoffreyBooth: Could we just add `+ 1` here to make all the ranges always exclusive?

helixbass: @GeoffreyBooth I think you're right that it makes sense to have ranges always be exclusive, as it's new for us and Babel (and seemingly all other JS ASTs, judging by a quick survey on [astexplorer.net](https://astexplorer.net)) use exclusive ranges

It seems like the single place where we specify a `range[1]` initially for our location data is in `makeLocationData()` (in `lexer.coffee`), so I updated there to use an exclusive end `range`

Then looks like Jison's `ranges` are totally unopinionated about whether ranges are inclusive/exclusive, it just passes through whatever we've set as `range`s on the tokens. And so then here we can just pass through `last.range[1]` which should already be exclusive

PR #5102, opened 2018-08-28
Port to Babel 7

bazineta:
Use the Babel 7 @babel/core transpiler, falling back to the older babel-core version if possible.

### DIFF
--- a/lib/coffeescript/command.js
+++ b/lib/coffeescript/command.js
@@ -619,16 +619,20 @@
         // on us to transpile for them; we assume most users will probably either
         // run CoffeeScript’s output without transpilation (modern Node or evergreen
         // browsers) or use a proper build chain like Gulp or Webpack.
-        require('babel-core');
+        require('@babel/core');
       } catch (error) {
-        // Give appropriate instructions depending on whether `coffee` was run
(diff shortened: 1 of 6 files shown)


rdeforest:
Cakefile needs some love too I think?

```diff
@@ -67,7 +67,11 @@ build = (callback) ->
   buildExceptParser callback

 transpile = (code) ->
-  babel = require 'babel-core'
+  try
+    babel = require '@babel/core'
+  catch
+    babel = require 'babel-core'
+
   presets = []
   # Exclude the `modules` plugin in order to not break the `}(this));`
   # at the end of the `build:browser` code block.
```

GeoffreyBooth:
> Cakefile needs some love too I think?

`Cakefile` is only run by someone who’s checked out the CoffeeScript repo, and therefore will have whatever dependencies and `devDependencies` are defined in CoffeeScript’s `package.json`. We should update CoffeeScript itself to use Babel 7 and `@babel/core`, and then also update `Cakefile` accordingly. We don’t need to support either `babel-core` or `@babel/core` in the `Cakefile`, though, as either one or the other will be specified in `package.json`.

PR #5106, opened 2018-09-13
Fix #4609: support new.target

### LINKED ISSUE #4609: Proposal: Derived class constructors and new.target
There are times when you may need to access static class members before super has been called:

```javascript

/* contrived example */
class Abstract {

    static checkArgs(args) {
        if (0 === args.length) {
            throw new Error('BadArgs')
        }    
    }

    constructor (...args) {
        this.args = args
    }
}

class Concrete extends Abstract {
    constructor(...args) {
        // Cannot access uninitialized variable 'this'
        // this.constructor.checkArgs(args)
        // so use new.target instead
        new.target.checkArgs(args)
        super(...args)
    }
}

class Dense extends Concrete {
    static checkArgs () {
        // no-op
    }
}

try {
    new Concrete()
}
catch (thrown) {
    // 'BadArgs' as expected
}


new Dense() // ok as expected
```

In CoffeeScript2, the Abstract and Dense class implementations are fine

```coffee
class Abstract

    @checkArgs: (args) ->
        if 0 is args.length
            throw new Error("BadArgs")

    constructor: (args...) ->
        this.args = args
```

But none of the following implementations of Concrete are possible as new.target is not available:

```coffee
class Concrete

    constructor: (args...) ->
        ###
        error: unexpected .
                new.target.checkArgs(args)
                   ^
        ###
        new.target.checkArgs(args)
        super(args...)
```

```coffee
class Concrete

    constructor: (args...) ->
        # error: Can't reference 'this' before calling super in derived class constructors
        @constructor.checkArgs(args)
        super(args...)

```

The result is that Concrete must reference itself to access checkArgs, but this prevents subclasses from override the static checkArgs method.
```coffee
class Concrete

    constructor: (args...) ->
        # subclasses can no longer override checkArgs
        Concrete.checkArgs(args)
        super(args...)
```

```coffee
class Dense extends Concrete
    @checkArgs: -> # no -op

new Dense() # throws "BadArgs" via Concretes reference to Concrete.checkArgs
```

If I'm not missing something, this looks like a bit of problem to me. How is it possible to reference the constructor and its properties before super in CS2?

helixbass:
Fixes #4609 

Allow `new.target` inside functions/methods

Throw an error if `new.target` is used outside of a function

Throw an error that mimics eg Babel/`espree` if you try to reference a property other than `target` on `new`

@GeoffreyBooth seemed to make sense to target this PR against `master`? Then I'll add AST support separately

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -196,6 +196,8 @@ exports.Lexer = class Lexer
          @tokens.length > 1 and @tokens[@tokens.length - 2][0] not in ['.', '?.', '@']
         @error "'#{prev[1]}' cannot be used as a keyword, or as a function call
         without parentheses", prev[2]
+      else if prev[0] is '.' and @tokens.length > 1 and @tokens[@tokens.length - 2][1] is 'new'
+        @tokens[@tokens.length - 2][0] = 'IDENTIFIER'
       else if @tokens.length > 2
         prevprev = @tokens[@tokens.length - 2]
         if prev[0] in ['@', 'THIS'] and prevprev and prevprev.spaced and
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -196,8 +196,8 @@ exports.Lexer = class Lexer
          @tokens.length > 1 and @tokens[@tokens.length - 2][0] not in ['.', '?.', '@']
         @error "'#{prev[1]}' cannot be used as a keyword, or as a function call
         without parentheses", prev[2]
-      else if prev[0] is '.' and @tokens.length > 1 and @tokens[@tokens.length - 2][1] is 'new'
-        @tokens[@tokens.length - 2][0] = 'IDENTIFIER'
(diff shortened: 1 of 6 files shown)


src/lexer.coffee
```diff
@@ -196,6 +196,8 @@ exports.Lexer = class Lexer
          @tokens.length > 1 and @tokens[@tokens.length - 2][0] not in ['.', '?.', '@']
         @error "'#{prev[1]}' cannot be used as a keyword, or as a function call
         without parentheses", prev[2]
+      else if prev[0] is '.' and @tokens.length > 1 and @tokens[@tokens.length - 2][1] is 'new'
```
GeoffreyBooth: Elsewhere in this section we check for the token tag, not its value. So in other words, `@tokens[@tokens.length - 2][0] is 'NEW'`

helixbass: @GeoffreyBooth looks like that doesn't uniquely identify it though, its `[0]` is not `NEW` but `UNARY`

So added a check for `UNARY` type but kept secondary check for `new` value

GeoffreyBooth: Huh. Okay, I guess that works.

PR #5107, opened 2018-09-17
Value AST

helixbass:
@GeoffreyBooth here's a PR for ASTs for `Value`s with properties (eg `a.b`, `a?::b[c]`)

As mentioned in https://github.com/helixbass/copheescript/pull/13, I've based this on your `ast-architecture-refactor` branch. [Here](https://github.com/GeoffreyBooth/coffeescript/compare/ast-architecture-refactor...helixbass:value-ast)'s just the diff against that branch


src/grammar.coffee
```diff
@@ -392,11 +392,11 @@ grammar =
   # or by array index or slice.
   Accessor: [
     o '.  Property',                            -> new Access $2
-    o '?. Property',                            -> new Access $2, 'soak'
-    o ':: Property',                            -> [LOC(1)(new Access new PropertyName('prototype')), LOC(2)(new Access $2)]
-    o '?:: Property',                           -> [LOC(1)(new Access new PropertyName('prototype'), 'soak'), LOC(2)(new Access $2)]
-    o '::',                                     -> new Access new PropertyName 'prototype'
-    o '?::',                                    -> new Access new PropertyName('prototype'), 'soak'
+    o '?. Property',                            -> new Access $2, soak: yes
+    o ':: Property',                            -> [LOC(1)(new Access new PropertyName('prototype'), shorthand: yes), LOC(2)(new Access $2)]
```
helixbass: Tell the `Access` whether it was `shorthand` (ie `::`) or not so that it can be exposed on the AST

src/nodes.coffee
```diff
@@ -1081,6 +1150,25 @@ exports.Value = class Value extends Base
 
     fragments
 
+  ast: ->
```
helixbass: `Value`s don't fit neatly into the base `ast()` method's "shape" since it could have different "top-level" `type`s (`MemberExpression` if it has properties, otherwise whatever its `@base` is). I'd had this as an override of `_toAst()`, overriding `ast()` is just as easy here

This is one of the more interesting transformations, we store `Value`s "linearly" as the base with its list of properties, whereas the AST expects nested `MemberExpressions`, eg `a.b.c` ends up as:
```
type: 'MemberExpression'
object:
  type: 'MemberExpression'
  object:
    type: 'Identifier'
    name: 'a'
  property:
    type: 'Identifier'
    name: 'b'
property:
  type: 'Identifier'
  name: 'c'
```

src/nodes.coffee
```diff
@@ -1081,6 +1150,25 @@ exports.Value = class Value extends Base
...
+    ret = @base.ast()
+    for prop, propIndex in @properties
+      ret =
+        mergeAstLocationData(
```
helixbass: Use new `mergeAstLocationData()` helper to "expand" the location data of the nested `MemberExpression`s as we go. I'd had this helper in `helpers.coffee` but it's only being used in `nodes.coffee` so moved it there

src/nodes.coffee
```diff
@@ -1081,6 +1150,25 @@ exports.Value = class Value extends Base
...
+          merge {
+            type: 'MemberExpression'
+            object: ret
+            property: prop.ast()
+            computed: prop instanceof Index or prop.name?.unwrap() not instanceof PropertyName
+            optional: !!prop.soak
```
helixbass: So I modeled this `optional` field after what I was seeing in Babel ASTs that used the proposed JS version of the soak operator ("nullish coalescing"). It looks like maybe by the time Babel 7 landed they switched to using different `type`s like `OptionalMemberExpression` rather than just an `optional` field on `MemberExpression`. The `optional` field seems more convenient but it's worth looking into why they chose to model them as unique types

src/nodes.coffee
```diff
@@ -1081,6 +1150,25 @@ exports.Value = class Value extends Base
...
+            shorthand: !!prop.shorthand
```
helixbass: `shorthand` is the only "extension" field that you wouldn't find on a Babel-style `MemberExpression` node (with the above caveat about `optional`). It's being used to be able to distinguish shorthand `::` from `.prototype`

src/nodes.coffee
```diff
@@ -1081,6 +1150,25 @@ exports.Value = class Value extends Base
...
+          }, prop.astLocationData()
+          ret
+        )
+      if propIndex is 0 and @base instanceof Parens and @base.locationData?
```
helixbass: This is a case I ran into when doing some Prettier tests: typically, the location data for the overall `MemberExpression` just needs to span the location of its `object` and its `property` (eg for `a.b` it spans `a`'s location through `b`'s location). But if the `object` is parenthesized, we also need to include the location of the parens in the `MemberExpression`'s data (eg `( a ).b`). Included a generated location data test case

src/nodes.coffee
```diff
@@ -1414,6 +1501,9 @@ exports.Access = class Access extends Base
     else
       [@makeCode('['), name..., @makeCode(']')]
 
+  ast: ->
```
helixbass: Both `Access` and `Index` "forward" their AST generation on to a child. Again, just overriding `ast()` is as easy what I was doing previously (overriding `_toAst()` or `toAst()`)

src/nodes.coffee
```diff
@@ -3553,6 +3646,9 @@ exports.Parens = class Parens extends Base
     return @wrapInBraces fragments if @csxAttribute
     if bare then fragments else @wrapInParentheses fragments
 
+  ast: ->
```
helixbass: Parens don't generate AST nodes. It looks like Babel includes some paren-related info under `extra` (eg `extra: {parenthesized: true, parenStart: <range start>}`) of the parenthesized node, but `espree` doesn't (ESLint examines the source code/tokens for paren-related stuff). I haven't run into a compelling reason to try and include something like the Babel-style `parenthesized` in the AST yet. But it's something we may eventually want to do

src/nodes.coffee
```diff
@@ -4066,3 +4162,29 @@ makeDelimitedLiteral = (body, options = {}) ->
...
   "#{options.delimiter}#{body}#{options.delimiter}"
+
+# Extends the location data of an AST node to include the location data from
+# another AST node.
+mergeAstLocationData = (intoNode, fromNode) ->
+  {range: intoRange} = intoNode
+  {range: fromRange} = fromNode
+  return intoNode unless intoRange and fromRange
+  if fromRange[0] < intoRange[0]
+    intoNode.range = intoRange = [
```
helixbass: Again, treat `range`s as immutable (otherwise location data bugginess ensues from shared mutated `range`s)

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -0,0 +1,379 @@
...
+    loc:
+      start:
+        line: 1
+        column: 0
+      end:
+        line: 1
+        column: 4
+
+test "AST location data as expected for Access node", ->
+  testAstLocationData 'obj.prop',
```
helixbass: @GeoffreyBooth these tests currently aren't checking the nested nodes as desired because of your refactor of the shared AST testing helpers in helixbass#13

src/nodes.coffee
GeoffreyBooth: See https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md#changes

src/helpers.coffee
```diff
@@ -136,11 +136,11 @@ buildTokenDataDictionary = (parserState) ->
 # This returns a function which takes an object as a parameter, and if that
 # object is an AST node, updates that object's locationData.
 # The object is returned either way.
-exports.addDataToNode = (parserState, first, last) ->
+exports.addDataToNode = (parserState, first, last, {forceUpdateLocation} = {}) ->
```
GeoffreyBooth: Let’s not mix styles, of arguments and then an options object:

```coffee
exports.addDataToNode = (parserState, first, last, forceUpdateLocation = yes) ->
```

Any reason not to make it default to true? Then you don’t need it in `getAddDataToNodeFunctionString`.

src/grammar.coffee
GeoffreyBooth: Sounds good. Probably not in this PR, but we need to think about the style we’re using for these inline objects in `grammar.coffee`. I try to avoid seemingly ambiguous cases like `fn a, b: c, d: e`. Reading it carefully, it’s clear that it needs to be `fn a, {b: c, d: e}` but it’s easy for a quick reader to not be aware of where the object starts and ends. I don’t know what a better style is; just adding the braces? Making all such objects indented on new lines? I’m not sure, but I don’t like how the current style can be confusing. (And I know you didn’t come up with this, that it was this way before.)

src/nodes.coffee
GeoffreyBooth: First, I’ve been trying to always use `Object.assign` instead of `merge` and `extends`. Those helper functions aren’t needed anymore, and should really be refactored out of the codebase; but until that happens, I’m trying not to create more references to them.

Second, I hadn’t imagined `ast` as ever getting overridden. The AST you quote would imply to me that `Value`‘s `astProperties` method would return an object with two keys, `object` and `property`:
- The value of `object` is what we already have on the current instance of the class: `type: 'Identifier', name: 'a'`, from the class’s `base` child node.
- The value of `property` is a child node, which I presume is the class’s `properties` child. So `property: @properties.ast()`, with variations for when `properties` is an array, etc.

And so on. You don’t need to write code to drill down to all the children of the `Value`/`MemberExpression`, because you’re recursively calling `ast()` on the node class’s children. Or at least, that should be how it works.

src/nodes.coffee
helixbass: In that link I see a description of the `optional` field but don't see any references to `OptionalMemberExpression`?

Cakefile
```diff
@@ -279,7 +279,10 @@ buildDocTests = (watch = no) ->
   outputFolder      = "docs/v#{majorVersion}"
 
   # Included in test.html
-  testHelpers = fs.readFileSync('test/support/helpers.coffee', 'utf-8').replace /exports\./g, '@'
+  readHelpersFile = (basename) ->
+    fs.readFileSync("test/support/#{basename}.coffee", 'utf-8').replace /exports\./g, '@'
+  testHelpers = readHelpersFile('helpers') + '\n'
+  testHelpers += readHelpersFile('abstract_syntax_tree_helpers')
```
helixbass: I'm not familiar with `doc:test`, the generated `docs/v2/test.html` looked visually correct (included the AST test helpers after the existing test helpers) but when I loaded that file in a browser I saw an `Uncaught SyntaxError: missing "` in the console

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -0,0 +1,392 @@
+# Astract Syntax Tree location data
+# ---------------------------------
+
+testAstLocationData = (code, expected) ->
```
helixbass: Refactored these to support nested location data testing. Preserved the existing custom error messaging for location data fields (with additional "path" info to help identify which node was failing)

src/grammar.coffee
GeoffreyBooth: The issue is that the object often isn’t the only argument of the function, or even the last argument: `fn arg1, obj, arg2`. I agree, the “options argument” pattern is way more flexible than dealing with sequential function arguments, and I would be in favor if refactoring the codebase to standardize on that potentially; but in its own PR 😄For now I’m trying to just follow the style of the functions we have, and not _mix_ styles. I’m hoping that once all this Prettier stuff is done we do a big cleanup pass for style, assisted by Prettier reformatting 😄

src/nodes.coffee
GeoffreyBooth: Object spread only came to Node in 8.6, part of ES2018: https://node.green/#ES2018-features-object-rest-spread-properties. We need to support back to Node 6, at least for now. I got a lot of pushback dropping Node 4, I don’t think we can drop Node 6 just yet.

I’ll have to look into what you’re saying later. In general I want to follow the object-oriented style of the node classes. Not sure what the best way to do that is, with also needing to deal with this weird shape.

src/nodes.coffee
GeoffreyBooth: Oh I don’t know about that specifically, I just noticed it had some spec changes listed.

src/nodes.coffee
helixbass: > I don’t think we can drop Node 6 just yet

My question is sort of the inverse, not whether we should consider dropping Node 6 but whether we should consider keeping the polyfill for object spread even though it's reached stage (4?) so that it's still easily usable against Node 6. In my mind it's worth entertaining that even just for the sake of being able to use object spread syntax in the Coffeescript codebase itself

> In general I want to follow the object-oriented style of the node classes

I don't think overriding `ast()` is violating object-oriented style. The base `ast()` is a template method, so subclasses that fit that "template" just override the individual subparts of the template as necessary. But for subclasses that don't fit the template, you just override the template method itself

In this instance, `Value`'s `ast()` implementation is encapsulating/taking responsibility for arranging its base value and properties into the right shape. It's still delegating responsibility in the typical way to its children (to generate their own AST representations), it's just then assembling those child ASTs in its `ast()` method (rather than `astProperties()`, which would be a very awkward fit since it's natural in this case to couple specifying the `type` and location data of the AST nodes the `Value` is creating with specifying its other properties)

src/grammar.coffee
helixbass: Ya to be clear, I'm saying that my tendency is to only omit object braces if the object is trailing, so a convention where `fn arg1, arg2, obj` gets formatted `fn arg1, arg2, a: ..., b: ...` but `fn arg1, obj, arg2` gets formatted `fn arg1, {a: ..., b: ...}, arg2`

If you follow that convention, then you can assume that you're seeing a single trailing object argument if you see an implicit object amongst call args, which makes it significantly visually easier to parse (than if you had to scan whether it was in fact a trailing arg, a single object, ...)

That convention is like Ruby in the sense that (from my understanding, maybe this was only pre-Ruby 2.0) Ruby only allows you to use a syntax that looks like implicit object args if it's a single trailing object (hash)

Speaking of Prettier, that's actually the current default behavior of the Prettier formatter is that it'll put explicit braces around a non-trailing object arg but leave it implicit if it's trailing

src/nodes.coffee
```diff
@@ -4139,3 +4165,29 @@ makeDelimitedLiteral = (body, options = {}) ->
     when ps        then '\\u2029'
     when other     then (if options.double then "\\#{other}" else other)
   "#{options.delimiter}#{body}#{options.delimiter}"
+
+# Extends the location data of an AST node to include the location data from
+# another AST node.
+mergeAstLocationData = (intoNode, fromNode) ->
+  {range: intoRange} = intoNode
+  {range: fromRange} = fromNode
```
GeoffreyBooth: Can you explain “into” and “from” here? I would think that this function would take input like `locationDataA`, `locationDataB` and generate a new `locationData` that includes the data from A + B. So like look at A’s `start` value and then look at B’s `start` value, and the lower (earlier) of the two becomes the merged `start` value. Ditto for `end`, and for the start/end values of the ranges.

helixbass: @GeoffreyBooth this function is mutating `intoNode`, whose existing location data should be expanded to include `fromNode`'s location data

Since we're dealing with AST nodes here (where the individual location data fields are at the same top level as other fields), returning a new "`locationData`" object seems less natural than if we were dealing with eg node class objects or tokens (both of which store `locationData` as a standalone object)

And I think it's a cleaner calling pattern for this to mutate the first node's location data fields. Otherwise it seems like you'd end up having to call it like eg
```
someExistingAstNode = Object.assign {},
  someExistingAstNode,
  mergeAstLocationData(someExistingAstNode, otherAstNode)
```
rather than just `mergeAstLocationData(someExistingAstNode, otherAstNode)`

I'm not sure what the upside would be of keeping it "immutable" and then forcing the caller to reassemble a new AST node like that, the typical case is that there's no reason to preserve the previous location data

PR #5108, opened 2018-09-17
Fix #5086: Don’t generate unnecessary interpolations in JSX tags around herecomments

### LINKED ISSUE #5086: Bug: Inline Comments in JSX cause unnecessary expression wrapping
Adding Comments in JSX cause unnecessary expression wrapping. I'm working on a JSX transpiler that uses Flow-like annotations so this is completely broken in Coffeescript currently. 

```coffee
<div>
  <div ###comment### attribute={value} />
</div>
```
Produces
```js
<div>
  {<div /*comment*/attribute={value} />}
</div>;
```
Where it should produce
```js
<div>
  <div /*comment*/attribute={value} />
</div>;
```
Also true for attribute comments like
```coffee
<div>
  <div attribute={###attr comment###value} />
</div>
```

Produces
```js
<div>
  {<div attribute={/*attr comment*/value} />}
</div>
```

EDIT: Actually it does this in a nested manner so:
```coffee
<div>
  <div>
    <div>
      <div ###comment### attribute={value} />
    </div>
  </div>
</div>
```
Produces:
```js
<div>
  {<div>
    {<div>
      {<div /*comment*/attribute={value} />}
    </div>}
  </div>}
</div>
```

* CoffeeScript version: CS 2.3.1
* Node.js version: 8.11.1

GeoffreyBooth:
Fix #5086: Don’t generate unnecessary interpolations in JSX tags when the tags contain only here (`/* ... */`) comments.

@ryansolid can you please verify this against your codebase? And please look at https://github.com/GeoffreyBooth/coffeescript/blob/22ec8d91d81a6f22e8a8fe073b52d69419fe630e/test/csx.coffee#L745-L784 and let me know if the expected output is correct. I copied from your issue.


ryansolid:
Yes the test output looks correct. I tested this PR against a few examples I had, but admittedly I abandoned this approach of comment annotating a month or so back so I don't have many examples. I didn't hit any issues with what I do have.

Thank you for the fix.

PR #5109, opened 2018-09-17
Op AST

helixbass:
@GeoffreyBooth PR for `Op` ASTs

Again, found switching to the new AST methods smooth and in fact an improvement

### DIFF
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -99,7 +99,7 @@
     }
     sources[filename].push(code);
     if (generateSourceMap) {
-      map = new SourceMap;
+      map = new SourceMap();
     }
     tokens = lexer.tokenize(code, options);
     // Pass a list of referenced variables, so that generated variables won’t get
@@ -239,7 +239,7 @@
   };
 
   // Instantiate a Lexer for our use here.
-  lexer = new Lexer;
+  lexer = new Lexer();
 
   // The real Lexer produces a generic stream of tokens. This object provides a
   // thin wrapper around it, compatible with the Jison API. We can then pass it
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3402,10 +3402,6 @@ exports.Op = class Op extends Base
   isInOperator: ->
     @originalOperator is 'in'
 
-  checkDeleteOperand: (o) ->
-    if @operator is 'delete' and o.scope.check(@first.unwrapAll().value)
-      @error 'delete operand may not be argument or var'
-
   checkUpdateAssignability: ->
     if @operator in ['--', '++']
       message = isUnassignable @first.unwrapAll().value
@@ -3423,7 +3419,8 @@ exports.Op = class Op extends Base
     # In chains, there's no need to wrap bare obj literals in parens,
     # as the chained expression is wrapped.
     @first.front = @front unless isChain
-    @checkDeleteOperand o
+    if @operator is 'delete' and o.scope.check(@first.unwrapAll().value)
+      @error 'delete operand may not be argument or var'
     @checkUpdateAssignability()
     return @compileContinuation o if @isYield() or @isAwait()
     return @compileUnary        o if @isUnary()
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3308,6 +3308,10 @@ exports.Op = class Op extends Base
       if ((firstCall = unwrapped = first.unwrap()) instanceof Call or (firstCall = unwrapped.base) instanceof Call) and not firstCall.do and not firstCall.isNew
         return new Value firstCall.newInstance(), if firstCall is unwrapped then [] else unwrapped.properties
       first = new Parens first if unwrapped instanceof Op and unwrapped.operator is 'do'
+      call = new Call first, []
+      call.locationData = @locationData
+      call.isNew = yes
+      return call
 
     @operator = CONVERSIONS[op] or op
     @first    = first
@@ -3507,7 +3511,6 @@ exports.Op = class Op extends Base
 
   astType: ->
     switch @operator
-      when 'new'           then 'NewExpression'
       when '||', '&&', '?' then 'LogicalExpression'
       when '++', '--'      then 'UpdateExpression'
       else
@@ -3518,10 +3521,6 @@ exports.Op = class Op extends Base
     firstAst = @first.ast()
     secondAst = @second?.ast()
     switch
-      when @operator is 'new'
-        return
-          callee: firstAst
-          arguments: []
       when @isUnary()
         return
           argument: firstAst
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1077,7 +1077,7 @@ exports.Value = class Value extends Base
   isBoolean      : -> @bareLiteral(BooleanLiteral)
   isAtomic       : ->
     for node in @properties.concat @base
-      return no if node.soak or node instanceof Call
+      return no if node.soak or node instanceof Call or node instanceof Op and node.operator is 'do'
     yes
 
   isNotCallable  : -> @isNumber() or @isString() or @isRegex() or
@@ -3307,7 +3307,7 @@ exports.Op = class Op extends Base
     if op is 'new'
       if ((firstCall = unwrapped = first.unwrap()) instanceof Call or (firstCall = unwrapped.base) instanceof Call) and not firstCall.do and not firstCall.isNew
         return new Value firstCall.newInstance(), if firstCall is unwrapped then [] else unwrapped.properties
-      first = new Parens first if unwrapped instanceof Op and unwrapped.operator is 'do'
+      first = new Parens first unless first instanceof Parens or first.unwrap() instanceof IdentifierLiteral or first.hasProperties?()
       call = new Call first, []
       call.locationData = @locationData
       call.isNew = yes
@@ -3471,9 +3471,9 @@ exports.Op = class Op extends Base
     if o.level >= LEVEL_ACCESS
       return (new Parens this).compileToFragments o
     plusMinus = op in ['+', '-']
-    parts.push [@makeCode(' ')] if op in ['new', 'typeof', 'delete'] or
+    parts.push [@makeCode(' ')] if op in ['typeof', 'delete'] or
                       plusMinus and @first instanceof Op and @first.operator is op
-    if (plusMinus and @first instanceof Op) or (op is 'new' and @first.isStatement o)
+    if plusMinus and @first instanceof Op
       @first = new Parens @first
     parts.push @first.compileToFragments o, LEVEL_OP
     parts.reverse() if @flip
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3313,6 +3313,11 @@ exports.Op = class Op extends Base
     @first    = first
     @second   = second
     @flip     = !!flip
+
+    if @operator in ['--', '++']
+      message = isUnassignable @first.unwrapAll().value
+      @first.error message if message
+
     return this
 
   # The map of conversions from CoffeeScript to JavaScript symbols.
@@ -3402,11 +3407,6 @@ exports.Op = class Op extends Base
   isInOperator: ->
     @originalOperator is 'in'
 
-  checkUpdateAssignability: ->
-    if @operator in ['--', '++']
-      message = isUnassignable @first.unwrapAll().value
-      @first.error message if message
-
   compileNode: (o) ->
     if @isInOperator()
       inNode = new In @first, @second
@@ -3421,7 +3421,6 @@ exports.Op = class Op extends Base
     @first.front = @front unless isChain
     if @operator is 'delete' and o.scope.check(@first.unwrapAll().value)
       @error 'delete operand may not be argument or var'
-    @checkUpdateAssignability()
     return @compileContinuation o if @isYield() or @isAwait()
     return @compileUnary        o if @isUnary()
     return @compileChain        o if isChain
@@ -3506,10 +3505,6 @@ exports.Op = class Op extends Base
   toString: (idt) ->
     super idt, @constructor.name + ' ' + @operator
 
-  ast: ->
-    @checkUpdateAssignability()
-    super()
-
   astType: ->
     switch @operator
       when 'new'           then 'NewExpression'
@@ -3524,16 +3519,19 @@ exports.Op = class Op extends Base
     secondAst = @second?.ast()
     switch
       when @operator is 'new'
-        callee: firstAst
-        arguments: []
+        return
+          callee: firstAst
+          arguments: []
       when @isUnary()
-        argument: firstAst
-        operator: @originalOperator
-        prefix: !@flip
+        return
+          argument: firstAst
+          operator: @originalOperator
+          prefix: !@flip
       else
-        left: firstAst
-        right: secondAst
-        operator: "#{if @invertOperator then "#{@invertOperator} " else ''}#{@originalOperator}"
+        return
+          left: firstAst
+          right: secondAst
+          operator: "#{if @invertOperator then "#{@invertOperator} " else ''}#{@originalOperator}"
 
 #### In
 exports.In = class In extends Base
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3437,12 +3437,12 @@ exports.Op = class Op extends Base
 
   astType: ->
     switch @operator
-      when 'new'         then 'NewExpression'
-      when '||', '&&'    then 'LogicalExpression'
-      when '++', '--'    then 'UpdateExpression'
+      when 'new'           then 'NewExpression'
+      when '||', '&&', '?' then 'LogicalExpression'
+      when '++', '--'      then 'UpdateExpression'
       else
-        if @isUnary()    then 'UnaryExpression'
-        else                  'BinaryExpression'
+        if @isUnary()      then 'UnaryExpression'
+        else                    'BinaryExpression'
 
   ast: ->
     @checkUpdateAssignability()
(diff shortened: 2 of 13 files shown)


test/abstract_syntax_tree.coffee
```diff
@@ -889,80 +1002,211 @@ test "AST as expected for BooleanLiteral node", ->
...
+    type: 'BinaryExpression'
+    operator: '<<'
+    left:
+      type: 'NumericLiteral'
+      value: 1
+    right:
+      type: 'NumericLiteral'
+      value: 2
+
+  testExpression 'new Old',   # NOTE: `new` with params is a `Call` node.
```
helixbass: ...more evidence? (I didn't write this comment)

src/nodes.coffee
```diff
@@ -3367,6 +3398,35 @@ exports.Op = class Op extends Base
         answer = [].concat lhs, @makeCode(" #{@operator} "), rhs
         if o.level <= LEVEL_OP then answer else @wrapInParentheses answer
 
+  astType: ->
+    switch @operator
+      when 'new'           then 'NewExpression'
+      when '||', '&&', '?' then 'LogicalExpression'
```
helixbass: Had realized that binary `?` really behaved more like a `LogicalExpression` (ie `and`/`or`/`&&`/`||`) than a `BinaryExpression` in that it has similar control flow semantics. And then saw that the proposed JS version of this operator (`??`) does in fact generate a `LogicalExpression` eg in Babel

test/error_messages.coffee
```diff
@@ -4,12 +4,15 @@
 # Ensure that errors of different kinds (lexer, parser and compiler) are shown
 # in a consistent way.
 
-assertErrorFormat = (code, expectedErrorFormat) ->
-  throws (-> CoffeeScript.run code), (err) ->
+assertErrorFormat = (code, expectedErrorFormat, {ast} = {}) ->
```
helixbass: It seems like a nice goal would be to have as many errors that occur when compiling to JS also get thrown when generating an AST as possible. The abstraction of `Op::checkUpdateAssignability()` that gets called from both `Op::compileNode()` and `Op::ast()` is an example of what would probably be a typical pattern for doing that

So then for testing that those errors occur when generating an AST, I tried to come up with a lightweight way to specify that existing tests should be run in the context of AST generation (in addition to compilation to JS) and expect the same errors to be thrown

test/strict.coffee
```diff
@@ -17,8 +17,10 @@
 # * `eval` or `arguments` as the operand of a post/pre-fix inc/dec-rement expression
 
 # helper to assert that code complies with strict prohibitions
-strict = (code, msg) ->
+strict = (code, msg, {ast} = {}) ->
```
helixbass: ...same idea here

src/nodes.coffee
```diff
@@ -3367,6 +3398,35 @@ exports.Op = class Op extends Base
...
+      when '++', '--'      then 'UpdateExpression'
+      else
+        if @isUnary()      then 'UnaryExpression'
+        else                    'BinaryExpression'
+
+  ast: ->
```
helixbass: Here's another nice pattern for overriding `ast()`: do some error checking, then call out to `super()` like usual

src/nodes.coffee
```diff
@@ -3498,6 +3506,35 @@ exports.Op = class Op extends Base
...
+  astType: ->
+    switch @operator
+      when 'new'           then 'NewExpression'
+      when '||', '&&', '?' then 'LogicalExpression'
+      when '++', '--'      then 'UpdateExpression'
+      else
+        if @isUnary()      then 'UnaryExpression'
+        else                    'BinaryExpression'
+
+  astProperties: ->
```
helixbass: Here, `astProperties()` was a definite improvement on the version using `astChildren()` and `astProps()`, which had parallel conditional structures (so if eg you wanted to see what  a `new` AST looked like, you'd have to scan both `astChildren()` and `astProps()` and mentally combine the results)

src/nodes.coffee
```diff
@@ -3498,6 +3506,35 @@ exports.Op = class Op extends Base
...
+    firstAst = @first.ast()
+    secondAst = @second?.ast()
+    switch
+      when @operator is 'new'
+        callee: firstAst
```
helixbass: Again, this "smells" like `new` should be a `Call`, since here we're mimicking the AST of a call (and there'll end up being duplication of specifying `type: 'NewExpression'` since `new` with args is in fact a `Call`)

src/nodes.coffee
```diff
@@ -3498,6 +3506,35 @@ exports.Op = class Op extends Base
...
     super idt, @constructor.name + ' ' + @operator
 
+  ast: ->
+    @checkUpdateAssignability()
+    super()
+
+  astType: ->
+    switch @operator
+      when 'new'           then 'NewExpression'
+      when '||', '&&', '?' then 'LogicalExpression'
```
helixbass: Had realized that binary `?` really behaved more like a `LogicalExpression` (ie `and`/`or`/`&&`/`||`) than a `BinaryExpression` in that it has similar control flow semantics. And then saw that the proposed JS version of this operator (`??`) does in fact generate a `LogicalExpression` eg in Babel

src/nodes.coffee
```diff
@@ -3498,6 +3506,35 @@ exports.Op = class Op extends Base
   toString: (idt) ->
     super idt, @constructor.name + ' ' + @operator
 
+  ast: ->
```
helixbass: Here's another nice pattern for overriding `ast()`: do some error checking, then call out to `super()` like usual

src/nodes.coffee
```diff
@@ -3238,12 +3267,10 @@ exports.Op = class Op extends Base
   constructor: (op, first, second, flip, {@invertOperator, @originalOperator = op} = {}) ->
     super()
 
-    if op is 'do'
-      return Op::generateDo first
     if op is 'new'
-      if (firstCall = first.unwrap()) instanceof Call and not firstCall.do and not firstCall.isNew
-        return firstCall.newInstance()
-      first = new Parens firstCall if firstCall instanceof Code and firstCall.bound or firstCall.do
+      if ((firstCall = unwrapped = first.unwrap()) instanceof Call or (firstCall = unwrapped.base) instanceof Call) and not firstCall.do and not firstCall.isNew
```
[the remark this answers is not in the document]
GeoffreyBooth: > It feels like the thing that would clean this up would be to always transform `new` from an `Op` to a `Call` (which is what it really is) up front

Yes. I feel like we could do a better job with this in the grammar, so that there’s less hackiness here. For example why are we updating the location data to expand to the left to pull in the `new `? We don’t need such shenanigans for any other node. The grammar should handle that kind of thing I would think. (Not saying you did anything wrong, this is very old code, you’re just making it obvious why it’s bad 😄)

src/nodes.coffee
GeoffreyBooth: That is a good pattern, but why not check in the constructor? Is that not possible?

src/nodes.coffee
```diff
@@ -1260,12 +1278,18 @@ exports.Call = class Call extends Base
     if @locationData and @needsUpdatedStartLocation
       @locationData.first_line = locationData.first_line
       @locationData.first_column = locationData.first_column
-      @locationData.range[0] = locationData.range[0]
+      @locationData.range = [
```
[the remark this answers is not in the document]
GeoffreyBooth: What rule is that? This whole function is mutating `locationData`, just look at what it’s doing to `first_line` etc.

src/nodes.coffee
GeoffreyBooth: I’ve been putting `return` before indented objects, to make clearer what’s happening:
```coffee
    switch
      when @operator is 'new'
        return
          callee: firstAst
          arguments: []
```

I know it’s not strictly necessary, but we don’t have any “loose” objects getting implicitly returned anywhere else in the codebase as far as I’m aware.

test/abstract_syntax_tree.coffee
```diff
@@ -1006,80 +1006,211 @@ test "AST as expected for Index node", ->
...
-#     flip: no
+  testExpression '1 << 2',
+    type: 'BinaryExpression'
+    operator: '<<'
+    left:
+      type: 'NumericLiteral'
+      value: 1
+    right:
+      type: 'NumericLiteral'
+      value: 2
```
GeoffreyBooth: It’s a bad idea to use 1 (or 0) as values in tests, because they can also equal true or false. Just pick any other numbers: 3, 42, whatever.

test/parser.coffee
```diff
@@ -52,7 +52,7 @@ test "new calls have a range including the new", ->
     eq node.locationData.last_column, lastColumn
 
   [assign] = block.expressions
-  outerCall = assign.value
+  outerCall = assign.value.base
```
GeoffreyBooth: Can you please explain the changes in this file?

test/strict.coffee
GeoffreyBooth: The idea is fine, but I’m not a fan of mixing styles. We can refactor to move to options arguments later:

```coffee
strict = (code, msg, ast = no) ->
```

test/strict.coffee
```diff
@@ -17,8 +17,10 @@
 # * `eval` or `arguments` as the operand of a post/pre-fix inc/dec-rement expression
 
 # helper to assert that code complies with strict prohibitions
-strict = (code, msg) ->
+strict = (code, msg, ast) ->
```
helixbass: > The idea is fine, but I’m not a fan of mixing styles. We can refactor to move to options arguments later

Ok first of all in looking more closely at this I realized that it wasn't actually triggering the `ast` versions of the tests, and it's very related to options-style arguments:

I was making calls like `strict "#{keyword}++", ast: yes`, but with the existing signature, `ast: yes` was being treated as the `msg` argument. So if we just add a sequential argument for `ast`, it'll be called like
```
strict "#{keyword}++", null, yes
```
That's pretty ugly and you basically have to go back and read the definition of `strict()` to know how to interpret the arguments

I know that you'd rather postpone these types of things until later refactors, so I updated it to use the above sequential calling style for now

For when the time comes, let me make an argument against "mixing styles" (always) being a bad thing:

I'm understanding "mixing styles" to mean using a combination of sequential and options-object arguments

To me, it's too inflexible to say "always all sequential arguments or just an options argument". I think the determining factor should be what the "natural" style is to call the function. Often times the first (couple) arguments are self-evident and required so it's more awkward to label them as options than just to supply them as sequential arguments

Take `CoffeeScript.compile()` as an existing example of "mixed style". You call it like eg
```
CoffeeScript.compile code, ast: yes
```
To insist on non-mixed style, you'd have to change the signature to eg
```
CoffeeScript.compile {code, ast: yes}
```
This isn't bad (especially with the succintness of object shorthand), it's far preferable to a fully sequential calling style. But I think there's a reason this API was designed the way it was, `code` is a rather obvious first argument to `compile()` so doesn't particularly benefit from being an option arg. But then everything after it should definitely be an option

Again (thinking of the discussion about trailing implicit option object calling style), I think Ruby is perhaps the strongest influence on Coffeescript "style" in general and calling styles in particular with its emphasis on natural, uncluttered readability. And in Ruby an implicit call with trailing "implicit" options (after possibly one or more sequential arguments) would be a typical style afaik. I feel like that community would say "choose the calling interface that makes calling it easiest and most readable" rather than something as dogmatic as "all sequential or all options, always"

test/parser.coffee
helixbass: The `new` `Call` now ends up wrapped in a `Value` (related to the `new`-wrangling in `Op::constructor()`). I guess in theory the test could use `.unwrap()` to avoid explicitly "hardcoding" whether there's a `Value` wrapper or not, but there were only two tests that needed updating

test/strict.coffee
GeoffreyBooth: Yeah maybe sequential here is being too dogmatic. Though I think the original authors would instead make a new helper like `strictAst` that could have the same signature as `strict`.

test/strict.coffee
```diff
@@ -19,6 +19,9 @@
 # helper to assert that code complies with strict prohibitions
 strict = (code, msg) ->
   throws (-> CoffeeScript.compile code), null, msg ? code
+strictAst = (code, msg) ->
```
helixbass: > I think the original authors would instead make a new helper like `strictAst` that could have the same signature as `strict`

You're right that does seem in keeping with the existing style. Updated to use `strictAst()` here and `assertErrorMessageAst()` in `test/error_messages.coffee`

GeoffreyBooth:
So I think this PR is mostly done except for the comments I didn’t click “mark resolved” on above. Namely, I feel like we’re doing something wrong in the grammar if we need to patch the location data to rope in a `new` to the left of a `Call`. I know this is a problem in the non-AST codebase as well, but is it something we can fix?

That would bypass the issue of mutating versus overwriting `locationData`, though I agree with you that we should update that method that mutates it to make it an overwrite (`@locationData = (new location data object)`).

test/abstract_syntax_tree.coffee
GeoffreyBooth: Not sure what you mean?

lib/coffeescript/coffeescript.js
```diff
@@ -99,7 +99,7 @@
     }
     sources[filename].push(code);
     if (generateSourceMap) {
-      map = new SourceMap;
+      map = new SourceMap();
```
helixbass: Since `new` always ends up being a `Call` now, we're generating eg `new SourceMap()` from `new SourceMap`, which is in line with eg Prettier's formatting of simple no-arg `new`

src/nodes.coffee
```diff
@@ -1077,7 +1077,7 @@ exports.Value = class Value extends Base
   isBoolean      : -> @bareLiteral(BooleanLiteral)
   isAtomic       : ->
     for node in @properties.concat @base
-      return no if node.soak or node instanceof Call
+      return no if node.soak or node instanceof Call or node instanceof Op and node.operator is 'do'
```
helixbass: This had to be added for `Parens` not to decide to unparenthesize its output for eg `new (do ->).b`

I didn't totally grasp conceptually what `isAtomic()` means but this seems to make sense as `do` is semantically equivalent to a `Call`

src/nodes.coffee
```diff
@@ -3298,17 +3304,24 @@ exports.Op = class Op extends Base
...
 
-    if op is 'do'
-      return Op::generateDo first
     if op is 'new'
-      if (firstCall = first.unwrap()) instanceof Call and not firstCall.do and not firstCall.isNew
-        return firstCall.newInstance()
-      first = new Parens firstCall if firstCall instanceof Code and firstCall.bound or firstCall.do
+      if ((firstCall = unwrapped = first.unwrap()) instanceof Call or (firstCall = unwrapped.base) instanceof Call) and not firstCall.do and not firstCall.isNew
+        return new Value firstCall.newInstance(), if firstCall is unwrapped then [] else unwrapped.properties
+      first = new Parens first unless first instanceof Parens or first.unwrap() instanceof IdentifierLiteral or first.hasProperties?()
```
helixbass: Compared to the previous version
```
first = new Parens firstCall if firstCall instanceof Code and firstCall.bound or firstCall.do
```
, this takes a more conservative approach and defaults to parenthesizing the argument to "simple" (no-arg/no-call-parens) `new` unless it's clearly something that doesn't need to be parenthesized in the output eg `new A` or `new A.B`

Added a test case checking that a more complex argument (`new try A`) parenthesizes its output correctly; this test would have failed with the new "always convert `new` to a `Call`" approach without this

src/nodes.coffee
```diff
@@ -3460,9 +3471,9 @@ exports.Op = class Op extends Base
     if o.level >= LEVEL_ACCESS
       return (new Parens this).compileToFragments o
     plusMinus = op in ['+', '-']
-    parts.push [@makeCode(' ')] if op in ['new', 'typeof', 'delete'] or
```
helixbass: Removed the parts of `Op::compileUnary()` that were specific to compiling `new`

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -443,3 +434,193 @@ test "AST location data as expected for Parens node", ->
...
+    loc:
+      start:
+        line: 1
+        column: 0
+      end:
+        line: 1
+        column: 11
+
+# test "AST location data as expected for Call node", ->
+#   testAstLocationData 'new Old',
```
helixbass: Commented out `new` AST test cases, as these will now be handled by `Call`'s AST code (which we don't have in place yet)

src/nodes.coffee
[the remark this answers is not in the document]
GeoffreyBooth: So if we’re following the “immutable `locationData`“ rule, don’t we need a line `@locationData = {}`? Like

```coffee
if @locationData and @needsUpdatedStartLocation
  @locationData = {}
  # ... then same as above
```
So that we don’t mutate the old object originally referenced by `@locationData`, in case it’s shared by another node.

GeoffreyBooth:
This looks great. Just that last note about immutable `locationData` and I think this is done.

src/nodes.coffee
```diff
@@ -1305,14 +1305,24 @@ exports.Call = class Call extends Base
   # expands the range on the left, but not the right.
   updateLocationDataIfMissing: (locationData) ->
     if @locationData and @needsUpdatedStartLocation
-      @locationData.first_line = locationData.first_line
-      @locationData.first_column = locationData.first_column
-      @locationData.range[0] = locationData.range[0]
+      @locationData = Object.assign {},
```
helixbass: Ok updated to not mutate `@locationData`

I think by starting from eg `@locationData = {}` we would've lost `last_line`/`last_column`, so I used an `Object.assign()`-based pattern

vendethiel: Don't we support `{...o}`?

GeoffreyBooth: CoffeeScript does but Node 6 doesn't. The compiler is keeping compatibility with Node 6+.

PR #5113, opened 2018-09-29
Fix #5112: A string of `, ` in an array should not be detected as an elision

### LINKED ISSUE #5112: Bug: string interpolation introduces invalid newline in nested object
There is a bug in how string interpolation in an object within an array are handled if they are followed by a comma. In short, the following issue should be self-explanatory.

### Input Code
```coffee
[
  foo: "#{x}, #{y}"
]
```

### Expected Behavior
```js
[
  {
    foo: `${x}, ${y}`
  }
];
```

### Current Behavior
```js
[
  {
    foo: `${x},
  ${y}`
  }
];
```

### Possible Solution
unkown, it seems that the parser misidentifiers the content of the string as part of an object?

### Context
the newline causes bugs in the following code

### Environment
coffeescript v2.3.2

GeoffreyBooth:
Fix #5112: A string of `, ` in an array should not be detected as an elision

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2469,10 +2469,10 @@
           return [this.makeCode('[]')];
         }
         o.indent += TAB;
-        fragmentIsElision = function(fragment) {
-          return fragmentsToText(fragment).trim() === ',';
+        fragmentIsElision = function([fragment]) {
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
@zdenko can you please review this? Elision was your addition I think.

zdenko:
@GeoffreyBooth LGTM

PR #5117, opened 2018-10-07
Call AST

helixbass:
@GeoffreyBooth AST for `Call`s (including `new`)

Based on `op-ast`, [here](https://github.com/helixbass/copheescript/compare/op-ast...call-ast) is just the diff against that branch


src/grammar.coffee
```diff
@@ -520,7 +520,7 @@ grammar =
   # The list of arguments to a function call.
   Arguments: [
     o 'CALL_START CALL_END',                    -> []
-    o 'CALL_START ArgList OptComma CALL_END',   -> $2
+    o 'CALL_START ArgList OptComma CALL_END',   -> $2.implicit = $1.generated; $2
```
helixbass: Preserve whether the call was implicit through the grammar

src/nodes.coffee
```diff
@@ -1416,6 +1423,19 @@ exports.Call = class Call extends Base
...
+    if @isNew
+      'NewExpression'
+    else
+      'CallExpression'
+
+  astProperties: ->
+    return
+      callee: @variable.ast()
+      arguments: arg.ast() for arg in @args
+      optional: !!@soak
```
helixbass: `optional` is similar to what we saw with `MemberExpression` (this is how Babel was representing the proposed JS version of soak operators, though they now seem to have switched to distinguishing via `type` (here would be `OptionalCallExpression`) as of Babel 7)

src/nodes.coffee
```diff
@@ -1416,6 +1423,19 @@ exports.Call = class Call extends Base
...
+      implicit: !!@implicit
```
helixbass: `implicit` is the only extension to the standard `CallExpression`/`NewExpression` AST

src/nodes.coffee
helixbass: Probably, but I'd like to see if I can find an explanation as to why they switched

src/rewriter.coffee
```diff
@@ -655,7 +655,7 @@ exports.Rewriter = class Rewriter
   # primitive string and separately passing any expected token data properties
   exposeTokenDataToGrammar: ->
     @scanTokens (token, i) ->
-      if token.data and Object.keys(token.data).length or token[0] is 'JS' and token.generated
+      if token.data and Object.keys(token.data).length or token[0] in ['JS', 'CALL_START'] and token.generated
```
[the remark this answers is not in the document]
helixbass: From my understanding, `generated` is a preexisting pattern that sets `generated` directly on `token` rather than using the token data mechanism. So this wasn't catching it

I guess one option would be to update all current uses of `generated` to be set as token data? Not sure if that's the type of refactor you'd rather treat separately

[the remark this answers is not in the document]
GeoffreyBooth: Maybe rather than looking for token tag, should we look for `.generated`?

```coffee
if token.generated or (token.data and Object.keys(token.data).length isnt 0)
```

I guess the question is, what’s the harm in applying this for all generated tokens rather than only generated `JS` and `CALL_START` ones. Are those two special because they’re generated in the rewriter? Are they the only token types generated in the rewriter, and if not, should we include the other types (or if not, why not)?

Just wanting to make sure we understand what’s going on here. A comment explaining it might be good to add.

src/nodes.coffee
helixbass: If you're curious, https://github.com/babel/babel/issues/7256 discusses the switch (to `OptionalMemberExpression`/`OptionalCallExpression`). Haven't finished reading through it yet, but one thing that has implications if we want to mimic it is that `OptionalMemberExpression` "cascades" (so eg in `a?.b.c`, both the inner (ie `a?.b`) and outer (`a?.b.c`) are `OptionalMemberExpression`s - the inner has `optional: true` corresponding to it actually having a `?`, while the outer has `optional: false`)

helixbass: @GeoffreyBooth yes I think so. But I'd like to tackle that as a separate PR as there's some complexity there. And in the meantime I'll keep putting up individual node class PRs

helixbass: @GeoffreyBooth sure added a TODO comment on a test that should end up using `OptionalMemberExpression` (unless you meant commit a failing test?)

GeoffreyBooth: I was thinking we'd add a commented out failing test, but either way works.

PR #5119, opened 2018-10-13
Range/Slice AST

helixbass:
@GeoffreyBooth AST for `Range`/`Slice`

Based on `call-ast`, [here](https://github.com/helixbass/copheescript/compare/call-ast...range-ast) is just the diff against that branch


src/nodes.coffee
```diff
@@ -1683,6 +1697,12 @@ exports.Range = class Range extends Base
     args   = ', arguments' if hasArgs(@from) or hasArgs(@to)
     [@makeCode "(function() {#{pre}\n#{idt}for (#{body})#{post}}).apply(this#{args ? ''})"]
 
+  astProperties: -> {
```
helixbass: `Range` is a Coffeescript-specific node type, but its AST representation seems pretty clear-cut

src/nodes.coffee
```diff
@@ -1719,6 +1739,9 @@ exports.Slice = class Slice extends Base
           "+#{fragmentsToText compiled} + 1 || 9e9"
     [@makeCode ".slice(#{ fragmentsToText fromCompiled }#{ toStr or '' })"]
 
+  ast: ->
```
helixbass: `Slice` just forwards AST generation to `Range`

test/abstract_syntax_tree.coffee
```diff
@@ -561,90 +561,121 @@ test "AST as expected for Index node", ->
...
-#   testExpression 'x[...]',
-#     properties: [
-#       range:
-#         type: 'Range'
-#         exclusive: yes
-#         equals: ''
-#         from: undefined
-#         to: undefined
-#     ]
+  # testExpression 'for x in [42...43] then',
```
GeoffreyBooth: I’m assuming these tests are commented out because we don’t yet support AST output for `for` or `in`?

GeoffreyBooth:
Just one question, otherwise this looks good.

test/abstract_syntax_tree.coffee
helixbass: Yup exactly so just included tests for "standalone" ranges and slices for now

PR #5120, opened 2018-10-13
Arr/Elision/Splat AST

helixbass:
@GeoffreyBooth AST for `Arr`/`Elision`/`Splat` nodes

Based on `range-ast`, [here](https://github.com/helixbass/copheescript/compare/range-ast...array-ast) is the diff against that branch


src/grammar.coffee
```diff
@@ -351,7 +351,7 @@ grammar =
   # A splat that occurs outside of a parameter list.
   Splat: [
     o 'Expression ...',                         -> new Splat $1
-    o '... Expression',                         -> new Splat $2
+    o '... Expression',                         -> new Splat $2, {postfix: no}
```
helixbass: Preserve whether the splat dots were prefix or postfix

There will be similar updates in the grammar for param and object rest splats as we get to them

src/nodes.coffee
```diff
@@ -1948,6 +1971,13 @@ exports.Arr = class Arr extends Base
       obj = obj.unwrapAll()
       obj.eachName iterator
 
+  astType: -> 'ArrayExpression'
```
helixbass: Once we've added AST support for `Assign`, this will have additional logic based on whether it's an "assigned-to" (ie left-hand-side) array or not, since those are `type: 'ArrayPattern'`

Similar goes for `Splat`, which as a left-hand-side will be `type: 'RestElement'`

src/nodes.coffee
```diff
@@ -3204,6 +3234,13 @@ exports.Splat = class Splat extends Base
 
   unwrap: -> @name
 
+  astType: -> 'SpreadElement'
+
+  astProperties: -> {
+    argument: @name.ast()
+    @postfix
```
helixbass: `postfix` is an extension to the Babel-style `SpreadElement` AST

On a related note, you've mentioned in previous code reviews that we prefer only using postfix `...` in the Coffeescript codebase. In the ESLint integration, I created a `spread-direction` rule (which this exposing of `postfix` in the AST made super easy) that enforces such a preference. So when we get to the point of setting that up in this repo, we can turn on that rule and have ESLint do the enforcement for us

src/nodes.coffee
```diff
@@ -3242,6 +3279,9 @@ exports.Elision = class Elision extends Base
 
   eachName: (iterator) ->
 
+  ast: ->
+    null
```
helixbass: `Elision`s just generate `null` ASTs in their parent `Arr`'s list of `elements`

src/nodes.coffee
```diff
@@ -1962,6 +1972,13 @@ exports.Arr = class Arr extends Base
       obj = obj.unwrapAll()
       obj.eachName iterator
 
+  astType: -> 'ArrayExpression'
+
+  astProperties: ->
+    return
+      elements:
+        object.ast() for object in @objects
```
GeoffreyBooth: This is very elegant 🥇

src/nodes.coffee
GeoffreyBooth: Yeah, that’s kind of surprising, at least to me. I guess Babel has their reasons, I would’ve expected an `Elision` node type or something.

GeoffreyBooth:
This one looks fine to me, though we need to merge in #5119 first. This can get merged in as soon as that one is.

helixbass:
@GeoffreyBooth ok rebased so this should be ready to merge

src/nodes.coffee
GeoffreyBooth: The left-hand-side `...` didn’t exist until CoffeeScript 2, when it was added as an equivalent option to the `...` that’s been around for years. Hence the preference for `...`, because naturally that’s what the CoffeeScript compiler codebase itself uses since it predates CoffeeScript 2.

There are definitely times for the left side version though, such as when using Vue:

```coffee
export default
  name: 'my-component'

  computed: {
  }
```

Without the option for the left-hand `...`, the `mapGetters` here would need to be written as `mapGetters('user', ['profile'])...` which isn’t terrible, but less elegant than parentheses-less; and less idiomatic with JS Vue code.

PR #5123, opened 2018-10-19
Import/export AST

helixbass:
@GeoffreyBooth AST for `import`/`export`


src/grammar.coffee
```diff
@@ -485,6 +485,7 @@ grammar =
     o 'EXPORT DEFAULT Expression',                           -> new ExportDefaultDeclaration $3
     o 'EXPORT DEFAULT INDENT Object OUTDENT',                -> new ExportDefaultDeclaration new Value $4
     o 'EXPORT EXPORT_ALL FROM String',                       -> new ExportAllDeclaration new Literal($2), $4
+    o 'EXPORT { } FROM String',                              -> new ExportNamedDeclaration new ExportSpecifierList([]), $5
```
helixbass: Seems like this case was missing

src/nodes.coffee
```diff
@@ -874,6 +875,12 @@ exports.StringLiteral = class StringLiteral extends Literal
     unquoted = unquoted.replace /\\n/g, '\n' if csx
     unquoted
 
+  astProperties: ->
```
helixbass: There will be more to string/heredoc ASTs but this gets a basic version in place

src/nodes.coffee
```diff
@@ -979,6 +986,12 @@ exports.BooleanLiteral = class BooleanLiteral extends Literal
     value: if @value is 'true' then yes else no
     name: @originalValue
 
+exports.DefaultLiteral = class DefaultLiteral extends Literal
```
helixbass: Added a `DefaultLiteral` subclass so that it can have the proper AST (look like an `Identifier`)

src/nodes.coffee
```diff
@@ -2344,6 +2367,12 @@ exports.ImportClause = class ImportClause extends Base
 
     code
 
+  ast: ->
+    compact flatten [
+      @defaultBinding?.ast()
+      @namedImports?.ast()
+    ]
```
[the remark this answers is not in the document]
GeoffreyBooth: @helixbass wrote:

`ImportClause`‘s AST ends up as `specifiers` on `ImportDeclaration` and is expected to be an array of specifier ASTs. `@defaultBinding` is an `ImportDefaultDeclaration` whose AST is a single object. `@namedImports` is either an `ImportNamespaceDeclaration` (whose AST is a single object) or an `ImportSpecifierList` (whose AST is a list of `ImportSpecifier` objects)

So `flatten` is a convenient way to iron that out into the expected shape (a non-nested list of specifier objects)

`compact` combined with the `?.ast()` guards is a way to avoid building up the list manually via conditionals (it’ll get rid of the `undefined`s from when `@defaultBinding` or `@namedImports` doesn’t exist)

Without either you’d end up with something like
```coffee
ast: ->
  ret = []
  if @defaultBinding?
    ret.push @defaultBinding.ast()
  if @namedImports?
    if @namedImports instanceof ImportNamespaceSpecifier
      ret.push @namedImports.ast()
    else
      ret.push @namedImports.ast()...
  ret
```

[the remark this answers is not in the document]
GeoffreyBooth: @helixbass wrote:

`ImportClause`‘s AST ends up as `specifiers` on `ImportDeclaration` and is expected to be an array of specifier ASTs. `@defaultBinding` is an `ImportDefaultDeclaration` whose AST is a single object. `@namedImports` is either an `ImportNamespaceDeclaration` (whose AST is a single object) or an `ImportSpecifierList` (whose AST is a list of `ImportSpecifier` objects)

So `flatten` is a convenient way to iron that out into the expected shape (a non-nested list of specifier objects)

`compact` combined with the `?.ast()` guards is a way to avoid building up the list manually via conditionals (it’ll get rid of the `undefined`s from when `@defaultBinding` or `@namedImports` doesn’t exist)

Without either you’d end up with something like
```coffee
ast: ->
  ret = []
  if @defaultBinding?
    ret.push @defaultBinding.ast()
  if @namedImports?
    if @namedImports instanceof ImportNamespaceSpecifier
      ret.push @namedImports.ast()
    else
      ret.push @namedImports.ast()...
  ret
```

PR #5124, opened 2018-10-23
Object AST

helixbass:
@GeoffreyBooth AST for `Obj`

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2583,7 +2583,7 @@ exports.ExportSpecifier = class ExportSpecifier extends ModuleSpecifier
 exports.Assign = class Assign extends Base
   constructor: (@variable, @value, @context, options = {}) ->
     super()
-    {@param, @subpattern, @operatorToken, @moduleDeclaration, @shorthand, @originalContext = @context} = options
+    {@param, @subpattern, @operatorToken, @moduleDeclaration, @originalContext = @context} = options
 
   children: ['variable', 'value']
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -930,7 +930,7 @@ exports.ComputedPropertyName = class ComputedPropertyName extends PropertyName
     [@makeCode('['), @value.compileToFragments(o, LEVEL_LIST)..., @makeCode(']')]
 
   ast: ->
-    Object.assign @value.ast(), @astLocationData()
+    @value.ast()
 
 exports.StatementLiteral = class StatementLiteral extends Literal
   isStatement: YES
@@ -1923,15 +1923,9 @@ exports.Obj = class Obj extends Base
     {variable, value, shorthand} = property
     isComputedPropertyName = variable instanceof Value and variable.base instanceof ComputedPropertyName
 
-    key =
-      if isComputedPropertyName
-        variable.base.value
-      else
-        variable.unwrap()
-
     Object.assign
       type: 'ObjectProperty'
-      key: key.ast()
+      key: variable.ast()
       value: value.ast()
       shorthand: !!shorthand
       computed: !!isComputedPropertyName
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -929,6 +929,9 @@ exports.ComputedPropertyName = class ComputedPropertyName extends PropertyName
   compileNode: (o) ->
     [@makeCode('['), @value.compileToFragments(o, LEVEL_LIST)..., @makeCode(']')]
 
+  ast: ->
+    Object.assign @value.ast(), @astLocationData()
+
 exports.StatementLiteral = class StatementLiteral extends Literal
   isStatement: YES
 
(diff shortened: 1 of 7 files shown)


src/nodes.coffee
```diff
@@ -1894,6 +1897,48 @@ exports.Obj = class Obj extends Base
       answer.push @makeCode join
     if @front then @wrapInParentheses answer else answer
 
+  expandProperty: (property) ->
+    {variable, context, operatorToken} = property
+    key = if property instanceof Assign and context is 'object'
```
helixbass: This has some duplication with `Obj::compileNode()` but I don't see an easy way to share code (however, see comments re a missing `ObjectProperty` abstraction, which might change that)

src/nodes.coffee
```diff
@@ -1894,6 +1897,48 @@ exports.Obj = class Obj extends Base
...
+    @expandProperty(property) for property in @properties
+
+  getPropertyAst: (property) ->
+    return property.ast() if property instanceof Splat
+
+    {variable, value, shorthand} = property
+    isComputedPropertyName = variable instanceof Value and variable.base instanceof ComputedPropertyName
+
+    Object.assign
+      type: 'ObjectProperty'
```
helixbass: You'll see this as a code smell. And I agree. In theory I guess this could be delegated to `Assign`'s AST generation (which we haven't gotten to yet). But that feels overloaded/messy and would seem to require making `Assign`s more self-aware of whether they're in fact an object "assignment" or not

To me this screams "missing abstraction"/new node class(/subclass). The logic that I tried to encapsulate in `expandProperty()`/`expandProperties()` to convert the various types that may be in an `Obj`'s `@properties` into a more normalized form (and the similar logic in `Obj::compileNode()`) combined with then having to overload `Assign` for object properties would presumably all disappear if there was a separate node-class-level abstraction for object properties. What is the upside of the existing reuse of `Assign` for object properties?

But (a) I'm not sure exactly how/where to introduce that new abstraction (b) that would be a major refactor, not sure it's worth trying to introduce now?

src/nodes.coffee
```diff
@@ -2519,7 +2564,7 @@ exports.ExportSpecifier = class ExportSpecifier extends ModuleSpecifier
 exports.Assign = class Assign extends Base
   constructor: (@variable, @value, @context, options = {}) ->
     super()
-    {@param, @subpattern, @operatorToken, @moduleDeclaration, @originalContext = @context} = options
+    {@param, @subpattern, @operatorToken, @moduleDeclaration, @shorthand, @originalContext = @context} = options
```
helixbass: This "making `Assign` more self-aware of whether/how it's an object property" via `@shorthand` would also seem to be evidence of a missing abstraction

src/nodes.coffee
```diff
@@ -1894,6 +1897,48 @@ exports.Obj = class Obj extends Base
...
+    Object.assign
+      type: 'ObjectProperty'
+      key: variable.ast()
+      value: value.ast()
+      shorthand: !!shorthand
+      computed: !!isComputedPropertyName
+      method: no
+    , property.astLocationData()
+
+  astType: -> 'ObjectExpression'
```
helixbass: As with arrays, this will get more logic once we have ASTs for assignments (and thus left-hand-side objects)

src/nodes.coffee
GeoffreyBooth: This reminds me of when I was working on `import`/`export`, and I created a `ModuleDeclaration` class to be the parent of `ImportDeclaration` and `ExportDeclaration`. Here we could have something like `BaseAssign` and then `Assign` and `ObjectAssign`. At first `BaseAssign` would have everything that `Assign` does now and `Assign` just extends it without adding anything, and `ObjectAssign` has the extra code specific to objects. I would get that working first, then try to separate out code that only `Assign` needs and `ObjectAssign` doesn’t and gradually move all the “only needed by one of the children but not the other” code out from the base and into the children. I think it’s worth some experimentation.

helixbass: @GeoffreyBooth ok starting to think through this, the biggest thing that makes me question whether subclassing `Assign` is the right abstraction is that there are object properties that aren't assignments/`Assign`s (eg shorthand properties (`{a, [b], @c}`) or splats)

But I should look at `Assign`'s implementation and see what it might take to extract the object-property stuff out completely

src/nodes.coffee
```diff
@@ -1894,6 +1897,66 @@ exports.Obj = class Obj extends Base
...
+      variable
+    else if property instanceof Assign
+      operatorToken.error "unexpected #{operatorToken.value}" unless @lhs
+      variable
+    else
+      property
+    if key instanceof Value and key.hasProperties()
+      key.error 'invalid object key' unless context isnt 'object' and key.this
+      if property instanceof Assign
+        return new ObjectProperty fromAssign: property
```
helixbass: Ok I think I came up with a lightweight way to "normalize" this

Now `Obj::expandProperty()` returns either an `ObjectProperty` (new node class) or a `Splat` (mimicking the target AST shape, where `ObjectProperty`s live side-by-side with `SpreadElement`s)

Ultimately we might want to extend this back through eg the grammar and use `ObjectProperty`s during compile-to-JS, but for now the conversion to `ObjectProperty`s is encapsulated here in `Obj::expandProperty()` so only used when generating AST (avoiding potential breakage of existing code/messier refactor)

src/nodes.coffee
```diff
@@ -1894,6 +1897,66 @@ exports.Obj = class Obj extends Base
...
+
+  astType: -> 'ObjectExpression'
+
+  astProperties: ->
+    properties:
+      property.ast() for property in @expandProperties()
+    implicit: !!@generated
+
+exports.ObjectProperty = class ObjectProperty extends Base
+  constructor: ({key, fromAssign}) ->
```
helixbass: `ObjectProperty::constructor()` is a little more "clever" than a typical node class constructor, knowing how to deconstruct an `Assign` and self-initializing its own `@locationData`

The target shape is basically:
- `@key` is always initialized
- `@value` is initialized for everything except "simple" shorthand properties (eg `{a, [b], @c}`)
- `@shorthand` is set to `true` for shorthand properties (ie when `shorthand: true` in the AST)
- `@locationData` is initialized

This makes the conversion to AST (ie `ObjectProperty::astProperties()`) straightforward

src/nodes.coffee
```diff
@@ -1894,6 +1897,66 @@ exports.Obj = class Obj extends Base
...
+    new ObjectProperty key: property
+
+  expandProperties: ->
+    @expandProperty(property) for property in @properties
+
+  astType: -> 'ObjectExpression'
+
+  astProperties: ->
+    properties:
+      property.ast() for property in @expandProperties()
```
helixbass: So now to generate `properties` on an `Obj`'s AST we can just delegate to children's `.ast()` methods in the typical way

PR #5126, opened 2018-10-28
Assign AST

helixbass:
@GeoffreyBooth AST for `Assign`

The `ObjectProperty` abstraction I introduced in `obj-ast` (#5124) looks like it worked correctly for left-hand-side object destructuring 👍 

Based on `obj-ast`, [here](https://github.com/helixbass/copheescript/compare/obj-ast...assign-ast) is the diff against that branch

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2583,7 +2583,7 @@ exports.ExportSpecifier = class ExportSpecifier extends ModuleSpecifier
 exports.Assign = class Assign extends Base
   constructor: (@variable, @value, @context, options = {}) ->
     super()
-    {@param, @subpattern, @operatorToken, @moduleDeclaration, @shorthand, @originalContext = @context} = options
+    {@param, @subpattern, @operatorToken, @moduleDeclaration, @originalContext = @context} = options
 
   children: ['variable', 'value']
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -930,7 +930,7 @@ exports.ComputedPropertyName = class ComputedPropertyName extends PropertyName
     [@makeCode('['), @value.compileToFragments(o, LEVEL_LIST)..., @makeCode(']')]
 
   ast: ->
-    Object.assign @value.ast(), @astLocationData()
+    @value.ast()
 
 exports.StatementLiteral = class StatementLiteral extends Literal
   isStatement: YES
@@ -1923,15 +1923,9 @@ exports.Obj = class Obj extends Base
     {variable, value, shorthand} = property
     isComputedPropertyName = variable instanceof Value and variable.base instanceof ComputedPropertyName
 
-    key =
-      if isComputedPropertyName
-        variable.base.value
-      else
-        variable.unwrap()
-
     Object.assign
       type: 'ObjectProperty'
-      key: key.ast()
+      key: variable.ast()
       value: value.ast()
       shorthand: !!shorthand
       computed: !!isComputedPropertyName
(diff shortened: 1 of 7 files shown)


src/nodes.coffee
```diff
@@ -1760,7 +1763,7 @@ exports.Slice = class Slice extends Base
 
 # An object literal, nothing fancy.
 exports.Obj = class Obj extends Base
-  constructor: (props, @generated = no, @lhs = no) ->
```
helixbass: The `@lhs` arg to `Obj::constructor()` was unused

src/nodes.coffee
```diff
@@ -1812,14 +1815,7 @@ exports.Obj = class Obj extends Base
...
     # are too.
-    if @lhs
-      for prop in props when prop instanceof Assign
-        {value} = prop
-        unwrappedVal = value.unwrapAll()
-        if unwrappedVal instanceof Arr or unwrappedVal instanceof Obj
-          unwrappedVal.lhs = yes
-        else if unwrappedVal instanceof Assign
-          unwrappedVal.nestedLhs = yes
+    @propagateLhs()
```
helixbass: The biggest thing introduced here is to make the process of nodes (eg `Obj`/`Arr`/`Assign`) knowing whether they're on the left-hand-side of an assignment (ie setting `.lhs`/`.nestedLhs`) more shared/normalized/complete

Previously the recursive "LHS-self-awareness propagation" was being done in `compileNode()` (and was missing some cases eg splats). But we always need to know whether we're left-hand-side for AST generation as well (since different `type`s are used eg `ObjectExpression` vs `ObjectPattern`)

So `propagateLhs()` methods are used here to normalize this and allow the LHS-propagation to occur elsewhere besides just `compileNode()`

src/nodes.coffee
```diff
@@ -2546,18 +2640,11 @@ exports.Assign = class Assign extends Base
   compileNode: (o) ->
     isValue = @variable instanceof Value
     if isValue
-      # When compiling `@variable`, remember if it is part of a function parameter.
-      @variable.param = @param
```
helixbass: I don't know exactly when I removed this. But went back through and it appears to make good sense - `.param` is only used on `Assign`s, so here setting `@variable.param` wouldn't have an effect since `@variable` wouldn't be an `Assign`

So this is a cleanup of unused LHS-propagation

src/nodes.coffee
GeoffreyBooth: It seems a little odd that we need a method to _set_ the left-handedness on this node’s children. I would think that the nodes would all have an `isOnLeftHandSide` method or similar, that first looks at itself for an `isLHS` property and then walks up the tree looking for such a property on each parent until we hit the first non-`Obj`/`Arr`/`Assign` node. But maybe I’m not fully understanding what’s going on here.

helixbass: @GeoffreyBooth I can picture that. But the existing way seems to be to go "top-down" and have parents responsible for telling their children that they're on the left-hand-side. This is just converting it from a more "ad-hoc" approach mixed into the traversal of children during `compileNode()` into a more clean, comprehensive method-based approach

And I'd guess that this way is a bit more compact (and definitely more efficient to the extent that matters)

Also our nodes don't have a reference to their parents (so no way to traverse up the tree) no?

GeoffreyBooth: I guess you’re right. I’m looking at the traversal methods and they’re top-down like `traverseChildren`. So I guess this is the way it’s done.

Issue #5122, opened 2018-10-19
Proposal: allow caching of compilation results (to disk) when using coffeescript/register


adrian-gierakowski:
feature request

When working with large codebases and utilising `coffeescript/register` ( for example when running tests with `mocha --require coffeescript/register test.coffee` ) the startup time can increase significantly due to compilation of coffeescript files.

In a codebase I've been working on I was able to cut the startup time by 60-70%  by caching results of `CoffeeScript.compile` to disk, in a similar way as its done in [node-coffee-cache](https://github.com/trello/node-coffee-cache) or [coffee-register-cache](https://github.com/guillaume86/coffee-register-cache). 

The way those modules do it is by overwriting `require.extensions` for coffee files, in order to cache results of `CoffeeScript.compile` to disk and load the cached results when appropriate. 

However, there is a problem with this approach since when the cached results are loaded (bypassing the call to `CoffeeScript.compile`), we end up without an entry in the [sourceMaps](https://github.com/jashkenas/coffeescript/blob/7cf739e2b6da654dc5e2782a0ab860ca04f4c4e1/src/coffeescript.coffee#L55) ( nor [sources](https://github.com/jashkenas/coffeescript/blob/7cf739e2b6da654dc5e2782a0ab860ca04f4c4e1/src/coffeescript.coffee#L53)), and hence [prepareStackTrace](https://github.com/jashkenas/coffeescript/blob/7cf739e2b6da654dc5e2782a0ab860ca04f4c4e1/src/coffeescript.coffee#L335) can't do its job properly for files loaded from cache. Which is a shame since the whole point of this affair is to speedup test driven development workflow, but currently when a test fail, we often end up with invalid line numbers in stack traces.

### Current Behaviour

Currently there is no option to cache compilation results to disk and it is impossible to implement it properly (externally to the coffeescript module) since both [sourceMaps](https://github.com/jashkenas/coffeescript/blob/7cf739e2b6da654dc5e2782a0ab860ca04f4c4e1/src/coffeescript.coffee#L55) and [sources](https://github.com/jashkenas/coffeescript/blob/7cf739e2b6da654dc5e2782a0ab860ca04f4c4e1/src/coffeescript.coffee#L53) are hidden from the outside world.

### Possible Solution
[...]

PR #5130, opened 2018-11-01
implement coffeescript.registerCompiled method

adrian-gierakowski:
This method enables an external module to implement caching of
compilation results. When the compiled js source is loaded from cache,
the original coffee code should be added with this method in order to
enable the Error.prepareStackTrace below to correctly adjust the stack
trace for the corresponding file (the source map will be generated on
demand).

This is the minimal solution proposed in https://github.com/jashkenas/coffeescript/issues/5122

I have not added any new tests since the code path is already covered by existing tests (the method was created by extracting existing in-line code into a function).

I have not marked this PR as closing the above issue since I would still be inclined to implement the caching internally if anyone is interested.

### DIFF
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -4,7 +4,7 @@
   // on Node.js/V8, or to run CoffeeScript directly in the browser. This module
   // contains the main entry functions for tokenizing, parsing, and compiling
   // source CoffeeScript into JavaScript.
-  var FILE_EXTENSIONS, Lexer, SourceMap, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, sourceMaps, sources, withPrettyErrors,
+  var FILE_EXTENSIONS, Lexer, SourceMap, addSource, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, sourceMaps, sources, withPrettyErrors,
     indexOf = [].indexOf;
 
   ({Lexer} = require('./lexer'));
@@ -74,6 +74,18 @@
   // Also save source maps if generated, in form of `(source)`: [`(source map)`].
   sourceMaps = {};
 
+  // This is exported to enable an external module to implement caching of
+  // compilation results. When the compiled js source is loaded from cache, the
+  // original coffee code should be added with this method in order to enable the
+  // Error.prepareStackTrace below to correctly adjust the stack trace for the
+  // corresponding file (the source map will be generated on demand).
+  exports._addSource = addSource = function(filename, code) {
+    if (sources[filename] == null) {
+      sources[filename] = [];
+    }
+    return sources[filename].push(code);
+  };
+
   // Compile CoffeeScript code to JavaScript, using the Coffee/Jison compiler.
 
   // If `options.sourceMap` is specified, then `options.filename` must also be
@@ -94,10 +106,7 @@
     generateSourceMap = options.sourceMap || options.inlineMap || (options.filename == null);
     filename = options.filename || '<anonymous>';
     checkShebangLine(filename, code);
-    if (sources[filename] == null) {
-      sources[filename] = [];
-    }
-    sources[filename].push(code);
+    addSource(filename, code);
     if (generateSourceMap) {
       map = new SourceMap;
     }
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -79,7 +79,7 @@
   // original coffee code should be added with this method in order to enable the
   // Error.prepareStackTrace below to correctly adjust the stack trace for the
   // corresponding file (the source map will be generated on demand).
-  exports.registerCompiled = registerCompiled = function(filename, source, output, sourcemap) {
+  exports.registerCompiled = registerCompiled = function(filename, source, sourcemap) {
     if (sources[filename] == null) {
       sources[filename] = [];
     }
@@ -206,7 +206,7 @@
       sourceURL = `//# sourceURL=${(ref1 = options.filename) != null ? ref1 : 'coffeescript'}`;
       js = `${js}\n${sourceMapDataURI}\n${sourceURL}`;
     }
-    registerCompiled(filename, code, js, map);
+    registerCompiled(filename, code, map);
     if (options.sourceMap) {
       return {
         js,
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -4,7 +4,7 @@
   // on Node.js/V8, or to run CoffeeScript directly in the browser. This module
   // contains the main entry functions for tokenizing, parsing, and compiling
   // source CoffeeScript into JavaScript.
-  var FILE_EXTENSIONS, Lexer, SourceMap, addSource, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, sourceMaps, sources, withPrettyErrors,
+  var FILE_EXTENSIONS, Lexer, SourceMap, base64encode, checkShebangLine, compile, formatSourcePosition, getSourceMap, helpers, lexer, packageJson, parser, registerCompiled, sourceMaps, sources, withPrettyErrors,
     indexOf = [].indexOf;
 
   ({Lexer} = require('./lexer'));
@@ -79,11 +79,17 @@
   // original coffee code should be added with this method in order to enable the
   // Error.prepareStackTrace below to correctly adjust the stack trace for the
   // corresponding file (the source map will be generated on demand).
-  exports._addSource = addSource = function(filename, code) {
+  exports.registerCompiled = registerCompiled = function(filename, source, output, sourcemap) {
     if (sources[filename] == null) {
       sources[filename] = [];
     }
-    return sources[filename].push(code);
+    sources[filename].push(source);
+    if (sourcemap != null) {
+      if (sourceMaps[filename] == null) {
+        sourceMaps[filename] = [];
+      }
+      return sourceMaps[filename].push(sourcemap);
+    }
   };
 
   // Compile CoffeeScript code to JavaScript, using the Coffee/Jison compiler.
@@ -106,7 +112,6 @@
     generateSourceMap = options.sourceMap || options.inlineMap || (options.filename == null);
     filename = options.filename || '<anonymous>';
     checkShebangLine(filename, code);
-    addSource(filename, code);
     if (generateSourceMap) {
       map = new SourceMap;
     }
@@ -171,10 +176,6 @@
     }
     if (generateSourceMap) {
       v3SourceMap = map.generate(options, code);
-      if (sourceMaps[filename] == null) {
-        sourceMaps[filename] = [];
-      }
-      sourceMaps[filename].push(map);
     }
     if (options.transpile) {
       if (typeof options.transpile !== 'object') {
@@ -205,6 +206,7 @@
       sourceURL = `//# sourceURL=${(ref1 = options.filename) != null ? ref1 : 'coffeescript'}`;
       js = `${js}\n${sourceMapDataURI}\n${sourceURL}`;
     }
+    registerCompiled(filename, code, js, map);
     if (options.sourceMap) {
       return {
         js,
--- a/lib/coffeescript/index.js
+++ b/lib/coffeescript/index.js
@@ -152,9 +152,8 @@
     }
   }
 
-  CoffeeScript._compileFile = function(filename, options = {}) {
-    var answer, err, raw, stripped;
-    raw = fs.readFileSync(filename, 'utf8');
+  CoffeeScript._compileRawFileContent = function(raw, filename, options = {}) {
+    var answer, err, stripped;
     // Strip the Unicode byte order mark, if this file begins with one.
     stripped = raw.charCodeAt(0) === 0xFEFF ? raw.substring(1) : raw;
     options = Object.assign({}, options, {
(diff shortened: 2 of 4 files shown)


GeoffreyBooth:
I think this makes sense to potentially implement internally. One concern that comes to mind is the cache on disk. How would we ensure that it didn’t grow out of control, and that it was cleaned up? Are there situations where a user wouldn’t want caching? How do other transpilation tools like Babel or TypeScript handle caching?

jashkenas:
In general, I don't think that CoffeeScript ought to try to handle compilation caching. For a serious scale project, you should compile your code to JS, and run the app with `node`. Running code directly from `coffee` is more for experimentation and exploration.

If you compile your code, you can use makefiles (or `--watch`, or your build solution of choice), to ensure that source is never recompiled unnecessarily.

GeoffreyBooth:
@jashkenas Please take a look at #5122. The use case isn’t running an app from `coffee`, it’s for running tests. In general, test runners like `mocha` use `coffeescript/register` to transpile files for testing, and this is where caching can speed things up.

jashkenas:
In my opinion, it doesn't matter so much what the use case is for — you can compile your tests just like you compile your code.

If we try to handle compilation caching, you open up a whole new world of bugs:

* Where are the files stored?
* What if the user wants to keep the cache somewhere else?
* How is the cache keyed? 
* Keyed between different versions of CoffeeScript?
* How is the cache expired?
* What if the user leaks a secret key, that is now stashed in the invisible cache?
* What about cache hits between different commits, using the `master` branch of the CoffeeScript compiler, where the version number hasn't changed (yet)?

It really doesn't seem worth getting into (again, IMO).

adrian-gierakowski:
@GeoffreyBooth 

> One concern that comes to mind is the cache on disk. How would we ensure that it didn’t grow out of control, and that it was cleaned up? 

I would leave it to the user.

> Are there situations where a user wouldn’t want caching? 

I guess in cases where the speed-up is not significant.

> How do other transpilation tools like Babel or TypeScript handle caching?

I'd have to look into it.

In my case, I only use caching when running tests locally. On CI caching is disabled so there is no problem with managing disk usage. I've been using one of the modules mention in the proposal for months and never bothered to clean my local cache. I use a directory at the top level of each project to store the cache files so it would theoretically be possible to scan all the files in the project and prune all cache files which no longer have a corresponding source. But tbh I think deleting the entire cache directory once in a while should be good enough when working locally.

The 2 modules use different approaches to creating the cache files: one computes hashes of the sources contents and stores all cache files in the same dir using hash as the file name. The other mirrors the directory structure of the project, names cache files after the sources and checks file modification date to determine if the cache file is out of date. It seems like in the first case the cache will grow quicker since files are modified more often than moved around. However hashing I think is a more sound strategy since you can also add compile options to the content before hashing (I've just implemented it [here](https://github.com/adrian-gierakowski/coffee-register-cache/blob/master/src/index.coffee#L42))

@jashkenas 

you make some very good points, as you have by creating coffeescript itself 🙏:
- there already exists a work-flow which provides a solution to the problem
- implementing caching properly is non-trivial
- it might better to limit the scope of the core project to necessary minimum

And I guess for a large project, compiling to disk first might offer additional benefits, like reducing runtime memory footprint etc.

However, for the last few years, since I first started using your beautiful language, I have been happily living in a lalaland in which I could pretend that the language called JavaScript doesn't exits 🤘🎉💥. I never felt a need to look at the compiled version of my CS code (apart from one time when I stumbled upon a compiler bug). I also did not need to worry about any build steps or checking in compilation output to source control. I do run code in production with `coffee app.coffee` (never had any issues) and it feels nice being able to run any of my test files with a simple `mocha tests/test.coffee`. 

I understand if you don't support implementing caching within the coffeescript but I would be super grateful if you could accept this small and insignificant change which would allow me to experiment with my own implementation and happily continue using my current work-flow.

adrian-gierakowski:
@jashkenas

Some thoughts on the points you've raised:
> Where are the files stored?

directory at the top of projects root by default

> What if the user wants to keep the cache somewhere else?

that could be configured with an env var, or a command line argument

> How is the cache keyed?

hash of the: source + options + compiler version

> Keyed between different versions of CoffeeScript?

see above

> How is the cache expired?

up to the user, a tool could be provided to prune cache based on current state of the project

> What if the user leaks a secret key, that is now stashed in the invisible cache?

its up to the user to keep their secrets safe. My text editor keeps a cache of all searches I made. My shell keeps command history. If I'm not careful I might commit database credentials to a public git repo.

> What about cache hits between different commits, using the master branch of the CoffeeScript compiler, where the version number hasn't changed (yet)?

if I understand correctly, the issue here is due to switching between different dev versions of CoffeeScript compiler, which have the same version string stored in source code, but might produce different output? Again, I would simply warn the user. If you want to be super safe, you could add
a file to the CoffeeScript project which would contains the sha of the previous commit (or hash of the compiler source), and is updated on each commit (you already run a build step to compile cs->js so it could be done at the same time). The sha could be used when calculating the cache key.

GeoffreyBooth:
Now that you've spelled out all that's involved, I'm leaning toward @jashkenas and thinking that this is something best left out of core. Especially since it appears that there are other modules that can wrap CoffeeScript and handle it?

The change in this PR though is quite minimal, just exposing a previously private method. If that makes the external caching modules work better, that seems like it would be fine with me. @jashkenas, what do you think of this PR specifically?

adrian-gierakowski:
@jashkenas thanks! 

I've implemented the `registerCompiled` as you suggested, however I am not sure if there is anything I should do with the `output` argument. Doesn't seem like its persisted in the original `compile` method. Do you have something in mind for the future?

Btw. below are my thoughts regarding caching the `SourceMap`:

Initially I did think about saving `SourceMap` for each file, but after looking into it I realised that it would require more work to implement and in majority of cases might actually slow things down (assuming that we don't need the source map for most files most of the time - reasoning inspired by [this comment](https://github.com/jashkenas/coffeescript/blob/294bb4754ec566d4fb10fd80fa8056be3d715bd7/src/coffeescript.coffee#L46)). Moreover most of the time when running files with `coffee` or using `coffeescript/register` the `SourceMap` is not returned from [`.compile`](https://github.com/jashkenas/coffeescript/blob/294bb4754ec566d4fb10fd80fa8056be3d715bd7/src/coffeescript.coffee#L161) since `options.sourceMap` is undefined (we get the inline map since [`_compileFile`](https://github.com/jashkenas/coffeescript/blob/294bb4754ec566d4fb10fd80fa8056be3d715bd7/src/index.coffee#L122) always sets `options.inlineMap` to true). To set `options.sourceMap` someone would have to do `module.options.sourceMap = yes` somewhere in their source since running `coffee --map a.coffee` executes in "compilation mode" and results in `a.js.map` file being saved but doesn't actually run the code in `a.coffee`.

So I think in my external caching implementation I'd only cache the `js` output and omit the `sourcemap` arg from the call to `registerCompiled`.

adrian-gierakowski:
@jashkenas btw would you mind if I also added a step to in the `build` cake task which would complete a `sha256` of concatenated sources the `src` folder and store it in the a file at the top level of the repo? Then if this file could be distributed with he coffeescript package I should be able to locate it and include the sha when calculating cache keys. Or even better, the sha could be exported by the coffeescript module :)

adrian-gierakowski:
also please let me know what you think about: b47c4bc61f2cd39102b5fd7108f850a01908a205 which 
allows me to avoid a duplicate call to `fs.readFileSync` in case of a cache miss. 

see:
https://github.com/adrian-gierakowski/coffee-register-cache/blob/187a37cc60023cd3aec36b04f117c071b4e077e7/src/index.coffee#L75

https://github.com/adrian-gierakowski/coffee-register-cache/blob/187a37cc60023cd3aec36b04f117c071b4e077e7/src/index.coffee#L63

https://github.com/jashkenas/coffeescript/blob/294bb4754ec566d4fb10fd80fa8056be3d715bd7/src/index.coffee#L114

GeoffreyBooth:
@adrian-gierakowski The `coffeescript` package already exports the version:

```
$ coffee --eval 'console.log require("coffeescript").VERSION'
2.3.2
```

NPM won’t let me republish an already-published version, so this basically a unique hash. Could this suffice instead of a sha?

adrian-gierakowski:
@GeoffreyBooth the sha was just to insure against cases when someone is using a dev version of `coffeescript` where the behaviour might change between commits while the `.VERSION` stays the same.

As @jashkenas poited out above:
> What about cache hits between different commits, using the master branch of the CoffeeScript compiler, where the version number hasn't changed (yet)?

GeoffreyBooth:
@adrian-gierakowski Sure, then a hash could be generated. Please make sure it doesn’t noticeably slow down the `except-parser` version of the build process.

GeoffreyBooth:
@adrian-gierakowski This is quite good, thanks for making the effort, especially with the tests. A few minor notes:

- The codebase’s filenames use underscores, not slashes. Maybe just rename `src-sha.js` to `sha.js`?
- Please put the output SHA file in `lib/coffeescript` with the rest of the CoffeeScript generated files. I don’t see a reason for it to be at the package root, especially since people should be able to retrieve the SHA via `require('coffeescript').SHA`.
- The documentation is generated from the comments, so we write them in complete sentences with Markdown formatting. Please update your comments to match.
- The code style is to use single quotes unless the string has an interpolation (`"foo#{bar}"`).

The `Cakefile` changes don’t seem to cache the `fs.readdirSync` results, and if I’m reading it correctly we read the folder contents twice (once to get the files for the SHA, again to pass to the CoffeeScript compiler). I’m not sure it’s worth the hassle to cache the lookup, as I feel like it must take milliseconds for a modern computer to read a directory that has a dozen files in it, but it’s something to consider.

To answer your question about Jison, yes, technically that’s true; but the [last commit](https://github.com/zaach/jison/commits/master) to Jison was in 2017, and the last commit before that was in 2016. So we’re not likely to be updating that dependency version anytime soon; _however_ you would cover this case by including `package.json` in the SHA calculations, since the Jison version number would be in that `package.json`. It’s maybe not a bad idea to include `package.json` or `package-lock.json` in the SHA calculations anyway, to invalidate caches in case CoffeeScript’s dependency versions change, though the dependencies are only dev dependencies so it really shouldn’t matter.

adrian-gierakowski:
@GeoffreyBooth I have addressed your comments. Opted for including the `package-lock.json` just in case. Also added a `sha` cake task for cases when `package-lock.json` changes (but not the sources). 

I left `""` in the test titles as it seems like all other tests use them even though there is no string interpolation.

I am still not sure if we should keep the `output` argument in the `registerCompiled` method as it is not used for anything.

adrian-gierakowski:
@GeoffreyBooth @jashkenas any thoughts?

jashkenas:
(I know it's frustrating, but ...) My (abbreviated) thoughts are that I don't think we should merge this change.

As a minor thing — the SHA hashing makes me uncomfortable. As you pointed out above, just because two CoffeeScript source files hash to the same value, doesn't mean that their JS output necessarily does (across different versions of CoffeeScript, or just different versions of the Jison dependency).

But the more major things are that I still don't fully understand how a user would be expected to understand or able to use these hooks to easily achieve cached compilation. And they're now a more invasive change than they were previously ... and accordingly something that demands a more compelling use case.

If it were up to me, I would say: "Compile your tests as you write them, just like you compile your code", and avoid this whole dilemma.

But if there's still a minimal, non-invasive, change here that would let you satisfy your use case, then that would be fine.

vendethiel:
> (across different versions of CoffeeScript, or just different versions of the Jison dependency).

or across `-l`, `-b` and `--no-header` flags.

GeoffreyBooth:
Rereading this thread, it looks like the hashing was only done to support development versions of CoffeeScript, e.g. someone like me working on a checkout of the CoffeeScript repo generating new builds. I don’t think the three or so of us currently working on CoffeeScript need complication caching support, and the rest of the world should be fine with caching compilation based on `CoffeeScript.VERSION`. What if we go back to using that, and get rid of all the hashing stuff?

adrian-gierakowski:
@jashkenas @GeoffreyBooth 

The only change I really need is 333b78c (or its revised form 51947304).

And b47c4bc6 would be really nice to have.

Re hashing of compiler source: I am happy to remove it from this PR. Since the overhead of it seems to be very low, I can just as well do it when my module loads (I'd locate and hash the js files in this case). 

@jashkenas
> But the more major things are that I still don't fully understand how a user would be expected to understand or able to use these hooks to easily achieve cached compilation. 

Here's a working example: https://github.com/adrian-gierakowski/coffee-register-cache

> If it were up to me, I would say: "Compile your tests as you write them, just like you compile your code"

I'll be sure to try this in my next project. But I think it's nice to have options :) 


@vendethiel 
> or across -l, -b and --no-header flags.

The implementation linked above includes  flags when generating cache keys.

GeoffreyBooth:
@adrian-gierakowski Do you mind changing this branch back to the ideal version without the hashing? And then it’s up to the consuming tool to hash if they wish, however it wants (SHA of source files, CoffeeScript version number, whatever). Especially if the compiler won’t be doing the caching, it makes some sense to leave as many of the specifics of the caching to the downstream tool.

adrian-gierakowski:
@GeoffreyBooth @jashkenas 
I have deleted the commits related to hashing. The PR as it is now would make it possible for me to implement caching in an external module efficiently and without hacks. 

Let me know if you want me to remove or keep the `output` argument in `registerCompiled` as it is not used at the moment (as mentioned above https://github.com/jashkenas/coffeescript/pull/5130#issuecomment-435919589 )

jashkenas:
Sure. If you're not using it — let's try with `output` out.

adrian-gierakowski:
I have now removed the `output` arg from `registerCompiled`. Let me know if you are happy with current state and if you want me to squash the commits. Cheers!

GeoffreyBooth:
Nice work @adrian-gierakowski, thanks for bearing with us 😀

adrian-gierakowski:
Thank you guys, I appreciate it!

PR #5135, opened 2018-11-15
Existence/Throw/Expansion AST

helixbass:
@GeoffreyBooth PR for ASTs for Existence, Throw and Expansion nodes

Sorry to have fallen off the pace. Was getting crushed on a two-week solo project. But through with that, and finished getting `ast` (ie the new AST methods) merged into my branch (with Prettier plugin and ESLint plugin tests passing against the updated branch) so set up for smooth sailing 👍 

That being said, these are the last of the "easy" leaf node type ASTs. I think it makes sense next to introduce block ASTs so that we can start getting into functions/control structures/etc

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3444,6 +3444,11 @@ exports.Expansion = class Expansion extends Base
 
   eachName: (iterator) ->
 
+  astType: -> 'RestElement'
+
+  astProperties: ->
+    argument: null
+
 #### Elision
 
 # Array elision element (for example, [,a, , , b, , c, ,]).
@@ -3869,6 +3874,11 @@ exports.Throw = class Throw extends Base
     fragments.push @makeCode ';'
     fragments
 
+  astType: -> 'ThrowStatement'
+
+  astProperties: ->
+    argument: @expression.ast()
+
 #### Existence
 
 # Checks a variable for existence -- not `null` and not `undefined`. This is
@@ -3910,6 +3920,13 @@ exports.Existence = class Existence extends Base
         if @negated then '===' else '!=='
       code = "#{code} #{cmp} #{@comparisonTarget}"
     [@makeCode(if o.level <= LEVEL_COND then code else "(#{code})")]
+
+  astType: -> 'UnaryExpression'
+
+  astProperties: ->
+    argument: @expression.ast()
+    operator: '?'
+    prefix: no
 
 #### Parens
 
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -3444,6 +3444,11 @@ exports.Expansion = class Expansion extends Base
 
   eachName: (iterator) ->
 
+  astType: -> 'RestElement'
```
helixbass: While `Expansion`s are Coffeescript-specific, it makes sense to model them like their close cousins splats, so `type: 'RestElement'` just with a `null` `argument` field

src/nodes.coffee
```diff
@@ -3911,6 +3921,13 @@ exports.Existence = class Existence extends Base
       code = "#{code} #{cmp} #{@comparisonTarget}"
     [@makeCode(if o.level <= LEVEL_COND then code else "(#{code})")]
 
+  astType: -> 'UnaryExpression'
```
helixbass: Initially I'd exposed `Existence` as its own unique Coffeescript-specific `type: 'Existence'`

But then realized that conceptually it's a `UnaryExpression` (like eg `!a`), just with a postfix operator. `UnaryExpression` ASTs already have a `prefix` field, so this doesn't even require extending the Babel-style AST at all

src/nodes.coffee
GeoffreyBooth: Unless `RestElement` with a `null` `argument` might someday be used for something else in the Babel spec?

helixbass: That's an interesting consideration. Conceptually I think this representation makes sense (an expansion basically is a "spread without an argument") and I'd tend not to worry too hard about potential future evolution of the Babel spec (as we've already seen, it's a bit of a moving target)

src/nodes.coffee
helixbass: Sorry I didn't mean literally the same as `!a`, just using `!` as an example where the general semantics is similar to `Existence` (a unary operator whose output is a boolean based on whether its argument is some form of "truthy" or not)

So to be clear, the AST for `!a` looks like
```
type: 'UnaryExpression'
operator: '!'
argument:
  type: 'Identifier'
  name: 'a'
prefix: yes
```
whereas the AST for `a?` looks like
```
type: 'UnaryExpression'
operator: '?'
argument:
  type: 'Identifier'
  name: 'a'
prefix: no
```

GeoffreyBooth: That makes much more sense. Sounds good 👍

PR #5136, opened 2018-11-16
JSX element AST

helixbass:
@GeoffreyBooth PR for JSX elements

This doesn't include fragments (`<></>`), attributes (`<div a="b" />`) or child content (`<div>abc</div>`)

Based on `existence-ast`, [here](https://github.com/helixbass/copheescript/compare/existence-ast...helixbass:jsx-element-ast) is the diff against that branch


GeoffreyBooth:
> Is it really worth not parsing it legitimately?

What would that look like?

GeoffreyBooth:
Thanks @helixbass. I just wanted to get the to-do list documented while it was still fresh in mind. @xixixao do you have any thoughts on this?

@helixbass if you think the refactor is a good idea, and it sounds like you do (and that it is), perhaps we should open an issue with the list from your last comment? And we could potentially tackle it after the AST work is done.

src/nodes.coffee
```diff
@@ -1443,6 +1460,88 @@ exports.Call = class Call extends Base
...
+      #   @CSXFragmentToAst {tagName, attributes, content}
+    ,
+      children: []
+        # if content and not content.base.isEmpty?()
+        #   content.base.csx = yes
+        #   compact flatten [
+        #     content.ast()
+        #   ]
+        # else
+        #   []
```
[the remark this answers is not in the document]
helixbass: This is where we'll actually generate the AST nodes for JSX content, but I kept this PR minimal

test/error_messages.coffee
```diff
@@ -1587,9 +1587,9 @@ test "CSX error: non-matching tag names", ->
     <div><span></div></span>
   ''',
   '''
-    [stdin]:1:7: error: expected corresponding CSX closing tag for span
+    [stdin]:1:6: error: expected corresponding CSX closing tag for span
     <div><span></div></span>
-          ^^^^
+         ^^^^^
```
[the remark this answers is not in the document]
helixbass: Sure restored this behavior/test

src/lexer.coffee
```diff
@@ -555,18 +555,40 @@ exports.Lexer = class Lexer
...
+      [input, id] = match
+      fullId = id
+      if '.' in id
+        [id, properties...] = id.split '.'
+      else
+        properties = []
+      tagToken = @token 'CSX_TAG', id,
+        length: id.length + 1
+        data:
+          openingBracketToken: @makeToken '<', '<'
```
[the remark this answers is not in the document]
helixbass: > Why is `<` a “sub-token” rather than a delimiter like strings’ `"`?

Strings we tokenize as strings and parse as strings. What JSX exposes as tokens for parsing doesn't resemble the actual source code - we put in fake tokens so that JSX gets parsed as nonexistent calls, arrays etc

eg `<div a="b" />` gets "transformed" so that to the parser it looks like `div([{a: 'b'}])`

But in the process we'd be losing track of all the actual original source code location data

So we need to preserve the original JSX location data as we're doing the transformation

> Or if we need to present it as a node in the AST, why not generate it in the node classes rather instead of producing this unparsed psuedo-token here?

By that time, we'd have lost the original location data. It's not about generating AST nodes using the sub-tokens, it's about generating accurate AST location data

src/nodes.coffee
```diff
@@ -1234,20 +1238,33 @@ exports.Value = class Value extends Base
...
       optional: !!property.soak
       shorthand: !!property.shorthand
 
+  astLocationData: ->
+    return super() unless @isCSXTag()
+    # don't include leading < of JSX tag in location data
+    mergeAstLocationData(
+      locationDataToAst @base.tagNameLocationData
+      locationDataToAst @properties[@properties.length - 1].locationData
+    )
```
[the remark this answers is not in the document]
GeoffreyBooth: That’s fine, at least the commas imply the continuation.

src/nodes.coffee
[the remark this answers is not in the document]
GeoffreyBooth: Okay. Do you mind adding a comment explaining that this will be uncommented when X?

xixixao:
I agree with the assessment.

src/nodes.coffee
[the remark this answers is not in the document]
helixbass: Sure added TODO comments

src/lexer.coffee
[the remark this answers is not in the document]
GeoffreyBooth: >  It’s not about generating AST nodes using the sub-tokens, it’s about generating accurate AST location data

What about the location data for JSX code in the normal compiler, not for AST output? Are we losing a big chunk of the source maps location data right now, that we should be preserving? Should we perhaps make these lexer changes on a new branch off of `master`, to improve JSX location data in the normal compiler?

src/nodes.coffee
```diff
@@ -1443,6 +1460,88 @@ exports.Call = class Call extends Base
...
+        #   content.base.csx = yes
+        #   compact flatten [
+        #     content.ast()
+        #   ]
+        # else
+        #   []
+    )
+
+  ast: ->
+    return @CSXToAst() if @csx
```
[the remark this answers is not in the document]
GeoffreyBooth: Is it worth refactoring `Call` into `CallBase` or something, with child classes `Call` and `CSX`? Or maybe the base could be `Call`, with children `FunctionCall` and `CSX` (and the current `new Call` code would be refactored to use `new FunctionCall`).

src/nodes.coffee
```diff
@@ -4525,3 +4627,20 @@ mergeAstLocationData = (nodeA, nodeB) ->
     ]
     start: lesser  nodeA.start, nodeB.start
     end:   greater nodeA.end,   nodeB.end
+
+# Convert Jison-style node class location data to Babel-style location data
+locationDataToAst = ({first_line, first_column, last_line, last_column, range}) ->
```
GeoffreyBooth: Can we rename this to something like `jisonLocationDataToAstLocationData` to be clearer?

src/lexer.coffee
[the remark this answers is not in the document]
helixbass: I don't think I'm in a position to (easily) assess how badly source maps suffer due to the way we're handling JSX, or how big of a deal that is. But my rough understanding is that if source maps are not handling JSX accurately, we'd probably have to fully parse JSX in order to have accurate JSX location data propagated to source maps

Given that most of the lexer changes here are about exposing the "sub-tokens" (which wouldn't affect source maps) I don't think that it'd be worth trying to target this specifically for that purpose

src/nodes.coffee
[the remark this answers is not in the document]
helixbass: Ok updated to use a new `CSXElement` class

I took the approach of returning a `CSXElement` instance from `Call::constructor()`, as this was the only way I saw to cleanly "swap out" what were instances of `Call` for instances of this new class. Doing it this way, `CSXElement` has its own `compileNode()` method too, getting rid of the similar pattern in `Call::compileNode()` where it was immediately calling `compileCSX()` for CSX elements

src/nodes.coffee
helixbass: Sure updated

That name is perhaps a little long, would any similar alternatives be good enough in your opinion? Eg `jisonToAstLocationData` or `jisonLocationDataToAst`?

src/nodes.coffee
```diff
@@ -1296,6 +1314,115 @@ exports.LineComment = class LineComment extends Base
 
 #### Call
 
+exports.CSXElement = class CSXElement extends Base
+  constructor: ({@tagName, @attributes, @content}) ->
+    super()
+
+  children: ['tagName', 'attributes', 'content']
```
helixbass: I'm not clear enough on how exactly `children` gets used to know if this is necessary? But seems like the right value for `children` here

src/nodes.coffee
```diff
@@ -1296,6 +1314,115 @@ exports.LineComment = class LineComment extends Base
...
+        fragments.push obj.compileToFragments(o, LEVEL_PAREN)...
+    if @content
+      fragments.push @makeCode('>')
+      fragments.push @content.compileNode(o, LEVEL_LIST)...
+      fragments.push [@makeCode('</'), tag..., @makeCode('>')]...
+    else
+      fragments.push @makeCode(' />')
+    fragments
+
+  ast: ->
```
helixbass: While this is still "hairier" than most ASTs, I'd say this is a big improvement: we get to use the standard AST methods for constructing the outer `JSXElement` AST (getting rid of `Call::CSXToAst()`/`Call::CSXElementToAst()`)

src/nodes.coffee
```diff
@@ -1296,6 +1314,115 @@ exports.LineComment = class LineComment extends Base
...
+    if @content?
+      @closingElementLocationData = mergeAstLocationData(
+        jisonLocationDataToAstLocationData tagName.closingTagOpeningBracketLocationData
+        jisonLocationDataToAstLocationData tagName.closingTagClosingBracketLocationData
+      )
+
+    super()
+
+  astType: ->
+    'JSXElement'
```
helixbass: If I were doing this "from scratch", I'd just name the class `JSXElement` and not have to override `astType()`

But I named it `CSXElement` since we seem to be using the naming convention `CSX` (as opposed to `JSX`)

I'm not convinced that's worth doing (it really _is_ JSX, so not sure what the benefit is of labeling it with a different name "CSX"), but not going to push to rename everything `CSX` -> `JSX`

GeoffreyBooth: Yeah I’ve been planning on renaming all the CSX stuff to JSX for a long time now. The original author was fond of CSX as the name for “JSX in CoffeeScript” and hence the code was written that way and the PR for it described it as such, but when we wrote the docs for it we quickly realized that we had to call it JSX or else no one would have any idea what we were talking about; especially since it exactly matched the syntax for JSX. So anyway I’m 100% on board with just renaming all the CSX references to JSX if you want. We could maybe do it on a PR against `master` if that makes sense (and then merge that into this).

src/nodes.coffee
GeoffreyBooth: What I didn’t like about ones that ended in `ToAst` was that it implied that it was perhaps returning more than just the location data. Let’s leave it long, I want to discourage this function’s use anyway 😉because it’s only needed when we’re not following the pattern of one node class per one AST node.

src/nodes.coffee
GeoffreyBooth: `children` is used by the node traversal methods, like `unwrap` and `traverseChildren` and so on. So basically any nodes that this element contains would be children.

src/nodes.coffee
GeoffreyBooth: All the `compileNode` stuff that shows up as new code above is really only “new” in that it moved, right?

src/nodes.coffee
```diff
@@ -1314,6 +1314,115 @@ exports.LineComment = class LineComment extends Base
...
+      fragments.push @makeCode(' />')
+    fragments
+
+  ast: ->
+    # The location data spanning the opening element < ... > is captured by
+    # the generated Arr which contains the element's attributes
+    @openingElementLocationData = locationDataToAst @attributes.base.locationData
+
+    tagName = @tagName.base
+    tagName.locationData = tagName.tagNameLocationData
```
GeoffreyBooth: I would think with a name like `tagName` it would be a string, but I guess this is the node for the JSX identifier? Like `<MyComponent>`?

src/nodes.coffee
```diff
@@ -1314,6 +1314,115 @@ exports.LineComment = class LineComment extends Base
...
+          node.loc.end = {
+            line: node.loc.end.line
+            column: node.loc.end.column + rangeDiff
+          }
+        currentExpr = closingElement.name
+        while currentExpr.type is 'JSXMemberExpression'
+          shiftAstLocationData currentExpr unless currentExpr is closingElement.name
+          shiftAstLocationData currentExpr.property
+          currentExpr = currentExpr.object
+        shiftAstLocationData currentExpr
```
GeoffreyBooth: So all of this `openingElement` and `closingElement` code is because we don’t have node classes for those?

src/nodes.coffee
```diff
@@ -1324,7 +1433,12 @@ exports.Call = class Call extends Base
     if @variable instanceof Value and @variable.isNotCallable()
       @variable.error "literal is not a function"
 
-    @csx = @variable.base instanceof CSXTag
+    if @variable.base instanceof CSXTag
+      return new CSXElement(
+        tagName: @variable
+        attributes: @args[0]
+        content: @args[1]
+      )#.withLocationDataFrom @
```
GeoffreyBooth: What’s `#.withLocationDataFrom @`?

GeoffreyBooth:
I think this is probably ready, please just answer the questions (I assume you’ll have good explanations for everything) and we can merge this. I’m going to trust you on all the lexer and node class stuff, I don’t really have the time to give a more thorough review than basically just looking for trivial style complaints and code smell sniffing.

If you want to rename all CSX stuff to JSX I would also support that, though maybe we should do that in its own PR so it’s an isolated commit. Maybe after we finish the JSX-related AST PRs?

src/nodes.coffee
helixbass: Yes, exactly, it was `Call::compileCSX()`

src/nodes.coffee
helixbass: Ya that makes sense but can't think of a better name - I'd think "tag" refers to everything between (and including) `<`...`>` (eg `<MyComponent a="b">`), while the "tag name" is just eg `MyComponent`

src/nodes.coffee
helixbass: Ah ok. My instinct is to do that in a separate commit/PR (after we're through the JSX AST stuff like you suggested) but do it against this branch (`ast`) rather than `master` - there will be enough updates to JSX-related code in this branch that I think it'd be more merging pain that it was worth to temporarily additionally target `master`

src/nodes.coffee
helixbass: Right, now with the introduction of `CSXElement` we can use the normal AST methods for the outermost JSX element AST node (`JSXElement`), but the immediate AST children `JSXOpeningElement`/`JSXClosingElement` still don't have corresponding node classes so their AST nodes have to be constructed manually

I don't see a way to introduce separate node classes for the opening/closing elements since there aren't corresponding structures in our pseudo-representation of JSX (everything between and including an opening and closing element, ie a whole `JSXElement`, is represented by a single `Call`)

PR #5138, opened 2018-11-23
JSX fragment/attribute AST

helixbass:
@GeoffreyBooth PR for JSX fragments (eg `<></>`) and attributes (eg `<div a="b" c={d} {...e} />`)

So the only other remaining JSX piece is JSX content, which I can address once we've introduced ASTs for string interpolations

Based on `jsx-element-ast`, [here](https://github.com/helixbass/copheescript/compare/jsx-element-ast...jsx-fragment-ast) is the diff against that branch

### DIFF
--- a/test/csx.coffee
+++ b/test/csx.coffee
@@ -58,8 +58,8 @@ test 'object attribute', ->
     <div x={{y: 42}} />
   ''', '''
     <div x={{
-        y: 42
-      }} />;
+      y: 42
+    }} />;
   '''
 
 test 'attribute without value', ->
@@ -230,9 +230,9 @@ test 'nested CSX within an attribute, with object attr value', ->
   ''', '''
     <Company>
       <Person name={<NameComponent attr3={{
-          'a': {},
-          b: '{'
-        }} />} />
+      'a': {},
+      b: '{'
+    }} />} />
     </Company>;
   '''
 
@@ -241,11 +241,11 @@ test 'complex nesting', ->
     <div code={someFunc({a:{b:{}, C:'}{}{'}})} />
   ''', '''
     <div code={someFunc({
-        a: {
-          b: {},
-          C: '}{}{'
-        }
-      })} />;
+      a: {
+        b: {},
+        C: '}{}{'
+      }
+    })} />;
   '''
 
 test 'multiline tag with nested CSX within an attribute', ->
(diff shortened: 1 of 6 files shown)


GeoffreyBooth:
This is trivial, but I feel like we should treat “CSX” like we do “AST” in `astProperties` or `toAst`, so it would be like `csxFragmentToAst` or `checkValidCsxAttribute`.

The `getCSXAttributeAst` and related stuff is definitely smelly. I guess this is because we’re skipping the refactor you were describing, of creating proper nodes for all these parts? So this is the best we can do unless we bite the bullet and do that refactor?

GeoffreyBooth:
Resolving the merge conflicts for this will take some thought, as I guess this was done before you created the CSXElement class.

helixbass:
> Resolving the merge conflicts for this will take some thought, as I guess this was done before you created the `CSXElement` class.

@GeoffreyBooth yup I tried to give you a [heads up](https://github.com/jashkenas/coffeescript/pull/5136#pullrequestreview-178349147) so you wouldn't have to waste time on the merge conflicts!

Rebased and updated to use `CSXElement`

> The `getCSXAttributeAst` and related stuff is definitely smelly

Let me try and see if we can apply a similar technique here like we did with `CSXElement` to avoid these patterns on eg `Obj` and `Assign` serving "double-duty" as JSX stand-ins

> This is trivial, but I feel like we should treat “CSX” like we do “AST” in `astProperties` or `toAst`, so it would be like `csxFragmentToAst` or `checkValidCsxAttribute`

Sure - if we're able to avoid the "double-duty" overloading of non-JSX node classes, that'll actually get rid of the need to specify "CSX" in method names since they'll be on a class that itself is CSX/JSX-specific (eg `getCSXAttributeAst()` became just `getAttributeAst()` on `CSXElement`)

test/csx.coffee
```diff
@@ -230,9 +230,9 @@ test 'nested CSX within an attribute, with object attr value', ->
   ''', '''
     <Company>
       <Person name={<NameComponent attr3={{
-          'a': {},
-          b: '{'
-        }} />} />
+      'a': {},
+      b: '{'
+    }} />} />
```
GeoffreyBooth: Can you please explain why the indentation levels changed in these tests?

helixbass: Looks like it's because there's no longer the
```
idt = o.indent += TAB
```
which was set before calling `@compileCSXAttributes()` in `Obj::compileNode()`

Either way we wouldn't have consistent indentation of multiline JSX attribute values (with respect to their enclosing JSX element) since nested JSX elements aren't indented in a structured way when printing (we're just preserving whitespace), but these are mostly an improvement

PR #5140, opened 2018-11-29
Try AST

helixbass:
@GeoffreyBooth PR for `Try` AST (introduces first non-root `Block`s)

Based on `root-ast`, [here](https://github.com/helixbass/copheescript/compare/root-ast...helixbass:try-ast) is the diff against that branch


src/grammar.coffee
```diff
@@ -629,16 +629,16 @@ grammar =
...
+    o 'TRY Block FINALLY Block',                -> new Try $2, null, $4, LOC(3)(new Literal $3)
+    o 'TRY Block Catch FINALLY Block',          -> new Try $2, $3, $5, LOC(4)(new Literal $4)
   ]
 
   # A catch clause names its error and runs a block of code.
   Catch: [
-    o 'CATCH Identifier Block',                 -> [$2, $3]
-    o 'CATCH Object Block',                     -> [LOC(2)(new Value($2)), $3]
-    o 'CATCH Block',                            -> [null, $2]
+    o 'CATCH Identifier Block',                 -> new Catch $3, $2
```
helixbass: By introducing a new `Catch` node class, we are able to use the standard AST methods as well as not have to do any additional work to preserve the location data of the `catch` keyword

src/grammar.coffee
```diff
@@ -629,16 +629,16 @@ grammar =
   # The variants of *try/catch/finally* exception handling blocks.
   Try: [
     o 'TRY Block',                              -> new Try $2
-    o 'TRY Block Catch',                        -> new Try $2, $3[0], $3[1]
-    o 'TRY Block FINALLY Block',                -> new Try $2, null, null, $4
-    o 'TRY Block Catch FINALLY Block',          -> new Try $2, $3[0], $3[1], $5
+    o 'TRY Block Catch',                        -> new Try $2, $3
+    o 'TRY Block FINALLY Block',                -> new Try $2, null, $4, LOC(3)(new Literal $3)
```
helixbass: For the `finally` keyword we do have to do extra work to preserve its location data

src/nodes.coffee
```diff
@@ -270,7 +270,7 @@ exports.Base = class Base
   # as JSON. This is what the `ast` option in the Node API returns.
   # We try to follow the [Babel AST spec](https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md)
   # as closely as possible, for improved interoperability with other tools.
-  ast: ->
+  ast: (o) ->
```
helixbass: This PR introduces the `o` param, as it's needed as an argument for `.isStatement()`, which we need in order to be able to decide whether to wrap a block's expressions/statements in `ExpressionStatement` AST nodes or not

test/abstract_syntax_tree.coffee
```diff
@@ -278,13 +291,13 @@ test "AST as expected for ComputedPropertyName node", ->
     implicit: yes
 
 test "AST as expected for StatementLiteral node", ->
-  testExpression 'break',
+  testStatement 'break',
```
helixbass: By being explicit about whether we're testing an "expression" (which is wrapped in an `ExpressionStatement`) or a statement, we're testing that the AST is correctly doing that wrapping or not for different node types

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -2438,3 +2438,249 @@ test "AST location Data as expected for CSXTag node", ->
...
+        range: [4, 14]
+        loc:
+          start:
+            line: 1
+            column: 4
+          end:
+            line: 1
+            column: 14
+      ]
+      # TODO: the location data for the block is wrong (wrong location data on generated INDENT)
```
helixbass: This gives a taste of the kind of generated `Block`/`INDENT`/`OUTDENT`/comments/location data stuff we'll run into. Left this particular case as a TODO as we can deal with this stuff in its own later PR

src/nodes.coffee
GeoffreyBooth: Do you mind please explaining this a little more thoroughly? I've never fully understood the `o` param.

helixbass: Sure, I think you could call `o` a "context" param, basically an object that holds various configuration/other stuff relevant to the currently executing code (in this case, during AST generation inside `nodes.coffee`)

Whether doing AST generation (ie calling `nodes.ast(options)`) or transpile-to-JS (`nodes.compileToFragments(options)`), its initial value is the outer `options` argument to `compile()` in `coffeescript.coffee` (so basically the command-line options or equivalent per my understanding, eg `options.bare`)

Some of the big things that get passed around on `o` inside `nodes.coffee` are the current scope (`o.scope`) and the current compilation "level" (eg `LEVEL_LIST`). We haven't needed to pass around `o` yet while doing AST generation since we haven't introduced any logic yet that relies on things like scope or compilation level (but we will)

But now we need to use the existing `isStatement()` method to help determine the block-level AST (where non-statement "expressions" need to get wrapped in an `ExpressionStatement` AST node while statement nodes (eg `ExportNamedDeclaration`) don't)

Since `isStatement()` expects `o` as a param (to allow node classes to dynamically determine whether they're a statement or not based on things like current compilation level), we need to start passing `o` around now

So basically it gets passed down the same way as when transpiling to JS - the initial value of `o` gets passed in as the `options` from `compile()` in `coffeescript.coffee`, and then it just keeps getting passed down (and modified as appropriate) while we traverse the node tree (so primarily via calls to `.ast()`/`.compileNode()` etc)

GeoffreyBooth:
@helixbass do you mind resolving conflicts?

helixbass:
@GeoffreyBooth sure merged in `ast`

test/abstract_syntax_tree_location_data.coffee
GeoffreyBooth: Everything here looks really good, this is my only hesitation. Why is this location data incorrect? Is it something particular to `try` or rather to all blocks?

helixbass: It applies to blocks in general - basically in the rewriter (eg `fixOutdentLocationData()`) there's some stuff going on with assigning location data for generated tokens like `INDENT`/`OUTDENT` that isn't accurate enough for our AST purposes

Now that we're getting into blocks, I could do a PR soon for this - I do think it'll be cleaner to do in its own PR since it applies to blocks in general and (like most things in the rewriter) takes a bit of mental energy to think through

GeoffreyBooth: Okay I’ll merge this in, and then do you mind doing the blocks PR next? I feel like we should fix that sooner than later, before we do a bunch of other PRs for the other indented node classes like `If` and so on.

helixbass: @GeoffreyBooth sure that makes sense

PR #5145, opened 2019-01-07
Fix 5085

brewingcode:
Thanks for all the great work with CoffeeScript, I was just trolling through the issues and this looked like a simple fix. The fix is based on an assumption that I could be _completely_ wrong about, but according to the tests (including the new test for the bug itself), the assumption is correct.

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -2968,7 +2968,7 @@
         return this.body.traverseChildren(false, (node) => {
           if (node instanceof ThisLiteral) {
             return node.value = this.name;
-          } else if (node instanceof Code && node.bound && node.isStatic) {
+          } else if (node instanceof Code && node.bound && (node.isStatic || !node.name)) {
             return node.context = this.name;
           }
         });
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1989,7 +1989,7 @@ exports.ExecutableClassBody = class ExecutableClassBody extends Base
     @body.traverseChildren false, (node) =>
       if node instanceof ThisLiteral
         node.value   = @name
-      else if node instanceof Code and node.bound and node.isStatic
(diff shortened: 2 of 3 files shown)


brewingcode:
Force-pushed small change the wording of the commit: "Assuming that a **bound** Code block..."

brewingcode:
....and one more force-push to remove an unnecessary step from the test.

GeoffreyBooth:
Thanks for this! @connec do you care to take a look?

GeoffreyBooth:
@connec or @zdenko or @helixbass, anyone familiar with the `Class` node have a minute to review a one-line fix?

vendethiel:
Can you please add a test for `class A then fn(b: => ...)`?

brewingcode:
@vendethiel Sorry, I don't follow: what do you want the test to do?

vendethiel:
Make sure the function inside the call inside an object is bound to the class (we shouldn’t attribute a name outside of class properties IIRC so the test should already work).

brewingcode:
I'm not totally sure if this is what you meant...?

222067d

If it's not, could you just write the exact code you mean? Sorry that I'm just not getting it.

vendethiel:
```
myvar = 0
fn = (o) -> o.prop()
class Klass
  @foo = 1
  fn prop: => myvar = @foo
```

brewingcode:
```
class Klass
  fn prop: => myvar = @foo
```
I don't understand what this is doing at all, or what the intent of writing it would be, but OK, test added. :-)

Is `myvar` _supposed_ to be 1 after this...?

vendethiel:
There was an old bug where we considered object properties outside of the class top-level and gave them names, just adding a test.

GeoffreyBooth:
So @vendethiel is this okay to merge?

vendethiel:
LGTM

GeoffreyBooth:
Thanks @brewingcode!

brewingcode:
You're welcome!

PR #5152, opened 2019-01-22
AST location data fixes

helixbass:
@GeoffreyBooth PR to fix the known block location data issue (see #5140) and a few other location data bugs


src/nodes.coffee
```diff
@@ -4803,20 +4803,46 @@ makeDelimitedLiteral = (body, options = {}) ->
...
+  return no unless a.last_line is b.last_line
+  a.last_column > b.last_column
+
 # Take two nodes’ location data and return a new `locationData` object that
 # encompasses the location data of both nodes. So the new `first_line` value
 # will be the earlier of the two nodes’ `first_line` values, the new
 # `last_column` the later of the two nodes’ `last_column` values, etc.
 mergeLocationData = (locationDataA, locationDataB) ->
-  return
-    first_line:   lesser locationDataA.first_line,   locationDataB.first_line
```
helixbass: This logic was buggy (in `mergeLocationData()` and similarly in `mergeAstLocationData()`) - we can't individually compare line and column, we need to treat line/column as a single object for comparison (eg line 4 column 0 is greater than line 3 column 2, so we want to end up with line 4 column 0, not line 4 column 2)

src/rewriter.coffee
```diff
@@ -512,18 +512,19 @@ exports.Rewriter = class Rewriter
...
         line = column = 0
-        range = [0, 0]
+        rangeIndex = 0
       token[2] = {
         first_line:   line
         first_column: column
         last_line:    line
         last_column:  column
-        range
+        range: [rangeIndex, rangeIndex]
```
helixbass: When I looked at the stuff on my branch in the rewriter related to `INDENT`/`OUTDENT`/block location data, it was pretty much just related to comments (which we haven't gotten to yet). It turned out that the issue we were seeing with the block location data was due to this bug where I wasn't being consistent with `range`

src/rewriter.coffee
```diff
@@ -512,18 +512,19 @@ exports.Rewriter = class Rewriter
       return 1 if     token[2]
       return 1 unless token.generated or token.explicit
       if token[0] is '{' and nextLocation=tokens[i + 1]?[2]
-        {first_line: line, first_column: column, range} = nextLocation
+        {first_line: line, first_column: column, range: [rangeIndex]} = nextLocation
       else if prevLocation = tokens[i - 1]?[2]
-        {last_line: line, last_column: column, range} = prevLocation
+        {last_line: line, last_column: column, range: [, rangeIndex]} = prevLocation
+        column += 1
```
helixbass: This is another place where we could suffer from "column + 1" being not always accurate (ie where we really want to know whether this token ends with a newline and thus incrementing its `last_line`/`last_column` should give us the first column of the next line, not the next column in the current line)

I tried playing around with tracking this when creating tokens so that we could correctly handle cases where "column + 1" isn't correct. It seemed to be working, but I didn't see any current use cases to test against where a token/node that gets used for AST location data ends with an included newline. So I saved that branch and can revisit it if we see actual test cases where column + 1 is causing inaccuracy in AST location data

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -2695,17 +2695,16 @@ test "AST as expected for Try node", ->
...
-      # end: 14
-      # range: [4, 14]
-      # loc:
-      #   start:
-      #     line: 1
-      #     column: 4
-      #   end:
-      #     line: 1
-      #     column: 14
+      start: 3
```
helixbass: This is one character before what might be expected (3 vs 4). This is because when assigning the artificial location for the generated `INDENT` token, it pretends that it's a zero-width token one character after the preceding token

So in this example (`try cappuccino`), rather than assigning the block the exact location data of `cappuccino`, it's assigning the block the location data including the preceding space (` cappuccino`). I think this is fine, if we end up seeing any issues with this we can look more closely at other strategies for assigning the `INDENT` location data I guess

GeoffreyBooth:
Thank you for fixing this! This all looks great.

PR #5154, opened 2019-01-29
Switch AST

helixbass:
@GeoffreyBooth PR for Switch AST

### DIFF
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -47,9 +47,14 @@ o = (patternString, action, options) ->
     getAddDataToNodeFunctionString = (first, last) ->
       "yy.addDataToNode(yy, @#{first}#{if last then ", @#{last}" else ''})"
 
+    returnsLoc = /^LOC/.test action
     action = action.replace /LOC\(([0-9]*)\)/g, getAddDataToNodeFunctionString('$1')
     action = action.replace /LOC\(([0-9]*),\s*([0-9]*)\)/g, getAddDataToNodeFunctionString('$1', '$2')
-    performActionFunctionString = "$$ = #{getAddDataToNodeFunctionString(1, patternCount)}(#{action});"
+    performActionFunctionString =
+      if returnsLoc
+        "$$ = #{action};"
+      else
+        "$$ = #{getAddDataToNodeFunctionString(1, patternCount)}(#{action});"
   else
     performActionFunctionString = '$$ = $1;'
 
@@ -786,21 +791,21 @@ grammar =
   Switch: [
     o 'SWITCH Expression INDENT Whens OUTDENT',                -> new Switch $2, $4
     o 'SWITCH ExpressionLine INDENT Whens OUTDENT',            -> new Switch $2, $4
-    o 'SWITCH Expression INDENT Whens ELSE Block OUTDENT',     -> new Switch $2, $4, $6
-    o 'SWITCH ExpressionLine INDENT Whens ELSE Block OUTDENT', -> new Switch $2, $4, $6
+    o 'SWITCH Expression INDENT Whens ELSE Block OUTDENT',     -> new Switch $2, $4, LOC(5,6) $6
+    o 'SWITCH ExpressionLine INDENT Whens ELSE Block OUTDENT', -> new Switch $2, $4, LOC(5,6) $6
     o 'SWITCH INDENT Whens OUTDENT',                           -> new Switch null, $3
-    o 'SWITCH INDENT Whens ELSE Block OUTDENT',                -> new Switch null, $3, $5
+    o 'SWITCH INDENT Whens ELSE Block OUTDENT',                -> new Switch null, $3, LOC(4,5) $5
   ]
 
   Whens: [
-    o 'When'
+    o 'When',                                   -> [$1]
     o 'Whens When',                             -> $1.concat $2
   ]
 
   # An individual **When** clause, with action.
   When: [
-    o 'LEADING_WHEN SimpleArgs Block',            -> [[$2, $3]]
-    o 'LEADING_WHEN SimpleArgs Block TERMINATOR', -> [[$2, $3]]
+    o 'LEADING_WHEN SimpleArgs Block',            -> new SwitchWhen $2, $3
+    o 'LEADING_WHEN SimpleArgs Block TERMINATOR', -> LOC(1, 3) new SwitchWhen $2, $3
   ]
 
   # The most basic form of *if* is a condition and an action. The following
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -4575,7 +4575,7 @@ exports.Switch = class Switch extends Base
 
   astType: -> 'SwitchStatement'
 
-  getCasesAst: (o) ->
+  casesAst: (o) ->
     cases = []
 
     for kase, caseIndex in @cases
@@ -4612,7 +4612,7 @@ exports.Switch = class Switch extends Base
   astProperties: (o) ->
     return
       discriminant: @subject?.ast(o) ? null
-      cases: @getCasesAst o
+      cases: @casesAst o
 
 exports.SwitchWhen = class SwitchWhen extends Base
   constructor: (@conditions, @block) ->
@@ -4849,6 +4849,7 @@ makeDelimitedLiteral = (body, options = {}) ->
 # Helpers for `mergeLocationData` and `mergeAstLocationData` below.
 lesser  = (a, b) -> if a < b then a else b
 greater = (a, b) -> if a > b then a else b
+
 isAstLocGreater = (a, b) ->
   return yes if a.line > b.line
   return no unless a.line is b.line
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -4543,12 +4543,12 @@ exports.Switch = class Switch extends Base
   isStatement: YES
 
   jumps: (o = {block: yes}) ->
-    for [conds, block] in @cases
+    for {block} in @cases
       return jumpNode if jumpNode = block.jumps o
     @otherwise?.jumps o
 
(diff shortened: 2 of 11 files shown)


src/grammar.coffee
```diff
@@ -47,9 +47,14 @@ o = (patternString, action, options) ->
     getAddDataToNodeFunctionString = (first, last) ->
       "yy.addDataToNode(yy, @#{first}#{if last then ", @#{last}" else ''})"
 
+    returnsLoc = /^LOC/.test action
```
helixbass: The existing grammar-rule function always wrapped the return value (the newly constructed node class object) in a call to `LOC()` (the location-data-attaching function)

But we now want to have a grammar rule that uses a non-default `LOC()` for the returned node class object (see below). So here introduce extra logic that only wraps the return value in `LOC()` if the grammar rule isn't already explicitly doing so

src/grammar.coffee
```diff
@@ -786,21 +791,21 @@ grammar =
...
+    o 'When',                                   -> [$1]
     o 'Whens When',                             -> $1.concat $2
   ]
 
   # An individual **When** clause, with action.
   When: [
-    o 'LEADING_WHEN SimpleArgs Block',            -> [[$2, $3]]
-    o 'LEADING_WHEN SimpleArgs Block TERMINATOR', -> [[$2, $3]]
+    o 'LEADING_WHEN SimpleArgs Block',            -> new SwitchWhen $2, $3
+    o 'LEADING_WHEN SimpleArgs Block TERMINATOR', -> LOC(1, 3) new SwitchWhen $2, $3
```
helixbass: Introducing a new `SwitchWhen` class (rather than representing the `when`s as an array) makes it much more convenient to preserve its location data (including the leading `when` token)

Also, this is the rule that explicitly wraps its return value in `LOC()` (here, to not include the location data of the trailing `TERMINATOR`)

src/nodes.coffee
```diff
@@ -4543,12 +4543,12 @@ exports.Switch = class Switch extends Base
   isStatement: YES
 
   jumps: (o = {block: yes}) ->
-    for [conds, block] in @cases
+    for {block} in @cases
```
helixbass: So now `@cases` is a list of `SwitchWhen` objects rather than a list of lists

src/nodes.coffee
```diff
@@ -4573,6 +4573,53 @@ exports.Switch = class Switch extends Base
     fragments.push @makeCode @tab + '}'
     fragments
 
+  astType: -> 'SwitchStatement'
+
+  casesAst: (o) ->
```
helixbass: Switch is a pretty interesting AST transformation

We want to expand our multi-`when` cases into multiple `SwitchCase` AST nodes (so basically the structure of the `switch` statement that we output in JS) for compatibility/reuse with JS switch ASTs

Eg this:
```
switch yes
  when a, b then c
```
should generate an AST like:
```
type: 'SwitchStatement'
cases: [
  type: 'SwitchCase'
  test:
    type: 'Identifier'
    name: 'a'
  consequent: []
,
  type: 'SwitchCase'
  test:
    type: 'Identifier'
    name: 'b'
  consequent: [
    type: 'ExpressionStatement'
    expression:
      type: 'Identifier'
      name: 'c'
  ]
]
```
which is the same AST generated for this JS:
```
switch (true) {
  case a:
  case b:
    c;
}
```

So it's not a one-to-one mapping between our `SwitchCase`s and the AST `SwitchCase`s. I didn't see a great way to do that mapping into our node classes (which would allow us to use the typical call-`ast()`-on-the-children approach) so here there's a lot of manual assembling of the AST objects and their location data

src/nodes.coffee
```diff
@@ -4848,25 +4895,46 @@ mergeLocationData = (locationDataA, locationDataB) ->
 # location data object that encompasses the location data of both nodes. So the
 # new `start` value will be the earlier of the two nodes’ `start` values, the
 # new `end` value will be the later of the two nodes’ `end` values, etc.
-mergeAstLocationData = (nodeA, nodeB) ->
+mergeAstLocationData = (nodeA, nodeB, {justLeading, justEnding} = {}) ->
```
helixbass: Extended `mergeAstLocationData()` to accept optional `justLeading`/`justEnding` modifiers for when we only want to consider `nodeB`'s start or end location data (ie when we only want to extend `nodeA`'s location data in a single direction rather than potentially in both directions)

src/grammar.coffee
```diff
@@ -47,9 +47,14 @@ o = (patternString, action, options) ->
...
 
+    returnsLoc = /^LOC/.test action
     action = action.replace /LOC\(([0-9]*)\)/g, getAddDataToNodeFunctionString('$1')
     action = action.replace /LOC\(([0-9]*),\s*([0-9]*)\)/g, getAddDataToNodeFunctionString('$1', '$2')
-    performActionFunctionString = "$$ = #{getAddDataToNodeFunctionString(1, patternCount)}(#{action});"
+    performActionFunctionString =
+      if returnsLoc
+        "$$ = #{action};"
+      else
+        "$$ = #{getAddDataToNodeFunctionString(1, patternCount)}(#{action});"
```
GeoffreyBooth: Do you mind please explaining this change? Why was this necessary?

src/nodes.coffee
GeoffreyBooth: Should the cases perhaps be their own node class?

src/nodes.coffee
GeoffreyBooth: Do you mind adding a comment explaining `justLeading` and `justEnding` and when they should be used?

src/grammar.coffee
helixbass: This is part of the `LOC()`-wrapping logic described in the comment above (not sure if you started this review before I'd posted the comments)

Here, we don't do the typical wrapping in `getAddDataToNodeFunctionString()` (ie `LOC()`) if the grammar rule already calls `LOC()` on its return value

src/nodes.coffee
helixbass: So that's the thing I think would be tricky

This PR introduces a new `SwitchCase` node class, which basically corresponds to the node class we'd want here in order to be able to do a more typical child-node AST generation

The issue is that it's not a one-to-one mapping from our `SwitchCase` node class objects to AST `SwitchCase` nodes. So in order to try and set it up for typical child-node AST generation, we'd need to transform our `SwitchCase` node class objects into something, I guess:
1. They could map to different instances of our same `SwitchCase` node class, but that seems a little weird/confusing perhaps?
1. Or we could introduce a separate eg `SwitchCaseAst` node class, but that also seems kind of weird/overkill?

Additionally, whichever node class we were mapping to, we'd also need to map our `@otherwise` into one of them since it needs to generate a `SwitchCase` AST node (that's how the `default:` case is represented in Babel-style ASTs)

And either way, it seems possible that the mapping might be even messier than the AST generation logic below (with its location data wrangling)

But I could try one of these approaches and we could compare, we certainly do usually end up feeling like it's cleaner to avoid manually assembling AST nodes whenever possible

src/nodes.coffee
helixbass: Sure, added to the existing function comment and added tests for `mergeAstLocationData()` to `test/abstract_syntax_tree_location_data.coffee`

src/nodes.coffee
```diff
@@ -4573,6 +4573,57 @@ exports.Switch = class Switch extends Base
...
+  constructor: (@test, @block, {@trailing} = {}) ->
+    super()
+
+  children: ['test', 'block']
+
+  astProperties: (o) ->
+    return
+      test: @test?.ast(o) ? null
+      consequent: @block?.ast(o).body ? []
+      trailing: !!@trailing
```
helixbass: `trailing` is an extension to the Babel-style `SwitchCase` AST. It gets used by the ESLint plugin to be able to do Coffeescript-specific code path analysis on switch statements (basically it sees this and acts like there's an implicit `break` statement, which has code path/control flow implications)

src/nodes.coffee
```diff
@@ -4822,51 +4873,107 @@ isLocationDataEndGreater = (a, b) ->
 # encompasses the location data of both nodes. So the new `first_line` value
 # will be the earlier of the two nodes’ `first_line` values, the new
 # `last_column` the later of the two nodes’ `last_column` values, etc.
-mergeLocationData = (locationDataA, locationDataB) ->
+# 
+# If you only want to extend the first node’s location data with the start or
```
helixbass: Now `mergeLocationData()` (as opposed to `mergeAstLocationData()`) needs `justLeading`/`justEnding`, so updated its comment and added tests like I'd done for `mergeAstLocationData()`

src/grammar.coffee
GeoffreyBooth: Yes our comments passed each other in flight. So you’re saying that `getAddDataToNodeFunctionString` isn’t necessary for rules that invoke `LOC`, since `LOC` already adds the missing location data? I _think_ that could be correct, if all that `getAddDataToNodeFunctionString` did was add location data, but doesn’t it also add comments data?

GeoffreyBooth:
So looking at the `test/abstract_syntax_tree.coffee` file and what node class tests are still commented out, I see:

- StringLiteral
- RegexLiteral
- PassthroughLiteral
- PropertyName
- ComputedPropertyName
- Return
- YieldReturn
- AwaitReturn
- Value
- SuperCall
- Super
- RegexWithInterpolations
- Extends
- Class
- ExecutableClassBody
- Code (a.k.a. function)
- Param
- While
- Parens
- StringWithInterpolations
- For
- If

Many of these feel like they’re probably quite small, but a few like Class and Code and While and For and If seem likely to be significant. Anyway so I guess we have a ways to go, but the end is in sight.

PR #5155, opened 2019-02-05
Code AST

helixbass:
@GeoffreyBooth PR for basic Code AST

There will be other considerations related to functions, eg correctly handling directives, tracking scope variables, and handling class methods differently

But this PR includes basic AST generation for functions (and `await`/`yield`)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3557,7 +3557,11 @@ exports.Code = class Code extends Base
     for {name} in @params when name instanceof Arr or name instanceof Obj
       name.propagateLhs yes
 
-  astType: -> 'FunctionExpression'
+  astType: ->
+    if @bound
+      'ArrowFunctionExpression'
+    else
+      'FunctionExpression'
 
   paramForAst: (param) ->
     return param if param instanceof Expansion
@@ -3577,7 +3581,8 @@ exports.Code = class Code extends Base
       body: @body.ast o
       generator: !!@isGenerator
       async: !!@isAsync
-      bound: !!@bound
+      # We never generate named functions, so specify `id` as `null`, which
+      # matches the Babel AST for anonymous function expressions/arrow functions
       id: null
 
 #### Param
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3262,6 +3262,8 @@ exports.Code = class Code extends Base
       if node instanceof For and node.isAwait()
         @isAsync = yes
 
+    @propagateLhs()
+
   children: ['params', 'body']
 
   isStatement: -> @isMethod
@@ -3551,6 +3553,33 @@ exports.Code = class Code extends Base
 
     seenSuper
 
+  propagateLhs: ->
+    for {name} in @params when name instanceof Arr or name instanceof Obj
+      name.propagateLhs yes
+
+  astType: -> 'FunctionExpression'
+
+  paramForAst: (param) ->
+    return param if param instanceof Expansion
+    {name, value, splat} = param
+    if splat
+      new Splat name, lhs: yes, postfix: splat.postfix
+      .withLocationDataFrom name
+    else if value?
+      new Assign name, value, null, param: yes
+      .withLocationDataFrom locationData: mergeLocationData name.locationData, value.locationData
+    else
+      name
+
+  astProperties: (o) ->
+    return
+      params: @paramForAst(param).ast(o) for param in @params
+      body: @body.ast o
+      generator: !!@isGenerator
+      async: !!@isAsync
+      bound: !!@bound
+      id: null
+
 #### Param
 
 # A parameter in a function definition. Beyond a typical JavaScript parameter,
@@ -3661,7 +3690,7 @@ exports.Param = class Param extends Base
 # A splat, either as a parameter to a function, an argument to a call,
 # or as part of a destructuring assignment.
 exports.Splat = class Splat extends Base
-  constructor: (name, {@postfix = true} = {}) ->
+  constructor: (name, {@lhs, @postfix = true} = {}) ->
     super()
     @name = if name.compile then name else new Literal name
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -4044,6 +4044,8 @@ exports.Op = class Op extends Base
     super idt, @constructor.name + ' ' + @operator
 
   astType: ->
+    return 'AwaitExpression' if @isAwait()
+    return 'YieldExpression' if @isYield()
     switch @operator
       when '||', '&&', '?' then 'LogicalExpression'
       when '++', '--'      then 'UpdateExpression'
@@ -4056,10 +4058,21 @@ exports.Op = class Op extends Base
     secondAst = @second?.ast o
     switch
       when @isUnary()
-        return
-          argument: firstAst
+        argument =
+          if @isYield() and @first.unwrap().value is ''
+            null
+          else
+            firstAst
+        return {argument} if @isAwait()
+        return {
+          argument
+          delegate: @operator is 'yield*'
+        } if @isYield()
+        return {
+          argument
           operator: @originalOperator
           prefix: !@flip
+        }
       else
         return
           left: firstAst
--- a/test/abstract_syntax_tree.coffee
+++ b/test/abstract_syntax_tree.coffee
@@ -1503,12 +1503,99 @@ test "AST as expected for Code node", ->
     bound: no
     id: null
 
-  # testExpression '(@a) ->',
-  # testExpression '(@a = 1) ->',
-  # testExpression '({@a}) ->',
-  # testExpression '({[a]}) ->',
-  # testExpression '(...a) ->',
-  # testExpression '(a...) ->',
+  testExpression '(@a) ->',
+    type: 'FunctionExpression'
+    params: [
+      type: 'MemberExpression'
+      object:
+        type: 'ThisExpression'
+        shorthand: yes
+      property: ID 'a'
+    ]
+    body: EMPTY_BLOCK
+    generator: no
+    async: no
+    bound: no
+    id: null
+
+  testExpression '(@a = 1) ->',
+    type: 'FunctionExpression'
+    params: [
+      type: 'AssignmentPattern'
+      left:
+        type: 'MemberExpression'
+      right: NUMBER 1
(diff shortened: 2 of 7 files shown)


src/nodes.coffee
```diff
@@ -3551,6 +3553,33 @@ exports.Code = class Code extends Base
 
     seenSuper
 
+  propagateLhs: ->
+    for {name} in @params when name instanceof Arr or name instanceof Obj
+      name.propagateLhs yes
```
helixbass: Make destructured array/object params aware that they're left-hand-side (so that they'll eg generate AST nodes of type `ObjectPattern` instead of `ObjectExpression`) with the existing pattern using `propagateLhs()`

src/nodes.coffee
```diff
@@ -3551,6 +3553,33 @@ exports.Code = class Code extends Base
...
+
+  astType: -> 'FunctionExpression'
```
helixbass: There's an argument to be made that bound (fat-arrow) functions should be type `ArrowFunctionExpression` instead of `FunctionExpression` since we transpile them to ES6 arrow functions

Currently instead all (non-class-method) functions are type `FunctionExpression` and there's an additional `bound` AST property to distinguish fat-arrow from skinny-arrow functions

I can try and make them `ArrowFunctionExpression` and check how well the Prettier and ESLint plugins handle that

test/abstract_syntax_tree.coffee
```diff
@@ -91,6 +91,22 @@ test 'Confirm functionality of `deepStrictIncludeExpectedProperties`', ->
   check 'Prop is missing', throws,
     missingProp: {}
 
+# Shorthand helpers for common AST patterns.
+
+EMPTY_BLOCK =
```
helixbass: Added some shorthand helpers to make some of the common AST patterns more concise in the tests

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -3349,3 +3349,178 @@ test "AST location data as expected for Switch node", ->
...
+      start: 2
+      end: 2
+      range: [2, 2]
+      loc:
+        start:
+          line: 1
+          column: 2
+        end:
+          line: 1
+          # column: 2 TODO: make this accurate/should match range?
```
helixbass: The end column location data is off-by-one for the function body block since its end location is demarcated by a generated zero-length token:

Basically, its location data would look something like:
```
range: [2, 2]
first_line: 0
first_column: 2
last_line: 0
last_column: 2
```

But then since Jison line/column data is inclusive (while Babel's is exclusive), we do "column + 1" when converting to AST location data. For a zero-length token, that causes wonkiness/disagreement between `range` and the line/column info in the AST location data, eg here you'll end up with:
```
range: [2, 2]
loc:
  start:
    line: 1
    column: 2
  end:
    line: 1
    column: 3
```

I'd started trying to combat the "column + 1" hack by giving tokens an `endsLine` property to know whether "column + 1" is correct (as opposed to "line + 1, column 0" for a token that ends a line (including the actual newline)) - that's on my [ast-block-location-data](https://github.com/helixbass/copheescript/commits/ast-block-location-data) branch if you're curious

However, this case makes me think that approach won't work/be enough, since it wouldn't do anything about this zero-width token case. So then the other approach I'd had in mind was just to go ahead and add additional exclusive end line/column properties to token location data (eg `last_line_exclusive`/`last_column_exclusive`) - from my understanding we can't just replace the existing `last_line`/`last_column` (or convert them to be exclusive) since Jison expects them to be present and to be inclusive

So maybe I'll explore that approach in a different branch. Tracking the exclusive last line/column of tokens would in theory get rid of both the potential inaccuracy of "column + 1" for line-ending tokens (that we've already discussed in previous PRs) as well as this zero-length generated token end line/column location data issue

src/nodes.coffee
GeoffreyBooth: In astexplorer.net `const fn = () => {};` translates to `ArrowFunctionExpression`, so I guess we should be doing that. It shouldn’t be that hard, since we just look at whether it’s bound or not?

src/nodes.coffee
```diff
@@ -3551,6 +3553,33 @@ exports.Code = class Code extends Base
...
+      name
+
+  astProperties: (o) ->
+    return
+      params: @paramForAst(param).ast(o) for param in @params
+      body: @body.ast o
+      generator: !!@isGenerator
+      async: !!@isAsync
+      bound: !!@bound
+      id: null
```
GeoffreyBooth: Why `id: null`?

test/abstract_syntax_tree_location_data.coffee
GeoffreyBooth: Sounds good. Could you perhaps track in `grammar.coffee` when a zero-width token is passed in? Or maybe in the lexer, see if a token is followed by a zero-width token and preserve that info through Jison? Just trying to think of other solutions.

src/nodes.coffee
helixbass: Yup exactly. Before making that change here I'd just like to see what breaks in the Prettier and ESLint plugins when introducing `ArrowFunctionExpression`

src/nodes.coffee
helixbass: `id` is used for named functions - since we never generated named functions, we always want `id: null` (which matches the Babel AST for both arrow functions `() => {}` and anonymous function expressions `function() {}`)

test/abstract_syntax_tree_location_data.coffee
helixbass: Ya just always explicitly tracking the exclusive last line/column location data is the conceptually simplest approach I can see. I started implementing that and wasn't too bad

The issue (that I was afraid I might run into) that I found is that Jison doesn't just blindly pass along the entire location data object, it explicitly "forwards" `first_line`/`first_column`/`last_line`/`last_column` (and optionally `range`). So Jison is thwarting our attempts to pass through `last_line_exclusive`/`last_column_exclusive` every time `LOC()` gets called

Only a minimal change (two lines if you hard-code `last_line_exclusive`/`last_column_exclusive` rather than eg adding a new option to specify them) to Jison is required to pass those through as well - on my [exclusive-end-line-column](https://github.com/helixbass/copheescript/compare/code-ast...exclusive-end-line-column) branch, I have all the tests passing (including updating these AST location data tests to have the correct end column) against a version of the parser compiled against a temporarily patched version of Jison

So... what is the best way to proceed? Jison looks like it's not actively maintained. So submitting a PR to add an option allowing for passing through additional location data keys doesn't seem like it would get merged. Would there be a downside to forking it and using the forked version as our dependency?

Otherwise I'm not sure if there's a more devious way to smuggle the `last_line_exclusive`/`last_column_exclusive` through the parser

GeoffreyBooth: Why not smuggle them through like you do the other data, as properties on the string object? Or you can use `addDataToNode`, the method I’m using to pass through the comments: https://github.com/jashkenas/coffeescript/blob/master/src/helpers.coffee#L135-L153

src/nodes.coffee
GeoffreyBooth: That sounds fine, do you mind please adding a comment explaining this?

src/nodes.coffee
helixbass: Sure, updated to use `ArrowFunctionExpression` if it's bound. And removed the `bound` AST property since it would be redundant (I think we'll want to re-add it to distinguish bound vs non-bound class methods, but we can do that when we get there)

It's on me to update those plugins anyway, I can do that independently of how we handle this PR

src/nodes.coffee
helixbass: Sure, added a comment

test/abstract_syntax_tree_location_data.coffee
helixbass: I'm not seeing any visibility of the complete token location data objects in `addDataToNode()` - the `first`/`last` args are the "filtered" versions of the location data that don't include `last_line_exclusive`/`last_column_exclusive`. And I didn't see anything relevant sniffing around on the `parserState` arg

So let's think about if we could use that typical smuggling technique. We could modify the call to `addDataToNode()` in the grammar to also include args for eg `$1` (the token value) in addition to `@$1` (the filtered location data that is the current `first`/`last` args)

At that point I believe every single token would need to use the smuggling-via-`String()`-object technique, since any token could get used in a `LOC()` call for its location data

So the question I think is whether it's feasible to use that technique *everywhere*. If that would mean needing to manually call `.toString()` on every single grammar rule, that seems ugly enough that I'd question what the downside is of forking Jison to avoid having to do that

Although here's a thought - our grammar `o()` function is already doing some magic of rewriting the actual function body. Maybe we could generate the `.toString()` calls on any "bare" references to eg `$1`...?

Inve1951:
Wouldn't it make sense to go with something like `BoundFunctionExpression` (or stick with `FunctionExpression` and keep the flag) instead of `ArrowFunctionExpression`, and translate it on export?
I feel it's confusing as in CS all function expressions use arrows. This is AST for CS after all.

GeoffreyBooth:
> Wouldn’t it make sense to go with something like

We’re explicitly following the Babel AST, so we have to go with what they went with. We’re only deviating where we have things that JavaScript doesn’t, like Expansion.

Inve1951:
>We’re explicitly following the Babel AST

I do get that. Still, IMO it makes sense (in this case specifically) to not blindly copy the name when it's ambiguous. A final transform step / alias resolving, would nicely avoid such situation.

helixbass:
@GeoffreyBooth I'll tackle the last line/column stuff in a separate PR

@Inve1951 if we were designing the AST from scratch that might make sense, but there are massive upsides to generating AST that's as compatible with JS ASTs as possible (eg integration with ESLint and other tools that consume JS AST). So for anything that can be cleanly mapped to existing Babel AST nodes, that outweighs any aesthetic considerations of whether someone reading the AST might find it less obvious than an alternative representation

> A final transform step / alias resolving, would nicely avoid such situation

To have Coffeescript export a less-compatible AST and then require an external transform step in order to align it with Babel AST would largely defeat the purpose of trying for compatibility

GeoffreyBooth:
Thanks for the effort, looking forward to the other PR for the last line/column issue.

PR #5156, opened 2019-02-07
AST: Track exclusive end line/column

helixbass:
@GeoffreyBooth here's a PR for explicitly tracking token exclusive end line/column location data to avoid the issues discussed in previous PRs


GeoffreyBooth:
So you found a way to fix it without patching/forking Jison?

GeoffreyBooth:
I'm convinced. Great work!

PR #5157, opened 2019-02-09
Return AST

helixbass:
@GeoffreyBooth PR for `Return`/`YieldReturn`/`AwaitReturn` AST

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1123,51 +1123,42 @@ exports.Return = class Return extends Base
   astProperties: (o) ->
     argument: @expression?.ast(o) ? null
 
+# Parent class for `YieldReturn`/`AwaitReturn`.
+exports.FuncDirectiveReturn = class FuncDirectiveReturn extends Return
+  constructor: (expression, {@returnKeyword}) ->
+    super expression
+
+  compileNode: (o) ->
+    @checkScope o
+    super o
+
+  checkScope: (o) ->
+    unless o.scope.parent?
+      @error "#{@keyword} can only occur inside functions"
+
+  isStatementAst: NO
+
+  ast: (o) ->
+    @checkScope o
+
+    new Op @keyword,
+      new Return @expression
+      .withLocationDataFrom(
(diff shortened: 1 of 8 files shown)


src/nodes.coffee
```diff
@@ -4048,6 +4097,10 @@ exports.Op = class Op extends Base
   toString: (idt) ->
     super idt, @constructor.name + ' ' + @operator
 
+  ast: (o) ->
+    @checkContinuation o if @isYield() or @isAwait()
```
helixbass: This `Op` update isn't related to `YieldReturn`/`AwaitReturn`, it just allows us to test for "simple" `yield`/`await` outside of a function throwing an error when generating AST (in addition to the comparable tests for `yield return`/`await return` that I was adding)

test/error_messages.coffee
```diff
@@ -942,14 +942,14 @@ test "#4130: unassignable in destructured param", ->
   '''
 
 test "`yield` outside of a function", ->
-  assertErrorFormat '''
+  assertErrorFormatAst '''
```
GeoffreyBooth: Why are we changing these tests into `assertErrorFormatAst`? The same error should still be thrown for regular code being compiled into JS.

helixbass: `assertErrorFormatAst()` tests that *both* compile-to-JS and AST generation throw the error. I can see how the name might imply *just* AST, but a name that is more explicit that it's testing both would probably be really long?

GeoffreyBooth: Yeah . . . `assertErrorFormatJsAndAst`? Or make the combined check the default, and JS only an option?

helixbass: I like the idea of making the combined the default behavior of `assertErrorFormat()`

I think doing that refactor after we've gone through all the node classes makes sense since then most of them should be able to be tested against both (JS and AST) and it'll probably catch some that should be throwing during AST generation

GeoffreyBooth: Sounds good.

PR #5158, opened 2019-02-10
Remove extra word in comment in src/nodes.coffee

rdeforest:
Just a correction to a comment to add the "___to___" near the end of this sentence:

> Occasionally it may be useful to make an expression behave as if it was 'hoisted', whereby the result of the expression is available before its location in the source, but the expression's variable scope corresponds ___to___ the source position.


GeoffreyBooth:
Thanks!

Issue #5114, opened 2018-10-01
Proposal: multiline should be documented inside REPL


sr105:
I eventually found the `Ctrl-V` answer in StackOverflow, but not before I'd used the REPL for about 2 weeks thinking that multiline was broken in the REPL. Also, while StackOverflow told me to use `Ctrl-V` to enter multiline mode, it didn't mention to use that keycode to exit. So, as a Unix user, I hit Ctrl-D and the REPL quit. It would be nice if the REPL told people about `Ctrl-V` to enter and exit multiline mode. Also, it would be nice if it told people about `.help`.

CoffeeScript version 2.3.1
Node v10.4.0

GeoffreyBooth:
It’s mentioned briefly in https://coffeescript.org/#cli, but I agree a section under Usage dedicated to the REPL would be nice. Please feel free to submit a pull request.

PR #5159, opened 2019-02-10
Issue 5114 - Add greeting and Ctrl-V help to REPL

rdeforest:
An implementation of #5114.

The REPL now starts with a greeting:
```
$ bin/coffee

CoffeeScript v2.3.2 REPL, running on NodeJS v11.9.0.
Press Ctrl-V to enter and exit multi-line mode.
Enter '.help' to see a list of repl commands.

coffee>
```
And the first time the user presses Ctrl-V they get a reminder to press it again to leave multi-line mode:
```
coffee> console.log """
(Press Ctrl-V again to exit multi-line mode.)
------> console.log """
.......   Hello world!
....... """
Hello world!
undefined
------>
```
In the example above, the multi-line prompt after `undefined` demonstrates that the `coffee>` prompt was changed in-place the second time, just like the original behavior.

I didn't add any tests since this is essentially a cosmetic change. I also only did enough testing to make sure it did what I expected on NodeJS versions 6 and 11.

Is there a better place to put UI strings such as the greeting and hint? (I'm used to having them kept in non-code files elsewhere in projects, like `${PROJECT_ROOT}/data/messages`.)


jashkenas:
Just a preference — but I generally prefer no "greeting" on my command-line tools. If you use them regularly, it gets old incredibly fast.

rdeforest:
Reasonable. If someone wants to know more about the tool they can check `coffee -v`, `coffee -h` and `node -v`, and if they want a more verbose startup they can wrap or modify the tool themselves. I've amended the PR accordingly.

edemaine:
The downside to removing the greeting is that it's pretty non-obvious that CTRL-V exists at all. (I learned of it only a few weeks ago, after years of CoffeeScript.)  Without some greeting, this doesn't fix #5114.  Would a shorter greeting be reasonable?

GeoffreyBooth:
So what does this PR do now? Show a hint about Ctrl-V once the user already presses Ctrl-V, to tell them to repeat the key combination to turn off multiline mode? If someone already knows about Ctrl-V to enter multiline mode, wouldn’t they also know how to toggle it back off?

Inve1951:
>So what does this PR do now? Show a hint about Ctrl-V once the user already presses Ctrl-V, to tell them to repeat the key combination to turn off multiline mode?

Yes. First time they enter it each REPL session.

>If someone already knows about Ctrl-V to enter multiline mode, wouldn’t they also know how to toggle it back off?

Certainly shouldn't take more than a few seconds to figgure it out. At least that's how I remember it.

Btw, this addition from line 5 is redundant: `process = require 'process'`

ranska:
Hello this is open since 2019 and all check have passed. Is there any specifics reasons to not merge it ?

GeoffreyBooth:
> Hello this is open since 2019 and all check have passed. Is there any specifics reasons to not merge it ?

There are some usability concerns raised on this thread that I don't think have yet been addressed.

One further one is, how does the regular Node REPL handle this and can we take the same approach?

rdeforest:
> how does the regular Node REPL handle this?

```
$ node
Welcome to Node.js v14.13.1.
Type ".help" for more information.
> .help
.break    Sometimes you get stuck, this gets you out
.clear    Alias for .break
.editor   Enter editor mode
.exit     Exit the REPL
.help     Print this help message
.load     Load JS from a file into the REPL session
.save     Save all evaluated commands in this REPL session to a file

Press Ctrl+C to abort current expression, Ctrl+D to exit the REPL
> .editor
// Entering editor mode (Ctrl+D to finish, Ctrl+C to cancel)

>
```

Even though I'm the author of this PR, I find the comments of @jashkenas and @GeoffreyBooth compelling and am in favor of keeping the REPL quiet. It is more in line with The Unix Way.

GeoffreyBooth:
> Even though I’m the author of this PR, I find the comments of @jashkenas and @GeoffreyBooth compelling and am in favor of keeping the REPL quiet.

So should we close this PR and the accompanying issue? We can always reopen if there’s a proposal for a UX that everyone’s happy with.

rdeforest:
Yes, and #5114.

mrmowgli:
Just a 2 cents here, but wouldn't it be good to include these messages to ease newbie usage?  Having to google something esoteric three seconds after using it seems like a non-starter for a lot of people.  Python is a great example of tons of usage and still including the messages for their repl.

edemaine:
+1 for the original version of this PR (as documented in the original post). 

I use CoffeeScript every day, but I use the REPL pretty rarely, and I still struggle to remember the right key for CTRL-V. So I think it would help veterans too, not just new users.

I don't see the message being annoying; it's pretty short. Maybe folks would be happier if there were an environment variable and/or a command-line option to disable it?

PR #5161, opened 2019-02-17
While AST

helixbass:
@GeoffreyBooth PR for `While` AST (including `loop`/`until`)

### DIFF
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -679,14 +679,14 @@ grammar =
   While: [
     o 'WhileSource Block',                      -> $1.addBody $2
     o 'WhileLineSource Block',                  -> $1.addBody $2
-    o 'Statement  WhileSource',                 -> $2.addBody LOC(1) Block.wrap([$1])
-    o 'Expression WhileSource',                 -> $2.addBody LOC(1) Block.wrap([$1])
+    o 'Statement  WhileSource',                 -> (Object.assign $2, postfix: yes).addBody LOC(1) Block.wrap([$1])
+    o 'Expression WhileSource',                 -> (Object.assign $2, postfix: yes).addBody LOC(1) Block.wrap([$1])
     o 'Loop',                                   -> $1
   ]
 
   Loop: [
-    o 'LOOP Block',                             -> new While(LOC(1) new BooleanLiteral 'true').addBody $2
-    o 'LOOP Expression',                        -> new While(LOC(1) new BooleanLiteral 'true').addBody LOC(2) Block.wrap [$2]
+    o 'LOOP Block',                             -> new While(LOC(1)(new BooleanLiteral 'true'), isLoop: yes).addBody $2
+    o 'LOOP Expression',                        -> new While(LOC(1)(new BooleanLiteral 'true'), isLoop: yes).addBody LOC(2) Block.wrap [$2]
   ]
 
   # Array, object, and range comprehensions, at the most generic level.
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3832,11 +3832,8 @@ exports.Elision = class Elision extends Base
 # it, all other loops can be manufactured. Useful in cases where you need more
 # flexibility or more speed than a comprehension can provide.
 exports.While = class While extends Base
-  constructor: (condition, options) ->
+  constructor: (@condition, {@invert, @guard, @isLoop} = {}) ->
     super()
-
-    @condition = if options?.invert then condition.invert() else condition
-    @guard     = options?.guard
 
   children: ['condition', 'guard', 'body']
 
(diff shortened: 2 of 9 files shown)


src/grammar.coffee
```diff
@@ -679,14 +679,14 @@ grammar =
   While: [
     o 'WhileSource Block',                      -> $1.addBody $2
     o 'WhileLineSource Block',                  -> $1.addBody $2
-    o 'Statement  WhileSource',                 -> $2.addBody LOC(1) Block.wrap([$1])
-    o 'Expression WhileSource',                 -> $2.addBody LOC(1) Block.wrap([$1])
+    o 'Statement  WhileSource',                 -> (Object.assign $2, postfix: yes).addBody LOC(1) Block.wrap([$1])
+    o 'Expression WhileSource',                 -> (Object.assign $2, postfix: yes).addBody LOC(1) Block.wrap([$1])
```
helixbass: Preserve the fact that this was a postfix `while`/`until`

src/grammar.coffee
```diff
@@ -679,14 +679,14 @@ grammar =
...
     o 'Loop',                                   -> $1
   ]
 
   Loop: [
-    o 'LOOP Block',                             -> new While(LOC(1) new BooleanLiteral 'true').addBody $2
-    o 'LOOP Expression',                        -> new While(LOC(1) new BooleanLiteral 'true').addBody LOC(2) Block.wrap [$2]
+    o 'LOOP Block',                             -> new While(LOC(1)(new BooleanLiteral 'true'), isLoop: yes).addBody $2
```
helixbass: Preserve the fact that this used `loop`

src/nodes.coffee
```diff
@@ -3832,12 +3832,9 @@ exports.Elision = class Elision extends Base
 # it, all other loops can be manufactured. Useful in cases where you need more
 # flexibility or more speed than a comprehension can provide.
 exports.While = class While extends Base
-  constructor: (condition, options) ->
+  constructor: (@condition, {@invert, @guard, @isLoop} = {}) ->
     super()
 
-    @condition = if options?.invert then condition.invert() else condition
```
helixbass: Similar to `If`, delay inverting the condition until `compileNode()` gets called so that it's preserved for AST generation purposes

src/nodes.coffee
```diff
@@ -3878,12 +3875,26 @@ exports.While = class While extends Base
...
+  processedCondition: ->
+    @processedConditionCache ?= if @invert then @condition.invert() else @condition
+
+  astType: -> 'WhileStatement'
+
+  astProperties: (o) ->
+    return
+      test: @condition.ast o, LEVEL_PAREN
+      body: @body.ast o, LEVEL_TOP
+      guard: @guard?.ast(o) ? null
```
helixbass: `guard`, `inverted`, `postfix` and `loop` are all extensions to Babel's `WhileStatement` AST

src/rewriter.coffee
```diff
@@ -511,6 +511,9 @@ exports.Rewriter = class Rewriter
     @scanTokens (token, i, tokens) ->
       return 1 if     token[2]
       return 1 unless token.generated or token.explicit
+      if token.fromThen and token[0] is 'INDENT'
```
helixbass: The test case with an "empty `then` block" (`while 1 < 2 then`) called out the need to include this logic - without it, the while body block was being assigned empty location data before the `then`, whereas it should include the `then`

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -3928,13 +3928,13 @@ test "AST as expected for If node", ->
             line: 1
             column: 17
       ]
-      start: 8
+      start: 9
```
helixbass: Making the change to `then` location data handling in the rewriter actually made some of these existing tests including `then` "more correct": rather than the block location data starting a character before the `then` (which I'd pointed out as perhaps slightly unexpected but not problematic), it now starts on the first character of `then`

test/abstract_syntax_tree_location_data.coffee
```diff
@@ -4456,3 +4456,384 @@ test "AST as expected for If node", ->
       end:
         line: 5
         column: 1
+
+test "AST as expected for While node", ->
+  testAstLocationData 'loop 1',
+    type: 'WhileStatement'
+    test:
+      start: 0
```
helixbass: So I'd consider the inclusion of the generated `BooleanLiteral` (`true`) in the AST for `loop` to be worth questioning:

In general, we don't want to include any AST nodes that aren't actually present in the original source since we can't output AST that violates location data nesting/ordering (its location data needs to look "real"/sequential or else both Prettier and ESLint will get upset)

In this case, we can get away with assigning the location data of the `loop` keyword (so here, `range: [0, 4]`) to the generated `BooleanLiteral` and that still results in an AST with location data that makes sense

And the upside of including the generated `true` condition in the AST is that it's more compatible - it both avoids potential incompatibilities due to having a `null` `test` (which would never happen in a real Babel `while` AST) and lets things that might be interested in inspecting the `while`'s condition (eg ESLint code path analysis) know that it's effectively a constant `true` condition

So it's a question of weighing the upsides against the general downsides of including any generated nodes in the AST. In this case, I think it's worth including since it makes the generated AST for `loop` statements pretty much compatible with their corresponding Babel `while` ASTs

test/abstract_syntax_tree.coffee
```diff
@@ -1775,22 +1775,109 @@ test "AST as expected for Elision node", ->
...
+    postfix: no
+    loop: yes
 
-#   testExpression 'while 1 < 2 then',
-#     type: 'While'
-#     condition:
-#       type: 'Op'
-#     body:
-#       type: 'Block'
+  testStatement 'while 1 < 2 then',
```
GeoffreyBooth: Should we add another test where there’s some logic after the `then`? Like `while 1 < 2 then fn()`?

helixbass: Sure, added to AST and location data tests

PR #5167, opened 2019-03-14
Fix #5128: parens around default param

### LINKED ISSUE #5128: Bug: function default parameter wrapped in extraneous parenthesis when part of boolean expression
### Input Code

```coffee
false or (a, b = {}) -> null
```

### Expected Behavior
```coffee
false || function(a, b = {}) {
  return null;
};
```

### Current Behavior
```coffee
false || function(a, (b = {})) {
  return null;
};
```

### Environment

* CoffeeScript version: 2.3.2
* Node.js version: 10.12.0

helixbass:
Fixes #5128 

The problem appeared to me to be that functions params weren't being compiled at an explicit `level`, so in the example from #5128 the outer `LEVEL_OP` was "bleeding through" to the param's compilation level

So made params explicitly compile at `LEVEL_PAREN` (which seems correct)


GeoffreyBooth:
Thanks!

PR #5169, opened 2019-03-15
Dynamic import

### LINKED ISSUE #4834: [Awaiting Stage 4] Grammar: Dynamic imports
```coffeescript
loadStuff = ->
  await Promise.all [ import './assets', import './stuff' ]
```
Yields...
```
[stdin]:2:40: error: unexpected ,
  await Promise.all [ import './assets', import './stuff' ]
                                       ^
```
Not sure how to help with this sort of thing, but I know that dynamic imports are useful.

helixbass:
Fixes #4834 

@GeoffreyBooth I came across dynamic `import()` while working on the Prettier plugin and read #4834 

This PR adds support for dynamic `import()` (requiring explicit call parentheses per the apparent consensus in #4834, which I tend to agree with)

It mimics Babel's error message ("import() requires exactly one argument") when supplying less than or more than one argument

It looks like I should do a follow-on PR against `ast` since Babel appears to use `type: 'Import'` for dynamic import


src/lexer.coffee
```diff
@@ -667,6 +667,9 @@ exports.Lexer = class Lexer
         @error message, origin[2] if message
       return value.length if skipToken
 
+    if prev?[0] is 'IMPORT' and value is '('
+      prev[0] = 'IDENTIFIER'
```
GeoffreyBooth: This is technically cheating; the `import` here is a special keyword like `super`, so it really should be its own token type. Should we make a new token like `DYNAMIC_IMPORT` or something?

helixbass: Ah ok, I was just copying my approach (for `new.target`) from #5106. But I think you're right that (in both cases) it's cleaner to specify new token types/grammar constructs

So updated to use new `DYNAMIC_IMPORT` token and `DynamicImport`/`DynamicImportCall` node classes

This should set it up to be easy to generate the appropriate AST for dynamic imports (which looks like a regular call AST with a "variable"/`callee` of `type: 'Import'`)

As far as `new.target`, I checked Babel AST and it actually should be generating a special AST node for `new.target` (of `type: 'MetaProperty'`). So I think that's another argument that `new.target` should similarly be updated to use a special token type/node class. So for `new.target` I'd plan on updating it in a PR against `ast` unless you'd rather I separate out a PR against `master` for the token type/node class and then a second PR against `ast` for just the `MetaProperty` AST generation

src/grammar.coffee
```diff
@@ -240,6 +240,7 @@ grammar =
     o 'Super'
     o 'This'
     o 'SUPER Arguments',               -> new SuperCall LOC(1)(new Super), $2, no, $1
+    o 'DYNAMIC_IMPORT Arguments',      -> new DynamicImportCall LOC(1)(new DynamicImport), $2
```
helixbass: Technically it looks like dynamic imports are allowed to be object spread (eg `{...import('module')}` compiles in Babel)

Added a corresponding test

src/grammar.coffee
```diff
@@ -891,7 +893,7 @@ operators = [
   ['right',     'YIELD']
   ['right',     '=', ':', 'COMPOUND_ASSIGN', 'RETURN', 'THROW', 'EXTENDS']
   ['right',     'FORIN', 'FOROF', 'FORFROM', 'BY', 'WHEN']
-  ['right',     'IF', 'ELSE', 'FOR', 'WHILE', 'UNTIL', 'LOOP', 'SUPER', 'CLASS', 'IMPORT', 'EXPORT']
+  ['right',     'IF', 'ELSE', 'FOR', 'WHILE', 'UNTIL', 'LOOP', 'SUPER', 'CLASS', 'IMPORT', 'EXPORT', 'DYNAMIC_IMPORT']
```
helixbass: I don't know if this is technically necessary but it looked like `DYNAMIC_IMPORT` belonged at this precedence level (with `SUPER`, `IMPORT` and `EXPORT`)

src/grammar.coffee
GeoffreyBooth: It evaluates in browsers too:

image

Though it makes no sense, as the `import()` returns a Promise. Not sure how that’s supposed to interact with object spread, and apparently it doesn’t 😄but still the code compiles even if it’s useless.

src/lexer.coffee
```diff
@@ -667,6 +667,9 @@ exports.Lexer = class Lexer
         @error message, origin[2] if message
       return value.length if skipToken
 
+    if prev?[0] is 'IMPORT' and value is '('
```
GeoffreyBooth:
```suggestion
    if value is '(' and prev?[0] is 'IMPORT'
```

Like the next `if` statement, I think this would be faster if we put the `value` check first (as it’ll return false a lot more often than `prev?` will).

helixbass: @GeoffreyBooth sure

"Committed suggestion" - this is the first time I've seen that Github feature, neat! The only thing that feels weird about it (in general) is committing without checking locally that tests are passing - but I guess with this workflow you just watch CI afterwards?

helixbass: Actually, realized (in pulling the updated branch) that since we commit `lib/`, the "Commit suggestion" workflow also isn't complete since I needed to recompile locally and push an updated `lib/`

But it's still a cool way of specify suggestions/changes 👍

test/error_messages.coffee
```diff
@@ -1925,3 +1925,29 @@ test "`new.target` is only allowed meta property", ->
...
+    [stdin]:1:1: error: import() requires exactly one argument
+    import('x', {})
+    ^^^^^^^^^^^^^^^
+  '''
+
+test "#4834: dynamic import requires explicit call parentheses", ->
+  assertErrorFormat '''
+    promise = import 'foo'
+  ''', '''
+    [stdin]:1:23: error: unexpected end of input
```
GeoffreyBooth: This isn’t a deal-breaker, but I suppose we can’t get a better error message here? Probably not, I’m assuming, because this is a grammar fail?

helixbass: Ya I think we're at the mercy of the grammar there

vendethiel: Maybe a fake import rule which throws an error?
We don’t do that too often, but...

GeoffreyBooth: @vendethiel can you point to an example? I think it might be worth it in this case because the required parentheses will likely be surprising.

helixbass: @GeoffreyBooth @vendethiel I agree it'd be nice. But I don't see an obvious way to implement it

Part of why I went with required parentheses is that @jashkenas' [suggestion](https://github.com/jashkenas/coffeescript/issues/4834#issuecomment-465714019) seemed tricky to implement (though I additionally also agree with the apparent consensus that required parentheses, though perhaps unprecedented, are actually a nice disambiguator readability-wise)

Recall that `import 'a'` is valid static import syntax. You couldn't add implicit call parens since at rewriter time you have no idea if you're in statement or "nested expression" context, so the token sequence (visible to the grammar) for dynamic `import 'a'` and static `import 'a'` would be identical

So I don't think you could have a separate non-conflicting grammar rule for the *expression* `import 'a'` (since `Expression`s and `Statement`s are both `Line`s)

So I think that implies that you'd have to parse dynamic `import 'a'` and static `import 'a'` with the same grammar rule (ie instantiate the same node class, which would have to be a hack that represented both dynamic imports and static imports)...and then only at compile time would you know reliably (based on `o.level`) if you were a "nested expression" and therefore a dynamic import (whether for purposes of allowing implicit-call-paren dynamic imports or just throwing a nicer error message)?

If I'm correct, I think that's ugly enough that it wouldn't be worth doing period, let alone just for the sake of a nicer error message

And I can't think of any other "heuristic" way of catching many common cases - I guess in theory you could add a rewriter rule to look for eg the token sequence `AWAIT` `IMPORT` [something other than `(`] and throw a custom error message then? But even that doesn't necessarily seem worth it imo?

GeoffreyBooth: If there's no easy way to do it, it's not worth doing. I was thinking a new grammar rule that just threw an error might do it, then that's a one line addition. The error isn't with new node classes and so on.

vendethiel: > ...and then only at compile time would you know reliably (based on o.level)

`ModuleDeclaration`'s `checkScope` already has error checking depending on `o.indent.length` (which also relies on `--bare` implied for modules, but that's fine). 

> If I'm correct, I think that's ugly enough that it wouldn't be worth doing period, let alone just for the sake of a nicer error message

At the same time, the current impl. impacts lexer, grammar and nodes. -- but yes, it's better than moving this down somewhere else.

> So I don't think you could have a separate non-conflicting grammar rule for the _expression_ `import 'a'` (since `Expression`s and `Statement`s are both `Line`s)

I was thinking to check for a `TERMINATOR` before the `(`-less `IMPORT`, but this breaks cases such as:

```coffee
f(1; import("a"))
# or the long form:
f 1,
  import("a")
```

GeoffreyBooth: @vendethiel So do you think there’s a not-too-invasive way to add the error?

We can also just merge this and add a better error as a follow-on PR.

helixbass: I'm not totally clear on what case @vendethiel is suggesting we could catch - again, since "`(`-less `IMPORT`" is indistinguishable from the token sequence of a static import as I understand it, you couldn't just check for a preceding `TERMINATOR` since that would be a valid token sequence for static imports

vendethiel: > you couldn't just check for a preceding `TERMINATOR` since that would be a valid token sequence for static imports

If it was only a valid token for static imports, it'd be easy-ish to generate a correct error message in some cases, but that's not the case.
The one feature usable is jison' start conditions, but we don't do that at all, and it's probably better not to add context for such a thing.

GeoffreyBooth: Okay so @vendethiel @helixbass I’m assuming you both agree that we should merge this as is?

vendethiel: Yes. I see no simple way to make the error message more palatable right now.

helixbass: 👍

GeoffreyBooth:
Okay, so before we cut a new release we need to update the docs not just to describe `import()` but to also describe the new policy.

If we’re going to add provisional support for Stage 3 things, class fields could be next.

jtenner:
@GeoffreyBooth This is a pull request years in the making. Thank you!

voxsoftware:
when will be this realeased?

helixbass:
@GeoffreyBooth do you mind merging latest `master` into `ast`? Then I can open a PR for dynamic import AST

GeoffreyBooth:
@helixbass Done.

PR #5170, opened 2019-03-16
MetaProperty AST (new.target)

helixbass:
@GeoffreyBooth as mentioned in https://github.com/jashkenas/coffeescript/pull/5169#discussion_r266192600, `new.target` should generate a `MetaProperty` AST node


src/nodes.coffee
```diff
@@ -1377,6 +1368,34 @@ exports.Value = class Value extends Base
...
     )
 
+exports.MetaProperty = class MetaProperty extends Base
+  constructor: (@meta, @property) ->
+    super()
+
+  children: ['meta', 'property']
+
+  checkValid: (o) ->
+    if @meta.value is 'new'
```
helixbass: When reading about `MetaProperty`, I came across [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta). So makes sense to make `MetaProperty` a little bit "generic" so that we can plan on adding other meta-properties

test/error_messages.coffee
```diff
@@ -1929,10 +1929,20 @@ test "`new.target` outside of a function", ->
...
-  assertErrorFormat '''
+  assertErrorFormatAst '''
     -> new.something
   ''', '''
     [stdin]:1:4: error: the only valid meta property for new is new.target
     -> new.something
        ^^^^^^^^^^^^^
   '''
+
+test "`new.target` cannot be assigned", ->
```
helixbass: Before, it wasn't throwing a compilation error when eg assigning to `new.target` (since it was just being treated like a normal object-property access)

So this is a benefit of using a separate grammar construct for `MetaProperty`, since `new.target` shouldn't be able to be used as an "assignable"

GeoffreyBooth:
Looks good. We should add support for `import.meta` to the main branch too, after `import()`.

PR #5171, opened 2019-03-16
If AST refinement: conditional else expression

helixbass:
@GeoffreyBooth in merging the latest `ast` into my branch and updating to use that as the dependency for the Prettier plugin, I ran into failures based on what I'd consider to be an incorrect "asymmetry" in the AST being generated for `If`:

An `If` with simple consequent and `else` clauses that generates a `ConditionalExpression` (ie JS ternary) AST was treating the consequent as a simple expression but the `else` as a `BlockStatement`

Eg for this
```
f(if a then b else c)
```
the generated AST looked like
```
...
type: 'ConditionalExpression'
consequent:
  type: 'Identifier'
  name: 'b'
alternate:
  type: 'BlockStatement'
  body: [
    type: 'ExpressionStatement'
    expression:
      type: 'Identifier'
      name: 'c'
  ]
```
But it'd be preferable/more symmetrical for it to generate
```
...
type: 'ConditionalExpression'
consequent:
  type: 'Identifier'
  name: 'b'
alternate:
  type: 'Identifier'
  name: 'c'
```
So this PR updates the `If` AST generation logic accordingly


GeoffreyBooth:
@helixbass Updating my [previous comment](https://github.com/jashkenas/coffeescript/pull/5154#issuecomment-460088751), looking at the tests these are the remaining nodes that need to be done to finish the AST work:

- StringLiteral
- RegexLiteral
- PassthroughLiteral
- PropertyName
- Value
- SuperCall
- Super
- RegexWithInterpolations
- TaggedTemplateCall
- Extends
- Class
- ExecutableClassBody
- Param
- Parens
- StringWithInterpolations
- For

Much shorter than last time, but still a lot to go. How many of these are already done on your Prettier branch? You’ve pretty much got the pattern down so that I usually don’t have any notes, so if they’re already done on another branch we should just move them over 😄

helixbass:
@GeoffreyBooth per that list of remaining types, we've effectively already covered `Parens`, `Param`, and `Value`/`PropertyName`. So after `For` (which I've started preparing a PR for) there's really only two big remaining "groups":
- interpolated strings/regexen
- classes

There's full coverage on my Prettier branch. But let's keep putting them up for individual review, it's a valuable exercise for me as far as taking another look at my first-pass implementations (from the perspective of some of the patterns we've landed on) and extracting out just the stuff that's relevant to AST generation (as opposed to some of the other stuff on that branch)

We should be able to move through them pretty quickly, I've just been in the busiest period of a project the last couple weeks. And I'll try and keep leaving comments pointing out some of the unique/debatable aspects of the implementation to minimize the ask on your time as a reviewer

GeoffreyBooth:
Sounds good and thanks. I really appreciate the explanatory comments, that makes reviewing these much easier. I'm eager to get to the finish line, we're so close!

PR #5173, opened 2019-03-17
For AST

helixbass:
@GeoffreyBooth PR for `For` AST

### DIFF
--- a/src/grammar.coffee
+++ b/src/grammar.coffee
@@ -699,8 +699,8 @@ grammar =
   # Comprehensions can either be normal, with a block of expressions to execute,
   # or postfix, with a single expression.
   For: [
-    o 'Statement    ForBody',  -> $2.addBody $1
-    o 'Expression   ForBody',  -> $2.addBody $1
+    o 'Statement    ForBody',  -> $2.postfix = yes; $2.addBody $1
+    o 'Expression   ForBody',  -> $2.postfix = yes; $2.addBody $1
     o 'ForBody      Block',    -> $1.addBody $2
     o 'ForLineBody  Block',    -> $1.addBody $2
   ]
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -4537,6 +4537,9 @@ exports.For = class For extends While
 
   addBody: (body) ->
     @body = Block.wrap [body]
+    {expressions} = @body
+    if expressions.length
+      @body.locationData ?= mergeLocationData expressions[0].locationData, expressions[expressions.length - 1].locationData
     this
 
   addSource: (source) ->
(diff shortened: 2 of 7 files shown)


src/grammar.coffee
```diff
@@ -699,8 +699,8 @@ grammar =
   # Comprehensions can either be normal, with a block of expressions to execute,
   # or postfix, with a single expression.
   For: [
-    o 'Statement    ForBody',  -> $2.addBody $1
-    o 'Expression   ForBody',  -> $2.addBody $1
+    o 'Statement    ForBody',  -> $2.postfix = yes; $2.addBody $1
+    o 'Expression   ForBody',  -> $2.postfix = yes; $2.addBody $1
```
helixbass: Preserve whether this was a postfix `for`

src/nodes.coffee
```diff
@@ -4537,6 +4537,9 @@ exports.For = class For extends While
 
   addBody: (body) ->
     @body = Block.wrap [body]
+    {expressions} = @body
+    if expressions.length
+      @body.locationData ?= mergeLocationData expressions[0].locationData, expressions[expressions.length - 1].locationData
```
helixbass: Ideally this would live inside `Block.wrap()` so that it never returned a `Block` with no location data. But that will have to wait as there are currently other node class objects getting passed to it that don't have location data to use in deriving the new `Block`'s location data

src/nodes.coffee
```diff
@@ -4551,6 +4554,7 @@ exports.For = class For extends While
     @awaitTag.error 'await must be used with for-from' if @await and not @from
     @range   = @source instanceof Value and @source.base instanceof Range and not @source.properties.length and not @from
     @pattern = @name instanceof Value
+    @name.unwrap().propagateLhs?(yes) if @pattern
```
helixbass: When there's a for variable "pattern" eg `for {a} in b`, it needs to know that it's a "left-hand-side" so that it can eg generate `ObjectPattern` AST nodes

src/nodes.coffee
```diff
@@ -4664,6 +4668,25 @@ exports.For = class For extends While
     fragments.push @makeCode(returnResult) if returnResult
     fragments
 
+  astType: -> 'For'
```
helixbass: I chose to make `for` represented as a new (non-Babel) AST type `For`

For context, the possible corresponding Babel AST types would be `ForStatement`, `ForInStatement` and `ForOfStatement`

Our `for` constructs don't correspond to the parts expected by `ForStatement` (eg `for (i = 0; i < n; i++) {...}`), so that's not an option

In theory we could try and use `ForInStatement` for `for ... of` and `ForOfStatement` for `for ... from` since those correlate. But:
- we'd still have other `for` constructs (`for ... in` and `for <range>`) that we'd need to make up new types for
- it's confusing that the names don't match up, ie `for ... of` isn't `ForOfStatement`, it's `ForInStatement`

All of the different `for` constructs have similar "extension" children/properties (eg guard, step, postfix) so to me it made sense to unify them as a single AST node type and then indicate which construct (ie `for ... in`, `for ... of`, `for ... from`, or `for <range>`) it is via an AST property (specifically `style`, which can be `in`, `of`, `from` or `range`)

So that's why I went with a unique `For` AST node type that encompasses all of them

GeoffreyBooth: Well it sounds like Babel's version is a lower level of ours, like what ours transpiles into. Not sure we have much of a choice if we want to preserve the original content.

PR #5175, opened 2019-03-18
String interpolation AST

helixbass:
@GeoffreyBooth PR for basic interpolated strings (`"""` heredoc or `"`-quoted)

There is a fair amount of location-data related fixing here (necessary for ending up with correct AST location data on the whole interpolated string and its sub-pieces)

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -115,7 +115,7 @@ buildLocationData = (first, last) ->
     ]
 
 buildLocationHash = (loc) ->
-  "#{loc.first_line}x#{loc.first_column}-#{loc.last_line}x#{loc.last_column}"
+  "#{loc.range[0]}-#{loc.range[1]}"
 
 # Build a dictionary of extra token properties organized by tokens’ locations
 # used as lookup hashes.
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -114,6 +114,10 @@ buildLocationData = (first, last) ->
       last.range[1]
     ]
 
+# Get a lookup hash for a token based on its location data.
+# Multiple tokens might have the same location hash, but using exclusive
+# location data distinguishes e.g. zero-length generated tokens from
+# actual source tokens.
 buildLocationHash = (loc) ->
   "#{loc.range[0]}-#{loc.range[1]}"
 
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -820,6 +820,11 @@ exports.Lexer = class Lexer
         [open, ..., close] = nested
         open[0]  = 'INTERPOLATION_START'
         open[1]  = '('
+        open[2].first_column -= interpolationOffset
+        open[2].range = [
+          open[2].range[0] - interpolationOffset
+          open[2].range[1]
+        ]
         close[0]  = 'INTERPOLATION_END'
         close[1] = ')'
         close.origin = ['', 'end of interpolation', close[2]]
@@ -972,7 +977,7 @@ exports.Lexer = class Lexer
     [locationData.last_line, locationData.last_column, endOffset] =
       @getLineAndColumnFromChunk offsetInChunk + lastCharacter
     [locationData.last_line_exclusive, locationData.last_column_exclusive] =
-      @getLineAndColumnFromChunk offsetInChunk + lastCharacter + 1
+      @getLineAndColumnFromChunk offsetInChunk + lastCharacter + (if length > 0 then 1 else 0)
     locationData.range[1] = if length > 0 then endOffset + 1 else endOffset
 
     locationData
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -301,7 +301,7 @@ exports.Lexer = class Lexer
         indent = attempt if indent is null or 0 < attempt.length < indent.length
 
     delimiter = quote.charAt(0)
-    @mergeInterpolationTokens tokens, {quote, indent}, (value) =>
+    @mergeInterpolationTokens tokens, {quote, indent, endOffset: end}, (value) =>
       @validateUnicodeCodePointEscapes value, delimiter: quote
 
     if @atCSXTag()
@@ -355,8 +355,7 @@ exports.Lexer = class Lexer
       # this comment to; and follow with a newline.
       commentAttachments[0].newLine = yes
       @lineToken @chunk[comment.length..] # Set the indent.
-      placeholderToken = @makeToken 'JS', ''
-      placeholderToken.generated = yes
+      placeholderToken = @makeToken 'JS', '', generated: yes
       placeholderToken.comments = commentAttachments
       @tokens.push placeholderToken
       @newlineToken 0
@@ -417,7 +416,7 @@ exports.Lexer = class Lexer
         @token 'REGEX_START', '(',    {length: 0, origin}
         @token 'IDENTIFIER', 'RegExp', length: 0
         @token 'CALL_START', '(',      length: 0
-        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}}, (str) =>
+        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end}, (str) =>
           @validateUnicodeCodePointEscapes str, {delimiter}
         if flags
           @token ',', ',',                    offset: index - 1, length: 0
@@ -613,7 +612,7 @@ exports.Lexer = class Lexer
         @token ',', 'JSX_COMMA', generated: yes
         {tokens, index: end} =
           @matchWithInterpolations INSIDE_CSX, '>', '</', CSX_INTERPOLATION
-        @mergeInterpolationTokens tokens, {}, (value) =>
+        @mergeInterpolationTokens tokens, {endOffset: end}, (value) =>
           @validateUnicodeCodePointEscapes value, delimiter: '>'
         match = CSX_IDENTIFIER.exec(@chunk[end...]) or CSX_FRAGMENT_IDENTIFIER.exec(@chunk[end...])
         if not match or match[1] isnt "#{csxTag.name}#{(".#{property}" for property in csxTag.properties).join ''}"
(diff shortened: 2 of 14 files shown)


src/nodes.coffee
```diff
@@ -4518,6 +4526,36 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
...
+    {expressions, quasis, @quote}
+
+exports.TemplateElement = class TemplateElement extends Base
+  constructor: (@value, {@tail} = {}) ->
+    super()
+
+  astProperties: ->
+    return
+      value:
+        raw: @value
```
helixbass: There is also a `value.cooked` property in Babel-style `TemplateElement` AST, but I'm not sure the difference between `raw` and `cooked` so omitted at least for now. I'm guessing it's related to eg escapes and that's something that deserves its own investigation/PR

src/lexer.coffee
```diff
@@ -982,16 +977,17 @@ exports.Lexer = class Lexer
...
-      @getLineAndColumnFromChunk offsetInChunk + lastCharacter + 1
+      @getLineAndColumnFromChunk offsetInChunk + lastCharacter + (if length > 0 then 1 else 0)
     locationData.range[1] = if length > 0 then endOffset + 1 else endOffset
 
     locationData
 
   # Same as `token`, except this just returns the token without adding it
   # to the results.
-  makeToken: (tag, value, {offset: offsetInChunk = 0, length = value.length, origin} = {}) ->
+  makeToken: (tag, value, {offset: offsetInChunk = 0, length = value.length, origin, generated} = {}) ->
```
helixbass: This was actually "missing" in the sense that I've added some `@token()` or `@makeToken()` calls with `generated: yes` and it wasn't doing anything since this wasn't wired up to accept that option

src/rewriter.coffee
```diff
@@ -539,7 +539,7 @@ exports.Rewriter = class Rewriter
   fixOutdentLocationData: ->
     @scanTokens (token, i, tokens) ->
       return 1 unless token[0] is 'OUTDENT' or
-        (token.generated and token[0] is 'CALL_END') or
+        (token.generated and token[0] is 'CALL_END' and not token.data?.closingTagNameToken) or
```
helixbass: After correctly wiring up the `generated` option in `Lexer::makeToken()`, location data for JSX elements was breaking since the rewriter wanted to reassign "real" location data to the generated JSX pseudo-tokens

So need to tell it to ignore this particular type of JSX pseudo-token and respect its (pseudo) location data

Basically this is patching a seam related to the fact that we don't really parse JSX and instead pass through tokens that don't actually exist

src/lexer.coffee
```diff
@@ -821,6 +820,11 @@ exports.Lexer = class Lexer
         [open, ..., close] = nested
         open[0]  = 'INTERPOLATION_START'
         open[1]  = '('
+        open[2].first_column -= interpolationOffset
```
helixbass: This adjusts the location data of the opening `#{` to include the `#`

src/lexer.coffee
```diff
@@ -845,31 +849,17 @@ exports.Lexer = class Lexer
     unless str[...closingDelimiter.length] is closingDelimiter
       @error "missing #{closingDelimiter}", length: delimiter.length
 
-    [firstToken, ..., lastToken] = tokens
```
helixbass: This was adjusting the location data of the first and last tokens of the interpolation contents to include the surrounding delimiters (eg the interpolated string `"` quotes)

That's not actually what we want AST-location-data wise

This change is what necessitates the updates to token location data tests in `tests/location.coffee`

src/lexer.coffee
```diff
@@ -900,6 +890,19 @@ exports.Lexer = class Lexer
           addTokenData token, {heregex} if heregex
           token[0] = 'STRING'
           token[1] = '"' + converted + '"'
+          if tokens.length is 1 and quote?
```
helixbass: So here we're sort of "re-adding" the code that was removed above - only in the case that this is a non-interpolated string do we want the token's location data to include the surrounding quotes/delimiters (since at that point the token represents an entire string literal rather than just a sub-piece of an interpolated string)

src/lexer.coffee
```diff
@@ -919,15 +922,7 @@ exports.Lexer = class Lexer
...
-      rparen = @token 'STRING_END', ')'
-      rparen[2] =
-        first_line:            lastToken[2].last_line
-        first_column:          lastToken[2].last_column
-        last_line:             lastToken[2].last_line
-        last_column:           lastToken[2].last_column
-        last_line_exclusive:   lastToken[2].last_line_exclusive
-        last_column_exclusive: lastToken[2].last_column_exclusive
-        range:                 lastToken[2].range
+      rparen = @token 'STRING_END', ')', offset: endOffset - (quote ? '').length, length: quote?.length ? 0
```
helixbass: By passing `endOffset` into all calls to `mergeInterpolationTokens()`, we're able to accurately encode the location data of the closing quote/delimiter

src/helpers.coffee
```diff
@@ -115,7 +115,7 @@ buildLocationData = (first, last) ->
     ]
 
 buildLocationHash = (loc) ->
-  "#{loc.first_line}x#{loc.first_column}-#{loc.last_line}x#{loc.last_column}"
+  "#{loc.range[0]}-#{loc.range[1]}"
```
helixbass: With the fixes to interpolation location data in `lexer.coffee`, a JSX comment test was breaking
 
Basically by using *exclusive* location data (eg `range`) for generating the location hashes used for tracking comments, the test passed since it distinguished two tokens' location data

And conceptually it makes sense to use exclusive location data here since it's a slightly stronger identifier of different location data (for tokens in this case)

src/nodes.coffee
```diff
@@ -4483,6 +4478,18 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
         delete node.comments
       return yes
 
+    elements
+
+  compileNode: (o) ->
+    @comments ?= @startQuote?.comments
```
helixbass: After fixing the one test by using exclusive location data when mapping comments (see previous comment in `helpers.coffee`), a couple other comment-related tests were breaking

They were able to pass by preserving comments that got attached to the "start quote" (`STRING_START`) of interpolated strings

So that's why `startQuote` gets passed in from the grammar now when constructing `StringWithInterpolations`, so that its comments can be preserved and output here

src/helpers.coffee
GeoffreyBooth: Can you add some version of this to a comment in the code? It would be good to remember this when working with comments in the future.

src/lexer.coffee
GeoffreyBooth: Does this match what Babel would produce for source maps for strings?

src/lexer.coffee
GeoffreyBooth: Why not do the same for comments?

src/lexer.coffee
helixbass: I don't know enough about source maps to fully answer this (though I'd be curious to know more if you want to explain). I'd dug as deep as to see that yes this is more similar to the tokens generated by the Babel parser:

For a simple string literal eg `"a"`, it represents that as a single token of type `string`

For an interpolated string (template literal) eg `` `a${b}` ``, its token stream looks like `` ` ``, `a`, `${`, `b`, `}`, [empty text], `` ` ``, where `a` and the empty text are a `template` token type

src/lexer.coffee
helixbass: I don't think I understand, are you saying allow comments to be passed in as an option to the `token()`/`makeToken()` "constructor"?

src/lexer.coffee
GeoffreyBooth: Sorry I just meant does it match in terms of the line/column numbers. Like is Babel inclusive or exclusive? Since you’re changing CoffeeScript’s source maps for strings from inclusive to exclusive, I’m wondering which way Babel is.

src/lexer.coffee
GeoffreyBooth: Yes, currently you need to make the token and then add the comments afterward as a property. That used to be the case for `generated` but it isn’t now because of this change, I was asking if it makes sense to make the same change for comments.

src/lexer.coffee
helixbass: I believe the source mapping (based on location data) should stay consistent - basically the only thing that should be changing is the location data for a token that represents a "text chunk" of an interpolated string. The location data for a string literal (both its token and its resulting `StringLiteral` node class instance) and for an interpolated string (`StringWithInterpolations` instance) should be the same as they were previously

> Since you’re changing CoffeeScript’s source maps for strings from inclusive to exclusive

So if I'm understanding correctly, this change in behavior isn't about inclusive/exclusive, it's that location data for leading/trailing "text chunk" tokens in an interpolated string will no longer include the preceding/following quote in its location data (which in the existing behavior overlapped their location data with the adjacent `STRING_START`/`STRING_END` tokens, which are what actually represent the start/end quotes of the interpolated string)

Eg in `"a#{b}c"`, the location data range for the `a` token will now be `[1, 2]` instead of `[0, 2]` and for the `c` token will be `[6, 7]` instead of `[6, 8]`

So I think this won't have consequences for source maps. But if you think it deserves deeper investigation we could compare sourcemaps before and after this PR, I'm just not familiar with how to do eg a visual diff of the mappings created by sourcemap files

src/lexer.coffee
helixbass: That would make sense to me as a pattern

Looking at the implementation of `commentToken()`, it looks like typically comments get attached to an existing previous token so wouldn't be available yet as an argument at the time of construction (of that previous token)

src/lexer.coffee
GeoffreyBooth: I’m just a) nervous that this might be a breaking change to the non-AST side of things, and b) wanting to make sure that if our `a` in your example is `[1, 2]`, than the Babel source map for `` `a${b}c` `` would have its `a` also as `[1, 2]`. Like if we’re going to change at all, we should align with Babel here.

helixbass: The mapping of `a` to `[1, 2]` is definitely more correct (and matches Babel's token location data) as that's the actual location of that token

As far as sourcemaps, I did a visual diff of the sourcemap for `"a#{b}c"` pre- and post-this PR using [this tool](https://sokra.github.io/source-map-visualization/)

The only difference in the mapping is that `a` (in the output `` `a${b}c`;``) now gets mapped to `a` in the source (ie line 1, column 1) whereas previously it was getting mapped to the leading `"` in the source (line 1, column 0)

So it's a slightly more accurate sourcemap that's being generated

PR #5181, opened 2019-03-28
Heregex AST

helixbass:
@GeoffreyBooth PR for basic heregex AST

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -416,13 +416,13 @@ exports.Lexer = class Lexer
         @token 'REGEX_START', '(',    {length: 0, origin}
         @token 'IDENTIFIER', 'RegExp', length: 0
         @token 'CALL_START', '(',      length: 0
-        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end}, (str) =>
+        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end - flags.length, quote: '///'}, (str) =>
           @validateUnicodeCodePointEscapes str, {delimiter}
         if flags
           @token ',', ',',                    offset: index - 1, length: 0
           @token 'STRING', '"' + flags + '"', offset: index - 1, length: flags.length
-        @token ')', ')',                      offset: end - 1,   length: 0
-        @token 'REGEX_END', ')',              offset: end - 1,   length: 0
+        @token ')', ')',                      offset: end,       length: 0
+        @token 'REGEX_END', ')',              offset: end,       length: 0
 
     end
 
@@ -889,7 +889,11 @@ exports.Lexer = class Lexer
           converted = fn.call this, token[1], i
           addTokenData token, initialChunk: yes if i is 0
           addTokenData token, finalChunk: yes   if i is $
-          addTokenData token, {indent, quote, double}
+          addTokenData token, {
+            indent
+            quote: quote unless quote is '///'
+            double
+          }
           addTokenData token, {heregex} if heregex
           token[0] = 'STRING'
           token[1] = '"' + converted + '"'
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -889,11 +889,7 @@ exports.Lexer = class Lexer
           converted = fn.call this, token[1], i
           addTokenData token, initialChunk: yes if i is 0
           addTokenData token, finalChunk: yes   if i is $
-          addTokenData token, {
-            indent
-            quote: quote unless quote is '///'
-            double
-          }
+          addTokenData token, {indent, quote, double}
           addTokenData token, {heregex} if heregex
           token[0] = 'STRING'
           token[1] = '"' + converted + '"'
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1872,8 +1872,16 @@ exports.RegexWithInterpolations = class RegexWithInterpolations extends Base
   constructor: (@call) ->
     super()
 
+  children: ['call']
+
   compileNode: (o) ->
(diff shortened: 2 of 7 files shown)


src/lexer.coffee
```diff
@@ -416,13 +416,13 @@ exports.Lexer = class Lexer
         @token 'REGEX_START', '(',    {length: 0, origin}
         @token 'IDENTIFIER', 'RegExp', length: 0
         @token 'CALL_START', '(',      length: 0
-        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end}, (str) =>
+        @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end - flags.length, quote: '///'}, (str) =>
```
helixbass: Allow the `TemplateLiteral` (ie `StringWithInterpolations`) generated in the AST for a heregex to have `quote: '///'`

The `endOffset` change excludes flags location data from the location data for the `TemplateLiteral`, which makes sense (the `TemplateLiteral` corresponds to `///...///`, not the trailing flags)

src/lexer.coffee
```diff
@@ -416,13 +416,13 @@ exports.Lexer = class Lexer
...
           @validateUnicodeCodePointEscapes str, {delimiter}
         if flags
           @token ',', ',',                    offset: index - 1, length: 0
           @token 'STRING', '"' + flags + '"', offset: index - 1, length: flags.length
-        @token ')', ')',                      offset: end - 1,   length: 0
-        @token 'REGEX_END', ')',              offset: end - 1,   length: 0
+        @token ')', ')',                      offset: end,       length: 0
```
helixbass: Without this change the location data for heregexes was one character too short

This change makes sense, the generated tokens denoting the end of the heregex should be zero-length tokens between last character of the heregex and the following character

src/lexer.coffee
```diff
@@ -889,7 +889,11 @@ exports.Lexer = class Lexer
           converted = fn.call this, token[1], i
           addTokenData token, initialChunk: yes if i is 0
           addTokenData token, finalChunk: yes   if i is $
-          addTokenData token, {indent, quote, double}
+          addTokenData token, {
+            indent
+            quote: quote unless quote is '///'
```
helixbass: After updating the containing `StringWithInterpolations` of a heregex to have `quote: '///'` (see above comment), the string-chunks (`StringLiteral`s) inside of it were deciding that their quote was `/` (since they do something like `@quote = quote.charAt 0`)

This was causing test breakage and doesn't make sense, so don't "pass through" `quote` when it's `///`

src/nodes.coffee
```diff
@@ -1872,9 +1872,17 @@ exports.RegexWithInterpolations = class RegexWithInterpolations extends Base
   constructor: (@call) ->
     super()
 
+  children: ['call']
```
helixbass: This seemed to be missing

src/nodes.coffee
```diff
@@ -1872,9 +1872,17 @@ exports.RegexWithInterpolations = class RegexWithInterpolations extends Base
...
+
   compileNode: (o) ->
     @call.compileNode o
 
+  astType: -> 'InterpolatedRegExpLiteral'
```
helixbass: This is a non-Babel AST type. Its name vaguely mimics Babel's `RegExpLiteral` type for normal regexes as well as its properties (`flags`, `interpolatedPattern` instead of `pattern`)

test/location.coffee
```diff
@@ -595,18 +595,21 @@ test "Verify heregexes with interpolations have the right ending position", ->
...
+  assertColumn = (token, column, width = 0) ->
     eq token[2].first_line, 0
     eq token[2].first_column, column
     eq token[2].last_line, 0
     eq token[2].last_column, column
+    eq token[2].last_column_exclusive, column + width
 
   arrayEndColumn = arrayEnd[2].first_column
-  for token in [comma, flagsString]
+  for token in [comma]
```
helixbass: This test needed its location data updated to match the refined location data for generated heregex tokens

src/lexer.coffee
GeoffreyBooth: Wouldn’t `quote` be inconsistent then? Why not update `@quote =` to be something like
```coffee
@quote = if quote.startsWith('///') then `'///'` else quote.charAt 0
```

helixbass: Ok to clarify, there's two different places where `quote` is being applied here:

On the "outer" `StringWithInterpolations` (corresponding to the whole interpolated heregex), it's now getting passed `quote: '///'` (whereas before it didn't get passed `quote` at all)

Then on the inner `StringLiteral`s representing "text chunks" inside the interpolated heregex, `quote` really doesn't apply at all (this is the same idea that I was discussing in the previous `StringWithInterpolations` PR, that actual string literals with quotes are sort of different from text-chunks inside of an interpolated string). With the "outer" change to `quote: '///'`, this "inner" was also updating to `quote: '///'` since here the existing logic was for them both to use the same `quote`

So I think this is the cleanest place to specify that "when we're dealing with text chunks inside an interpolated heregex, make sure that they don't decide that their quote is `'///'`/`'/'`" (and then in `StringLiteral::constructor()`, `quote` will default to `'"'`)

GeoffreyBooth: Right but isn’t it weird that the other similar tokens have a `quote` property and this one doesn’t? Shouldn’t they be consistent, and we handle all the various quote possibilities (both 1-character and 3-character) on the other side?

helixbass: Sure, removed the special case here and shifted into `nodes.coffee`

PR #5183, opened 2019-03-29
Regex AST

helixbass:
@GeoffreyBooth PR for simple regex/non-interpolated heregex

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -966,6 +966,22 @@ exports.RegexLiteral = class RegexLiteral extends Literal
     val = replaceUnicodeCodePointEscapes val, {@flags}
     @value = "#{makeDelimitedLiteral val, delimiter: '/'}#{@flags}"
 
+  REGEX_REGEX: /// ^ / (.*) / \w* $ ///
+
+  astType: -> 'RegExpLiteral'
+
+  astProperties: ->
+    [, pattern] = @REGEX_REGEX.exec @value
+    return {
+      value: undefined
+      pattern, @flags, @delimiter
+      originalPattern: @originalValue
+      extra:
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -966,6 +966,22 @@ exports.RegexLiteral = class RegexLiteral extends Literal
     val = replaceUnicodeCodePointEscapes val, {@flags}
     @value = "#{makeDelimitedLiteral val, delimiter: '/'}#{@flags}"
 
+  REGEX_REGEX: /// ^ / (.*) / \w* $ ///
+
+  astType: -> 'RegExpLiteral'
+
+  astProperties: ->
```
helixbass: Here we're going for semantic compatibility with Babel (the AST generated for a regex/non-interpolated heregex should match the AST Babel would generate for the corresponding JS-compiled regex)

Then we extend the Babel type with additional properties (`originalPattern`, `extra.originalRaw`) that preserve the original source formatting (eg whitespace in heregex, unicode code point escapes in a non-`u` regex)

src/nodes.coffee
```diff
@@ -966,6 +966,22 @@ exports.RegexLiteral = class RegexLiteral extends Literal
...
+    [, pattern] = @REGEX_REGEX.exec @value
+    return {
+      value: undefined
```
helixbass: The `undefined` for `value`/`extra.rawValue` are a little unusual, but that's what Babel generates

I looked at the `babel-parser` source code, and it seems to treat `RegExpLiteral`s like other literal types that expose a `value`/`extra.rawValue`, but for some reason it passes a `value` that's undefined (at least for all regexes I've looked at)

PR #5184, opened 2019-03-29
Index AST refinement: compilation level

helixbass:
@GeoffreyBooth I ran into an inconsistency with the way `Index` ASTs were being generated while working on the Prettier plugin

Basically, the "compilation" level needed to be explicitly set when generating AST corresponding to `Index` nodes

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1362,12 +1362,14 @@ exports.Value = class Value extends Base
   astProperties: (o) ->
     [..., property] = @properties
     property.name.csx = yes if @isCSXTag()
-    return
+    computed = property instanceof Index or property.name?.unwrap() not instanceof PropertyName
+    return {
       object: @object().ast o, LEVEL_ACCESS
-      property: property.ast o
-      computed: property instanceof Index or property.name?.unwrap() not instanceof PropertyName
+      property: property.ast o, (LEVEL_PAREN if computed)
+      computed
       optional: !!property.soak
       shorthand: !!property.shorthand
+    }
 
   astLocationData: ->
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1362,12 +1362,14 @@ exports.Value = class Value extends Base
   astProperties: (o) ->
     [..., property] = @properties
     property.name.csx = yes if @isCSXTag()
-    return
+    computed = property instanceof Index or property.name?.unwrap() not instanceof PropertyName
+    return {
       object: @object().ast o, LEVEL_ACCESS
-      property: property.ast o
-      computed: property instanceof Index or property.name?.unwrap() not instanceof PropertyName
+      property: property.ast o, (LEVEL_PAREN if computed)
```
helixbass: Set `LEVEL_PAREN` when this is a "computed" access eg `a[b]` - this corresponds to the `LEVEL_PAREN` used in `Index::compileNode()`

test/abstract_syntax_tree.coffee
```diff
@@ -838,6 +838,17 @@ test "AST as expected for Index node", ->
     optional: no
     shorthand: no
 
+  testExpression 'a[if b then c]',
+    type: 'MemberExpression'
+    object: ID 'a'
+    property:
+      type: 'ConditionalExpression'
```
helixbass: This test would have previously been breaking since the if statement AST was being generated at `LEVEL_TOP` (which was "bleeding through" from the outer context) so was producing a block with an `IfStatement` instead of a `ConditionalExpression`

PR #5185, opened 2019-03-29
Tagged template literal AST

helixbass:
@GeoffreyBooth PR for `TaggedTemplateLiteral` AST

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -949,6 +949,12 @@ exports.StringLiteral = class StringLiteral extends Literal
     unquoted = unquoted.replace /\\n/g, '\n' if csx
     unquoted
 
+  # `StringLiteral`s can represent either entire literal strings
+  # or pieces of text inside of e.g. an interpolated string.
+  # When parsed as the former but needing to be treated as the latter
+  # (e.g. the string part of a tagged template literal), this will return
+  # a copy of the `StringLiteral` with the quotes trimmed from its location
+  # data (like it would have if parsed as part of an interpolated string).
   withoutQuotesInLocationData: ->
     endsWithNewline = @originalValue[-1..] is '\n'
     locationData = Object.assign {}, @locationData
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -949,6 +949,28 @@ exports.StringLiteral = class StringLiteral extends Literal
     unquoted = unquoted.replace /\\n/g, '\n' if csx
     unquoted
 
+  withoutQuotesInLocationData: ->
+    endsWithNewline = @originalValue[-1..] is '\n'
+    locationData = Object.assign {}, @locationData
+    locationData.first_column          += @quote.length
+    if endsWithNewline
+      locationData.last_line -= 1
+      locationData.last_column =
+        if locationData.last_line is locationData.first_line
+          locationData.first_column + @originalValue.length - '\n'.length
+        else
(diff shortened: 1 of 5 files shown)


src/nodes.coffee
```diff
@@ -949,6 +949,28 @@ exports.StringLiteral = class StringLiteral extends Literal
     unquoted = unquoted.replace /\\n/g, '\n' if csx
     unquoted
 
+  withoutQuotesInLocationData: ->
```
helixbass: This is more of the "actual string literal" vs "text chunk inside interpolation" business

Basically, a `TaggedTemplateLiteral` always wants its "string" to be a `StringWithInterpolations`, but if it's a simple non-interpolated string eg `a"b"`, it'll have been parsed as a `StringLiteral` so a conversion from `StringLiteral` -> `StringWithInterpolations` is required

And the existing version of that conversion didn't assign accurate location data

src/nodes.coffee
```diff
@@ -949,6 +949,28 @@ exports.StringLiteral = class StringLiteral extends Literal
...
+    endsWithNewline = @originalValue[-1..] is '\n'
+    locationData = Object.assign {}, @locationData
+    locationData.first_column          += @quote.length
+    if endsWithNewline
```
helixbass: This calculation of `last_line`/`last_column` gets a little hairy (it actually has become irrelevant to AST generation since `last_line_exclusive`/`last_column_exclusive` were introduced but still might as well be completely accurate with location data)

So I added some tests to `test/location.coffee`

src/nodes.coffee
```diff
@@ -1906,12 +1928,19 @@ exports.RegexWithInterpolations = class RegexWithInterpolations extends Base
 
 exports.TaggedTemplateCall = class TaggedTemplateCall extends Call
   constructor: (variable, arg, soak) ->
-    arg = new StringWithInterpolations Block.wrap([ new Value arg ]) if arg instanceof StringLiteral
+    arg = StringWithInterpolations.fromStringLiteral arg if arg instanceof StringLiteral
```
helixbass: This is the existing `StringLiteral` -> `StringWithInterpolations` conversion (which didn't assign correct location data)

I think modeling this as a static method "pseudo-constructor" `StringWithInterpolations::fromStringLiteral()` seems appropriate

And `StringWithInterpolations::fromStringLiteral()` will see reuse when we cover JSX text content

src/nodes.coffee
```diff
@@ -4469,6 +4498,12 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
   constructor: (@body, {@quote, @startQuote} = {}) ->
     super()
 
+  @fromStringLiteral: (stringLiteral) ->
+    updatedString = stringLiteral.withoutQuotesInLocationData()
+    updatedStringValue = new Value(updatedString).withLocationDataFrom updatedString
+    new StringWithInterpolations Block.wrap([updatedStringValue]), quote: stringLiteral.quote
+    .withLocationDataFrom stringLiteral
```
helixbass: Basically, the `StringWithInterpolations`'s location data should include the quotes but the inner "text-chunk" `StringLiteral`'s location data shouldn't include the quotes

src/nodes.coffee
GeoffreyBooth: I feel like some of this explanation should probably go in comments in the code? Like a reminder of why this method exists and when one should use it?

helixbass: Sure, added an explanatory comment

PR #5187, opened 2019-03-30
JSX content AST

helixbass:
@GeoffreyBooth PR for AST for JSX element content (text, JSX expressions, nested JSX elements)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1503,10 +1503,10 @@ exports.CSXIdentifier = class CSXIdentifier extends IdentifierLiteral
   astType: -> 'JSXIdentifier'
 
 exports.CSXExpressionContainer = class CSXExpressionContainer extends Base
-  constructor: (@expression) ->
+  constructor: (@expression, {locationData} = {}) ->
     super()
     @expression.csxAttribute = yes
-    @locationData = @expression.locationData
+    @locationData = locationData ? @expression.locationData
 
   children: ['expression']
 
@@ -1518,6 +1518,24 @@ exports.CSXExpressionContainer = class CSXExpressionContainer extends Base
   astProperties: ->
     return
       expression: @expression.ast()
+
+exports.CSXEmptyExpression = class CSXEmptyExpression extends Base
+  astType: -> 'JSXEmptyExpression'
+
+exports.CSXText = class CSXText extends Base
+  constructor: (stringLiteral) ->
+    super()
+    @value = stringLiteral.unquote yes, yes
+    @locationData = stringLiteral.locationData
+
+  astType: -> 'JSXText'
+
+  astProperties: ->
+    return {
+      @value
+      extra:
+        raw: @value
+    }
 
 exports.CSXAttribute = class CSXAttribute extends Base
   constructor: ({@name, value}) ->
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -1695,14 +1713,38 @@ exports.CSXElement = class CSXElement extends Base
 
     {openingFragment, closingFragment}
 
+  contentAst: (o) ->
```
helixbass: Took the approach of having the `CSXElement` parent be responsible (during AST generation) for extracting the relevant individual children and eg instantiating new `CSXText`/`CSXEmptyExpression` node classes (only used currently for AST generation)

src/nodes.coffee
```diff
@@ -1695,14 +1713,38 @@ exports.CSXElement = class CSXElement extends Base
...
+    return [] unless @content and not @content.base.isEmpty?()
+
+    content = @content.unwrapAll()
+    children =
+      if content instanceof StringLiteral
+        [new CSXText content]
+      else # StringWithInterpolations
+        for element in @content.unwrapAll().extractElements o, includeInterpolationWrappers: yes
```
helixbass: We need the `Interpolation` location data for the `CSXExpressionContainer` (whose location data needs to include the surrounding `{`/`}` of the JSX expression), so added an `includeInterpolationWrappers` option to `StringWithInterpolations::extractElements()` (as well as a `locationData` option to `CSXExpressionContainer::constructor()`)

src/nodes.coffee
GeoffreyBooth: Which feels like it should be the job of the lexer/parser, right? But refactoring to make that happen is probably way too big a scope.

test/abstract_syntax_tree.coffee
```diff
@@ -343,6 +343,91 @@ test "AST as expected for CSXTag node", ->
...
+    closingFragment:
+      type: 'JSXClosingFragment'
+    children: [
+      type: 'JSXText'
+      extra:
+        raw: 'abc'
+      value: 'abc'
+    ,
+      type: 'JSXEmptyExpression'
+    ]
```
GeoffreyBooth: I assume all these match Babel? Since there’s nothing custom about JSX in CoffeeScript.

src/nodes.coffee
helixbass: Yup, with JSX the options are:

1. Fully lex/parse as JSX rather than passing JSX through the grammar as pseudo-calls etc. We've discussed that this might ultimately be worth doing (it'd avoid some of the backflips we've done in these PRs to eg preserve JSX location data) but isn't something we're considering as in scope for this AST work

So then in order to avoid either manually assembling JSX AST nodes or having node classes like `Call` serve "double-duty" at AST generation time and have their `ast()` method check whether they're actually a JSX node, we've been either

2. Converting to JSX-specific node class instances "up front" (ie during node class construction), so that the JSX-specific node classes are responsible for both compile-to-JS (ie have `compileNode()` methods) and AST generation. This is what we've done with eg `CSXElement`

3. Converting to JSX-specific node class instances during AST generation (leaving the existing compile-to-JS code as-is)

In this case I took # 3, since it felt like it could be cleanly handled by the `CSXElement` parent without having to mess with the existing code in `StringWithInterpolations::compileNode()` for differentiating JSX content from actual interpolated string contents

test/abstract_syntax_tree.coffee
helixbass: Yup exactly, they should all match Babel's JSX AST. We've discussed converting all `CSX*` names to `JSX*` as part of this, which would simplify the JSX AST code a bit in that they'd all have node class names matching the AST type and so wouldn't need custom `astType()` methods

This PR is the last chunk of JSX AST coverage, so should I go ahead and make a PR for `CSX` -> `JSX` now? Or better to wait until after we've gotten through the rest of the AST stuff?

GeoffreyBooth: Do you mind following up this PR with another one right after it that just does the rename? That would be good to keep in its own commit for posterity, and for being able to see in the history what changed due to that versus the AST work.

PR #5188, opened 2019-03-31
AST: "CSX" -> "JSX"

helixbass:
@GeoffreyBooth PR to update all uses of "CSX" to use "JSX" instead


src/nodes.coffee
```diff
@@ -1034,27 +1034,11 @@ exports.IdentifierLiteral = class IdentifierLiteral extends Literal
     return
       name: @value
 
-exports.CSXTag = class CSXTag extends IdentifierLiteral
```
helixbass: Just had to move this class to below `JSXIdentifier` so that I could inherit from it (specifically to inherit its `astType: -> 'JSXIdentifier'`)

PR #5189, opened 2019-03-31
AST: JSX empty expression

helixbass:
@GeoffreyBooth in merging the latest `ast` into my branch and updating the Prettier plugin to use that dependency, I found that the AST structure being generated for JSX empty expressions (`{}`) was incorrect

Instead of a `JSXEmptyExpression` node *replacing* the typical `JSXExpressionContainer` node, it actually needs to be *nested inside* a `JSXExpressionContainer` node

So this PR updates the AST generated for JSX empty expressions to correctly match Babel AST

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1718,7 +1718,25 @@ exports.JSXElement = class JSXElement extends Base
           else # Interpolation
             {expression} = element
             unless expression?
-              new JSXEmptyExpression().withLocationDataFrom element
+              emptyExpression = new JSXEmptyExpression()
+              # We don’t currently have a token corresponding to the empty space
+              # between the JSX expression braces, so piece together the location
+              # data by trimming the braces from the Interpolation’s location data.
+              # Technically the last_line/last_column calculation here could be
+              # incorrect if the ending brace is preceded by a newline, but
+              # last_line/last_column aren’t used for AST generation anyway.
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -1718,7 +1718,25 @@ exports.JSXElement = class JSXElement extends Base
...
             unless expression?
-              new JSXEmptyExpression().withLocationDataFrom element
+              emptyExpression = new JSXEmptyExpression()
+              # We don’t currently have a token corresponding to the empty space
+              # between the JSX expression braces, so piece together the location
+              # data by trimming the braces from the Interpolation’s location data.
+              # Technically the last_line/last_column calculation here could be
+              # incorrect if the ending brace is preceded by a newline, but
+              # last_line/last_column aren’t used for AST generation anyway.
+              emptyExpression.locationData =
```
helixbass: The `JSXEmptyExpression`'s location data needs to correspond to the *empty space* in between the JSX expression `{`/`}`

As I explained in a code comment here, we don't have a token corresponding to that empty space that we could directly use for the `JSXEmptyExpression` location data

So rather than trying to do something rather messy to eg create an "empty space" token to represent that (and pass it through the grammar), we're able to piece together the necessary location data here from the JSX expression's location data by stripping the location of the starting `{` and closing `}`

PR #5190, opened 2019-04-01
AST: JSX element as expression

helixbass:
@GeoffreyBooth here's another case I realized wasn't being distinguished by the existing logic for JSX content:

Even though it's unnecessary to do so, you could have a JSX element explicitly wrapped in a JSX expression (`{`/`}`). Eg typically you'd write
```
<a><b /></a>
```
but you could write
```
<a>{<b />}</a>
```
and the AST should be different (the second should have a `JSXExpressionContainer` wrapping the nested `JSXElement`)

This PR covers that case

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1739,7 +1739,9 @@ exports.JSXElement = class JSXElement extends Base
               new JSXExpressionContainer emptyExpression, locationData: element.locationData
             else
               unwrapped = expression.unwrapAll()
-              if unwrapped instanceof JSXElement
+              if unwrapped instanceof JSXElement and
+                  # distinguish `<a><b /></a>` from `<a>{<b />}</a>`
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -1739,7 +1739,9 @@ exports.JSXElement = class JSXElement extends Base
               new JSXExpressionContainer emptyExpression, locationData: element.locationData
             else
               unwrapped = expression.unwrapAll()
-              if unwrapped instanceof JSXElement
+              if unwrapped instanceof JSXElement and
+                  # distinguish `<a><b /></a>` from `<a>{<b />}</a>`
+                  unwrapped.locationData.range[0] is element.locationData.range[0]
```
helixbass: Both the "unwrapped" and "wrapped" nested JSX element cases currently generate identical node class structures (they're both treated as `Interpolation`s), but we can distinguish them by inspecting location data

PR #5194, opened 2019-04-05
Basic class AST

helixbass:
@GeoffreyBooth first PR for class AST

This starts simple and just covers simple class/superclass name and simple methods (not static, not bound)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -792,6 +792,8 @@ exports.Block = class Block extends Base
   astType: ->
     if @isRootBlock
       'Program'
+    else if @isClassBody
+      'ClassBody'
     else
       'BlockStatement'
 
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -2734,6 +2739,27 @@ exports.Class = class Class extends Base
 
     null
 
+  isStatementAst: -> yes
+
+  ast: (o, level) ->
+    @body.isClassBody = yes
+    @body.locationData = zeroWidthLocationDataFromEndLocation @locationData if @hasGeneratedBody
```
helixbass: Supply missing location data for an empty class body

src/nodes.coffee
```diff
@@ -3779,15 +3807,34 @@ exports.Code = class Code extends Base
     else
       name
 
-  astProperties: (o) ->
+  methodAstProperties: (o) ->
     return
+      static: no
```
helixbass: In TDD style, hardcode `static` and `computed` for now and will make them dynamically calculated once we get there/have tests that require it

PR #5197, opened 2019-04-09
Object property AST refinement: compilation level

helixbass:
@GeoffreyBooth I ran into a missing compilation level when generating AST for `ObjectProperty` (similar to eg #5184)

So this explicitly generates AST for the object property key/value at `LEVEL_LIST` (similar to `Assign`)

### DIFF
--- a/test/abstract_syntax_tree.coffee
+++ b/test/abstract_syntax_tree.coffee
@@ -1245,6 +1245,36 @@ test "AST as expected for Obj node", ->
     ]
     implicit: yes
 
+  testExpression '''
+    a:
+      if b then c
+  ''',
+    type: 'ObjectExpression'
+    properties: [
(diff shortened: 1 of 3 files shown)


test/abstract_syntax_tree.coffee
```diff
@@ -1245,6 +1245,36 @@ test "AST as expected for Obj node", ->
...
+  testExpression '''
+    a:
+      if b then c
+  ''',
+    type: 'ObjectExpression'
+    properties: [
+      type: 'ObjectProperty'
+      key: ID 'a'
+      value:
+        type: 'ConditionalExpression'
```
helixbass: This tests the specific case I ran into working on the Prettier plugin - this was generating an `IfStatement` inside a block instead of a `ConditionalExpression`

PR #5199, opened 2019-04-14
Class AST: static methods/properties

helixbass:
@GeoffreyBooth PR for class static methods/static properties

Remaining parts of class ASTs include bound/computed methods/properties and executable class bodies

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -486,8 +486,9 @@ exports.Root = class Root extends Base
   # better not to generate them in the first place, but for now, clean up
   # obvious double-parentheses.
   compileNode: (o) ->
-    o.indent  = if o.bare then '' else TAB
-    o.level   = LEVEL_TOP
+    o.indent    = if o.bare then '' else TAB
+    o.level     = LEVEL_TOP
+    o.compiling = yes
     @initializeScope o
     fragments = @body.compileRoot o
     return fragments if o.bare
@@ -1272,8 +1273,10 @@ exports.Value = class Value extends Base
     lastProperty instanceof Slice
 
   looksStatic: (className) ->
-    (@this or @base instanceof ThisLiteral or @base.value is className) and
+    return no unless ((thisLiteral = @base) instanceof ThisLiteral or (name = @base).value is className) and
       @properties.length is 1 and @properties[0].name?.value isnt 'prototype'
+    return
+      staticClassName: thisLiteral ? name
 
   # The value can be unwrapped as its inner node, if there are no attached
   # properties.
@@ -2554,7 +2557,7 @@ exports.Class = class Class extends Base
 
   compileNode: (o) ->
     @name          = @determineName()
-    executableBody = @walkBody()
+    executableBody = @walkBody o
 
     # Special handling to allow `class expr.A extends A` declarations
     parentName    = @parent.base.value if @parent instanceof Value and not @parent.hasProperties()
@@ -2622,7 +2625,7 @@ exports.Class = class Class extends Base
       @variable.error message if message
     if name in JS_FORBIDDEN then "_#{name}" else name
 
-  walkBody: ->
+  walkBody: (o) ->
     @ctor          = null
     @boundMethods  = []
     executableBody = null
@@ -2640,7 +2643,7 @@ exports.Class = class Class extends Base
         pushSlice = -> exprs.push new Value new Obj properties[start...end], true if end > start
 
         while assign = properties[end]
-          if initializerExpression = @addInitializerExpression assign
+          if initializerExpression = @addInitializerExpression assign, o
             pushSlice()
             exprs.push initializerExpression
             initializer.push initializerExpression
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -486,8 +486,9 @@ exports.Root = class Root extends Base
   # better not to generate them in the first place, but for now, clean up
   # obvious double-parentheses.
   compileNode: (o) ->
-    o.indent  = if o.bare then '' else TAB
-    o.level   = LEVEL_TOP
+    o.indent    = if o.bare then '' else TAB
+    o.level     = LEVEL_TOP
+    o.compiling = yes
```
helixbass: Introduce a flag on `o` that lets us distinguish (in node class methods shared by AST generation and compile-to-JS) whether we're "compiling" (to JS) as opposed to generating AST

src/nodes.coffee
```diff
@@ -1272,8 +1273,10 @@ exports.Value = class Value extends Base
     lastProperty instanceof Slice
 
   looksStatic: (className) ->
-    (@this or @base instanceof ThisLiteral or @base.value is className) and
+    return no unless ((thisLiteral = @base) instanceof ThisLiteral or (name = @base).value is className) and
       @properties.length is 1 and @properties[0].name?.value isnt 'prototype'
+    return
+      staticClassName: thisLiteral ? name
```
helixbass: In order to be able to generate AST that include the actual "static class name reference" (eg `A` from `A.b`, `this`, or `@` in the following:
```
class A
  A.b = ->
  @c: 2
  this.d = 3
```
, return that static class name reference as part of the return value from `looksStatic()` (while preserving its existing calling interface which expects a truthy or falsy return value)

src/nodes.coffee
```diff
@@ -2684,11 +2687,13 @@ exports.Class = class Class extends Base
...
   # (e.g. getters and setters defined via the `get` and `set` keywords as
   # opposed to the `Object.defineProperty` method).
-  addInitializerExpression: (node) ->
+  addInitializerExpression: (node, o) ->
     if node.unwrapAll() instanceof PassthroughLiteral
       node
     else if @validInitializerMethod node
       @addInitializerMethod node
+    else if not o.compiling and @validClassProperty node
+      @addClassProperty node
```
helixbass: In Babel, static class properties generate `ClassProperty` AST nodes, so if we're generating AST, create instances of the new `ClassProperty` node class as we encounter static properties while walking the class body

src/nodes.coffee
```diff
@@ -2872,6 +2894,23 @@ exports.ExecutableClassBody = class ExecutableClassBody extends Base
...
+
+  isStatement: YES
+
+  astProperties: (o) ->
+    return
+      key: @name.ast o, LEVEL_LIST
+      value: @value.ast o, LEVEL_LIST
+      static: !!@isStatic
+      computed: no
+      operator: @operatorToken?.value ? '='
```
helixbass: `operator` and `staticClassName` are Coffeescript-specific extensions to the Babel `ClassProperty` AST node type (and similarly on `ClassMethod` below)

src/nodes.coffee
GeoffreyBooth: I’m not sure about `o.compiling`. This appears to be the only place you need it; are you using it elsewhere in your Prettier branch for other things?

It seems like you’re using this in order to create a different node tree for JS versus AST generation. I’m wondering, why should we have separate trees? That seems like it would just make the overall codebase harder to maintain.

If you need a new child class for class properties or static class properties, can you make that new class used for both JS and AST? I know that might mean a bigger refactor on the JS side, but I think there’s value in keeping the two node trees/structures the same.

helixbass: > are you using it elsewhere in your Prettier branch for other things?

Yes, and it may come in handy elsewhere in these AST PRs

> I’m wondering, why should we have separate trees? That seems like it would just make the overall codebase harder to maintain

Well, let's look at this from the perspective of what we've been doing so far. In some places we've introduced node classes that also get used during compile-to-JS (eg the new JSX node classes). In other places we've introduced node classes that only get used during AST generation (eg `ObjectProperty`). We've been basing those decisions on things like how straightforward it is to integrate the new node classes

So basically yes there's a convenience to having the AST and JS node trees match but we've already set precedent (for good reasons) for having them diverge

The only difference here is that that divergence is taking place inside a shared helper method (rather than eg an AST-specific method) - sharing helper methods between compile-to-JS and AST generation is good for similar reasons (more maintainable, DRYer)

And conceptually I don't have a problem with a flag on `o` indicating whether we're in "compile-to-JS" mode or "AST generation" mode - `o` is basically a context object for various things about keeping track of the current task (whether that's compiling to JS or AST generation), so differentiating those two tasks seems like part of that context

> If you need a new child class for class properties or static class properties, can you make that new class used for both JS and AST?

So in terms of this specific choice (not to try and reuse `ClassProperty` in compile-to-JS), I avoided it because class compilation is a rather complex process and I felt like it wasn't worth messing with its nuts and bolts (just to avoid an AST-specific node class)

PR #5200, opened 2019-04-18
Partially fix #5121: array literal ending with empty function

### LINKED ISSUE #5121: Bug: Array literals that contain function or implicit objects with function values
Choose one: is this a bug report or feature request?

Bug report

### Input Code

```coffee
[k: ->]
```
and
```coffee
[j: -> m is 0]
```
and
```coffee
[->]
```

### Expected Behavior
The above should compile.

### Current Behavior
It does not compile.

### Context
```coffee
[->a]
```
compiles and
```coffee
[->,a]
```
compiles

### Environment

* CoffeeScript version: 2 only, version 1 compiles
* Node.js version:

helixbass:
@GeoffreyBooth I'd noted in #5121 that at least the cases where the `->` is immediately followed by `]` should be a simple fix

And then ran into that case working on the Prettier plugin

So this is a one-line fix (extending the similar fix from #4588) for those cases (I'd leave #5121 open since there is another case there that this PR doesn't address)

### DIFF
--- a/lib/coffeescript/rewriter.js
+++ b/lib/coffeescript/rewriter.js
@@ -846,7 +846,7 @@
           return i + 2;
         };
         return this.scanTokens(function(token, i, tokens) {
-          var conditionTag, j, k, ref, ref1, tag;
+          var conditionTag, j, k, ref, ref1, ref2, tag;
           [tag] = token;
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
Should this target `master`?

helixbass:
Sure, retargeted to `master`

PR #5202, opened 2019-04-20
Fix #5201: allow leading breaking elision

### LINKED ISSUE #5201: Simple indented leading elisions aren't supported
@GeoffreyBooth I came across this while working on the Prettier plugin:

When an array with leading elisions breaks onto multiple lines, this would be the expected formatting (this matches what the Prettier JS formatter outputs):

Eg for `[, , b, c]`:

```
[
  ,
  ,
  b,
  c
]
```

But this doesn't parse

Will submit a fixing PR momentarily

helixbass:
Fixes #5201 

Enables simple indented leading elisions eg:
```
[
  ,
  ,
  a,
  b
]
```


src/grammar.coffee
```diff
@@ -604,7 +604,8 @@ grammar =
   ArgElisionList: [
     o 'ArgElision'
     o 'ArgElisionList , ArgElision',                                          -> $1.concat $3
-    o 'ArgElisionList OptElisions TERMINATOR ArgElision',                     -> $1.concat $2, $4
```
helixbass: Basically there was no grammar support for a `TERMINATOR` following a "leading" `Elision` (ie preceding any actual non-elision array elements)

The reason that things like this worked is because leading commas suppress `TERMINATOR`s:
```
[
  ,
  ,
  ,
]
```

So the main grammar change is to add the rule `Elision TERMINATOR` below

Then this rule needed to be split apart in order not to create a grammar conflict - basically this rule `ArgElisionList OptElisions TERMINATOR ArgElision` was covering all of these different cases:

- A typical indented array with no comma eg
```
[
  a
  b
]
```

- A typical indented array with a comma eg
```
[
  a,
  b
]
```

- A non-initial elision followed by a newline eg
```
[
  a
  ,
  ,
  b
]
```

So this just replaces the rule with a rule that only covers the first two of these cases, and then the third is handled by the addition of `Elision TERMINATOR` below

GeoffreyBooth:
Should this go on master?

helixbass:
Up to you, I figured since no one asked for a fix that it was a bit less "urgent"?

GeoffreyBooth:
Yeah but since it's a general bug, not related to AST stuff, I thought it should go on master. That way we keep the AST branch and eventual PR as only the changes for outputting an AST.

helixbass:
Ok retargeted to `master`

Issue #5204, opened 2019-04-26
Compiles to invalid JS: computed class property


helixbass:
@GeoffreyBooth I came across this while working on class AST:

This:
```
class A
  [b]: 3
```
compiles to:
```
var A;

A = (function() {
  class A {};

  A.prototype.[b] = 3;

  return A;

}).call(this);
```
which doesn't parse

Does it make sense to allow computed class properties like this?

If so, it's probably just a question of omitting the `.` when compiling (so here generate `A.prototype[b] = 3` instead)?

Fwiw, it looks like this didn't compile (as Coffeescript source) in Coffeescript 1.x. I didn't trace when it became allowed Coffeescript syntax

helixbass:
For the static case, it's currently inconsistent as far as allowed syntax:
```
# this compiles
class A
  @[b] = 3

# but this doesn't
class A
  @[b]: 3
```
but the version that compiles (`@[b] = 3`) generates valid JS it appears (`A[b] = 3`)

GeoffreyBooth:
I guess it might as well compile, since we support the same for objects since #4803. That might’ve been the PR where this broke for classes.

Can we fix this on `master`?

PR #5205, opened 2019-04-26
Class prototype property AST

helixbass:
@GeoffreyBooth PR for class "prototype property" AST

As far as I'm aware, there's no corresponding proposed JS syntax/Babel AST for class properties that get attached to the prototype eg:
```
  class A
    b: 3
```

So this PR introduces a new node class/AST type `ClassPrototypeProperty` to represent these

Also contains a fix for #5204 (the invalid JS for computed class prototype properties, not my second comment in that issue). If you'd prefer I could extract the fix to a separate PR against `master`

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2694,8 +2694,8 @@ exports.Class = class Class extends Base
       @addInitializerMethod node
     else if not o.compiling and @validClassProperty node
       @addClassProperty node
-    else if not o.compiling and @validClassInstanceProperty node
-      @addClassInstanceProperty node
+    else if not o.compiling and @validClassPrototypeProperty node
+      @addClassPrototypeProperty node
     else
       null
 
@@ -2738,11 +2738,11 @@ exports.Class = class Class extends Base
       operatorToken
     }).withLocationDataFrom assign
 
-  validClassInstanceProperty: (node) ->
+  validClassPrototypeProperty: (node) ->
     return no unless node instanceof Assign
     node.context is 'object' and not node.variable.hasProperties()
 
-  addClassInstanceProperty: (assign) ->
+  addClassPrototypeProperty: (assign) ->
     {variable, value} = assign
     new ClassPrototypeProperty({
       name: variable.base
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2694,6 +2694,8 @@ exports.Class = class Class extends Base
       @addInitializerMethod node
     else if not o.compiling and @validClassProperty node
       @addClassProperty node
+    else if not o.compiling and @validClassInstanceProperty node
+      @addClassInstanceProperty node
     else
       null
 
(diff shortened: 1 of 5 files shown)


src/nodes.coffee
```diff
@@ -2883,7 +2896,11 @@ exports.ExecutableClassBody = class ExecutableClassBody extends Base
         # The class scope is not available yet, so return the assignment to update later
         assign = @externalCtor = new Assign new Value, value
       else if not assign.variable.this
-        name      = new (if base.shouldCache() then Index else Access) base
+        name =
+          if base instanceof ComputedPropertyName
```
helixbass: This is the fix for #5204

src/nodes.coffee
```diff
@@ -2883,7 +2896,11 @@ exports.ExecutableClassBody = class ExecutableClassBody extends Base
...
+            new Index base.value
+          else
+            new (if base.shouldCache() then Index else Access) base
```
helixbass: If you remove the conditional logic here (so just `new Access base`) no tests break. But I don't feel completely confident removing it. I traced the code's origin back to #4354

src/nodes.coffee
```diff
@@ -2911,6 +2928,20 @@ exports.ClassProperty = class ClassProperty extends Base
...
 
+exports.ClassPrototypeProperty = class ClassPrototypeProperty extends Base
+  constructor: ({@name, @value}) ->
+    super()
+
+  children: ['name', 'value']
+
+  isStatement: YES
+
+  astProperties: (o) ->
```
helixbass: This mimics the properties of Babel AST types like `ClassProperty`/`ObjectProperty`

GeoffreyBooth:
This looks fine. One thing to consider is class fields: https://github.com/tc39/proposal-class-fields

Which are currently in Stage 3 but are already shipped in Chrome and Node 12, so we’ll probably need to support them soon.

helixbass:
Extracted the fix for #5204 to a PR against `master` (#5206). So once that gets merged, if you merge `master` into `ast`, I can add AST tests here for static computed class properties as well

> One thing to consider is class fields

Ok so my understanding is this:

"Class fields" in JS eg
```
class A {
  x = 3
}
```
represent properties set on new class *instances*, so basically correspond to the Coffeescript:
```
class A
  constructor: ->
    @x = 3
```
I personally find the JS syntax weird, I guess I'm more accustomed to Python/Ruby/Coffeescript-style initialization of instances inside the constructor

So I'd assume we don't need to provide a syntax that compiles to JS class fields (since you can achieve the equivalent by initializing inside the constructor)?

Then I'm also familiar with arrow function "class fields" (I don't know the actual JS nomenclature/proposals) eg
```
class A {
  b = () => 3
}
```
And this seems to correspond to Coffeescript bound methods eg
```
class A
  b: => 3
```

But then Coffeescript's "prototype properties" eg
```
class A
  b: 3
```
are a different thing that just get set once per class (on the prototype) rather than once per instance. If there's a JS proposal for this I haven't seen it, which is why I create a new (non-Babel) AST type `ClassPrototypeProperty` in this PR to represent them

PR #5208, opened 2019-04-29
Class AST: bound/computed properties, executable body

helixbass:
@GeoffreyBooth this covers the remaining class AST:

- bound methods
- computed properties/methods
- executable bodies

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3889,6 +3889,7 @@ exports.Code = class Code extends Base
           'method'
       operator: @operatorToken?.value ? '='
       staticClassName: @isStatic.staticClassName?.ast(o) ? null
+      bound: !!@bound
 
   astProperties: (o) ->
     return Object.assign
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2668,6 +2668,7 @@ exports.Class = class Class extends Base
       else if method.bound
         @boundMethods.push method
 
+    return unless o.compiling
     if initializer.length isnt expressions.length
       @body.expressions = (expression.hoist() for expression in initializer)
       new Block expressions
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2925,7 +2925,7 @@ exports.ClassProperty = class ClassProperty extends Base
       key: @name.ast o, LEVEL_LIST
       value: @value.ast o, LEVEL_LIST
       static: !!@isStatic
-      computed: no
+      computed: @name instanceof ComputedPropertyName
       operator: @operatorToken?.value ? '='
       staticClassName: @staticClassName?.ast(o) ? null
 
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -2668,6 +2668,7 @@ exports.Class = class Class extends Base
       else if method.bound
         @boundMethods.push method
 
+    return unless o.compiling
```
helixbass: I think the natural representation for executable class body statements is the equivalent AST as if they were just statements in a regular block

So here, we don't have to deal with `ExecutableClassBody`/`HoistTarget` at all, we just leave the class body `Block` as a mix of "class-specific" nodes (`ClassMethod`/`ClassProperty`/`ClassPrototypeProperty`) and other "executable body" nodes, and then the class body AST ends up looking like eg:
```
# this could correspond to eg:
# class A
#   b: ->
#   x = 3
#   @c: 2
#   if y
#     z()
type: 'ClassBody'
body: [
  {type: 'ClassMethod', ...}
  {type: 'ExpressionStatement', ...}
  {type: 'ClassProperty', ...}
  {type: 'IfStatement', ...}
]
```

GeoffreyBooth: The ones that match class fields should probably output as those Babel node types, I'd think?

helixbass: Can you give an example of what you mean by "class fields"? See my [comment](https://github.com/jashkenas/coffeescript/pull/5205#issuecomment-487392736) as far as my understanding of how Coffeescript constructs relate to the JS class fields proposal

So basically we are matching Babel for everything that corresponds semantically/structurally to a JS/Babel construct

Eg a static property corresponds to a JS "static class field", so for this:
```
class A
  @b: 2
```
we generate an AST node with:
```
type: 'ClassProperty'
static: true
```
like you'd see in the Babel AST for
```
class A {
  static b = 2
}
```

GeoffreyBooth: I just meant the class fields proposal. Though I guess it doesn’t apply here because these are all on the prototype?

See https://astexplorer.net/#/gist/152f7080b8be9dde369647933ab7747f/c7aea4d9cae9df2a4b6fbf61bbac604f6d8dac2e

```js
class Foo {
  a = 1
}
```

image

helixbass: Yup I don't think we have a structure that corresponds to non-static JS class fields

I read over #4552 for context. My understanding is that the answer to his first question there ("does the `b` property have different semantics?") is "yes", that since `b: 3` is set (once) on the prototype, it's not semantically the same as a JS class field eg `b = 3`

So that's why I introduced `ClassPrototypeProperty` AST nodes to represent eg `b: 3`, since they're a different construct that doesn't seem to have an equivalent in JS/Babel classes

PR #5209, opened 2019-04-30
Directive AST

helixbass:
@GeoffreyBooth now that we've covered various types of blocks including class bodies, here's a PR for AST for directives (eg `'use strict'`)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1002,6 +1002,10 @@ exports.StringLiteral = class StringLiteral extends Literal
   shouldGenerateTemplateLiteralAst: ->
     @quote.length is 3
 
+  ast: (o, level) ->
+    return StringWithInterpolations.fromStringLiteral(@).ast o if @shouldGenerateTemplateLiteralAst()
+    super o, level
+
   astProperties: ->
     return
       value: @originalValue
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -799,20 +799,27 @@ exports.Block = class Block extends Base
       'BlockStatement'
 
   astProperties: (o) ->
+    checkForDirectives = del o, 'checkForDirectives'
+
+    sniffDirectives @expressions, notFinalExpression: checkForDirectives if @isRootBlock or checkForDirectives
+    directives = []
     body = []
     for expression in @expressions
+      expressionAst = expression.ast o
+      if expression instanceof Directive
+        directives.push expressionAst
       # If an expression is a statement, it can be added to the body as is.
-      if expression.isStatementAst o
-        body.push expression.ast o
+      else if expression.isStatementAst o
+        body.push expressionAst
       # Otherwise, we need to wrap it in an `ExpressionStatement` AST node.
       else
         body.push Object.assign
             type: 'ExpressionStatement'
-            expression: expression.ast o
+            expression: expressionAst
           ,
             expression.astLocationData()
 
-    return
+    return {
       # For now, we’re not including `sourceType` on the `Program` AST node.
       # Its value could be either `'script'` or `'module'`, and there’s no way
       # for CoffeeScript to always know which it should be. The presence of an
@@ -826,8 +833,20 @@ exports.Block = class Block extends Base
       # then CoffeeScript can copy Node’s algorithm.
 
       # sourceType: 'module'
-      body: body
-      directives: [] # Directives like `'use strict'` are coming soon.
+      body, directives
+    }
+
+# A directive e.g. 'use strict'.
+# Currently only used during AST generation.
+exports.Directive = class Directive extends Base
+  constructor: (@value) ->
+    super()
+
+  astProperties: (o) ->
+    return
+      value: Object.assign {},
+        @value.ast o
+        type: 'DirectiveLiteral'
 
 #### Literal
 
@@ -979,6 +998,9 @@ exports.StringLiteral = class StringLiteral extends Literal
     copy = new StringLiteral @originalValue, {@quote, @initialChunk, @finalChunk, @indent, @double, @heregex}
     copy.locationData = locationData
     copy
+
+  shouldGenerateTemplateLiteralAst: ->
+    @quote.length is 3
 
   astProperties: ->
     return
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -799,20 +799,27 @@ exports.Block = class Block extends Base
       'BlockStatement'
 
   astProperties: (o) ->
+    checkForDirectives = del o, 'checkForDirectives'
+
+    sniffDirectives @expressions, notFinalExpression: checkForDirectives if @isRootBlock or checkForDirectives
```
helixbass: `sniffDirectives()` replaces expressions that should be treated as directives with `Directive` node class instances

src/nodes.coffee
```diff
@@ -980,6 +999,13 @@ exports.StringLiteral = class StringLiteral extends Literal
     copy.locationData = locationData
     copy
 
+  shouldGenerateTemplateLiteralAst: ->
+    @quote.length is 3
```
helixbass: I'd run into cases where a leading heredoc was being treated as a directive. I guess in theory someone could be using them this way (since in the generated JS the heredoc will just be a normal string) but it seems bizarre/unexpected to treat a heredoc as a directive

This relates to what we generate as AST for heredocs - where I landed was that we should generate `TemplateLiteral` AST nodes (ie JS `` `...` ``) for all heredocs (rather than `StringLiteral`). This seemed more appropriate to me since they're more similar to template literals with respect to things like newlines

src/nodes.coffee
```diff
@@ -2781,6 +2807,7 @@ exports.Class = class Class extends Base
     @body.isClassBody = yes
     @body.locationData = zeroWidthLocationDataFromEndLocation @locationData if @hasGeneratedBody
     @walkBody o
+    sniffDirectives @body.expressions
```
helixbass: We also allow class bodies to have leading directives

src/nodes.coffee
```diff
@@ -5342,6 +5369,17 @@ makeDelimitedLiteral = (body, options = {}) ->
     when other     then (if options.double then "\\#{other}" else other)
   "#{options.delimiter}#{body}#{options.delimiter}"
 
+sniffDirectives = (expressions, {notFinalExpression} = {}) ->
+  index = 0
+  lastIndex = expressions.length - 1
+  while index <= lastIndex and expression = expressions[index]
+    break if index is lastIndex and notFinalExpression
```
helixbass: We need a `notFInalExpression` option to make sure that in the case of a block that is only strings that look like directives, the final one is not considered a directive since it's implicitly returned

Eg:
```
->
  'not a directive'
```

src/nodes.coffee
GeoffreyBooth: Yes, see #5019. It’s been a leftover TODO from the CoffeeScript 2 effort that we need to output heredoc strings as template literals using backticks.

src/nodes.coffee
```diff
@@ -3895,7 +3922,7 @@ exports.Code = class Code extends Base
   astProperties: (o) ->
     return Object.assign
       params: @paramForAst(param).ast(o) for param in @params
-      body: @body.ast o, LEVEL_TOP
+      body: @body.ast (merge o, checkForDirectives: yes), LEVEL_TOP
```
GeoffreyBooth: Can we use `Object.assign`? I’m trying to phase out `merge` and other unnecessary helpers.

src/nodes.coffee
```diff
@@ -5342,6 +5369,17 @@ makeDelimitedLiteral = (body, options = {}) ->
     when other     then (if options.double then "\\#{other}" else other)
   "#{options.delimiter}#{body}#{options.delimiter}"
 
+sniffDirectives = (expressions, {notFinalExpression} = {}) ->
+  index = 0
+  lastIndex = expressions.length - 1
+  while index <= lastIndex and expression = expressions[index]
```
GeoffreyBooth: In general we try to avoid assignments in conditionals, as it looks like a typo. Can `expression = expressions[index]` be moved to the first line in the `while` block?

test/abstract_syntax_tree.coffee
```diff
@@ -164,21 +164,36 @@ test "AST as expected for NaNLiteral node", ->
...
-#     type: 'StringLiteral'
-#     value: "'cheese string'"
-#     quote: "'"
-
-# test "AST as expected for RegexLiteral node", ->
-#   testExpression '/^(?!.*(.).*\\1)[gimsuy]*$/',
-#     type: 'RegexLiteral'
-#     value: '/^(?!.*(.).*\\1)[gimsuy]*$/'
+test "AST as expected for StringLiteral node", ->
+  # Just a standalone string literal would be treated as a directive
```
GeoffreyBooth: “would be treated as directive”? You mean as `StringLiteral`?

src/nodes.coffee
helixbass: Sure, updated to use `Object.assign()`

src/nodes.coffee
helixbass: Sure, moved into the `while` block

test/abstract_syntax_tree.coffee
helixbass: Sorry, that comment probably needed more explanation

Expanded to "Just a standalone string literal would be treated as a directive, so embed the string literal in an enclosing expression (e.g. a call)."

Meaning that if this checked the AST for just eg `"string cheese"`, it wouldn't be able to test for a `StringLiteral` AST node because that single string would actually be "parsed" as a directive (so would generate a `Directive` AST node)

GeoffreyBooth:
Thanks, looks good. Any thoughts on outputting heredocs as backticked template literals with real newlines, rather than as strings with `\n`s? I think at some point we had discussed making that change as part of or after the AST effort, since as part of figuring out the `raw` value for `StringWithInterpolations` you’d have to fix the lexer and/or rewriter to preserve that newline vs. escaped newline information.

helixbass:
> Any thoughts on outputting heredocs as backticked template literals with real newlines, rather than as strings with `\n`s?

Ya that seems doable at this point, I'll take a closer look

PR #5211, opened 2019-05-10
Compile heredocs to template literals

helixbass:
@GeoffreyBooth here's a PR to generate JS template literals when compiling heredocs

This is WIP in the sense that:
1. A test is currently failing:
  `test/strict.coffee` won't run in Node, failing with `SyntaxError: Octal escape sequences are not allowed in strict mode.`

  I don't fully grasp what's going on here - it looks like the fact that eg `'''\08'''` now compiles to `` `\08` `` (instead of `'\08'`) is causing it not to run. Does the presence of a template literal trigger strict mode in Node or something?

  And it looks like we forbid octal escapes in strings but only involving the digits `0`-`7` (see `validateEscapes()` in the lexer)? So what exactly is `'\08'`? A non-octal octal escape?

  In ASTExplorer, the AST for `'\08'` (using `babylon7` parser) contains `value: "\u00008"`. So isn't that parsing it as a null escape (or whatever `\0` is called) followed by the character `8`?

2. I'm a little worried that there would be other obscure edge cases like this (that aren't covered by tests) that would make this a breaking change. So would appreciate your insight

But in general this is working

> if it found any newlines it would output the string as a template literal with real newlines

I'd also initially only generated template literal AST when the heredoc included a newline, but ended up always generating template literal AST for heredocs (regardless of whether there's a newline)

I don't remember exactly what triggered me making that change for AST, but it ends up being simpler/more symmetrical to not distinguish the includes-newlines case. Eg otherwise it'd be a little weird (at least from an AST perspective) that if you just added a linebreak to an existing heredoc, its AST would change from a `StringLiteral` to a `TemplateLiteral`

So similarly here it's always generating a template literal for heredocs (regardless of whether it includes a newline or not)


src/nodes.coffee
```diff
@@ -961,16 +961,25 @@ exports.StringLiteral = class StringLiteral extends Literal
       @double
     }
 
+    @unquotedValueForTemplateLiteral = makeDelimitedLiteral val, {
```
helixbass: Again, we get into the territory of actual-string-literal vs text-chunk-of-interpolated-string

I'm not sure that's fully untangled yet but I think this is an improvement - rather than always quoting the value here in `StringLiteral::constructor()` and then undoing that quoting in a couple places (in a mutating way), it creates the different unquoted variants it may need here and then uses those later as needed. I got rid of `StringLiteral::unquote()` entirely

src/nodes.coffee
```diff
@@ -961,16 +961,25 @@ exports.StringLiteral = class StringLiteral extends Literal
...
+    @unquotedValueForJSX = makeDelimitedLiteral val, {
+      @double
+      escapeNewlines: no
+      includeDelimiters: no
+      escapeDelimiter: no
+    }
+
   compileNode: (o) ->
-    return [@makeCode @unquote(yes, yes)] if @jsx
+    return StringWithInterpolations.fromStringLiteral(@).compileNode o if @shouldGenerateTemplateLiteral()
```
helixbass: Given that now we're converting to a `StringWithInterpolations` (in order to generate a template literal/template literal AST) in both `StringLiteral::compileNode()` and `StringLiteral::ast()`, it'd seem cleaner to just convert it to a `StringWithInterpolations` in the constructor of `StringLiteral`

But I *believe* we don't know at construction time yet which "variant" it is (ie actual string literal vs text chunk of template literal vs JSX text)?

src/nodes.coffee
```diff
@@ -2523,7 +2535,7 @@ exports.Arr = class Arr extends Base
       for fragment, fragmentIndex in answer
         if fragment.isHereComment
           fragment.code = "#{multident(fragment.code, o.indent, no)}\n#{o.indent}"
-        else if fragment.code is ', ' and not fragment?.isElision and fragment.type isnt 'StringLiteral'
+        else if fragment.code is ', ' and not fragment?.isElision and fragment.type not in ['StringLiteral', 'StringWithInterpolations']
```
helixbass: Luckily a test happened to catch this weird case where `Arr` is sniffing its fragments and deciding whether a fragment whose text is `', '` is an array comma that it wants to reformat or part of a string

src/nodes.coffee
```diff
@@ -4779,15 +4791,8 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
...
-        unless @jsx
-          # Backticks and `${` inside template literals must be escaped.
-          element.value = element.value.replace /(\\*)(`|\$\{)/g, (match, backslashes, toBeEscaped) ->
-            if backslashes.length % 2 is 0
-              "#{backslashes}\\#{toBeEscaped}"
-            else
-              match
-        fragments.push element.compileToFragments(o)...
+        unquotedElementValue = if @jsx then element.unquotedValueForJSX else element.unquotedValueForTemplateLiteral
+        fragments.push @makeCode unquotedElementValue
```
helixbass: So here when compiling a `StringWithInterpolations`'s text chunks:

Rather than mutating the `StringLiteral`'s `value` and then asking it to compile itself, we're taking full control over its compilation and just plucking its unquoted value that we need

src/nodes.coffee
```diff
@@ -5348,34 +5353,61 @@ unfoldSoak = (o, parent, name) ->
...
-    | \\?(?: (\n)|(\r)|(\u2028)|(\u2029) ) # (Possibly escaped) newlines.
+    #{
+      if escapeDelimiter
+        "|\\\\?(#{delimiterOption})"       # (Possibly escaped) delimiter.
+      else
+        ''
+    }
+    #{
+      if escapeTemplateLiteralCurlies
+        "|\\\\?(\\$\\{)"                   # `${` inside template literals must be escaped.
```
helixbass: Is there a nicer pattern for being able to still use heregex/regex syntax for a conditionally included piece of a heregex (rather than having to do eg double escaping of a string that's going to be included as part of the regex like this)?

src/nodes.coffee
```diff
@@ -5348,34 +5353,61 @@ unfoldSoak = (o, parent, name) ->
...
-    when delimiter then "\\#{delimiter}"
-    when lf        then '\\n'
-    when cr        then '\\r'
-    when ls        then '\\u2028'
-    when ps        then '\\u2029'
-    when other     then (if options.double then "\\#{other}" else other)
-  "#{options.delimiter}#{body}#{options.delimiter}"
+  body = body.replace regex, (match, backslash, nul, ...args) ->
+    delimiter =
+      args.shift() if escapeDelimiter
```
helixbass: I personally like this syntax (as opposed to single-line `delimiter = args.shift() if escapeDelimiter` which means something different) but if it makes you uneasy I could rewrite these as eg
```
delimiter =
  if escapeDelimiter
    args.shift()
  else
    null
```

GeoffreyBooth:
cc @lydell as he’s very knowledgeable about this issue. @lydell, please see the initial comment. Any thoughts on those questions?

lydell:
Octal escapes are not allowed in template literals (TIL). See the end of the [Template Literal Lexical Components](https://tc39.github.io/ecma262/#sec-template-literal-lexical-components):

> A conforming implementation must not use the extended definition of EscapeSequence described in [B.1.2](https://tc39.github.io/ecma262/#sec-additional-syntax-string-literals) when parsing a TemplateCharacter.

And it is the mentioned extension that adds LegacyOctalEscapeSequence.

So we have to either throw errors for them, or compile `\07` to `\u00007`, for example.

"\08" === "\0" + "8"` in non-strict mode. I cannot find _why_ `"\08"` seems to be treated this way in the spec. I also just learned that `"\08"` is invalid in strict mode. So `validateEscapes` (which I wrote back in the day) is wrong.

I think it's correct and practically safe to disallow `\0` followed by another digit everywhere.

I think it makes sense to always compile heredocs to template literals.

src/nodes.coffee
```diff
@@ -5348,34 +5353,61 @@ unfoldSoak = (o, parent, name) ->
...
 
 # Constructs a string or regex by escaping certain characters.
-makeDelimitedLiteral = (body, options = {}) ->
-  body = '(?:)' if body is '' and options.delimiter is '/'
+makeDelimitedLiteral = (body, {delimiter: delimiterOption, escapeNewlines, double, includeDelimiters = yes, escapeDelimiter = yes} = {}) ->
+  body = '(?:)' if body is '' and delimiterOption is '/'
+  escapeTemplateLiteralCurlies = delimiterOption is '`'
   regex = ///
       (\\\\)                               # Escaped backslash.
     | (\\0(?=[1-7]))                       # Null character mistaken as octal escape.
```
lydell: As far as I remember, this code is here to make sure that `"\0#{}7"` compiles to `` `\x007` ``. But id didn't seem to work in [try CoffeeScript](http://coffeescript.org/#try:%22%5C0%23%7B%7D7%22).

Either way, this should probably deal with `\08`, `\08` and `\09` as well, since they are invalid in strict mode and template literals.

src/nodes.coffee
lydell: `/\n/.source` and `/// \n ///.source` I think

lib/coffeescript/nodes.js
```diff
@@ -2732,7 +2740,7 @@
           }
           rite = new Call(rite, this.args);
           rite.isNew = this.isNew;
-          left = new Literal(`typeof ${left.compile(o)} === "function"`);
+          left = new Literal(`typeof ${left.compile(o)} === \"function\"`);
```
helixbass: This is slightly interesting (from `lib/coffeescript/` being recompiled using the heredoc -> template literal change):

Escaped quotes from the original string are now preserved - before, they were getting removed because we escaped quotes when storing the quoted version of the string chunk (as the `StringLiteral`'s `@value`) and then unescaped all quotes (whether originally escaped in the source or escaped by us) when `unquote()`ing the string chunk

This seems fine (I don't think there's any reason we'd need/want to strip unnecessary escapes (that are present in the original source) from the output)?

src/lexer.coffee
```diff
@@ -1264,7 +1264,7 @@ LINE_CONTINUER      = /// ^ \s* (?: , | \??\.(?![.\d]) | \??:: ) ///
 STRING_INVALID_ESCAPE = ///
   ( (?:^|[^\\]) (?:\\\\)* )        # Make sure the escape isn’t escaped.
   \\ (
-     ?: (0[0-7]|[1-7])             # octal escape
+     ?: (0\d|[1-7])                # octal escape
```
helixbass: > As far as I remember, this code is here to make sure that `"\0#{}7"` compiles to `` `\x007` ``. But id didn't seem to work in [try CoffeeScript](http://coffeescript.org/#try:%22%5C0%23%7B%7D7%22).
> 
> Either way, this should probably deal with `\08`, `\08` and `\09` as well, since they are invalid in strict mode and template literals.

Ok so it turns out this was here to guard against things like
```
"\0\
  7"
```
So extended these regexen per your suggestion to guard against `\00`/`\08`/`\09` as well

But your case `"\0#{}7"` is also good (and as you saw, currently generates invalid JS `` `\07` ``)

So my fix for that case was to add a similar guard to convert a trailing null escape (`\0`) to `\x00` in the `unquotedValueForTemplateLiteral` version of a `StringLiteral` (added tests cases)

It looks like we don't need a similar guard for the corresponding JSX case (eg `<div>\0{}8</div>`) since eg `<div>\08</div>` seems to be allowed and have its text content parsed as: a backslash (`\`), a zero (`0`), and an eight (`8`), not an (illegal) octal escape

src/nodes.coffee
```diff
@@ -5348,34 +5354,70 @@ unfoldSoak = (o, parent, name) ->
...
+  escapeTemplateLiteralCurlies = delimiterOption is '`'
   regex = ///
       (\\\\)                               # Escaped backslash.
-    | (\\0(?=[1-7]))                       # Null character mistaken as octal escape.
-    | \\?(#{options.delimiter})            # (Possibly escaped) delimiter.
-    | \\?(?: (\n)|(\r)|(\u2028)|(\u2029) ) # (Possibly escaped) newlines.
+    | (\\0(?=\d))                          # Null character mistaken as octal escape.
+    #{
+      if convertTrailingNullEscapes
+        /// | (\\0) $ ///.source           # Trailing null character that could be mistaken as octal escape.
```
helixbass: > `/\n/.source` and `/// \n ///.source` I think

Nice, updated

test/strict.coffee
```diff
@@ -45,8 +45,7 @@ test "octal escape sequences prohibited", ->
   strict    '"\\\\\\1"'
   strictOk  '"\\0"'
   eq "\x00", "\0"
-  strictOk  '"\\08"'
-  eq "\x008", "\08"
+  strict    '"\\08"'
```
helixbass: I guess rather than "flipping" this test from `strictOk()` to `strict()`, I could just remove it since technically it's not a strict-mode-specific test (it'll always fail to compile regardless of strict mode or not) and is already covered by the tests in `test/error_messages.coffee` above?

lydell: Sounds good to me!

helixbass: Ok removed

GeoffreyBooth:
This all looks great, thanks for a heroic effort @helixbass and for the expert assistance @lydell!

@helixbass do you want to backport the non-AST stuff here to a new PR against `master`? This would be a nice enhancement to the regular compiler too.

helixbass:
@GeoffreyBooth I think it'd be tricky to extract any meaningful part of this to target `master`, as it uses a lot of code/refactors from the `ast` branch eg `StringWithInterpolations::fromStringLiteral()`/`makeDelimitedLiteral()`

We're not that far off from landing `ast`!

PR #5220, opened 2019-07-04
Basic comments AST; PassthroughLiteral AST

helixbass:
@GeoffreyBooth PR for basic AST for comments and for PassthroughLiteral nodes

As discussed, the initial target for comments AST is a single top-level `comments` array on the root `File` AST node (since that is what both Prettier and ESLint expect)

Per our discussion, used a variation of the token-comment-dictionary technique to pass the list of all comments from the tokens -> node class tree

### DIFF
--- a/src/coffeescript.coffee
+++ b/src/coffeescript.coffee
@@ -113,6 +113,7 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
   # which might’ve gotten misaligned from the original source due to the
   # `clean` function in the lexer).
   if options.ast
+    nodes.allCommentsHash = helpers.buildTokenDataDictionary tokens
     sourceCodeNumberOfLines = (code.match(/\r?\n/g) or '').length + 1
     sourceCodeLastLine = /.*$/.exec(code)[0] # `.*` matches all but line break characters.
     ast = nodes.ast options
--- a/src/coffeescript.coffee
+++ b/src/coffeescript.coffee
@@ -113,7 +113,7 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
   # which might’ve gotten misaligned from the original source due to the
   # `clean` function in the lexer).
   if options.ast
-    nodes.allCommentsHash = helpers.buildTokenDataDictionary tokens
+    nodes.allCommentTokens = helpers.extractAllCommentTokens tokens
     sourceCodeNumberOfLines = (code.match(/\r?\n/g) or '').length + 1
     sourceCodeLastLine = /.*$/.exec(code)[0] # `.*` matches all but line break characters.
     ast = nodes.ast options
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -114,6 +114,17 @@ buildLocationData = (first, last) ->
       last.range[1]
     ]
 
+# Build a list of all comments attached to tokens.
+exports.extractAllCommentTokens = (tokens) ->
+  allCommentsObj = {}
+  for token in tokens when token.comments
+    for comment in token.comments
+      commentKey = comment.locationData.range[0]
+      allCommentsObj[commentKey] = comment
+  sortedKeys = Object.keys(allCommentsObj).sort (a, b) -> a - b
+  for key in sortedKeys
+    allCommentsObj[key]
+
 # Get a lookup hash for a token based on its location data.
 # Multiple tokens might have the same location hash, but using exclusive
 # location data distinguishes e.g. zero-length generated tokens from
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -123,9 +123,9 @@ buildLocationHash = (loc) ->
 
 # Build a dictionary of extra token properties organized by tokens’ locations
 # used as lookup hashes.
-buildTokenDataDictionary = (parserState) ->
+exports.buildTokenDataDictionary = buildTokenDataDictionary = (tokens) ->
   tokenData = {}
-  for token in parserState.parser.tokens when token.comments
+  for token in tokens when token.comments
     tokenHash = buildLocationHash token[2]
     # Multiple tokens might have the same location hash, such as the generated
     # `JS` tokens added at the start or end of the token stream to hold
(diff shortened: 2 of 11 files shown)


src/coffeescript.coffee
```diff
@@ -113,6 +113,7 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
   # which might’ve gotten misaligned from the original source due to the
   # `clean` function in the lexer).
   if options.ast
+    nodes.allCommentTokens = helpers.extractAllCommentTokens tokens
```
helixbass: I looked at how we might be able to pass the comment tokens directly within `grammar.coffee` - there's probably a convoluted way to do it involving eg adding a new option to the `o()` helper and only using that helper in the `Root` rules, but this is far simpler to just pass the comments at this "higher level" where we have visibility of both the `tokens` and the resulting `nodes`

I'm not sure whether it's a problem for the `nodes` tree not to be getting generated "purely" inside `grammar.coffee` - I guess the use case where this would be problematic is if someone wanted to manually get a handle on a nodes tree (eg via `CoffeeScript.nodes()` and then call `nodes.ast()` themselves?

If that is a use case that needs to be supported, I think the cleanup of the AST root's location data just below this would already be violating that (since it's creating a difference between what you get if you call `CoffeeScript.nodes()` -> `nodes.ast()` vs just `CoffeeScript.ast()`)

src/helpers.coffee
```diff
@@ -114,6 +114,17 @@ buildLocationData = (first, last) ->
       last.range[1]
     ]
 
+# Build a list of all comments attached to tokens.
+exports.extractAllCommentTokens = (tokens) ->
```
helixbass: I didn't literally reuse `buildTokenDataDictionary()` since it was more straightforward to just directly generate a sorted list of comments (rather than nest them/duplicate them under top-level token-associated keys)

But I guess if it was desirable to not do the additional loop over the tokens array, it might be possible to reuse the single `buildTokenDataDictionary()` data structure and extract the comments from there (depending on where we actually pass the comments to the AST root, per the previous comment)

src/lexer.coffee
```diff
@@ -314,55 +314,90 @@ exports.Lexer = class Lexer
   # everything has been parsed and the JavaScript code generated.
   commentToken: (chunk = @chunk) ->
     return 0 unless match = chunk.match COMMENT
-    [comment, here] = match
+    [withLeadingWhitespace, hereLeadingWhitespace, here, hereTrailingWhitespace, nonHere] = match
```
helixbass: There's a fair amount of cleanup/preservation of comment location data going on here

src/nodes.coffee
```diff
@@ -500,6 +500,15 @@ exports.Root = class Root extends Base
     # end up being declared on the root block.
     o.scope.parameter name for name in o.locals or []
 
+  commentsAst: ->
+    @allComments ?=
+      for commentToken in (@allCommentTokens ? [])
```
helixbass: The `Root` node is expecting to have `@allCommentTokens` set to a list of comment tokens

So when it needs to generate AST for the comments, it first converts the comment tokens to comment node class objects and then per the usual pattern just calls `.ast()` on them

src/nodes.coffee
```diff
@@ -806,7 +815,10 @@ exports.Block = class Block extends Base
     body = []
     for expression in @expressions
       expressionAst = expression.ast o
-      if expression instanceof Directive
+      # Ignore generated PassthroughLiteral
```
helixbass: Since there are "dummy" generated `PassthroughLiteral` nodes that just exist to get comments into the nodes tree, there are a couple places here where we have to explicitly ignore them for purposes of AST generation

src/nodes.coffee
```diff
@@ -4756,13 +4791,18 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
...
-          commentPlaceholder.comments = unwrapped.comments
-          (commentPlaceholder.comments ?= []).push node.comments... if node.comments
-          elements.push new Value commentPlaceholder
+          if o.compiling
+            commentPlaceholder = new StringLiteral('').withLocationDataFrom node
+            commentPlaceholder.comments = unwrapped.comments
+            (commentPlaceholder.comments ?= []).push node.comments... if node.comments
+            elements.push new Value commentPlaceholder
+          else
+            elements.push null
```
helixbass: Coffeescript supports empty interpolations (eg `"#{}"`) while JS does not

So I represented an empty interpolation as a `null` element in the `TemplateLiteral` AST `expressions` list (see examples in empty interpolation AST tests below)

src/helpers.coffee
helixbass: I'm not sure the best way to accurately time the performance, but when I just ran `time bin/cake build:except-parser` a few times with/without this loop included, there wasn't a noticeable difference between the two (there was significantly more variability between individual test runs for a given version than between the two versions, and there were individual passes of the version that included the loop that were faster than individual passes of the version that didn't include the loop)

And this additional loop is only being performed when they're building AST - I agree that it'd be nice to try and optimize the speed of the AST build for the reasons you mentioned. But I guess I'd picture that as a separate compiler-wide optimization pass once we have the AST functionality landed. I'd be curious to dig in and get a sense of how much time is typically spent in each of the compiler phases (eg rewriter passes)

src/lexer.coffee
helixbass: Sure, updated to use `hereComment` and `lineComment` instead (which are much better names, you're correct about what they contain)

src/lexer.coffee
```diff
@@ -314,55 +314,90 @@ exports.Lexer = class Lexer
...
+      placeholderToken = @makeToken 'JS', '', offset: withLeadingWhitespace.length, generated: yes
       placeholderToken.comments = commentAttachments
       @tokens.push placeholderToken
-      @newlineToken 0
+      @newlineToken withLeadingWhitespace.length
     else
       attachCommentsToNode commentAttachments, prev
 
-    comment.length
+    withLeadingWhitespace.length
```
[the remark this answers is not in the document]
helixbass: Sure, updated to `commentWithSurroundingWhitespace`

PR #5224, opened 2019-07-21
AST: OptionalMemberExpression/OptionalCallExpression

helixbass:
@GeoffreyBooth PR updating `Value`/`Call` AST to use Babel 7-style `OptionalMemberExpression`/`OptionalCallExpression` AST node types

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1414,6 +1414,16 @@ exports.Value = class Value extends Base
         mergeLocationData @base.locationData, initialProperties[initialProperties.length - 1].locationData
     object
 
+  containsSoak: ->
+    return no unless @hasProperties()
+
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -1414,6 +1414,16 @@ exports.Value = class Value extends Base
         mergeLocationData @base.locationData, initialProperties[initialProperties.length - 1].locationData
     object
 
+  containsSoak: ->
```
helixbass: Based on playing around on AST Explorer, the idea seems to be that everything "to the right of" the initial soak in a property/call chain should use the `OptionalMemberExpression`/`OptionalCallExpression` AST node type

But then its own `optional` AST field is still `true` or `false` depending on whether *that step* of the chain is soaked or not

PR #5225, opened 2019-07-23
Chained comparison AST

helixbass:
@GeoffreyBooth PR for AST for chained comparisons (eg `a < b < c`)

This construct doesn't exist in JS so I came up with this AST structure (with a peek at Python's AST for chained comparisons):

```
type: 'ChainedComparison'
operators: [
  '<'
  '<'
]
operands: [
  type: 'Identifier'
  name: 'a'
,
  type: 'Identifier'
  name: 'b'
,
  type: 'Identifier'
  name: 'c'
]
```


helixbass:
@GeoffreyBooth updated per your comment

PR #5226, opened 2019-07-23
Super AST

helixbass:
@GeoffreyBooth PR for AST for `Super`/`SuperCall`


src/grammar.coffee
```diff
@@ -394,9 +394,9 @@ grammar =
 
   # A `super`-based expression that can be used as a value.
   Super: [
-    o 'SUPER . Property',                                      -> new Super LOC(3)(new Access $3), [], no, $1
-    o 'SUPER INDEX_START Expression INDEX_END',                -> new Super LOC(3)(new Index $3),  [], no, $1
-    o 'SUPER INDEX_START INDENT Expression OUTDENT INDEX_END', -> new Super LOC(4)(new Index $4),  [], no, $1
+    o 'SUPER . Property',                                      -> new Super LOC(3)(new Access $3), LOC(1)(new Literal $1)
```
helixbass: The location data for the `super` keyword was not getting preserved for cases like `super.a` or `super[b]`

So pass a second argument to the `Super` constructor that preserves its location data (looks like there were some old unused arguments being passed that the `Super` constructor was ignoring)

src/nodes.coffee
```diff
@@ -2062,6 +2062,17 @@ exports.Super = class Super extends Base
     attachCommentsToNode salvagedComments, @accessor.name if salvagedComments
     fragments
 
+  ast: (o, level) ->
+    if @accessor?
+      return (
+        new Value(
```
helixbass: For things like `super.a`, the Babel-style AST is a `MemberExpression` whose `object` is a `Super`

So for those cases, transform the `Super` with an `@accessor` to a `Value` with a property

PR #5228, opened 2019-07-28
Start reconciling ast branch with Prettier plugin

helixbass:
@GeoffreyBooth I started plugging in the current `ast` branch as the Coffeescript dependency for the Prettier plugin and seeing what breaks

So here's a PR with a number of small updates to AST generation:
- making sure the root (`File`/`Program`) AST location data always starts with `0` (eg when there's a leading comment)
- exporting the `IDENTIFIER` regex
- generating a Babel-style `SequenceExpression` AST node for "sequences" eg `(a; b)`
- generating a new `EmptyInterpolation` AST node type for empty interpolations eg `"#{}"`
- adding a `hasIndentedBody` property to function AST nodes (eg `FunctionExpression`)

With these updates, it looks like most if not all of the remaining failing tests in the Prettier plugin (when using this branch as its Coffeescript dependency) are related to the parts of comments we haven't covered yet (eg looking at comment indentation to decide whether it's part of a block or not, heregex comments)

And then there will similarly be updates based on plugging in `ast` as the Coffeescript dependency of the ESLint plugin and seeing what breaks

### DIFF
--- a/src/coffeescript.coffee
+++ b/src/coffeescript.coffee
@@ -117,7 +117,11 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
     sourceCodeNumberOfLines = (code.match(/\r?\n/g) or '').length + 1
     sourceCodeLastLine = /.*$/.exec(code)[0] # `.*` matches all but line break characters.
     ast = nodes.ast options
-    ast.end = ast.range[1] = ast.program.end = ast.program.range[1] = code.length
+    range = [0, code.length]
+    ast.start = ast.program.start = range[0]
+    ast.end = ast.program.end = range[1]
+    ast.range = ast.program.range = range
+    ast.loc.start = ast.program.loc.start = {line: 1, column: 0}
     ast.loc.end.line = ast.program.loc.end.line = sourceCodeNumberOfLines
     ast.loc.end.column = ast.program.loc.end.column = sourceCodeLastLine.length
     return ast
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1187,7 +1187,7 @@ exports.JS_FORBIDDEN = JS_KEYWORDS.concat(RESERVED).concat(STRICT_PROSCRIBED)
 BOM = 65279
 
 # Token matching regexes.
-IDENTIFIER = /// ^
+exports.IDENTIFIER = IDENTIFIER = /// ^
   (?!\d)
   ( (?: (?!\s)[$\w\x7f-\uffff] )+ )
   ( [^\n\S]* : (?!:) )?  # Is this a property name?
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1187,7 +1187,7 @@ exports.JS_FORBIDDEN = JS_KEYWORDS.concat(RESERVED).concat(STRICT_PROSCRIBED)
 BOM = 65279
 
 # Token matching regexes.
-exports.IDENTIFIER = IDENTIFIER = /// ^
+IDENTIFIER = /// ^
   (?!\d)
   ( (?: (?!\s)[$\w\x7f-\uffff] )+ )
   ( [^\n\S]* : (?!:) )?  # Is this a property name?
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -4006,6 +4006,7 @@ exports.Code = class Code extends Base
       # We never generate named functions, so specify `id` as `null`, which
       # matches the Babel AST for anonymous function expressions/arrow functions
       id: null
+      hasIndentedBody: @body.locationData.first_line > @funcGlyph?.locationData.first_line
     ,
       if @isMethod then @methodAstProperties o else {}
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1819,7 +1819,7 @@ exports.JSXElement = class JSXElement extends Base
               emptyExpression.locationData = emptyExpressionLocationData {
                 interpolationNode: element
                 openingBrace: '{'
-                closingBrace: '{'
+                closingBrace: '}'
               }
               
               new JSXExpressionContainer emptyExpression, locationData: element.locationData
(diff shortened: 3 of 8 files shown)


src/coffeescript.coffee
```diff
@@ -117,7 +117,11 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
     sourceCodeNumberOfLines = (code.match(/\r?\n/g) or '').length + 1
     sourceCodeLastLine = /.*$/.exec(code)[0] # `.*` matches all but line break characters.
     ast = nodes.ast options
-    ast.end = ast.range[1] = ast.program.end = ast.program.range[1] = code.length
+    range = [0, code.length]
+    ast.start = ast.program.start = range[0]
```
helixbass: I was seeing that the location data on the root `File`/`Program` nodes was not always starting at `0` when the file started with a comment eg:
```
# comment
"use strict"
```
So similar to what we were already doing to make sure the *end* location data for the root nodes is correct, also make sure their *start* location data is always correct (ie `0`)

src/lexer.coffee
```diff
@@ -1187,7 +1187,7 @@ exports.JS_FORBIDDEN = JS_KEYWORDS.concat(RESERVED).concat(STRICT_PROSCRIBED)
 BOM = 65279
 
 # Token matching regexes.
-IDENTIFIER = /// ^
+exports.IDENTIFIER = IDENTIFIER = /// ^
```
helixbass: Export the `IDENTIFIER` regex to allow the Prettier plugin to import it and use it eg to decide whether it needs to quote object keys or not

src/nodes.coffee
```diff
@@ -799,6 +799,12 @@ exports.Block = class Block extends Base
     return nodes[0] if nodes.length is 1 and nodes[0] instanceof Block
     new Block nodes
 
+  ast: (o, level) ->
+    if (level? and level isnt LEVEL_TOP) and @expressions.length
+      return (new Sequence(@expressions).withLocationDataFrom @).ast o, level
```
helixbass: We represent things like `a; b` as a `Block`

But Babel-style ASTs distinguish these (eg the equivalent `a, b` in JS) as `SequenceExpression`s

So when generating the AST for a `Block`, generate a `SequenceExpression` AST node (using the new `Sequence` node class) if the current `level` isnt `LEVEL_TOP`

src/nodes.coffee
```diff
@@ -1803,30 +1809,19 @@ exports.JSXElement = class JSXElement extends Base
       if content instanceof StringLiteral
         [new JSXText content]
       else # StringWithInterpolations
-        for element in @content.unwrapAll().extractElements o, includeInterpolationWrappers: yes
+        for element in @content.unwrapAll().extractElements o, includeInterpolationWrappers: yes, isJsx: yes
```
helixbass: The changes here and in `StringWithInterpolations` below are related to how we handle AST generation for empty interpolations (either in strings eg `"#{}"` or in JSX expressions eg `<div>{}</div>`)

The main change is that for empty interpolations in strings, instead of generating a `null` AST (in the `expressions` list of the `TemplateLiteral` AST node), we generate a new `EmptyInterpolation` AST node type

`EmptyInterpolation` doesn't exist in Babel-style AST because JS doesn't allow empty interpolations eg `` `${}` `` is a syntax error

So it's Coffeescript-specific. But it mimics `JSXEmptyExpression`, which does exist in Babel-style ASTs (since empty JSX expressions are allowed)

The main reason it's better to have an actual AST node representing the empty interpolation (other than just for consistency with `JSXEmptyExpression`) is that it makes handling comments inside the empty interpolations (eg `"#{###here###}"`) much more straightforward (added AST tests for empty interpolations with/without comments for both strings and JSX expressions)

src/nodes.coffee
```diff
@@ -4006,6 +4001,7 @@ exports.Code = class Code extends Base
       # We never generate named functions, so specify `id` as `null`, which
       # matches the Babel AST for anonymous function expressions/arrow functions
       id: null
+      hasIndentedBody: @body.locationData.first_line > @funcGlyph?.locationData.first_line
```
helixbass: Added explicit `hasIndentedBody` flag to function AST nodes (eg `FunctionExpression`) since otherwise there's no reliable way to detect whether a function has an indented (vs inline) body just based on other existing AST location data

To illustrate that, look at eg:
```
(
  a
) -> b
```
vs:
```
(a) ->
  b
```
Since location data for the `)` and `->` tokens aren't surfaced in the AST, you'd have to try and figure out whether the body is inline by looking at whether the preceding AST element with location data (here, `a`) is on the same line as the beginning of the body's location data (here, `b`)

But in both of these examples, `b` starts on the next line after `a` ends, even though in one case we have an indented function body and in the other we have an inline function body

So it makes sense to explicitly expose that as a flag on the AST since when generating AST we still have access to the location data for `->` (`@funcGlyph`) so can reliably determine it

And it's something that the Prettier plugin wants to be able to determine in order to avoid inlining function bodies that weren't inline in the original source

src/lexer.coffee
GeoffreyBooth: We should probably export all the constants if we export any. Maybe something at the bottom like `exports.lexerConstants = { IDENTIFIER, ... }`

src/nodes.coffee
```diff
@@ -4927,8 +4923,20 @@ exports.StringWithInterpolations = class StringWithInterpolations extends Base
...
-        expressions.push element?.unwrap().ast(o) ? null
+      else # Interpolation
+        {expression} = element
+        node =
+          unless expression?
+            emptyInterpolation = new EmptyInterpolation()
+            emptyInterpolation.locationData = emptyExpressionLocationData {
+              interpolationNode: element
+              openingBrace: '#{'
+              closingBrace: '{'
```
GeoffreyBooth: Not `}`?

src/lexer.coffee
helixbass: Agreed this is sort of arbitrary to just export `IDENTIFIER`

I just updated the Prettier plugin to not import this but instead to just use the same utility that Prettier uses for determining whether an object key is a valid identifier name (and can thus be printed unquoted)

So correspondingly removed the export of `IDENTIFIER` here

src/nodes.coffee
helixbass: Good catch - it happened not to be breaking because it just uses the length of the brace string

Updated to `'}'`

PR #5231, opened 2019-07-31
JSX namespaced name AST

helixbass:
@GeoffreyBooth PR for AST for JSX namespaced names (tag names eg `<div:a />` or attribute names eg `<div a:b="c" />`)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1790,9 +1790,10 @@ exports.JSXElement = class JSXElement extends Base
           jisonLocationDataToAstLocationData @tagName.base.closingTagNameLocationData
         )
       }, @closingElementLocationData
-      if closingElement.name.type is 'JSXMemberExpression'
+      if closingElement.name.type in ['JSXMemberExpression', 'JSXNamespacedName']
         rangeDiff = closingElement.range[0] - openingElement.range[0] + '/'.length
-        shiftAstLocationData = (node) ->
+        columnDiff = closingElement.loc.start.column - openingElement.loc.start.column + '/'.length
+        shiftAstLocationData = (node) =>
           node.range = [
             node.range[0] + rangeDiff
             node.range[1] + rangeDiff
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -1750,9 +1768,15 @@ exports.JSXElement = class JSXElement extends Base
       'JSXElement'
 
   elementAstProperties: (o) ->
+    tagNameAst = =>
+      tag = @tagName.unwrap()
+      if tag?.value and ':' in tag.value
+        tag = new JSXNamespacedName tag
```
helixbass: Here the approach is just to continue to pass the namespaced name (containing `:`) as a single token through the grammar and then "parse" it (create new node class instances corresponding to the namespace prefix and name with correct location data) during AST generation

This avoids needing to add any new grammar rules

src/nodes.coffee
```diff
@@ -1762,31 +1786,36 @@ exports.JSXElement = class JSXElement extends Base
...
           node.range = [
             node.range[0] + rangeDiff
             node.range[1] + rangeDiff
           ]
           node.start += rangeDiff
           node.end += rangeDiff
           node.loc.start =
-            line: node.loc.start.line
-            column: node.loc.start.column + rangeDiff
+            line: @closingElementLocationData.loc.start.line
```
helixbass: There was a bug in this logic for generating adjusted closing element location data, it wasn't calculating closing element location data correctly when the closing element was on a different line than the opening element

So fixed and added AST location data tests for both `JSXMemberExpression` (`<a.b>`) and `JSXNamespacedName` (`<a:b>`) when the closing tag is on a different line than the opening tag

src/nodes.coffee
```diff
@@ -5714,6 +5743,26 @@ zeroWidthLocationDataFromEndLocation = ({range: [, endRange], last_line_exclusiv
   range: [endRange, endRange]
 }
 
+extractSameLineLocationDataFirst = (numChars) -> ({range: [startRange], first_line, first_column}) -> {
```
helixbass: Helpers for extracting the first/last `n` characters of the given location data

PR #5233, opened 2019-08-18
AST comments: include in blocks; heregex comments

helixbass:
@GeoffreyBooth PR for additional comments-related AST support:
- include/exclude comments in block location data based on indent
- exclude heregex comments from top-level AST comments

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -813,7 +813,7 @@ exports.Lexer = class Lexer
 
   # Close up all remaining open blocks at the end of the file.
   closeIndentation: ->
-    @outdentToken moveOut: @indent
+    @outdentToken moveOut: @indent, indentSize: 0
 
   # Match the contents of a delimited token and expand variables and expressions
   # inside it using Ruby-like notation for substitution of arbitrary
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -369,17 +369,28 @@ exports.Lexer = class Lexer
           comment
         .filter (comment) -> comment
 
+    getIndentSize = ({leadingWhitespace, nonInitial}) ->
+      lastNewlineIndex = leadingWhitespace.lastIndexOf '\n'
+      if hereComment? or not nonInitial
+        return null unless lastNewlineIndex > -1
+      else
+        lastNewlineIndex ?= -1
+      leadingWhitespace.length - 1 - lastNewlineIndex
     offsetInChunk = 0
     commentAttachments = for {content, length, leadingWhitespace, precededByBlankLine}, i in contents
       nonInitial = i isnt 0
       leadingNewlineOffset = if nonInitial then 1 else 0
       offsetInChunk += leadingNewlineOffset + leadingWhitespace.length
+      indentSize = getIndentSize {leadingWhitespace, nonInitial}
+      noIndent = not indentSize? or indentSize is -1
       commentAttachment = {
         content
         here: hereComment?
         newLine: leadingNewline or nonInitial # Line comments after the first one start new lines, by definition.
         locationData: @makeLocationData {offsetInChunk, length}
         precededByBlankLine
+        indentSize
+        outdented: not noIndent and indentSize < @indent
       }
       offsetInChunk += length
       commentAttachment
@@ -518,12 +529,12 @@ exports.Lexer = class Lexer
       @error 'missing indentation', offset: offset + indent.length
     else
       @indebt = 0
-      @outdentToken {moveOut: @indent - size, noNewlines, outdentLength: indent.length, offset}
+      @outdentToken {moveOut: @indent - size, noNewlines, outdentLength: indent.length, offset, indentSize: size}
     indent.length
 
   # Record an outdent token or multiple tokens, if we happen to be moving back
   # inwards past several recorded indents. Sets new @indent value.
-  outdentToken: ({moveOut, noNewlines, outdentLength = 0, offset = 0}) ->
+  outdentToken: ({moveOut, noNewlines, outdentLength = 0, offset = 0, indentSize}) ->
     decreasedIndent = @indent - moveOut
     while moveOut > 0
       lastIndent = @indents[@indents.length - 1]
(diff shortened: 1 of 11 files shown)


src/lexer.coffee
```diff
@@ -369,18 +369,31 @@ exports.Lexer = class Lexer
...
       offsetInChunk += leadingNewlineOffset + leadingWhitespace.length
+      indentSize = getIndentSize {leadingWhitespace, nonInitial}
+      noIndent = not indentSize? or indentSize is -1
       commentAttachment = {
         content
         here: hereComment?
         newLine: leadingNewline or nonInitial # Line comments after the first one start new lines, by definition.
         locationData: @makeLocationData {offsetInChunk, length}
         precededByBlankLine
+        indentSize
```
helixbass: We are giving the individual comments "indent-awareness" that can then be used by the rewriter (when determining how to adjust `INDENT`/`OUTDENT` location data)

src/nodes.coffee
```diff
@@ -502,7 +502,7 @@ exports.Root = class Root extends Base
 
   commentsAst: ->
     @allComments ?=
-      for commentToken in (@allCommentTokens ? [])
+      for commentToken in (@allCommentTokens ? []) when not commentToken.heregex
```
helixbass: Exclude heregex comments from the top-level AST comments

They shouldn't be included there since eg Prettier would want to print heregex comments explicitly/separately rather than having them get printed as part of printing the heregex itself

So currently heregex comments aren't explicitly represented in AST (they're just included as part of the raw source pattern string for heregex AST nodes). If we felt like they should get exposed as "parsed" comments, maybe we could give heregex AST nodes their own separate `comments` list (that Prettier would ignore)? But I'd want to think about that a bit more

src/rewriter.coffee
```diff
@@ -548,20 +548,86 @@ exports.Rewriter = class Rewriter
   # `OUTDENT` tokens should always be positioned at the last character of the
   # previous token, so that AST nodes ending in an `OUTDENT` token end up with a
   # location corresponding to the last “real” token under the node.
-  fixOutdentLocationData: ->
+  fixIndentationLocationData: ->
+    @allComments ?= extractAllCommentTokens @tokens
```
helixbass: We have to extract the comment tokens here inside the rewriter as well

Since we're concerned with performance of these additional passes over the tokens (to extract comments), this starts leading me back toward the way I implemented this on my `prettier` branch:

Instead of having to re-extract comments from the tokens, just keep track of the comments separately (in addition to the current behavior of attaching comments to tokens) in the lexer and then explicitly pass that list of comments around (between the lexer and rewriter and between the lexer/rewriter and nodes tree via the return value of the lexer)

Doing it that way we wouldn't have to do any additional passes over the tokens in order to extract comments since the comment list would get build up as we go in the lexer

I'd be inclined to do a separate refactor/optimization PR where we could try this pattern out (or see what else we come up with) if that works for you

src/rewriter.coffee
```diff
@@ -548,20 +548,86 @@ exports.Rewriter = class Rewriter
...
+        yes
+      if first
+        lastMatching = null
+        for comment in @allComments by -1
+          if matches comment
+            lastMatching = comment
+          else if lastMatching
+            return lastMatching
+        return lastMatching
+      for comment in @allComments when matches comment by -1
```
helixbass: There is also an optimization opportunity here when finding the closest preceding comment

Again I'd rather tackle optimization as a separate PR after this if that works for you

src/nodes.coffee
GeoffreyBooth: Comments within block regexes have been a tough case, since JavaScript itself doesn’t support comments within a regex literal. See https://coffeescript.org/#regexes: the comments there are just appended after the regex. Which isn’t ideal, and not very informative, but better (I think) than simply not outputting the comments at all. Can we do the AST equivalent of that?

src/lexer.coffee
```diff
@@ -369,18 +369,31 @@ exports.Lexer = class Lexer
           comment
         .filter (comment) -> comment
 
+    getIndentSize = ({leadingWhitespace, nonInitial}) ->
+      lastNewlineIndex = leadingWhitespace.lastIndexOf '\n'
```
GeoffreyBooth: We normalize line endings somewhere long before this, right? So we don’t need to worry about `\r` or the other variations?

src/rewriter.coffee
GeoffreyBooth: Definitely in favor of optimization as a phase 2 task, as I’m eager to get this out the door. I also wonder just how inefficient this really is. There is the `cake bench` command; how does this branch measure against the `ast` branch (i.e. how much slower does this PR make things)? The difference between those two values is the outer limit as to how much optimization could improve things.

GeoffreyBooth:
The latest commit looks great! Should I merge this or were you planning anything else?

src/lexer.coffee
helixbass: Yes `\r` has already been removed

We will however need to account for those normalizations in order to maintain accurate location data (I believe that there are ESLint plugin tests that will exercise that so will include it in the next AST PR, it's definitely something I added coverage for in my `prettier` branch)

src/nodes.coffee
helixbass: Ok I added support/tests for exposing a `comments` property on heregex AST nodes (interpolated/non-interpolated)

This involved attaching correct location data to heregex comments and then using the typical technique for passing them through the grammar so that they end up explicitly passed into `RegexLiteral`/`RegexWithInterpolations` nodes

src/rewriter.coffee
helixbass: Here are the results from `cake bench`:
`ast` branch:
```
Lex      630 ms (58227 tokens)
Rewrite  683 ms (64842 tokens)
Parse    977 ms
Compile  839 ms (494960 chars)
total   3129 ms
```
This branch:
```
Lex      639 ms (58914 tokens)
Rewrite  847 ms (65549 tokens)
Parse    980 ms
Compile  881 ms (501478 chars)
total   3347 ms
```
So a somewhat slower rewriter run

I believe that longer source files with lots of comments would gain the most from optimizing this rewriter pass

I'd be fine with trying to get in an optimization PR before we land `ast`, or deferring until phase 2

GeoffreyBooth: Up to you. 218 ms difference is I think 6% slower, which seems pretty trivial. I think it could wait until phase 2.

PR #5234, opened 2019-08-18
Reconcile ast branch with Prettier plugin: part 2

helixbass:
@GeoffreyBooth PR for reconciling remaining failing tests in the Prettier plugin when swapping in the `ast` branch as its Coffeescript dependency

Between this PR and #5233, it looks like at that point the `ast` branch could successfully be used as the Coffeescript dependency for the Prettier plugin

So then next step will be to do the same thing for the ESLint plugin

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1084,7 +1084,7 @@ exports.PassthroughLiteral = class PassthroughLiteral extends Literal
 
   astProperties: ->
     return {
-      @value
+      value: @originalValue
       here: !!@here
     }
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -3309,6 +3309,7 @@ exports.Assign = class Assign extends Base
   isAssignable: YES
 
   isStatement: (o) ->
+    return no unless o.compiling
     o?.level is LEVEL_TOP and @context? and (@moduleDeclaration or "?" in @context)
 
   checkAssignability: (o, varBase) ->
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1084,7 +1084,7 @@ exports.PassthroughLiteral = class PassthroughLiteral extends Literal
 
   astProperties: ->
     return {
-      @value
+      value: @originalValue
```
helixbass: Expose the original source version of `PassthroughLiteral` values in the AST as opposed to the "compiled" version that eg unescapes backticks

src/nodes.coffee
```diff
@@ -3054,7 +3054,7 @@ exports.ClassProperty = class ClassProperty extends Base
       key: @name.ast o, LEVEL_LIST
       value: @value.ast o, LEVEL_LIST
       static: !!@isStatic
-      computed: @name instanceof ComputedPropertyName
+      computed: @name instanceof Index or @name instanceof ComputedPropertyName
```
helixbass: There were a couple remaining cases where `computed` was not being set correctly on class methods/properties (added a test)

src/nodes.coffee
```diff
@@ -3639,6 +3639,8 @@ exports.Assign = class Assign extends Base
     # destructured variables.
     @variable.base.propagateLhs yes
 
+  isStatementAst: NO
```
helixbass: Need to override the default `isStatement()` logic for `Assign` so that the AST always treats assignments as non-statements and adds an `ExpressionStatement` wrapper

GeoffreyBooth:
Kind of amazing that this is the last(-ish) PR of the AST effort. I would’ve thought that `PassthroughLiteral`s would’ve been among the first things tackled 😄

PR #5237, opened 2019-10-10
Move sourceMappingURL to last line

mohd-akram:
This is more in line with the [spec](https://sourcemaps.info/spec.html) which says "the generated code may include a line at the end of the source, with the following form: `//# sourceMappingURL=<url>`" and "there exists a `//# sourceURL` comment in the generated code". It also looks better since the `sourceMappingURL` is very long. Also, add a trailing newline. This was motivated by mishoo/UglifyJS2#3441.

### DIFF
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -201,7 +201,7 @@
       encoded = base64encode(JSON.stringify(v3SourceMap));
       sourceMapDataURI = `//# sourceMappingURL=data:application/json;base64,${encoded}`;
       sourceURL = `//# sourceURL=${filename}`;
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
I guess? Would this be a breaking change for anyone?

vendethiel:
Going on a limb here - it's been a few years - but wasn't there some I.E. bug if the last line contained some specific characters that could pose an issue? Maybe we don't need to care, to be fair...

mohd-akram:
Is this good to merge?

mohd-akram:
Any update on this?

GeoffreyBooth:
I don’t know what to do with this. I worry that it could break some people, like bundlers that expect the current output, and I don’t really see why the change is needed.

PR #5242, opened 2019-11-19
Issue 5164: Allow capital E in scientific notation

### LINKED ISSUE #5164: Bug: Scientific notation should allow uppercase E
Choose one: is this a bug report or feature request? This is a Bug report.

### Input Code

```coffee
your (code) => Z = math.complex(1.6E-100)


```

### Expected Behavior
I expected a clean compile.

### Current Behavior
coffee --map --bare --compile --output lib/ ./
C:\Users\hsmyers\Documents\LilyPond\mug\drop.coffee:65:21: error: exponential notation in '1.6E-100' must be indicated
with a lowercase 'e'
Z = math.complex(1.6E-100)
............................................^

### Possible Solution
It would be better to accept either case as this would conform to both JS and normal scientific use patterns. It would also mean one less error message!

use of a case sensitive regex

modify the regex used  to be case insensitive
### Context

slight annoyance

Context is the normal give and take of the code/compile/deploy cycle.

### Environment

* CoffeeScript version: 2.3.2
* Node.js version: v11.2.0

rdeforest:
- Fixes #5164 
- Remove error for capital E from lexer
- Change tests to match


GeoffreyBooth:
The code looks fine, anyone have opinions on whether we should allow this change?

edemaine:
My intuition/recollection is that "E" is more popular than "e" in scientific communities, so I'd lean toward allowing it.  To confirm my feeling, here's some evidence:
* https://en.wikipedia.org/wiki/Scientific_notation#E-notation says "`6.022E23` (or `6.022e23`)", suggesting `E` is more common.
* Various educational websites seem to use "E" exclusively: https://www.mathsisfun.com/numbers/scientific-notation.html, https://chortle.ccsu.edu/Java5/Notes/chap11/ch11_5.html, https://www.nyu.edu/pages/mathmol/textbook/scinot.html

Of course, most other languages also support both `e` and `E` -- for example Python, though Python also supports `0O`, so maybe not the best example.

test/numbers.coffee
```diff
@@ -66,12 +66,15 @@ test "Python-style octal literal notation '0o777'", ->
...
-test "#2060: Disallow uppercase radix prefixes and exponential notation", ->
-  for char in ['b', 'o', 'x', 'e']
+test "#2060: Disallow uppercase radix prefixes", ->
+  for char in ['b', 'o', 'x']
     program = "0#{char}0"
     doesNotThrow -> CoffeeScript.compile program, bare: yes
     throws -> CoffeeScript.compile program.toUpperCase(), bare: yes
 
+test "#5164: Allow uppercase exponent notation", ->
+  doesNotThrow -> CoffeeScript.compile "0E0", bare: yes
```
GeoffreyBooth: Can we add a corresponding test that the lowercase version is also still permitted?

rdeforest: Done

STRd6:
JavaScript allows capital `E` these days: https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#prod-ExponentIndicator

PR #5245, opened 2019-11-28
AST: handle empty file

helixbass:
@GeoffreyBooth I came across the fact that an empty source file was crashing AST generation while working on reconciling the ESLint plugin

So this PR handles AST generation for empty source files

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -854,6 +854,10 @@ exports.Block = class Block extends Base
       body, directives
     }
 
+  astLocationData: ->
+    return if @isRootBlock and not @locationData?
+    super()
+
 # A directive e.g. 'use strict'.
 # Currently only used during AST generation.
 exports.Directive = class Directive extends Base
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -854,6 +854,10 @@ exports.Block = class Block extends Base
       body, directives
     }
 
+  astLocationData: ->
+    return if @isRootBlock and not @locationData?
```
helixbass: When there are no source tokens, the root `Block` ends up not having any `locationData` assigned to it (which was crashing `jisonLocationDataToAstLocationData()`)

So here we just avoid calling the root `astLocationData()` method for this case

And then the AST ends up with the correct location data on the `File` and `Program` AST nodes because of the [existing logic](https://github.com/jashkenas/coffeescript/blob/ast/src/coffeescript.coffee#L121) that sets the AST location data for those root nodes after-the-fact

PR #5247, opened 2019-11-30
AST: declarations

helixbass:
@GeoffreyBooth PR exposing whether variable references (ie `Identifier` AST nodes) represent a declaration of that variable or not

This represents one of the big chunks of reconciling with the ESLint plugin, since it enables the usage of variable declaration/scope-awareness-based ESLint rules like [`no-undef`](https://eslint.org/docs/rules/no-undef)/[`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars), which are probably the single most useful ESLint rules in practice

Based on `ast-with-tokens`, [here](https://github.com/helixbass/copheescript/compare/ast-with-tokens...ast-declaration) is just the diff against that branch

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -931,6 +931,7 @@ exports.InfinityLiteral = class InfinityLiteral extends NumberLiteral
   astProperties: ->
     return
       name: 'Infinity'
+      declaration: no
 
 exports.NaNLiteral = class NaNLiteral extends NumberLiteral
   constructor: ->
@@ -945,6 +946,7 @@ exports.NaNLiteral = class NaNLiteral extends NumberLiteral
   astProperties: ->
     return
       name: 'NaN'
+      declaration: no
 
 exports.StringLiteral = class StringLiteral extends Literal
   constructor: (@originalValue, {@quote, @initialChunk, @finalChunk, @indent, @double, @heregex} = {}) ->
@@ -1113,6 +1115,7 @@ exports.IdentifierLiteral = class IdentifierLiteral extends Literal
   astProperties: ->
     return
       name: @value
+      declaration: !!@isDeclaration
 
 exports.PropertyName = class PropertyName extends Literal
   isAssignable: YES
@@ -1126,6 +1129,7 @@ exports.PropertyName = class PropertyName extends Literal
   astProperties: ->
     return
       name: @value
+      declaration: no
 
 exports.ComputedPropertyName = class ComputedPropertyName extends PropertyName
   compileNode: (o) ->
@@ -1179,6 +1183,7 @@ exports.UndefinedLiteral = class UndefinedLiteral extends Literal
   astProperties: ->
     return
       name: @value
+      declaration: no
 
 exports.NullLiteral = class NullLiteral extends Literal
   constructor: ->
@@ -1199,6 +1204,7 @@ exports.DefaultLiteral = class DefaultLiteral extends Literal
   astProperties: ->
     return
       name: 'default'
+      declaration: no
 
 #### Return
 
@@ -3327,9 +3333,8 @@ exports.Assign = class Assign extends Base
   isStatement: (o) ->
     o?.level is LEVEL_TOP and @context? and (@moduleDeclaration or "?" in @context)
 
-  checkAssignability: (o, varBase) ->
-    if Object::hasOwnProperty.call(o.scope.positions, varBase.value) and
-       o.scope.variables[o.scope.positions[varBase.value]].type is 'import'
+  checkNameAssignability: (o, varBase) ->
+    if o.scope.type(varBase.value) is 'import'
       varBase.error "'#{varBase.value}' is read-only"
 
   assigns: (name) ->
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1113,6 +1115,7 @@ exports.IdentifierLiteral = class IdentifierLiteral extends Literal
   astProperties: ->
     return
       name: @value
+      declaration: !!@isDeclaration
```
helixbass: Any `IdentifierLiteral` node class objects that represent a variable declaration will have had their `isDeclaration` property set before AST generation, so here we expose that under a `declaration` field on `Identifier` AST nodes (this represents an extension to the Babel-style `Identifier` node, since in JS there are separate AST nodes for the variable declarations)

src/nodes.coffee
```diff
@@ -2566,7 +2572,11 @@ exports.ObjectProperty = class ObjectProperty extends Base
     keyAst = @key.ast o, LEVEL_LIST
 
     return
-      key: keyAst
+      key:
+        if keyAst?.declaration
```
helixbass: Avoid having a "double declaration" (on both the `key` and `value`) on left-hand-side shorthand object properties eg
```
{a} = ...
```

src/nodes.coffee
```diff
@@ -2918,9 +2928,15 @@ exports.Class = class Class extends Base
...
 
+  declareName: (o) ->
+    return unless (name = @variable?.unwrap()) instanceof IdentifierLiteral
+    alreadyDeclared = o.scope.find name.value
+    name.isDeclaration = not alreadyDeclared
+
   isStatementAst: -> yes
 
   ast: (o, level) ->
+    @declareName o
```
helixbass: Class names represent declarations (eg `class A`)

src/nodes.coffee
```diff
@@ -3338,6 +3353,48 @@ exports.Assign = class Assign extends Base
...
+        name.isDeclaration = yes
+      else if @param
+        o.scope.add name.value,
+          if @param is 'alwaysDeclare'
+            'var'
+          else
+            'param'
+      else
+        alreadyDeclared = o.scope.find name.value
+        name.isDeclaration ?= not alreadyDeclared
```
helixbass: The typical case, the left hand of an assignment represents a declaration unless it already exists

src/nodes.coffee
```diff
@@ -4733,6 +4757,13 @@ exports.Catch = class Catch extends Base
     [].concat @makeCode(" catch ("), placeholder.compileToFragments(o), @makeCode(") {\n"),
       @recovery.compileToFragments(o, LEVEL_TOP), @makeCode("\n#{@tab}}")
 
+  ast: (o, level) ->
+    @errorVariable?.eachName (name) ->
```
helixbass: A catch clause error variable represents a declaration

src/nodes.coffee
```diff
@@ -5169,6 +5200,14 @@ exports.For = class For extends While
     fragments.push @makeCode(returnResult) if returnResult
     fragments
 
+  ast: (o, level) ->
+    addToScope = (name) ->
```
helixbass: The loop variables of a `for` represent variable declarations (if they don't already exist)

This helper could probably be used to DRY up some of these other places where we're setting `isDeclaration` since it encapsulates the basic pattern of possibly adding to the scope (via `o.scope.find()`) and then setting `isDeclaration` based on whether it already existed in the scope

src/nodes.coffee
```diff
@@ -3327,9 +3343,8 @@ exports.Assign = class Assign extends Base
   isStatement: (o) ->
     o?.level is LEVEL_TOP and @context? and (@moduleDeclaration or "?" in @context)
 
-  checkAssignability: (o, varBase) ->
-    if Object::hasOwnProperty.call(o.scope.positions, varBase.value) and
-       o.scope.variables[o.scope.positions[varBase.value]].type is 'import'
```
GeoffreyBooth: Did we lose this `checkAssignability` method?

helixbass: Nope this diff is a combination of renaming `checkAssignability` -> `checkNameAssignability` (in order to disambiguate naming-wise from the added `checkAssignability` option below) as well as simplifying the scope checking by using `o.scope.type()`

GeoffreyBooth:
> > I feel like this is something we should write down somewhere, like in a comment near the start of the AST code, as it’s a pretty fundamental thing to remember for anyone working on the AST code in the future (unless you think it’s obvious)
> 
> Ya I think that's a good idea. It might make sense to try and include it as part of a more full-fledged "guide to working on the AST" that explains the internal API used by node classes etc
> 
> We could probably also bake some of those assumptions into our AST testing helpers for some automated testing that we're not violating them (eg check that all child AST nodes have location data within the parent's location data, check that all sibling nodes have non-overlapping location data)

Do you mind adding some of these to your AST to-do list? Maybe the place for the guide is the GitHub wiki pages, like [this one](https://github.com/jashkenas/coffeescript/wiki/%5BHowTo%5D-How-parsing-works).

PR #5248, opened 2019-12-01
Compound assign location data

helixbass:
@GeoffreyBooth this PR causes correct location data to be generated for "compound assign" (eg `or=`) tokens

This issue was surfaced by ESLint rules that inspect the location data of these tokens not behaving correctly against the `ast` branch eg [`space-infix-ops`](https://eslint.org/docs/rules/space-infix-ops)

Based on `ast-declaration`, [here](https://github.com/helixbass/copheescript/compare/ast-declaration...compound-assign-location-data) is just the diff against that branch

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -740,6 +740,12 @@ exports.Lexer = class Lexer
         prev[0] = 'COMPOUND_ASSIGN'
         prev[1] += '='
         prev.data.original += '=' if prev.data?.original
+        prev[2].range = [
+          prev[2].range[0]
+          prev[2].range[1] + 1
+        ]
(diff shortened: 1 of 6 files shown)


src/lexer.coffee
```diff
@@ -740,6 +740,12 @@ exports.Lexer = class Lexer
         prev[0] = 'COMPOUND_ASSIGN'
         prev[1] += '='
         prev.data.original += '=' if prev.data?.original
+        prev[2].range = [
```
helixbass: The existing logic failed to expand the location data of the existing eg `or` token when appending the `=` to it (in the case of `or=`)

PR #5249, opened 2019-12-01
Location data: compensate for stripped characters

helixbass:
@GeoffreyBooth PR making location data correct when "stripped" characters (BOM, carriage returns) are included in the source

Based on `compound-assign-location-data`, [here](https://github.com/helixbass/copheescript/compare/compound-assign-location-data...location-data-compensations) is just the diff against that branch

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1055,7 +1055,13 @@ exports.Lexer = class Lexer
     if lineCount > 0
       [..., lastLine] = string.split '\n'
       column = lastLine.length
-      columnCompensation = @getLocationDataCompensation @chunkOffset + offset - column, @chunkOffset + offset
+      previousLinesCompensation = @getLocationDataCompensation @chunkOffset, @chunkOffset + offset - column
+      # Don't recompensate for initially inserted newline.
+      previousLinesCompensation = 0 if previousLinesCompensation < 0
+      columnCompensation = @getLocationDataCompensation(
+        @chunkOffset + offset + previousLinesCompensation - column
+        @chunkOffset + offset + previousLinesCompensation
+      )
     else
       column += string.length
       columnCompensation = compensation
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -58,6 +58,8 @@ exports.Lexer = class Lexer
       opts.column or 0           # The start column of the current @chunk.
     @chunkOffset =
       opts.offset or 0           # The start offset for the current @chunk.
+    @locationDataCompensations =
+      opts.locationDataCompensations or {} # The location data compensations for the current @chunk.
     code = @clean code           # The stripped, cleaned original source code.
 
     # At every position, run through this list of attempted matches,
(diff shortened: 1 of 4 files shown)


src/lexer.coffee
```diff
@@ -93,11 +95,21 @@ exports.Lexer = class Lexer
   # returns, etc. If we’re lexing literate CoffeeScript, strip external Markdown
   # by removing all lines that aren’t indented by at least four spaces or a tab.
   clean: (code) ->
-    code = code.slice(1) if code.charCodeAt(0) is BOM
-    code = code.replace(/\r/g, '').replace TRAILING_SPACES, ''
+    thusFar = 0
+    if code.charCodeAt(0) is BOM
+      code = code.slice 1
+      @locationDataCompensations[0] = 1
```
helixbass: The basic idea is that we have to record the "compensations" for where we stripped out characters at certain offsets in the source

So here record a compensation for an initial BOM

src/lexer.coffee
```diff
@@ -93,11 +95,21 @@ exports.Lexer = class Lexer
...
+      thusFar += 1
     if WHITESPACE.test code
       code = "\n#{code}"
       @chunkLine--
+      @locationDataCompensations[0] ?= 0
+      @locationDataCompensations[0] -= 1
```
helixbass: In this case (of an initial indent) we prepended a newline to the source so need to record a *negative* compensation

src/lexer.coffee
```diff
@@ -93,11 +95,21 @@ exports.Lexer = class Lexer
...
+    code = code
+      .replace /\r/g, (match, offset) =>
+        @locationDataCompensations[thusFar + offset] = 1
```
helixbass: And here we record compensations for stripped carriage returns

src/lexer.coffee
```diff
@@ -1005,12 +1023,26 @@ exports.Lexer = class Lexer
   # Helpers
   # -------
 
+  # Compensate for the things we strip out initially (e.g. carriage returns)
+  # so that location data stays accurate with respect to the original source file.
+  getLocationDataCompensation: (start, end) ->
```
helixbass: Then here is a helper to calculate how much total compensation (ie adjustment) should be made for a given offset range

src/lexer.coffee
```diff
@@ -1023,10 +1055,18 @@ exports.Lexer = class Lexer
     if lineCount > 0
       [..., lastLine] = string.split '\n'
       column = lastLine.length
+      previousLinesCompensation = @getLocationDataCompensation @chunkOffset, @chunkOffset + offset - column
+      # Don't recompensate for initially inserted newline.
+      previousLinesCompensation = 0 if previousLinesCompensation < 0
+      columnCompensation = @getLocationDataCompensation(
```
helixbass: This gets a little hairy algorithmically but seems to be the correct behavior - if we're calculating the compensation for a column number of a chunk that includes a newline, calculate the compensation for just the part after the last newline, adjusting for the compensations in the part of the chunk *before* the last newline

GeoffreyBooth:
Whoa. This is heroic work. I’ve always wondered why we still seemed to have bugs in location data, maybe this explains it.

Please resolve the conflict then this looks good to merge.

helixbass:
Ok rebased onto `ast` :+1:

PR #5250, opened 2019-12-03
AST: implicit returns

helixbass:
@GeoffreyBooth PR exposing implicit returns in the AST

This enables ESLint rules like [`no-unused-expressions`](https://eslint.org/docs/rules/no-unused-expressions) to work since it allows distinguishing expressions that are implicitly returned as being "used"

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2962,6 +2962,7 @@ exports.Class = class Class extends Base
     @body.locationData = zeroWidthLocationDataFromEndLocation @locationData if @hasGeneratedBody
     @walkBody o
     sniffDirectives @body.expressions
+    @ctor?.noReturn = yes
 
     super o
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -235,10 +235,13 @@ exports.Base = class Base
   # Construct a node that returns the current node's result.
   # Note that this is overridden for smarter behavior for
   # many statement nodes (e.g. If, For)...
-  makeReturn: (res) ->
+  makeReturn: ({accumulator, mark} = {}) ->
+    if mark
+      @canBeReturned = yes
+      return
     me = @unwrapAll()
-    if res
-      new Call new Literal("#{res}.push"), [me]
+    if accumulator
+      new Call new Literal("#{accumulator}.push"), [me]
     else
       new Return me
 
@@ -273,6 +276,13 @@ exports.Base = class Base
   ast: (o, level) ->
     o = Object.assign {}, o
     o.level = level if level?
+    @makeReturn mark: yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
+
+    fullAst = @astFull o
+    return fullAst unless fullAst?
+    Object.assign fullAst, @astReturns()
+
+  astFull: (o) ->
     # Every abstract syntax tree node object has four categories of properties:
     # - type, stored in the `type` field and a string like `NumberLiteral`.
     # - location data, stored in the `loc`, `start`, `end` and `range` fields.
@@ -293,6 +303,13 @@ exports.Base = class Base
   # mutated into the structure that the Babel spec uses.
   astLocationData: ->
     jisonLocationDataToAstLocationData @locationData
+
+  # Mark AST nodes that correspond to expressions that (implicitly) return.
+  astReturns: ->
+    if @canBeReturned
+      returns: yes
+    else
+      {}
 
   # Determines whether an AST node needs an `ExpressionStatement` wrapper.
   # Typically matches our `isStatement()` logic but this allows overriding.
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -273,6 +276,13 @@ exports.Base = class Base
   ast: (o, level) ->
     o = Object.assign {}, o
     o.level = level if level?
+    @makeReturn mark: yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
```
helixbass: I didn't see a way around having to decouple the "call to create the AST for a node" method (`.ast()`) from the "override the default behavior for creating an AST node" method (formerly also `.ast()`, now the newly added `.astFull()`)

Because there needs to be a centralized place for triggering this marking of implicit returns - it doesn't seem like a good idea to try and force every node class that wants to override AST generation to separately invoke the implicit-return-marking behavior somehow

So the expected return value for `astFull()` is what was previously expected for node classes that overrode `ast()`: it should return the fully constructed AST node including location data (it just won't include the additional `returns` property flagging implicit returns, that gets managed by the invoking `Base::ast()` method here)

test/abstract_syntax_tree.coffee
```diff
@@ -2727,6 +2731,21 @@ test "AST as expected for While node", ->
...
+    type: 'AssignmentExpression'
+    right:
+      type: 'WhileStatement'
+      body:
+        type: 'BlockStatement'
+        body: [
+          type: 'ExpressionStatement'
+          expression:
+            type: 'CallExpression'
+            returns: yes
```
helixbass: The "return value" of accumulated `while`/`for` loops should be marked as `returns: yes`

src/nodes.coffee
GeoffreyBooth: Help me understand, why do we need a separate `astFull`? Are we trying to optimize returning the AST data for the node and skipping creating the `return` string? Could we be less optimized and have simpler code?

helixbass: If we just had `.ast()` (no `.astFull()`), then any node class that needed to override AST generation logic (the existing node classes that override `.ast()`/now `.astFull()`) would have to individually invoke the correct logic (that lives in `Base::ast()`) for marking implicit returns and adding `returns` to the generated AST node

It seems much simpler to just have that logic live in one place (`Base::ast()`), but that means that we can't have individual node classes overriding `.ast()` (without calling `super()`)

So to me this makes the most sense, I think you'd consider this a [template method](https://en.wikipedia.org/wiki/Template_method_pattern) (subclasses just override parts of the logic rather than the entire base method)

helixbass: > Are we trying to optimize returning the AST data for the node and skipping creating the `return` string?

No more the opposite, making sure that the `returns` logic gets invoked for all node classes (even ones that want to override the core logic of AST generation)

GeoffreyBooth: > So to me this makes the most sense, I think you'd consider this a [template method](https://en.wikipedia.org/wiki/Template_method_pattern) (subclasses just override parts of the logic rather than the entire base method)

I think that's what we had already with `astType`, `astProperties` etc. So `ast` is supposed to be the root one, that encapsulates all the others. Like if you look at its initial body:

```coffee
Object.assign {}, {type: @astType(o)}, @astProperties(o), @astLocationData()
```

I think it's confusing to have some differently named method be that 'root', and then a method named `ast`. How about the root (what you were proposing calling `astFull`) stays as `ast`, but we create new child methods `astNode` and `astReturns`? And then the core stuff about a node goes in `astNode` (what you were proposing as `ast`), and whether or not it does something unusual about returning goes in `astReturns`. Most nodes won't override `astReturns`, and few if any nodes would override `ast`, but if they did, it would be on them to reassemble the entire object (or call `super`).

helixbass: Ok sure, updated so that `astNode()` is what some node classes override (and `ast()` is the "root")

GeoffreyBooth: I'm looking at this again. What I was proposing as `astNode` we seem to already have in `astProperties`. Why do we need a separate method for the core properties of a node?

Can't we instead have a base method of `astReturns` that itself only gets overridden on the ones that do unusual stuff regarding returning? And everything else (`ast`, `astProperties`, etc.) stays the same.

GeoffreyBooth:
Can we please resolve this / https://github.com/helixbass/copheescript/pull/17 before we move on to further PRs? Ideally as a PR for returns and a separate PR for refactoring `ast` / `astNode` if that works for you?

helixbass:
Sure I didn't get notified of your replies on https://github.com/helixbass/copheescript/pull/17 for some reason

Will reply there to hash out how to proceed

helixbass:
Closing in favor of #5254 + #5255

PR #5251, opened 2019-12-03
AST: recognize quoted constructor

helixbass:
@GeoffreyBooth PR to correctly recognize a method named `"constructor"` (in quotes) as a constructor

This was causing a test for the ESLint rule [`no-useless-constructor`](https://eslint.org/docs/rules/no-useless-constructor) to fail

Based on `ast-mark-implicit-returns`, [here](https://github.com/helixbass/copheescript/compare/ast-mark-implicit-returns...ast-quoted-constructor) is just the diff against that branch


GeoffreyBooth:
This is fine once the other merges in.

PR #5252, opened 2019-12-05
AST: add params to scope

helixbass:
@GeoffreyBooth PR ensuring that function params get declared on the scope during AST generation. Basically in order for `declaration` (#5247) to always be accurate, everything needs to get added to the scope

Based on `ast-quoted-constructor`, [here](https://github.com/helixbass/copheescript/compare/ast-quoted-constructor...ast-add-params-to-scope) is just the diff against that branch


GeoffreyBooth:
👍 once the upstream PR merges in.

PR #5253, opened 2019-12-05
AST: object colon location data

helixbass:
@GeoffreyBooth PR ensuring the correct location data gets assigned to `:` tokens in object literals

This was causing tests to fail for ESLint rules like [`key-spacing`](https://eslint.org/docs/rules/key-spacing)

Based on `ast-add-params-to-scope`, [here](https://github.com/helixbass/copheescript/compare/ast-add-params-to-scope...ast-object-colon-location-data) is just the diff against that branch

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -246,7 +246,7 @@ exports.Lexer = class Lexer
         [poppedToken[2].first_line, poppedToken[2].first_column, poppedToken[2].range[0]]
     if colon
       colonOffset = input.lastIndexOf if inJSXTag then '=' else ':'
-      colonToken = @token ':', ':', offset: colonOffset, length: colon.length
+      colonToken = @token ':', ':', offset: colonOffset
       colonToken.jsxColon = yes if inJSXTag # used by rewriter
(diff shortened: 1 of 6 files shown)


src/lexer.coffee
```diff
@@ -246,7 +246,7 @@ exports.Lexer = class Lexer
         [poppedToken[2].first_line, poppedToken[2].first_column, poppedToken[2].range[0]]
     if colon
       colonOffset = input.lastIndexOf if inJSXTag then '=' else ':'
-      colonToken = @token ':', ':', offset: colonOffset, length: colon.length
+      colonToken = @token ':', ':', offset: colonOffset
```
helixbass: This was incorrectly assigning location data of eg length 2 to object colon tokens that were preceded by spaces (eg `a : b`)

GeoffreyBooth:
Good catch! We could just make this a PR against `master`, there’s nothing AST-related in here.

helixbass:
Given that `ast` will be landing soon in `master` and that I've been making these branches based one on the next in order to more easily have a single existing branch to temporarily use for testing the ESLint plugin, it's much simpler from my end to keep this targeted as-is

PR #5254, opened 2019-12-06
AST: Support marking implicit returns

helixbass:
@GeoffreyBooth here's the "first PR" for supporting marking implicit returns in the AST

It just includes the refactor to `makeReturn()` to accept a second `mark` argument (should match the version from your branch)

The next PR will include actually invoking the new `mark` functionality during AST generation

### DIFF
--- a/lib/coffeescript/nodes.js
+++ b/lib/coffeescript/nodes.js
@@ -67,13 +67,13 @@
   fragmentsToText = function(fragments) {
     var fragment;
     return ((function() {
-      var j, len1, results;
-      results = [];
+      var j, len1, results1;
+      results1 = [];
       for (j = 0, len1 = fragments.length; j < len1; j++) {
         fragment = fragments[j];
-        results.push(fragment.code);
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
Thanks, this looks great.

Another thought: is a node’s constructor called when `ast()` is called? Could we put the “AST initialization” code in there, if we wanted to? Presumably we would then need something similar to “mark” that would let us know that we’re in AST-generation mode rather than regular compile mode, but that could be something attached to `o`.

helixbass:
> Another thought: is a node’s constructor called when `ast()` is called? Could we put the “AST initialization” code in there, if we wanted to? Presumably we would then need something similar to “mark” that would let us know that we’re in AST-generation mode rather than regular compile mode, but that could be something attached to `o`.

I'm intrigued by this idea of leveraging constructors, but I'm not sure exactly what the pattern would be - the node class constructors get called at parsing time (ie by the grammar) but I don't think `o` exists/is exposed (yet) in that context?

PR #5255, opened 2019-12-06
AST: mark implicit returns

helixbass:
@GeoffreyBooth here's the "second PR" for marking implicit returns in the AST

It includes the refactor to use `astNode()`

Based on `support-marking-implicit-returns`, [here](https://github.com/helixbass/copheescript/compare/support-marking-implicit-returns...ast-mark-implicit-returns) is just the diff against that branch

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -279,7 +279,8 @@ exports.Base = class Base
   # Only override the component `ast*` methods as needed.
   ast: (o, level) ->
     o = @astInitialize o, level
-    @astAddReturns @astNode o
+    astNode = @astNode o
+    @astAddReturns astNode
 
   astInitialize: (o, level) ->
     o = Object.assign {}, o
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -523,7 +523,7 @@ exports.Root = class Root extends Base
           new LineComment commentToken
     comment.ast() for comment in @allComments
 
-  ast: (o) ->
+  astNode: (o) ->
     o.level = LEVEL_TOP
     @initializeScope o
     super o
@@ -938,10 +938,10 @@ exports.InfinityLiteral = class InfinityLiteral extends NumberLiteral
   compileNode: ->
     [@makeCode '2e308']
 
-  ast: (o, level) ->
+  astNode: (o) ->
     unless @originalValue is 'Infinity'
-      return new NumberLiteral(@value).withLocationDataFrom(@).ast o, level
-    super o, level
+      return new NumberLiteral(@value).withLocationDataFrom(@).ast o
+    super o
 
   astType: -> 'Identifier'
 
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -279,11 +279,7 @@ exports.Base = class Base
   # Only override the component `ast*` methods as needed.
   ast: (o, level) ->
     o = @astInitialize o, level
-    ast = @astNode o
-    return ast unless ast?
-    # Mark AST nodes that correspond to expressions that (implicitly) return.
-    ast.returns = yes if @canBeReturned
-    ast
+    @astAddReturns @astNode o
 
   astInitialize: (o, level) ->
     o = Object.assign {}, o
@@ -315,6 +311,12 @@ exports.Base = class Base
   # mutated into the structure that the Babel spec uses.
   astLocationData: ->
     jisonLocationDataToAstLocationData @locationData
+
+  # Mark AST nodes that correspond to expressions that (implicitly) return.
+  astAddReturns: (ast) ->
+    return ast unless ast?
+    ast.returns = yes if @canBeReturned
+    ast
 
   # Determines whether an AST node needs an `ExpressionStatement` wrapper.
   # Typically matches our `isStatement()` logic but this allows overriding.
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -278,18 +278,21 @@ exports.Base = class Base
   # **WARNING: DO NOT OVERRIDE THIS METHOD IN CHILD CLASSES.**
   # Only override the component `ast*` methods as needed.
   ast: (o, level) ->
+    o = @astInitialize o, level
+    ast = @astNode o
+    return ast unless ast?
+    # Mark AST nodes that correspond to expressions that (implicitly) return.
+    ast.returns = yes if @canBeReturned
+    ast
+
+  astInitialize: (o, level) ->
     o = Object.assign {}, o
     o.level = level if level?
     # `@makeReturn` must be called before `astProperties`, because the latter may call
     # `.ast()` for child nodes and those nodes would need the return logic from `makeReturn`
     # already executed by then.
     @makeReturn null, yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
-
-    ast = @astNode o
-    return ast unless ast?
-    # Mark AST nodes that correspond to expressions that (implicitly) return.
-    ast.returns = yes if @canBeReturned
-    ast
+    o
 
   astNode: (o) ->
     # Every abstract syntax tree node object has four categories of properties:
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -273,6 +278,15 @@ exports.Base = class Base
   ast: (o, level) ->
```
helixbass: I went through and updated to make sure that no node classes override `ast()`

src/nodes.coffee
```diff
@@ -273,6 +278,15 @@ exports.Base = class Base
...
     o = Object.assign {}, o
     o.level = level if level?
+    @makeReturn null, yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
```
GeoffreyBooth:
```suggestion
    # `@makeReturn` must be called before `astProperties`, because the latter may call
    # `.ast()` for child nodes and those nodes would need the return logic from `makeReturn`
    # already executed by then.
    @makeReturn null, yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
```

src/nodes.coffee
```diff
@@ -273,6 +278,15 @@ exports.Base = class Base
...
+
+    ast = @astNode o
+    return ast unless ast?
+    # Mark AST nodes that correspond to expressions that (implicitly) return.
+    ast.returns = yes if @canBeReturned
+    ast
```
GeoffreyBooth: This is only for organization/clarity; what do you think of this?

Something like:

```coffee
  ast: (o, level) ->
    o = astInitialize o, level
    astNode o # Calls `astReturn` to return final assembled object.

  # The “constructor” for AST nodes, this logic must be run as part of creating any
  # AST node.
  astInitialize: (o, level) ->
    o = Object.assign {}, o
    o.level = level if level?
    @makeReturn null, yes if @isStatement(o) and o.level isnt LEVEL_TOP and o.scope?
    o

  # If an AST node needs any unusual logic that doesn’t fit within the component
  # methods `astType` or `astProperties`, such as using a different AST class
  # entirely, that can go in `astNode`. If you override this, make sure to call
  # `@astReturn o` at the end of your method.
  astNode: (o) -> astReturn o

  # Assemble the serializable JSON-like AST object to be returned.
  astReturn: (o) ->
    # Mark AST nodes that correspond to expressions that (implicitly) return.
    returns = if @canBeReturned then {returns: yes} else {}

    # Every abstract syntax tree node object has four categories of properties:
    # - type, stored in the `type` field and a string like `NumberLiteral`.
    # - location data, stored in the `loc`, `start`, `end` and `range` fields.
    # - properties specific to this node, like `parsedValue`.
    # - properties that are themselves child nodes, like `body`.
    # These fields are all intermixed in the Babel spec; `type` and `start` and
    # `parsedValue` are all top level fields in the AST node object. We have
    # separate methods for returning each category, that we merge together here.
    Object.assign {}, {type: @astType(o)}, returns, @astProperties(o), @astLocationData()
```

helixbass: Ok I updated to use `astInitialize()`

The issue with nesting `astReturn()` inside `astNode()` is that then node classes that override `astNode()` wouldn't get the `returns` field added automatically - I think the adding of `returns` as the last step of assembling the AST node should remain a responsibility of `ast()` rather than `astNode()` so that it can stay centralized in a single place rather than being something that then every node class that overrides `astNode()` would have to separately know how to do

But so I updated to make the adding of `returns` a little more explicit by encapsulating in an `astAddReturns()` method (that gets applied as the last step of `Base::ast()`)

src/nodes.coffee
```diff
@@ -299,6 +313,12 @@ exports.Base = class Base
   astLocationData: ->
     jisonLocationDataToAstLocationData @locationData
 
+  # Mark AST nodes that correspond to expressions that (implicitly) return.
+  astAddReturns: (ast) ->
+    return ast unless ast?
```
GeoffreyBooth: When is `ast` falsy?

src/nodes.coffee
```diff
@@ -275,9 +275,23 @@ exports.Base = class Base
   # as JSON. This is what the `ast` option in the Node API returns.
   # We try to follow the [Babel AST spec](https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md)
   # as closely as possible, for improved interoperability with other tools.
+  # **WARNING: DO NOT OVERRIDE THIS METHOD IN CHILD CLASSES.**
+  # Only override the component `ast*` methods as needed.
   ast: (o, level) ->
+    o = @astInitialize o, level
+    astNode = @astNode o
+    @astAddReturns astNode
```
GeoffreyBooth: Why are we adding the returns _after_ assembling the AST object to be returned? In my mind, the `returns` key on the returned object is just one of the properties like `type` or all the other keys added by `astProperties` and `astLocationData`, hence that’s why I had it inserted as part of the `Object.assign` call.

Or in other words, why not refactor `astAddReturns` to return an object:

```coffee
astReturns: ->
  if @canBeReturned then {returns: yes} else {}
```

And then the `Object.assign` call just includes it:

```coffee
Object.assign {}, {type: @astType(o)}, @astProperties(o), @astReturns(), @astLocationData()
```

The only reason I can think of not to do it this way is if we somehow have to check that `ast` might not exist before we create an object that’s _only_ `{returns: true}`; my version leaves out your `unless ast?` check. But when/why is that check needed? I would think that `ast` would never return falsy, as it would always return an object with at least a `type` field or location data?

helixbass: Yes, `returns` is like `type` or the other keys

Like I said, the reason not to include it in the `Object.assign()` call is that the `Object.assign()` call lives inside `astNode()` (because that's conceptually what `astNode()` is overriding: the "core" AST generation for the node). So then node classes that override `astNode()` (and don't call `super()`) would have to all individually know that they need to invoke `astReturns()`. There's no need for that. As it stands, `returns` is much easier handled as a "special" key that only `Base::ast()` needs to understand and tags onto the assembled core AST node returned by `astNode()`

This was in fact the core reason for the refactor to `astNode()`, to be able to centralize the `makeReturn()`/`returns` logic in one place (`Base::ast()`) without having to worry about "delegating awareness" of it to individual node classes unnecessarily just because they happen to want to override `astNode()` for whatever reason

src/nodes.coffee
helixbass: Eg `Elision` and `PassthroughLiteral` can both return `null` for their AST

src/nodes.coffee
GeoffreyBooth: I still find this confusing, and think we can do better; but in the interest of time I’ll merge this in (and the PRs waiting on it) and propose my alternative, if I can think of something better, as a new PR.

helixbass: Ok thanks, I'm happy to continue the discussion

My impression is that you like the "symmetry" of having the single `Object.assign()` call to assemble the entire AST node. Which makes sense

But I'd argue that the typical use case in terms of future development against this API is that people might be adding new node classes and needing to support AST generation for those node classes. For those "consumers" of this API, I think it's cleaner/less confusing for them to not have to be aware of `returns` at all - their responsibility is to return the "standard AST" from `astNode()` (whether by overriding `astNode()` or just specifying `astType()` + `astProperties()`). And then `returns` is sort of a special case that is better governed by a single place in the code that has the awareness of how it works/that it exists

So basically I think that yes it's slightly more complex how the *provider* of the API is implemented (ie how `Base::ast()` works) but is less complex from the perspective of the *consumers* of the API - and that all the node classes except `Base` are just consumers of the API and we're giving them a nice clean interface for generating their AST nodes

GeoffreyBooth: > For those "consumers" of this API, I think it's cleaner/less confusing for them to not have to be aware of `returns` at all

I agree, but you could say the same thing about e.g. location data; most nodes shouldn’t care about it. I want returns to be at that level, where if you need/want to care about it you override the `astReturns` method, and if not, you don’t. _And_ the symmetry of one place where the object is assembled, yes, as that (to me) helps make clear that what we’re doing here is assembling and returning serializable objects, just as `compile`/`compileNode` assembles and returns a JS source code string.

PR #5256, opened 2019-12-07
AST: token cleanups

helixbass:
@GeoffreyBooth PR marking tokens as generated and other similar token-accuracy cleanups

This reduces the # of failing tests in the ESLint plugin from 300+ to 8. So we're making progress :+1:

Based on `ast-object-colon-location-data`, [here](https://github.com/helixbass/copheescript/compare/ast-object-colon-location-data...ast-generated-jsx-tokens) is just the diff against that branch

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -666,6 +666,10 @@ exports.Lexer = class Lexer
         return 2
       else if firstChar is '{'
         if prevChar is ':'
+          # This token represents the start of a JSX attribute value
+          # that’s an expression (e.g. the `{b}` in `<div a={b} />`).
+          # Our grammar represents the beginnings of expressions as `(`
+          # tokens, so make this into a `(` token that displays as `{`.
           token = @token '(', '{'
           @jsxObjAttribute[@jsxDepth] = no
           # tag attribute name as JSX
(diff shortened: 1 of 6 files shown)


src/lexer.coffee
```diff
@@ -666,8 +666,11 @@ exports.Lexer = class Lexer
         return 2
       else if firstChar is '{'
         if prevChar is ':'
-          token = @token '(', '('
+          token = @token '(', '{'
```
GeoffreyBooth: Can you please explain what’s going on here? And if the `(` version originally wasn’t just a mistake, then maybe should we add an explanation in a comment?

helixbass: This case appears to correspond to a JSX attribute value that's an expression (eg the `{b}` in `<div a={b} />`)

This is just another example of how we only fake-parse JSX but now need to preserve the real source tokens as well - here, to our grammar this is a `(` since we're just wrapping the attribute value expression in parens, but in reality it's a `{` in the source

GeoffreyBooth:
```suggestion
          # This token represents the start of a JSX attribute value
          # that’s an expression (e.g. the `{b}` in `<div a={b} />`).
          # Our grammar represents the beginnings of expressions as `(`
          # tokens, so make this into a `(` token that displays as `{`.
          token = @token '(', '{'
```

Issue #5241, opened 2019-11-12
Bug: Trailing comments after colons not output


Asc2011:
seems buggy to me ?
```coffee
CS = require 'coffeescript'
d = """
string : 'a string'  # string with a comment behind it
array : [] # array with trailing comment, this gets eaten
"""
n = CS.nodes d
n.compile bare: on
```
### Expected Behavior
```
({
  string: 'a string', // string with a comment behind it
  arr: [] // array with trailing comment, this gets eaten
});
```

### Current Behavior
```
({
  string: 'a string', // string with a comment behind it
  arr: []
});
```

### Context
working on a Atom-package to preserve comments in CSON/JSON

* CoffeeScript version: 2.4.1
* Node.js version: 10.16.3

Asc2011:
seems there is another one here
```coffee
d = """
string : 'a string'  # string with a comment behind it
object :   # object open with trailing comment, it looks as if this gets eaten, too
  member: 'string'
"""
```
```
({
  string: 'a string', // string with a comment behind it
  object: {
    member: 'string'
  }
});
```
sry for mentioning - i know how much effort went into comment-handling already...

GeoffreyBooth:
Comments are a tough case. Can you try with the `ast` branch and use a proper AST rather than the call to `nodes`?

Asc2011:
tried that, and it looks **much** better. Both missing comments show up in the AST.

GeoffreyBooth:
The issue is that comments aren't part of the grammar; they're like whitespace, in that they become part of the metadata of a nearby token (in whitespace's case, the indent level). When tokens get rewritten, such as to insert implied things like curly braces, metadata can get lost. We try to avoid that whenever possible, but it can happen; and it's hard to write tests to cover all edge cases.

helixbass:
> You could theoretically take the AST output and create generated JavaScript from that, via something like Babel or Prettier, if the AST output has better comments preservation than the generated JS output

That won't be true until we have a "transformed AST" (which we've talked about maybe being the next phase of the AST work - having a printer that consumes a transformed-to-JS-structure AST)

With the "Coffeescript AST", there are structures that don't correspond directly to JS output so you couldn't directly pass it to a JS printer like Prettier

PR #5257, opened 2019-12-07
AST: comment after semicolon

helixbass:
@GeoffreyBooth PR ensuring that a comment after a semicolon gets correctly included in the AST

As of this PR, there are only 2 failing tests in the ESLint plugin (out of 11,000+), neither of which I think are Coffeescript AST issues, so this should conclude reconciling the `ast` branch with the ESLint plugin (ie once this PR gets merged, I'll swap in the `ast` branch as the ESLint plugin's Coffeescript dependency)

Based on `ast-generated-jsx-tokens`, [here](https://github.com/helixbass/copheescript/compare/ast-generated-jsx-tokens...ast-semicolon-comment) is just the diff against that branch

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -117,8 +117,8 @@ buildLocationData = (first, last) ->
 # Build a list of all comments attached to tokens.
 exports.extractAllCommentTokens = (tokens) ->
   allCommentsObj = {}
-  for token in tokens when token.comments or token.prevToken?.comments
-    for comment in [(token.prevToken?.comments ? [])..., (token.comments ? [])...]
+  for token in tokens when token.comments
+    for comment in token.comments
       commentKey = comment.locationData.range[0]
       allCommentsObj[commentKey] = comment
   sortedKeys = Object.keys(allCommentsObj).sort (a, b) -> a - b
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -117,8 +117,8 @@ buildLocationData = (first, last) ->
 # Build a list of all comments attached to tokens.
 exports.extractAllCommentTokens = (tokens) ->
   allCommentsObj = {}
-  for token in tokens when token.comments
-    for comment in token.comments
+  for token in tokens when token.comments or token.prevToken?.comments
+    for comment in [(token.prevToken?.comments ? [])..., (token.comments ? [])...]
       commentKey = comment.locationData.range[0]
       allCommentsObj[commentKey] = comment
   sortedKeys = Object.keys(allCommentsObj).sort (a, b) -> a - b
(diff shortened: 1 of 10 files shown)


src/rewriter.coffee
```diff
@@ -685,6 +685,8 @@ exports.Rewriter = class Rewriter
           tokens.splice i, 1, @indentation()...
           return 1
         if @tag(i + 1) in EXPRESSION_CLOSE
+          if token[1] is ';' and @tag(i + 1) is 'OUTDENT'
```
helixbass: By storing a reference to the `;` token preceding an `OUTDENT`, we can preserve both its comments and location data

src/helpers.coffee
```diff
@@ -117,8 +117,8 @@ buildLocationData = (first, last) ->
 # Build a list of all comments attached to tokens.
 exports.extractAllCommentTokens = (tokens) ->
   allCommentsObj = {}
-  for token in tokens when token.comments
-    for comment in token.comments
+  for token in tokens when token.comments or token.prevToken?.comments
+    for comment in [(token.prevToken?.comments ? [])..., (token.comments ? [])...]
```
GeoffreyBooth: Why is this change necessary? It feels like a change that may have gotten the particular test to pass, but in a larger codebase it might introduce bugs.

In particular, there’s logic in the comments output code that avoids outputting a comment if that comment has already been output. That logic is there because sometimes the same comment gets attached to multiple tokens, such as when tokens are duplicated in the rewriter. So what I’m concerned about is that adding a condition here for looking for comments in not just the current token but also in the previous one too makes _every_ case of comments attached to tokens this “duplicated comments” case, that I’m working around by checking that we don’t double-output comments; but that’s really suboptimal because ideally comments wouldn’t be duplicated in the first place. Like if you removed the “don’t double-output comments” code, I bet that every comment would be double-output now, not just the ones where we expect that problem because of rewriting.

In other words, perhaps something needs to be done in the rewriter to move/copy the comments to/from the semicolon token as appropriate, rather than this global change looking for comments beyond where we “should” be looking.

helixbass: Ok I updated to avoid the need for this change by instead using `moveComments()` to shift the comment (that was getting lost) from the semicolon token to the outdent token

Which then also gets that comment to output during compile-to-JS (which it wasn't previously, eg this comment gets lost on `master`:
```
if a
  ;
  ### comment ###
```

GeoffreyBooth: Good! Thank you, that’s much better.

Doesn’t seem to fix #5241, but it’s on the right track.

GeoffreyBooth:
👍 when upstream PRs clear.

PR #5258, opened 2019-12-09
AST: reconcile Prettier plugin

helixbass:
@GeoffreyBooth I updated the Prettier plugin's Coffeescript dependency to the current `ast` branch and it surfaced a couple small things

So this PR addresses those and gets all tests passing again in the Prettier plugin

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1454,10 +1454,10 @@ exports.Value = class Value extends Base
         return new If new Existence(fst), snd, soak: on
       no
 
-  eachName: (iterator) ->
+  eachName: (iterator, {checkAssignability = yes} = {}) ->
     if @hasProperties()
       iterator @
-    else if @base.isAssignable()
+    else if not checkAssignability or @base.isAssignable()
       @base.eachName iterator
     else
       @error 'tried to assign to unassignable value'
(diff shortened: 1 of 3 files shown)


src/nodes.coffee
```diff
@@ -1454,10 +1454,10 @@ exports.Value = class Value extends Base
         return new If new Existence(fst), snd, soak: on
       no
 
-  eachName: (iterator) ->
+  eachName: (iterator, {checkAssignability = yes} = {}) ->
```
helixbass: We don't want `Value::eachName()` to always check that it's assignable

Specifically, `for`'s with "unassignable" loop variables were failing here during AST generation, eg
```
for [a..., b] in c
  d
```

So during `For` AST generation pass a `checkAssignability: no` flag to `eachName()`

src/nodes.coffee
```diff
@@ -1654,7 +1654,7 @@ exports.JSXExpressionContainer = class JSXExpressionContainer extends Base
 
   astProperties: (o) ->
     return
-      expression: @expression.ast o
+      expression: astAsBlockIfNeeded @expression, o
```
helixbass: The other issue was that "blocks" for JSX attribute expression values/string interpolation values were generating `SequenceExpression` AST's rather than `BlockStatement`'s (`SequenceExpression`'s should only be generated when there's actually a "sequence" in the source eg `(a; b)`)

So for example this:
```
<div a={
  b
  c
} />
```
or this:
```
a "#{
  b
  c
}"
```

So use a shared `astAsBlockIfNeeded()` helper that ensures that a `BlockStatement` AST gets generated if it's looking at a multi-statement block

PR #5259, opened 2019-12-14
AST: throw expected error messages

helixbass:
@GeoffreyBooth PR to throw (almost all) expected error messages during AST generation (in addition to during compile-to-JS)

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -1243,7 +1243,7 @@ exports.DefaultLiteral = class DefaultLiteral extends Literal
 
 # A `return` is a *pureStatement*—wrapping it in a closure wouldn’t make sense.
 exports.Return = class Return extends Base
-  constructor: (@expression) ->
+  constructor: (@expression, {@belongsToFuncDirectiveReturn} = {}) ->
     super()
 
   children: ['expression']
@@ -1277,6 +1277,11 @@ exports.Return = class Return extends Base
     answer.push @makeCode ';'
     answer
 
+  checkForPureStatementInExpression: ->
+    # don’t flag `return` from `await return`/`yield return` as invalid.
+    return if @belongsToFuncDirectiveReturn
+    super()
+
   astType: -> 'ReturnStatement'
 
   astProperties: (o) ->
@@ -1301,7 +1306,7 @@ exports.FuncDirectiveReturn = class FuncDirectiveReturn extends Return
     @checkScope o
 
     new Op @keyword,
-      new Return @expression
+      new Return @expression, belongsToFuncDirectiveReturn: yes
       .withLocationDataFrom(
         if @expression?
           locationData: mergeLocationData @returnKeyword.locationData, @expression.locationData
@@ -2643,12 +2648,12 @@ exports.Arr = class Arr extends Base
     return yes for obj in @objects when obj instanceof Elision
     no
 
-  isAssignable: ->
+  isAssignable: ({allowExpansion} = {}) ->
     return no unless @objects.length
 
     for obj, i in @objects
       return no if obj instanceof Splat and i + 1 isnt @objects.length
-      return no unless obj.isAssignable() and (not obj.isAtomic or obj.isAtomic())
+      return no unless (allowExpansion and obj instanceof Expansion) or (obj.isAssignable() and (not obj.isAtomic or obj.isAtomic()))
     yes
 
   shouldCache: ->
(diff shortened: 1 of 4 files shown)


src/nodes.coffee
```diff
@@ -3387,11 +3437,11 @@ exports.Assign = class Assign extends Base
   unfoldSoak: (o) ->
     unfoldSoak o, this, 'variable'
 
-  addScopeVariables: (o, {checkAssignability = yes} = {}) ->
+  addScopeVariables: (o, {allowAssignmentToExpansion = no} = {}) ->
```
helixbass: Had to refine the disabling of `checkAssignability` during AST generation to only specifically disable checking `Expansion`'s since it _should_ check assignability for things like `null` (eg `[null] = b`)

test/error_messages.coffee
```diff
@@ -8,10 +8,10 @@ errCallback = (expectedErrorFormat) -> (err) ->
   err.colorful = no
   eq expectedErrorFormat, "#{err}"
   yes
-assertErrorFormat = (code, expectedErrorFormat) ->
+assertErrorFormatNoAst = (code, expectedErrorFormat) ->
```
helixbass: Adjusted so the "default" `assertErrorFormat()` includes AST checking, and renamed the existing `assertErrorFormat()` to `assertErrorFormatNoAst()` (for cases that don't also throw an error during AST generation)

test/error_messages.coffee
```diff
@@ -1122,15 +1122,15 @@ test "cannot export * without a module to export from", ->
   '''
 
 test "imports and exports must be top-level", ->
```
helixbass: This is the only one (at least from `test/error_messages.coffee`) that still isn't being thrown during AST generation (but in theory could) - the logic for throwing an error on "non-top-level" imports/exports is based on `o.indent`, which doesn't get set during AST generation

Seems like there could be a different mechanism not based on indent level to detect when you're not at the top level in general?

GeoffreyBooth: Does `if o.level > LEVEL_TOP` not work?

src/nodes.coffee
```diff
@@ -3175,6 +3201,10 @@ exports.ImportDeclaration = class ImportDeclaration extends ModuleDeclaration
     code.push @makeCode ';'
     code
 
+  astNode: (o) ->
+    o.importedSymbols = []
```
GeoffreyBooth: Why is this here?

helixbass: This `o.importedSymbols` gets used to track imported names eg for throwing an error if you try and import the same name more than once (eg `import {foo, foo} from 'bar'`) in `ImportSpecifier::addIdentifierToScope()`

test/error_messages.coffee
helixbass: I think that wouldn't work since `level` really only reflects the "compilation level" *within a statement*

ie block-level statements always get compiled at `LEVEL_TOP`, eg here the import statement gets compiled at `LEVEL_TOP`:
```
if foo
  import { bar } from 'lib'
```

But it does seem like there could be a separate concept of "program top-level" (meaning I guess "a statement in the root `Block`"?)

It seems like that could get set on `o`. If we ended up tracking block scopes eg to be able to generate `const`/`let` where possible per your other comment, it seems like that would suffice

GeoffreyBooth: > It seems like that could get set on `o`. If we ended up tracking block scopes eg to be able to generate `const`/`let` where possible per your other comment, it seems like that would suffice

Is tracking block scopes something that’s easy to add? If so, absolutely let’s do that, I’d love to start outputting `let`/`const` 😄 

Does a node know its parents? Obviously it knows its children, but if it also can easily find its parents then you can see if there are any `Block`s among its parents.

This is also not the most important check. The runtime will throw if anyone tries to put an `import` in not-top-level code in executable JavaScript (as opposed to JS further transpiled by some other build tool). We could just stop checking for this in CoffeeScript entirely, or check only on JS output with a comment saying why it’s hard to do this check on AST output, and leave it at that.

src/nodes.coffee
GeoffreyBooth: I see, you’re duplicating what’s in `compileNode`. I guess we can’t just put this in the constructor to avoid the duplication?

I also don’t care all that much, it’s one line.

helixbass: Ya in general I took the approach of trying to DRY things up by sharing small helper methods between the `compileNode()` path and the AST generation path, but here this didn't seem worth trying to encapsulate

test/error_messages.coffee
helixbass: > Is tracking block scopes something that’s easy to add? If so, absolutely let’s do that, I’d love to start outputting `let`/`const` 😄

I'm not sure, but I'd be up for doing a deeper dive on how scopes work sometime soon

> Does a node know its parents?

Ya I was thinking something similar. Currently I don't think there's any way to walk "up" the node class object tree. And I can't picture a simple way to add "parent links" to enable that

> This is also not the most important check

Agreed, I think it's fine for this to remain something that only gets flagged during compile-to-JS for the time being

> or check only on JS output with a comment saying why it’s hard to do this check on AST output

Sure, added a comment to `ModuleDeclaration::checkScope()`

PR #5260, opened 2019-12-14
AST: test compile errors

helixbass:
@GeoffreyBooth PR testing expected compile errors

(so this PR updates tests of `throws()`/`doesNotThrow()` when compiling to also check during AST generation, building off of the previous PR which just updated `test/error_messages.coffee` to check for expected error messages during AST generation)

Based on `ast-throw-expected-error-messages`, [here](https://github.com/helixbass/copheescript/compare/ast-throw-expected-error-messages...ast-test-compile-errors) is just the diff against that branch


src/nodes.coffee
```diff
@@ -4181,9 +4321,21 @@ exports.Param = class Param extends Base
...
+    checkAssignabilityOfLiteral = (literal) ->
+      message = isUnassignable literal.value
+      if message
+        literal.error message
+      unless literal.isAssignable()
+        literal.error "'#{literal.value}' can't be assigned"
+
     atParam = (obj, originalObj = null) => iterator "@#{obj.properties[0].name.value}", obj, @, originalObj
+    if name instanceof Call
+      name.error "Function invocation can't be assigned"
```
helixbass: Technically this changes an existing compile-to-JS error message, eg this:
```
({a: b()}) ->
```
used to throw `error: 'b()' can't be assigned` but now throws `error: Function invocation can't be assigned`

Basically during compile-to-JS this error (which is needed to throw an error during AST generation) gets triggered before the existing  error

And the existing error uses eg `name.compile o ` to print `b()` in the error message, but it would be messy to try and have `o` be available here in order to call `name.compile o` (and it seems kind of weird to me to be calling `.compile()` to print out a snippet of (technically transpiled-to-JS) source code in an error message anyway)

test/assignment.coffee
```diff
@@ -278,18 +278,17 @@ test "#1024: destructure empty assignments to produce javascript-like results",
 
 test "#1005: invalid identifiers allowed on LHS of destructuring assignment", ->
   disallowed = ['eval', 'arguments'].concat CoffeeScript.RESERVED
-  throws (-> CoffeeScript.compile "[#{disallowed.join ', '}] = x"), null, 'all disallowed'
-  throws (-> CoffeeScript.compile "[#{disallowed.join '..., '}...] = x"), null, 'all disallowed as splats'
+  throwsCompileError "[#{disallowed.join ', '}] = x", null, null, 'all disallowed'
```
helixbass: The pattern is basically that I replaced `throws -> CoffeeScript.compile ...` with `throwsCompileError ...`, which checks that the error gets thrown both during compile-to-JS and during AST generation

And similarly with `doesNotThrow -> CoffeeScript.compile ...` -> `doesNotThrowCompileError ...`

src/nodes.coffee
```diff
@@ -2512,6 +2538,12 @@ exports.Obj = class Obj extends Base
     answer = @wrapInBraces answer
     if @front then @wrapInParentheses answer else answer
 
+  checkForMultipleSplats: ->
+    return unless @hasSplat() and @lhs
+    props = @properties
+    splatProps = (i for prop, i in props when prop instanceof Splat)
```
GeoffreyBooth: Can we not duplicate these lines? Maybe change `checkForMultipleSplats` to `splatProps` that both does the check and returns `splatProps`?

helixbass: Sure, updated to return `splatProps` and renamed to `getAndCheckSplatProps()`

helixbass:
@GeoffreyBooth rebased onto `ast`

src/nodes.coffee
```diff
@@ -3710,6 +3720,22 @@ exports.Assign = class Assign extends Base
...
+
+    # Count all `Splats`: [a, b, c..., d, e]
+    splats = (i for obj, i in objects when obj instanceof Splat)
+    # Count all `Expansions`: [a, b, ..., c, d]
+    expans = (i for obj, i in objects when obj instanceof Expansion)
+    # Combine splats and expansions.
+    splatsAndExpans = [splats..., expans...]
+    if splatsAndExpans.length > 1
+      # Sort 'splatsAndExpans' so we can show error at first disallowed token.
+      objects[splatsAndExpans.sort()[1]].error "multiple splats/expansions are disallowed in an assignment"
```
GeoffreyBooth: This is a rather big chunk to copy/paste.

src/nodes.coffee
```diff
@@ -4102,6 +4136,19 @@ exports.Code = class Code extends Base
...
+    for param, i in @params
+      # Was `...` used with this parameter? (Only one such parameter is allowed
+      # per function.) Splat/expansion parameters cannot have default values,
+      # so we need not worry about that.
+      if param.splat or param instanceof Expansion
+        if haveSplatParam
+          param.error 'only one splat or expansion parameter is allowed per function definition'
+        else if param instanceof Expansion and @params.length is 1
+          param.error 'an expansion parameter cannot be the only parameter in a function definition'
+        haveSplatParam = yes
```
GeoffreyBooth: This is also a pretty big duplicated section.

test/abstract_syntax_tree.coffee
```diff
@@ -2281,17 +2281,22 @@ test "AST as expected for Assign node", ->
       name: 'c'
       declaration: no
 
-  testExpression 'a ?= b',
```
GeoffreyBooth: Shouldn’t we also have this test?

src/nodes.coffee
helixbass: Ok DRYed this one up by having it return `{splats, expans, splatsAndExpans}` (and renamed to `getAndCheckSplatsAndExpansions()`)

src/nodes.coffee
helixbass: Ok for this one there wouldn't be a good way to fully DRY it up (the existing code still needs to track whether it's seen a splat param yet for purposes of compiling the params), but I trimmed down the comments to be more accurate to that particular context and adjusted the naming here to be more specific to this use case (`haveSplatParam` -> `seenSplatParam`). So made it less of a literal "copy-paste"

test/abstract_syntax_tree.coffee
helixbass: This no longer compiles now that we added the comparable error handling during AST generation (`a ?= b` throws an error during compile-to-JS), so I just adjusted this test to be valid Coffeescript (by embedding this expression inside a bigger expression that makes it valid)

PR #5262, opened 2019-12-16
AST: fix interpolated regex tokens

helixbass:
@GeoffreyBooth I turned on the ESLint plugin against the Coffeescript codebase (on a branch) and am using that to look for issues

Here's one on the Coffeescript side:

The tokens for interpolated regexes needed to be cleaned up, this was causing rules like [`keyword-spacing`](https://eslint.org/docs/rules/keyword-spacing) to break

Based on `ast-clean-up-tests`, [here](https://github.com/helixbass/copheescript/compare/ast-clean-up-tests...ast-fix-interpolated-regex-tokens) is just the diff against that branch


src/lexer.coffee
```diff
@@ -480,16 +480,16 @@ exports.Lexer = class Lexer
         @validateUnicodeCodePointEscapes body, {delimiter}
         @token 'REGEX', "/#{body}/#{flags}", {length: end, origin, data: {delimiter}}
       else
-        @token 'REGEX_START', '(',    {length: 0, origin}
-        @token 'IDENTIFIER', 'RegExp', length: 0
-        @token 'CALL_START', '(',      length: 0
+        @token 'REGEX_START', '(',    {length: 0, origin, generated: yes}
```
helixbass: Need to mark all these fake tokens from converting an interpolated regex into a nonexistent `RegExp()` call as generated

src/lexer.coffee
```diff
@@ -480,16 +480,16 @@ exports.Lexer = class Lexer
...
+        @token 'CALL_START', '(',      length: 0, generated: yes
         @mergeInterpolationTokens tokens, {double: yes, heregex: {flags}, endOffset: end - flags.length, quote: '///'}, (str) =>
           @validateUnicodeCodePointEscapes str, {delimiter}
         if flags
-          @token ',', ',',                    offset: index - 1, length: 0
-          @token 'STRING', '"' + flags + '"', offset: index - 1, length: flags.length
-        @token ')', ')',                      offset: end,       length: 0
-        @token 'REGEX_END', ')',              offset: end,       length: 0
+          @token ',', ',',                    offset: index - 1, length: 0, generated: yes
+          @token 'STRING', '"' + flags + '"', offset: index,     length: flags.length
```
helixbass: The location data generated for interpolated regex flags was off by one

src/lexer.coffee
```diff
@@ -1008,7 +1008,7 @@ exports.Lexer = class Lexer
           lastToken[2].range[1]
         ]
       ]
-      lparen[2] = lparen.origin[2]
+      lparen[2] = lparen.origin[2] unless quote?.length
```
helixbass: I traced this back [here](https://github.com/jashkenas/coffeescript/pull/4572/files#r130409263) to your comments PR

There I think you were running into `lparen` sometimes not having *any* location data set

But it generally does, and this line was wiping out the accurate location data (eg for the opening `///` in an interpolated regex) and replacing it with the location data for the entire interpolated string/regex

So updated it to only "overwrite" `lparen`'s location data if it doesn't already have legitimate location data

test/location.coffee
```diff
@@ -607,9 +607,9 @@ test "Verify heregexes with interpolations have the right ending position", ->
   arrayEndColumn = arrayEnd[2].first_column
   for token in [comma]
     assertColumn token, arrayEndColumn - 2
-  for token in [flagsString]
-    assertColumn token, arrayEndColumn - 2, 1
-  for token in [regexCallEnd, regexEnd, fnCallEnd]
+  for token in [flagsString, regexCallEnd]
```
helixbass: By updating the `regexCallEnd` token to be marked (correctly) as `generated`, its location data now gets updated by the rewriter (along with other generated tokens)

Its location data isn't particularly significant anyway as it's not a real token

src/lexer.coffee
GeoffreyBooth: Always nice to have my own bugs pointed out and fixed 😄

GeoffreyBooth:
✅ once conflicts resolved.

helixbass:
@GeoffreyBooth ok rebased

PR #5263, opened 2019-12-16
AST: allow assignment to nontrailing splat

helixbass:
@GeoffreyBooth here's another issue that applying the ESLint plugin to the Coffeescript codebase surfaced:

Assignment to a nontrailing splat (eg `[a..., b] = c`) was failing, so needed to extend the logic for the differences in what's "assignable" during AST generation vs compile-to-JS

Based on `ast-fix-interpolated-regex-tokens`, [here](https://github.com/helixbass/copheescript/compare/ast-fix-interpolated-regex-tokens...ast-allow-assignment-to-nontrailing-splat) is just the diff against that branch

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -2661,11 +2661,11 @@ exports.Arr = class Arr extends Base
     return yes for obj in @objects when obj instanceof Elision
     no
 
-  isAssignable: ({allowExpansion} = {}) ->
+  isAssignable: ({allowExpansion, allowNontrailingSplat} = {}) ->
     return no unless @objects.length
 
     for obj, i in @objects
-      return no if obj instanceof Splat and i + 1 isnt @objects.length
+      return no if not allowNontrailingSplat and obj instanceof Splat and i + 1 isnt @objects.length
       return no unless (allowExpansion and obj instanceof Expansion) or (obj.isAssignable() and (not obj.isAtomic or obj.isAtomic()))
     yes
 
@@ -3455,11 +3455,11 @@ exports.Assign = class Assign extends Base
   unfoldSoak: (o) ->
     unfoldSoak o, this, 'variable'
 
-  addScopeVariables: (o, {allowAssignmentToExpansion = no} = {}) ->
+  addScopeVariables: (o, {allowAssignmentToExpansion = no, allowAssignmentToNontrailingSplat = no} = {}) ->
     return unless not @context or @context is '**='
 
     varBase = @variable.unwrapAll()
-    if not varBase.isAssignable allowExpansion: allowAssignmentToExpansion
+    if not varBase.isAssignable allowExpansion: allowAssignmentToExpansion, allowNontrailingSplat: allowAssignmentToNontrailingSplat
       @variable.error "'#{@variable.compile o}' can't be assigned"
 
     varBase.eachName (name) =>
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
> Assignment to a nontrailing splat (eg `[a..., b] = c`) was failing

What do you mean by it failing? An error was being thrown during AST generation for this input source code?

helixbass:
> What do you mean by it failing? An error was being thrown during AST generation for this input source code?

Yes. This construct is considered "unassignable" (by `.isAssignable()`) since in JS you can't assign to a nontrailing splat like that. But during compile-to-JS it's already transformed this type of assignment (to valid JS) before it would flag it as unassignable

I think the status here is that (a) we previously were ignoring *all* assignability checking during AST generation, so this wouldn't have been getting flagged. But then when getting expected errors to be thrown during AST generation, that got refined so that things like `[null] = b` would still get flagged.  So this PR "re-unflags" a specific case that should be allowable during AST generation and (b) we didn't have AST test coverage for this case so it wasn't obvious that there was a "regression" when we refined the assignability checking

src/nodes.coffee
```diff
@@ -3455,11 +3455,15 @@ exports.Assign = class Assign extends Base
   unfoldSoak: (o) ->
     unfoldSoak o, this, 'variable'
 
-  addScopeVariables: (o, {allowAssignmentToExpansion = no} = {}) ->
+  addScopeVariables: (o, {allowAssignmentToExpansion = no, allowAssignmentToNontrailingSplat = no, allowAssignmentToEmptyArray = no} = {}) ->
```
helixbass: I ran into another similar case: assignment to an empty array (eg `[] = b`) was failing during AST generation

So added `allowAssignmentToEmptyArray` here as well

I think I slightly prefer this explicit style of enumerating the assignability "exceptions", but if it feels cumbersome I guess we could just pass eg a `forAst: yes` flag instead and have `Arr::isAssignable()` be aware of what exactly is "assignable" during AST generation (ie when `forAst: yes`)?

GeoffreyBooth: I think something like `allowAssignmentToEmptyArray` seems better because it’s clearer why behavior is branching. Maybe also add a comment as well explaining why this flag exists (i.e. your explanation above)?

GeoffreyBooth: I agree, I think something like `allowAssignmentToEmptyArray` seems better because it’s clearer why behavior is branching. Maybe also add a comment as well explaining why this flag exists (i.e. your explanation above)?

src/nodes.coffee
```diff
@@ -3455,11 +3455,24 @@ exports.Assign = class Assign extends Base
   unfoldSoak: (o) ->
     unfoldSoak o, this, 'variable'
 
-  addScopeVariables: (o, {allowAssignmentToExpansion = no} = {}) ->
+  addScopeVariables: (o, {
+    # During AST generation, we need to allow assignment to these constructs
```
helixbass: > Maybe also add a comment as well explaining why this flag exists (i.e. your explanation above)?

Sure, added an explanatory comment

src/nodes.coffee
```diff
@@ -3455,11 +3455,24 @@ exports.Assign = class Assign extends Base
...
+  } = {}) ->
     return unless not @context or @context is '**='
 
     varBase = @variable.unwrapAll()
-    if not varBase.isAssignable allowExpansion: allowAssignmentToExpansion
+    if not varBase.isAssignable {
+      allowExpansion: allowAssignmentToExpansion
+      allowNontrailingSplat: allowAssignmentToNontrailingSplat
+      allowEmptyArray: allowAssignmentToEmptyArray
+      allowComplexSplat: allowAssignmentToComplexSplat
```
helixbass: I ran into one more similar case where complex object splat assignments were failing eg `{{b...}...} = c`

PR #5264, opened 2019-12-17
WIP: Output JS-compatible line numbers

helixbass:
@GeoffreyBooth this is an interesting one I came across while running the ESLint plugin against the Coffeescript repo source code:

It was crashing running ESLint against `test/regex.coffee` because of the literal U+2028/U+2029 (Unicode linebreak) characters in that file

Looking into it, it turns out our location data/line numbers aren't exactly JS-compatible - per eg [here](https://books.google.com/books?id=4RChxt67lvwC&pg=PA22&lpg=PA22&dq=javascript+unicode+2028+line+number&source=bl&ots=ti07EhOVm8&sig=ACfU3U0w1HIaXZyPpEtaiQzjfvK56gqylg&hl=en&sa=X&ved=2ahUKEwir5ae5hrvmAhWXbs0KHRraDNgQ6AEwB3oECAoQAQ#v=onepage&q=javascript%20unicode%202028%20line%20number&f=false), JS considers newline (`\n`), carriage return (`\r`), carriage return + newline (`\r\n`), U+2028 and U+2029 to be line terminators

Whereas we strip carriage returns and then just consider newlines to be line terminators

So what was happening is that in a file with "unusual" line terminators according to JS (eg the U+2028 and U+2029 in `test/regex.coffee`), ESLint was doing the JS version of counting linebreaks and then getting confused when a certain line didn't have the number of columns reported by our location data

So thus far this PR swaps in the JS-compatible regex for detecting linebreaks in the obvious places. Which did get ESLint working against `test/regex.coffee`

But I'm probably inclined to not try and get this merged before initial release, as (a) it should only cause problems in files where someone's got "fake linebreak" characters floating around and (b) I think it's worth thinking through more fully

For one thing, I think that if someone had a `\r` in their source code that wasn't "attached" to a following `\n`, it'd need to be accounted for - since we strip out `\r`'s, I think we'd need to go a step further than the `locationDataCompensations` (that we added into the lexer to get accurate location data despite things like `\r`'s having been stripped out) and do something like additionally "remember" the source offsets of any "unattached" `\r`'s and then take those into account when generating location data

But ultimately it seems like it'd be a good idea to be compatible with JS's definition of linebreaks in our location data?

Based on `ast-splat-param-location-data`, [here](https://github.com/helixbass/copheescript/compare/ast-splat-param-location-data...ast-js-compatible-line-numbers) is just the diff against that branch

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1056,11 +1056,11 @@ exports.Lexer = class Lexer
     else
       string = @chunk[..offset-1]
 
-    lineCount = count string, '\n'
+    lineCount = (string.match(LINEBREAK) or []).length
 
     column = @chunkColumn
     if lineCount > 0
-      [..., lastLine] = string.split '\n'
+      [..., lastLine] = string.split LINEBREAK
       column = lastLine.length
       previousLinesCompensation = @getLocationDataCompensation @chunkOffset, @chunkOffset + offset - column
       # Don't recompensate for initially inserted newline.
@@ -1313,6 +1313,8 @@ OPERATOR   = /// ^ (
 ) ///
 
 WHITESPACE = /^[^\n\S]+/
+
+exports.LINEBREAK = LINEBREAK = /\r\n|[\r\n\u2028\u2029]/ug
 
 COMMENT    = /^(\s*)###([^#][\s\S]*?)(?:###([^\n\S]*)|###$)|^((?:\s*#(?!##[^#]).*)+)/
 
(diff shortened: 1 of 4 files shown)


src/lexer.coffee
```diff
@@ -1314,6 +1314,8 @@ OPERATOR   = /// ^ (
 
 WHITESPACE = /^[^\n\S]+/
 
+exports.LINEBREAK = LINEBREAK = /\r\n|[\r\n\u2028\u2029]/ug
```
helixbass: This regex came from [here](https://github.com/eslint/eslint/blob/master/lib/shared/ast-utils.js#L12)

GeoffreyBooth:
> But ultimately it seems like it'd be a good idea to be compatible with JS's definition of linebreaks in our location data?

Yes, I feel like this might fix a lot of the bugs we get reported about wrong line numbers in source maps and stack traces.

helixbass:
Reopened as #5277

PR #5269, opened 2019-12-29
AST: Fix lexing of BigInt and numbers with separators

GeoffreyBooth:
@helixbass I merged `master` into `ast` locally and ran the tests, and a bunch were failing related to numbers, which isn’t surprising since the recent changes on `master` involve support for `BigInt` and numeric separators. Basically, I think we need to revert two simplifications you made to the lexer's `numberToken` method:

1. We can’t use `Number.isFinite` to tell apart `Infinity` from other numbers. `BigInt`s are not finite, as they’re not numbers at all in JS; yet they’re still `NumericLiteral`s in our grammar.

2. We can’t use the `Number()` constructor to parse numbers, as it doesn’t support numeric separators (`Number('123_456')` returns `NaN`, not the `123456` we want/expect). The previous more complicated logic of detecting the base and using `parseInt` or `parseFloat` works, however, so I brought that back.

### DIFF
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -96,7 +96,7 @@
     // returns, etc. If we’re lexing literate CoffeeScript, strip external Markdown
     // by removing all lines that aren’t indented by at least four spaces or a tab.
     clean(code) {
-      var base, thusFar;
+      var base1, thusFar;
       thusFar = 0;
       if (code.charCodeAt(0) === BOM) {
         code = code.slice(1);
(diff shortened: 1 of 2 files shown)


helixbass:
@GeoffreyBooth ok this looks good

It seems like if `Number()` can no longer be used for parsing numbers we should get rid of the places where it's used in `nodes.coffee` as well (and I guess replace them with the same number-parsing logic that this PR restores to `lexer.coffee`)?

Eg this doesn't compile to the expected output on `master`:
```
[10_000...10_001]
```

helixbass:
Just a heads up I'm working on ☝️as part of a PR to add some other AST-related support for BigInt/numeric separators

GeoffreyBooth:
> Just a heads up I'm working on ☝️

Okay I'll merge this in first and can follow up with yours.

PR #5272, opened 2019-12-29
AST: numeric separators, BigInt

helixbass:
@GeoffreyBooth PR with some AST-related updates for numeric separators/BigInt

Based on `ast-update-from-master`, [here](https://github.com/GeoffreyBooth/coffeescript/compare/ast-update-from-master...helixbass:ast-numeric-separators-bigint) is just the diff against that branch

### DIFF
--- a/src/helpers.coffee
+++ b/src/helpers.coffee
@@ -268,6 +268,20 @@ exports.nameWhitespaceCharacter = (string) ->
     when '\t' then 'tab'
     else string
 
+exports.parseNumber = (string) ->
+  return NaN unless string?
+
+  base = switch string.charAt 1
+    when 'b' then 2
+    when 'o' then 8
+    when 'x' then 16
+    else null
+
+  if base?
+    parseInt string[2..].replace(/_/g, ''), base
+  else
+    parseFloat string.replace(/_/g, '')
+
 exports.isFunction = (obj) -> Object::toString.call(obj) is '[object Function]'
 exports.isNumber = isNumber = (obj) -> Object::toString.call(obj) is '[object Number]'
 exports.isString = isString = (obj) -> Object::toString.call(obj) is '[object String]'
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -12,7 +12,7 @@ Error.stackTraceLimit = Infinity
 {compact, flatten, extend, merge, del, starts, ends, some,
 addDataToNode, attachCommentsToNode, locationDataToString,
 throwSyntaxError, replaceUnicodeCodePointEscapes,
-isFunction, isPlainObject, isNumber} = require './helpers'
+isFunction, isPlainObject, isNumber, parseNumber} = require './helpers'
 
 # Functions required by parser.
 exports.extend = extend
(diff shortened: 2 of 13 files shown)


src/helpers.coffee
```diff
@@ -268,6 +268,20 @@ exports.nameWhitespaceCharacter = (string) ->
     when '\t' then 'tab'
     else string
 
+exports.parseNumber = (string) ->
```
helixbass: Per other [comment](https://github.com/jashkenas/coffeescript/pull/5269#issuecomment-569542323), extracted `parseNumber()` helper so that it can be used instead of `Number()` in the multiple places in `nodes.coffee` where `Number()` was being used

src/helpers.coffee
```diff
@@ -268,6 +268,20 @@ exports.nameWhitespaceCharacter = (string) ->
...
+  return NaN unless string?
+
+  base = switch string.charAt 1
+    when 'b' then 2
+    when 'o' then 8
+    when 'x' then 16
+    else null
+
+  if base?
+    parseInt string[2..].replace(/_/g, ''), base
```
helixbass: `parseInt()` doesn't understand numeric separators (`_`) apparently, so strip them out first

Otherwise the raw `value` in the AST was incorrect for eg `10_000`

src/nodes.coffee
```diff
@@ -936,15 +936,30 @@ exports.NumberLiteral = class NumberLiteral extends Literal
...
-        @parsedValue = Number @value
+        @parsedValue = parseNumber @value
 
-  astType: -> 'NumericLiteral'
+  isBigInt: ->
+    /n$/.test @value
+
+  astType: ->
+    if @isBigInt()
+      'BigIntLiteral'
```
helixbass: I hadn't realized this before but I guess BigInt's get their own `BigIntLiteral` type in Babel AST's

src/nodes.coffee
```diff
@@ -936,15 +936,30 @@ exports.NumberLiteral = class NumberLiteral extends Literal
...
+    else
+      'NumericLiteral'
 
   astProperties: ->
     return
-      value: @parsedValue
+      value:
+        if @isBigInt()
+          @parsedValue.toString()
```
helixbass: ...and the `value`/`extra.rawValue` for BigInt's is the quoted value (eg `"24"` for the source `"24n"`) in Babel AST's

test/numeric_literal_separators.coffee
```diff
@@ -0,0 +1,63 @@
...
+  eq 0b10_10, 0b1010
+
+test 'octal literals with separators', ->
+  eq 0o7_7_7, 0o777
+
+test 'infinity with separator', ->
+  eq 2e3_08, Infinity
+
+test 'range with separators', ->
+  range = [10_000...10_002]
```
helixbass: I guess I don't have a recent enough version of Node installed to actually run these tests but the compiled JS output was as expected

GeoffreyBooth: Node 13.5 handles them.

src/nodes.coffee
GeoffreyBooth: Should we have a separate grammar for them? Rather than overloading NumericLiteral.

helixbass: > Should we have a separate grammar for them? Rather than overloading NumericLiteral

That seems reasonable - I guess that would "trickle down" to needing different token types for BigInt's as well (in order to be able to distinguish them at the grammar level)?

Another option if they only behave differently during AST generation would be to start by only distinguishing them as a new eg `BigIntLiteral` node class during AST generation (rather than at the grammar level) like eg how we're currently only instantiating the `ObjectProperty` node class during AST generation

GeoffreyBooth: Since a BigInt apparently isn't a Number, I feel like we'll run into trouble long term by conflating them.

PR #5275, opened 2019-12-30
AST: Refactor marking nodes that return

GeoffreyBooth:
@helixbass what do you think of this minor refactor? We never override `astReturns`, so it doesn’t need to be its own method; and it seems more organized to me if that logic is in `ast`, which we also never override.

I still wish we could move the `returns` field to be set as part of `astNode`, and maybe I’ll make an attempt at trying to get that to work, but short of that I think this PR feels like an improvement in organization.

I also added some comments, please check that they’re correct. Is the returns thing because we need to wait for the children or the parents? Hopefully I got it right 😄

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -281,9 +281,15 @@ exports.Base = class Base
   # **WARNING: DO NOT OVERRIDE THIS METHOD IN CHILD CLASSES.**
   # Only override the component `ast*` methods as needed.
   ast: (o, level) ->
+    # Merge `level` into `o` and perform other universal checks.
     o = @astInitialize o, level
+    # Create serializable representation of this node.
     astNode = @astNode o
-    @astAddReturns astNode
+    # Mark AST nodes that correspond to expressions that (implicitly) return.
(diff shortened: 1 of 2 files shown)


src/nodes.coffee
```diff
@@ -281,9 +281,15 @@ exports.Base = class Base
...
   ast: (o, level) ->
+    # Merge `level` into `o` and perform other universal checks.
     o = @astInitialize o, level
+    # Create serializable representation of this node.
     astNode = @astNode o
-    @astAddReturns astNode
+    # Mark AST nodes that correspond to expressions that (implicitly) return.
+    # We can’t do this as part of `astNode` because we need to assemble child
+    # nodes first before marking the parent being returned.
+    astNode.returns = yes if @astNode? and @canBeReturned
```
helixbass: If you want this to be more obviously a modification of a JSON-style object/"`Object.assign()`-y", you could write like eg:
```
if @astNode? and @canBeReturned
  Object.assign astNode, {returns: yes}
```

Issue #5129, opened 2018-10-31
Bug: stack trace line numbers are incorrect


thisredone:
### Expected Behavior
When running code with `coffee` line numbers in stack traces should be correct

### Current Behavior
They're not

### Context
This goes pretty far in history of CoffeeScript. I don't think they were ever correct in 2+

* CoffeeScript version: 2.3.2
* Same situation in Windows and Linux
* Node.js version: 8.11.2 but others as well

GeoffreyBooth:
Please provide a reproducible example.

GeoffreyBooth:
Can you find an example that doesn’t involve another library?

thisredone:
I'm honestly not sure. Pretty much all code involves some libraries so is this really necessary? I might keep looking but it might take some time, Maybe tomorrow I'll find some

GeoffreyBooth:
If I change the first line of your example to `config = require './package.json'` (where that file exists) I get a correct stack trace:

```bash
coffee ./test.coffee
Error: bad thing
    at Foo.error (~/Sites/coffeescript/test.coffee:5:11)
    at Object.<anonymous> (~/Sites/coffeescript/test.coffee:7:11)
    at Object.<anonymous> (~/Sites/coffeescript/test.coffee:1:1)
```

So clearly not every import causes this.

If you can produce a minimal test case, with either no library involved or a very minimal (like “hello world”) library that doesn’t have any complexity, I’ll reopen the ticket.

Inve1951:
I've too seen wrong locations in stack traces but can't come up with a concrete example at the moment.
Only thing I got in a quick attempt to reproduce was line numbers being off by 1 line [#try](https://coffeescript.org/v2/#try) on subsequent runs of the code, first run they appear to be correct.

thisredone:
Maybe having this issue closed isn't such a good idea after all? Either I'm living in a bubble; am I the only coffeescript coder in the world; or everyone else is affected by this as well so keeping this issue open could lead to some external help.
The burden of finding the exact conditions in which this problem shows up shouldn't lay on me alone, considering we know nothing about it and it's non-obvious and potentially time-consuming to figure out.

PR #5277, opened 2019-12-30
WIP: Output JS-compatible line numbers

helixbass:
@GeoffreyBooth re-opening against `master`

See original PR: #5264

### DIFF
--- a/lib/coffeescript/coffeescript.js
+++ b/lib/coffeescript/coffeescript.js
@@ -4,10 +4,10 @@
   // on Node.js/V8, or to run CoffeeScript directly in the browser. This module
   // contains the main entry functions for tokenizing, parsing, and compiling
   // source CoffeeScript into JavaScript.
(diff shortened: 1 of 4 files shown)


thisredone:
Hey, are you still working on this? coffeescript really needs line numbers in stack traces fixed

helixbass:
This should only affect line numbers in files which use irregular linebreak characters, so not sure if it's likely to address your issue

Can you give an example though where it's currently incorrect?

thisredone:
There's something here: #5129
It's sort of strange. I guess maybe no one but me is using the `coffee` binary but I honestly don't remember the last time I've worked on a project (and I work on a _lot_ of projects of all sizes on both Linux and Windows) that had correct stack traces.

GeoffreyBooth:
I don't doubt there's an issue, but to fix it we need a minimal reproducible case. If you could provide as small an example as possible where the line numbers are wrong, either for stack traces or source maps, that would give us what we need to fix the bug.

wdavidw:
Just to confirm on my side, line numbers are incorrect with `.coffee.md` extensions, it has been the case for ages.

PR #5278, opened 2019-12-30
Allow linebreak/indent before param default

helixbass:
@GeoffreyBooth re-opening against `master`

See original PR: #5081


GeoffreyBooth:
> This seems like a good idea to me; it matches the rule of Assign.
> 
> (Hope you don’t mind the necro.)

Yes, it’s fine, do you mind rebasing against latest `main`?

PR #5284, opened 2019-12-31
2.5.0

GeoffreyBooth:
https://rawgit.com/geoffreybooth/coffeescript/v2.5.0/docs/v2/index.html

From the changelog:

*   The compiler now supports a new `ast` option, available via `--ast` on the command line or `ast` via the Node API. This option outputs an “abstract syntax tree,” or a JSON-like representation of the input CoffeeScript source code. This AST follows [Babel’s spec](https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md) as closely as possible, for compatibility with tools that work with JavaScript source code. Two tools that use this new AST output are [`eslint-plugin-coffee`](https://github.com/helixbass/eslint-plugin-coffee), a plugin to lint CoffeeScript via [ESLint](https://eslint.org/); and [`prettier-plugin-coffeescript`](https://github.com/helixbass/prettier-plugin-coffeescript), a plugin to reformat CoffeeScript source code via [Prettier](https://prettier.io/). _The structure and properties of CoffeeScript’s AST are not final and may undergo breaking changes between CoffeeScript versions; please [open an issue](https://github.com/jashkenas/coffeescript/issues) if you are interested in creating new integrations._
*   [Numeric separators](https://github.com/tc39/proposal-numeric-separator) are now supported in CoffeeScript, following the same syntax as JavaScript: `1_234_567`.
*   [`BigInt` numbers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) are now supported in CoffeeScript, following the same syntax as JavaScript: `42n`.
*   `'''` and `"""` strings are now output as more readable JavaScript template literals, or backtick (`` ` ``) strings, with actual newlines rather than `\n` escape sequences.
*   Classes can now contain computed properties, e.g. `[someVar]: ->` or `@[anotherVar]: ->`.
*   JSX tags can now contain XML-style namespaces, e.g. `<image xlink:href="data:image/png" />` or `<Something:Tag></Something:Tag>`.
*   Bugfixes for comments after colons not appearing the output; reserved words mistakenly being disallowed as JSX attributes; indented leading elisions in multiline arrays; and invalid location data in source maps.

___

~~@helixbass There seems to be a bug in the browser compiler. See https://rawgit.com/geoffreybooth/coffeescript/v2.5.0/docs/v2/index.html#try:a(). Any function call, e.g. `a()`, throws the error `TypeError: Cannot create property 'locationData' on boolean 'false'` in Chrome (and presumably other browsers that support the modern browser build, that uses `import`/`export` and therefore is in strict mode). The modern browser build works in 2.4.1, so this broke sometime since then. I think we need to fix this before releasing 2.5.0.~~


test/abstract_syntax_tree.coffee
```diff
@@ -49,6 +49,7 @@ testComments = (code, expected) ->
   testAgainstExpected ast.comments, expected
 
 test 'Confirm functionality of `deepStrictIncludeExpectedProperties`', ->
+  return if global.testingBrowser # TODO: Why do these checks fail in Chrome?
```
GeoffreyBooth: @helixbass If I edit `test.html` so that it uses the legacy compiler (by deleting the `nomodule` attribute from the legacy compiler `script` tag and commenting out the code in the `script type="module"` block) the [browser-based tests](https://rawgit.com/geoffreybooth/coffeescript/v2.5.0/docs/v2/test.html) run, and all pass except for this one. For now I'm just excluding this from the browser-based tests, but we should figure out why the checks in this test pass in Node but fail in Chrome, and update accordingly (or exclude only specific checks within this test as necessary, with comments explaining why).

helixbass: Glancing at this test `looseArray()` (which is only used in this test) looks like the most likely culprit?

It doesn't seem to be needed/used? Maybe we could just strip it and the `.loose` check in `deepStrictIncludeExpectedProperties()` out if they are in fact currently unused (and if it is in fact the culprit)?

GeoffreyBooth: So to test this, check out the `v2.5.0` branch and open `test/abstract_syntax_tree.coffee` and comment out this `return if global.testingBrowser` line. Then go to the `docs` folder and run `http-server` (`npm i -g http-server` if you don't already have it) and go to http://localhost:8080/test.html in Chrome. You should see this test fail. The error I see is:

```
Error: Expected Error: Property name: expected Name to equal "Name"
  Expected AST output to include:
  {
  "name": "\"Name\""
}
  but instead it was:
  {
  "name": "Name",
  "a": {
    "b": 1,
    "c": 2
  },
  "x": [
    1,
    2,
    3
  ]
} to equal Expected property does not match
```

So the comparison of `Name` to `"Name"`. If I comment out that check, it fails at another farther down. And so on.

Do you want to try to fix this before we publish 2.5.0?

helixbass: Ok tracing this through (I didn't write this code), it looks like the issue is:

The `check()` helper inside the ``'Confirm functionality of `deepStrictIncludeExpectedProperties`'`` test in question passes its `message` argument as the *second* argument to its `test` "tester" argument

That `test` argument is either `throws`/`doesNotThrow`. Looking at the Node.js [`assert` docs](https://nodejs.org/api/assert.html), it looks like this message argument is expected as the *third* argument to `throws()`/`doesNotThrow()`

So I'm not sure why these tests were passing in Node.js (maybe it ignores a string value for its second `error` argument, or deduces that it's supposed to be the message argument?), but to fix here:

In that `check()` helper (inside the ``'Confirm functionality of `deepStrictIncludeExpectedProperties`'`` test in `test/abstract_syntax_tree.coffee`), change this:
```
test (-> deepStrictIncludeExpectedProperties actual, expected), message
```
to:
```
test (-> deepStrictIncludeExpectedProperties actual, expected), null, message
```

At that point the test suite passes in the browser for me

GeoffreyBooth: Okay I'll take a look. FYI for the browser tests we mock some of Node's `assert` methods: https://github.com/jashkenas/coffeescript/blob/master/documentation/site/test.html#L103-L123. Maybe we mismatch a signature, or Node accepts a varying number of arguments?

helixbass: Right I think it basically boils down to different behavior between our mocked `assert` methods vs the actual `assert` methods

The fix I suggested seems to align it with the "canonical" usage of the actual `assert` methods (and works with our existing mocked `assert` methods)

GeoffreyBooth:
Okay I think this looks good now, the browser-based tests pass for me now:
https://rawgit.com/geoffreybooth/coffeescript/v2.5.0/docs/v2/test.html

@helixbass or anyone else, any other notes?

helixbass:
@GeoffreyBooth I have one more PR I'd like to get in if possible based on something I just ran into when running the Prettier plugin against the Coffeescript codebase

(incoming shortly)

GeoffreyBooth:
> @GeoffreyBooth I have one more PR I'd like to get in if possible based on something I just ran into when running the Prettier plugin against the Coffeescript codebase
> 
> (incoming shortly)

Okay, anything else?

laurentpayot:
Impressive work!

PR #5285, opened 2019-12-31
Browser compiler call parsing fix

helixbass:
@GeoffreyBooth this PR should fix the browser compiler issue you described in #5284

This was related to the workaround we used to avoid forking Jison, see [this comment](https://github.com/jashkenas/coffeescript/pull/5156#pullrequestreview-201353764) - interestingly, it looks like I even called out `OptFuncExist` (which was the culprit here) as violating the workaround but I didn't think it mattered... I didn't bother digging into why it in fact *did* matter when run in the browser compiler context, I just altered the `OptFuncExist` grammar rule to not violate the workaround

I ran `bin/cake build:browser` locally and was able to successfully compile function calls via the Try section of `docs/v2/index.html`

### DIFF
--- a/lib/coffeescript/grammar.js
+++ b/lib/coffeescript/grammar.js
@@ -1090,19 +1090,19 @@
       function() {
         return new TaggedTemplateCall($1,
       $3,
-      $2);
+      $2.soak);
       }),
       o('Value OptFuncExist Arguments',
       function() {
         return new Call($1,
       $3,
-      $2);
+      $2.soak);
       }),
       o('SUPER OptFuncExist Arguments',
       function() {
         return new SuperCall(LOC(1)(new Super()),
       $3,
-      $2,
+      $2.soak,
       $1);
       }),
       o('DYNAMIC_IMPORT Arguments',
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
It's been on my to-do list for a long time to add another `cake` task for running the browser tests in headless Chrome, now that that's possible. Then that can be part of our CI and these kinds of issues would be caught quickly, not at time of release. I'll try to prioritize that soon.

This particular error was due to all ES module code always running in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). We're now generating two browser builds: modern and legacy. The modern build is loaded via `import` and is therefore in Module and Strict mode; whereas the legacy one is not. Per MDN:

> Strict mode in ECMAScript 2015 forbids setting properties on primitive values. Without strict mode, setting properties is simply ignored (no-op), with strict mode, however, a `TypeError` is thrown.

I think this is the explanation: setting `locationData` on `false` is a no-op in sloppy mode but a `TypeError` in strict mode. Another thing we should do is add `'use strict'` to the top of the compiler, so put the whole thing in strict mode, even in Node CommonJS. Then these types of errors would be thrown in our Node tests too. The compiler obviously works in strict mode, as the 2.4.1 modern browser compiler runs that way.

Of course, we should also not be setting properties on primitives in the first place 😄 so this PR is a good fix.

helixbass:
Makes sense, thanks for the explanation

PR #5286, opened 2020-01-01
Allow linebreak/indent in destructured for variable

helixbass:
@GeoffreyBooth I ran into this when running the Prettier plugin against the Coffeescript codebase:

Things like this (with linebreaks inside the destructured variable of a `for` loop) currently aren't parsing:
```
for {
  something,
  somethingElse
}, y in z
  d
```

The grammar is already set up to accept this, it was just an issue of extending the logic for tagging tokens as eg `FORIN`

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -514,7 +514,7 @@ exports.Lexer = class Lexer
 
     prev = @prev()
     backslash = prev?[0] is '\\'
-    @seenFor = no unless backslash or @seenFor?.endsLength isnt @ends.length
+    @seenFor = no unless backslash or @seenFor?.endsLength < @ends.length
     @seenImport = no unless (backslash and @seenImport) or @importSpecifierList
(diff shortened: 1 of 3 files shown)


src/lexer.coffee
```diff
@@ -514,7 +514,7 @@ exports.Lexer = class Lexer
 
     prev = @prev()
     backslash = prev?[0] is '\\'
-    @seenFor = no unless backslash and @seenFor
+    @seenFor = no unless (backslash or @seenFor?.endsLength < @ends.length) and @seenFor
```
helixbass: There was existing logic to "continue the `for` line" in the case of a backslash line continuer, so just extended that logic to additionally continue the `for` line (ie preserve `@seenFor`) if the linebreak is occurring inside a pair of delimiters ie `{`/`}` or `[`/`]` (as indicated by something having been pushed onto `@ends` since we saw the `FOR` token)

PR #5294, opened 2020-01-26
Improve performance of location data compensation

### LINKED ISSUE #5290: Compilation very slow.
This is with node 12.14.1 and coffeescript 2.5.0 on windows

Compiling 12 files with ~300 lines each is taking 10-20 seconds each time.

I recently updated from coffee-script v1.12.7 which took 500ms to compile the same thing.

Any ideas on why it's so much slower?

helixbass:
Fixes #5290 

As discussed in that issue, performance for files with lots of "location data compensations" (eg stripped carriage returns, so all files on Windows) was blowing up because the algorithm used for calculating location data compensation was very inefficient

So this PR improves that algorithm to one that shouldn't have significant performance implications

To test, I made a copy of a decent-size source file eg `src/lexer.coffee` and converted it to have Windows-style line endings (`\r\n`), and ran `bin/coffee -b -p [converted-filename.coffee]`

Before this fix, compilation was taking many seconds, and with this fix, it took less than a second

I haven't done specific performance comparisons of a file with Windows line endings when you omit the location data compensation calculation entirely vs with this new algorithm, that might be interesting as a baseline

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -1033,14 +1033,17 @@ exports.Lexer = class Lexer
   # Compensate for the things we strip out initially (e.g. carriage returns)
   # so that location data stays accurate with respect to the original source file.
   getLocationDataCompensation: (start, end) ->
-    compensation = 0
+    totalCompensation = 0
     initialEnd = end
-    for index, length of @locationDataCompensations
-      index = parseInt index, 10
-      continue unless start <= index and (index < end or index is end and start is initialEnd)
-      compensation += length
-      end += length
-    compensation
+    current = start
+    while current <= end
+      break if current is end and start isnt initialEnd
+      compensation = @locationDataCompensations[current]
+      if compensation?
+        totalCompensation += compensation
+        end += compensation
+      current++
+    return totalCompensation
 
   # Returns the line and column number from an offset into the current chunk.
   #
--- a/lib/coffeescript/lexer.js
+++ b/lib/coffeescript/lexer.js
@@ -1504,20 +1504,22 @@
       // Compensate for the things we strip out initially (e.g. carriage returns)
     // so that location data stays accurate with respect to the original source file.
     getLocationDataCompensation(start, end) {
-      var compensation, index, initialEnd, length, ref;
-      compensation = 0;
+      var compensation, current, initialEnd, totalCompensation;
+      totalCompensation = 0;
       initialEnd = end;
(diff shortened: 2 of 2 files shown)


src/lexer.coffee
```diff
@@ -1033,14 +1033,17 @@ exports.Lexer = class Lexer
...
+    totalCompensation = 0
     initialEnd = end
-    for index, length of @locationDataCompensations
-      index = parseInt index, 10
-      continue unless start <= index and (index < end or index is end and start is initialEnd)
-      compensation += length
-      end += length
-    compensation
+    current = start
+    while current <= end
```
helixbass: This just changes the iteration when calculating the location data compensations for a given `start`/`end` range to iterate on `start..end` instead of iterating on the entire `@locationDataCompensations` data structure every time

GeoffreyBooth:
So you don't need Windows to test this, just a file with Windows line endings. To make things a bit more painful, I created a copy `nodes.coffee` with Windows endings:

```bash
 ✦ time cat src/nodes.coffee | coffee --stdio --compile > /dev/null
cat src/nodes.coffee  0.00s user 0.00s system 4% cpu 0.081 total
coffee --stdio --compile > /dev/null  2.21s user 0.17s system 158% cpu 1.500 total

 ✦ perl -p -e 's/\n/\r\n/' < src/nodes.coffee > test.coffee # https://superuser.com/a/71509/138751

 ✦ time cat test.coffee | coffee --stdio --compile > /dev/null
cat test.coffee  0.00s user 0.00s system 4% cpu 0.083 total
coffee --stdio --compile > /dev/null  248.14s user 1.87s system 102% cpu 4:04.77 total

 ✦ git checkout --track helixbass/iss5290-slow-location-data-compensation
Branch 'iss5290-slow-location-data-compensation' set up to track remote branch 'iss5290-slow-location-data-compensation' from 'helixbass'.
Switched to a new branch 'iss5290-slow-location-data-compensation'

 ✦ time cat test.coffee | ./bin/coffee --stdio --compile > /dev/null
cat test.coffee  0.00s user 0.00s system 2% cpu 0.127 total
./bin/coffee --stdio --compile > /dev/null  2.21s user 0.17s system 149% cpu 1.585 total
```

So in this test, compiling a version of `nodes.coffee` with Windows line endings dropped from _four minutes_ to 1.5 seconds, the same amount of time it took to compile the Unix line endings version. Seems like a success to me.

PR #5296, opened 2020-01-28
Handle implicit object as end of outer implicit object property value

### LINKED ISSUE #5292: Bug: Parser: implicit object as end of outer implicit object property value

helixbass:
Fixes #5292 

This PR updates `addImplicitBracesAndParens()` to correctly parse things like:
```
d: b or
  b: 1
```

### DIFF
--- a/src/lexer.coffee
+++ b/src/lexer.coffee
@@ -38,7 +38,7 @@ exports.Lexer = class Lexer
     @literate   = opts.literate  # Are we lexing literate CoffeeScript?
     @indent     = 0              # The current indentation level.
     @baseIndent = 0              # The overall minimum indentation level.
-    @indebt     = 0              # The over-indentation at the current level.
+    @continuationLineAdditionalIndent = 0 # The over-indentation at the current level.
     @outdebt    = 0              # The under-outdentation at the current level.
     @indents    = []             # The stack of all current indentation levels.
     @indentLiteral = ''          # The indentation.
@@ -531,15 +531,15 @@ exports.Lexer = class Lexer
       @error 'indentation mismatch', offset: indent.length
       return indent.length
 
-    if size - @indebt is @indent
+    if size - @continuationLineAdditionalIndent is @indent
       if noNewlines then @suppressNewlines() else @newlineToken offset
       return indent.length
 
     if size > @indent
       if noNewlines
-        @indebt = size - @indent unless backslash
-        if @indebt
-          prev.continuationLineIndent = @indent + @indebt
+        @continuationLineAdditionalIndent = size - @indent unless backslash
+        if @continuationLineAdditionalIndent
+          prev.continuationLineIndent = @indent + @continuationLineAdditionalIndent
         @suppressNewlines()
         return indent.length
       unless @tokens.length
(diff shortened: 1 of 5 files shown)


src/rewriter.coffee
```diff
@@ -342,10 +342,12 @@ exports.Rewriter = class Rewriter
         if stackTop()
           [stackTag, stackIdx] = stackTop()
           if (stackTag is '{' or stackTag is 'INDENT' and @tag(stackIdx - 1) is '{') and
-             (startsLine or @tag(s - 1) is ',' or @tag(s - 1) is '{')
+             (startsLine or @tag(s - 1) is ',' or @tag(s - 1) is '{') and
+             @tag(s - 1) not in UNFINISHED
```
helixbass: Getting it to parse the nested object as a separate object wasn't that hard...

src/rewriter.coffee
```diff
@@ -369,6 +371,12 @@ exports.Rewriter = class Rewriter
           break unless isImplicit stackItem
           stackItem[2].sameLine = no if isImplicitObject stackItem
 
+      # End indented-continuation-line implicit objects once that indentation is over.
+      if tag is 'TERMINATOR' and token.endsContinuationLineIndentation
```
helixbass: ...the hard part was to then parse additional properties of the outer object as belonging to the outer (not inner) object, eg:
```
a: b or
  b: 1
c: 2 # this belongs to the top-level object, not the nested b: 1 object
```

In order to accomplish that, I added extra token data for "indented line continuations" like this to allow comparing the indentation levels to know when the nested line continuation indent is done (thus when the nested object is done)

src/rewriter.coffee
```diff
@@ -843,3 +851,8 @@ DISCARDED = ['(', ')', '[', ']', '{', '}', ':', '.', '..', '...', ',', '=', '++'
   'INTERPOLATION_START', 'INTERPOLATION_END', 'LEADING_WHEN', 'OUTDENT', 'PARAM_END',
   'REGEX_START', 'REGEX_END', 'RETURN', 'STRING_END', 'THROW', 'UNARY', 'YIELD'
 ].concat IMPLICIT_UNSPACED_CALL.concat IMPLICIT_END.concat CALL_CLOSERS.concat CONTROL_IN_IMPLICIT
+
+# Tokens that, when appearing at the end of a line, suppress a following TERMINATOR/INDENT token
+exports.UNFINISHED = UNFINISHED = ['\\', '.', '?.', '?::', 'UNARY', 'DO', 'DO_IIFE', 'MATH', 'UNARY_MATH', '+', '-',
```
helixbass: I moved this here to avoid a circular dependency between `lexer.coffee` and `rewriter.coffee`

src/lexer.coffee
```diff
@@ -538,6 +538,8 @@ exports.Lexer = class Lexer
     if size > @indent
       if noNewlines
         @indebt = size - @indent unless backslash
+        if @indebt
```
helixbass: It might be nice to additionally rename `@indebt` to something more descriptive like `@continuationLineAdditionalIndent`?

GeoffreyBooth: Sure, `indebt` is a pretty bad name (too close to `indent`).

helixbass: Ok renamed `indebt` -> `continuationLineAdditionalIndent`

PR #5304, opened 2020-02-15
Update coffeescript.js

xen-text:
update the coffeescript.js file so it is easier to read the input errors


GeoffreyBooth:
I'm sorry what is this change?

PR #5306, opened 2020-02-21
CLI: Add --stream option to run script on each line

brewingcode:
This is almost Perl's -n argument: it only reads STDIN, and does not do
the magic filehandle `<>` trick of taking filenames from arguments.

I originally was going to write a shell script to do this, but so much of the other `coffee` arguments were useful (in particular, `-r`) that it made more sense to integrate it. I figure if CoffeeScript can borrow Perl's `/x` regex flag, then maybe it can borrow one more little thing. :-)


brewingcode:
Have a few small things to address, will remove wip: shortly.

GeoffreyBooth:
We try to keep the compiler's options to a minimum. Why should this be a feature of `coffee`, when there are [many existing ways](https://stackoverflow.com/questions/13939038/how-do-you-run-a-command-for-each-line-of-a-file) to run `coffee` for each line of a file?

brewingcode:
> We try to keep the compiler's options to a minimum. Why should this be a feature of coffee, when there are many existing ways to run coffee for each line of a file?

tl;dr nevermind this PR

Mostly because I didn't want to worry about the script that would wrangle all of:

- `readline.createInterface`
- preserve `coffee -r ...` behavior
- preserve `coffee -e ...f` and `coffee FILENAME`

...all of which is what `perl -n` offers.

HOWEVER, while I was flailing for justification and actually timing these changes: I'm tempted to just shut up and pretend this PR never happened. Holy shit, it's slow as hell processing each line this way....maybe calling `vm.runInContext` for each line is what's murdering performance, but yikes. The whole reason I was using this `-S` option was to use NPM modules to parse streaming logs, so performance wasn't an issue due to the logs being slow.

Thanks for looking anyway.

PR #5308, opened 2020-03-03
New no class assign compile option

SolidSnakeBA:
Added new compile option `--no-class-assign`, that will compile CS classes without assigning them to a variable, so the js version feels more natural in the way js classes are written. So to give an example:  
```coffeescript
class A

myFn class B
```
`coffee --bare --no-header --compile --no-class-assign` will be compiled to
```javascript
class A {};

myFn(class B {});
```
The transpiled code feels more natural. 
And one case that this could be usefull is when you have an old angularjs application where you want to use **ng-annotate** plugin. 
The way cs is transpiled this plugin does not work, because it could not handle this form `angular.module('app').service('MyService', MyService = class MyService {})`.
This was the case, why I created this feature.


GeoffreyBooth:
I’m sorry, but we keep our CLI and Node API options very minimal; we especially don’t provide options that alter the output, as then you can’t know what the CoffeeScript source is intended to be just by reading it (you’ll need to know what flags it’s meant to compile with). `bare` is a bit of an exception here and it really should go away, it’s a legacy holdover.

It might be an option to simply change the compilation output to always output classes that don’t assign to a variable. If you can prove that this is a safe change (it doesn’t break CoffeeScript’s tests, for starters) it might be a change that we can make in the interest of compatibility with Angular.

GeoffreyBooth:
Closing as _this_ PR won’t be moving forward, but a new version following the suggestion in https://github.com/jashkenas/coffeescript/pull/5308#issuecomment-594245864 might be considered in the future.

PR #5319, opened 2020-05-10
Support import.meta and import.meta.*

### LINKED ISSUE #5317: Grammar, Nodes: Support import.meta and import.meta.*
### Expected Behavior
`import.meta.url` should return module's full path in a `file://` protocol.

### Current Behavior
```
$ coffee -pe 'dirname = path.dirname import.meta.url'
[stdin]:1:24: error: unexpected import
dirname = path.dirname import.meta.url
                       ^^^^^^
$ coffee -pe 'dirname = path.dirname(import.meta.url)'
[stdin]:1:30: error: unexpected .
dirname = path.dirname(import.meta.url)
                             ^
```
### Workarrownd
We can workarrownd it with backticks
```coffee
dirname = path.dirname `import.meta.url`
```
...however it is not good to the language.

### Context
`import.meta.url` is necessary source of information to replace `__(dir|file)name` features.

### Environment
* CoffeeScript version: 2.5.1
* Node.js version: 14.1.0

aurium:
Allows to access `import.meta` and only if its property is `meta`.

The new token `IMPORT_META` was created to allow richer formatting and code analysis.
```
$ ./bin/coffee --tokens -e 'import.meta.something'
[IMPORT_META import] [. .] [PROPERTY meta] [. .] [PROPERTY something] [TERMINATOR \n]
```
closes #5317


vendethiel:
Maybe not important, but this breaks indented property access like
```coffeescript
a = import
 .meta

# or (forgive me...)
a = import.
  meta
```

aurium:
@vendethiel, now it can. Thanks!

GeoffreyBooth:
> this breaks indented property access

The same is true of `new.target`, so if we allow it for `import.meta` we should allow it there too.

I'm conflicted, as these aren't actually properties of an object the way that `obj.prop` is. `import` isn't an object, nor is it a function (even though it can appear like one e.g. `import()`). It's just a weird piece of syntax. Though I suppose the 'property access' is a part of grammar, and `import.meta`/`new.target` uses that grammar, so maybe it should follow that pattern?

src/grammar.coffee
```diff
@@ -396,6 +396,7 @@ grammar =
   # A "meta-property" access e.g. `new.target`
   MetaProperty: [
     o 'NEW_TARGET . Property',                  -> new MetaProperty LOC(1)(new IdentifierLiteral $1), LOC(3)(new Access $3)
+    o 'IMPORT_META . Property',                 -> new MetaProperty LOC(1)(new IdentifierLiteral $1), LOC(3)(new Access $3)
```
helixbass, line 399: Re your question in the issue thread re `LOC()`, it is used to make sure that newly created node class objects get correct location data assigned to them:

By default, the grammar will assign the location data spanned by *all* of the left-hand-side tokens to the "top-level" node returned by the grammar rule. So eg here, the `new MetaProperty()` will be assigned the correct location data corresponding to eg `import.meta` in the source code

But for "inner" node class objects created by grammar rules, they won't get correct location data assigned to them by default. So eg here in order for the `new IdentifierLiteral()` to get assigned the location data corresponding to `import` in the source code, we use `LOC(1)(new IdentifierLiteral ...)` to say "assign the location data of the *first* token of this grammar rule's left-hand-side (ie `IMPORT_META`) to this `new IdentifierLiteral()`"

You can see the implementation of `LOC()` in the `o()` function at the top of `src/grammar.coffee`, which does a little magic to allow the grammar rules to output Jison-compatible rules

src/lexer.coffee
```diff
@@ -186,7 +186,10 @@ exports.Lexer = class Lexer
       else if tag is 'UNLESS'
         tag = 'IF'
       else if tag is 'IMPORT'
-        @seenImport = yes
+        if match.input.match /^import\s*\.\s*[_a-zA-Z$][_a-zA-Z$0-9]*/
```
helixbass, line 189: Rather than "lexing ahead" via regex (which is best avoided if possible), I'd try and mimic how eg `NEW_TARGET` works and "retroactively adjust" to update a previously seen `IMPORT` token to instead be an `IMPORT_META` token

So ya if you grep `NEW_TARGET` in `src/lexer.coffee` you can see the pattern for that

test/modules.coffee
```diff
@@ -941,3 +941,57 @@ test "#4834: dynamic import", ->
...
+        .meta
+        .bar
+  """,
+  """
+    var foo;
+
+    foo = import.meta.bar;
+  """
+
+  throwsCompileError """
```
helixbass, line 995: This would be better expressed as a test in `test/error_messages.coffee` where you are specifying the exact error message thrown by the compiler

See the tests near the bottom of that file for `new.target` for examples - it looks to me like the `new.target outside of a function` test wouldn't apply to `import.meta`, but that the `new.target is only allowed meta property` (which is what you're testing here) and `new.target cannot be assigned` tests would apply to `import.meta`

vendethiel:
@aurium I meant the approach in general is different from how this is usually done. Might not matter in this specific case (because it's not gonna be an issue in practice), but you're always gonna break some code if you "revert" to regex scanning.

```coffeescript
b.
  # Here's a nice comment about the meta property
 c
```

src/lexer.coffee
aurium: Done!

test/modules.coffee
aurium: Done. Thanks!

GeoffreyBooth:
@helixbass I added your excellent [explanation](https://github.com/jashkenas/coffeescript/pull/5319#discussion_r423023138) as a comment in afab371 on this branch.

@aurium I merged the latest `master` into this branch. This adds a new and improved CI that includes testing the browser compiler, and uses GitHub Actions.

Have all of the review notes been addressed?

aurium:
I believe... yes.

GeoffreyBooth:
> I believe... yes.

Great! @helixbass and @vendethiel, do you have any more notes?

vendethiel:
LGTM

helixbass:
LGTM

gcxfd:
when where release this version ?

PR #5324, opened 2020-06-02
Fix call MetaProperty

### LINKED ISSUE #5323: Bug: import() and import.meta should be expressions
Neither of the following lines compile, but they both should:

```coffee
console.log import.meta.url
console.log import('./file.js')
```

```
[stdin]:1:13: error: unexpected import
console.log import.meta.url
            ^^^^^^
```

cc @aurium

GeoffreyBooth:
Fixes #5323. @aurium @helixbass


GeoffreyBooth:
@helixbass good catch, thanks. Added.

PR #5327, opened 2020-06-26
Remove Google Analytics

resynth1943:
Would you be willing to remove this Google Analytics 'functionality' from this website? There are better alternatives if you _must_ keep tabs on users' activity.

Some reputable alternatives can be seen here: https://switching.software/replace/google-analytics/

### DIFF
--- a/documentation/site/docs.coffee
+++ b/documentation/site/docs.coffee
@@ -1,17 +1,6 @@
 unless window.location.origin # Polyfill `location.origin` for IE < 11
   window.location.origin = "#{window.location.protocol}//#{window.location.hostname}"
 
-
-# Initialize Google Analytics
-window.GA_TRACKING_ID = 'UA-106156830-1'
-window.dataLayer ?= []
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
Sorry, but since we rely on GitHub Pages to host our site, we can't run custom analytics software like any of the alternatives you suggest. Google Analytics is one of the few insights we have into the CoffeeScript user base, so it would be a great loss for us to remove it.

resynth1943:
> Sorry, but since we rely on GitHub Pages to host our site, we can't run custom analytics software like any of the alternatives you suggest. Google Analytics is one of the few insights we have into the CoffeeScript user base, so it would be a great loss for us to remove it.

Excuse me, but why don't you try GoatCounter which is hosted, and allows **FREE ANALYTICS**?

GeoffreyBooth:
It wouldn't be free for CoffeeScript as we get more than 100k pageviews per month.

PR #5349, opened 2021-03-18
Fix documentation links to try/catch section

edemaine:
* Consistently use `#try-catch` for the section (and `#try` for the demo).
* The table of contents actually lacked this entry (and labeled `switch` to include it), making it hard to find.
* And another section linked to `#try` which went to the wrong place.
* This has been an issue for a long time, so I fixed both v1 and v2 docs.


GeoffreyBooth:
The `docs` folder is the generated output. You need to update the source files in `documentation`, e.g. https://github.com/jashkenas/coffeescript/blob/master/documentation/site/sidebar.html.

edemaine:
@GeoffreyBooth Oops, sorry about that. I'm obviously still new to the build process...

I've revised the source files to match.  I confirmed that `cake doc:site` produces the same HTML files that I originally committed and tested.

edemaine:
Just bumped into this again. Let me know if there's anything else that needs to be done for this to be merged.

GeoffreyBooth:
> Just bumped into this again. Let me know if there’s anything else that needs to be done for this to be merged.

Please just rebase or merge `master` in so that we get a passing CI.

PR #5353, opened 2021-04-22
Handle triple-quoted JSX attribute values

### LINKED ISSUE #5352: Bug: Tripple-double-quotes on CSX props produce invalid JSX
### Input Code
[#try](http://coffeescript.org/#try:example%20%3D%20%3CExample%20style%3D%22%22%22%0A%20%20color%3A%20blue%3B%0A%22%22%22%20%2F%3E)
```coffee
example = <Example style="""
  color: blue;
""" />
```
Same applies to
```
example = <Example style='''
  color: blue;
''' />
```

This only happens when there's no string interpolation. The following [compiles](http://coffeescript.org/#try:example%20%3D%20%3CExample%20style%3D%22%22%22%0A%20%20color%3A%20%23%7Bcolor%7D%3B%0A%22%22%22%20%2F%3E) as [expected](https://babeljs.io/repl#?code_lz=G4QwTgBApgHiC2AHANlA3AKA7BKoQF4IAeAUTiVQgGcAXAT1QIG8ADAYwHtlOwAuCABJmXHmAC-aVuIgB6AHyYgA&presets=react):
```coffee
example = <Example style="""
  color: #{color};
""" />
```

### Expected Behavior
Compiler additionally wraps the string in curlies:
```js
var example;

example = <Example style={`color: blue;`} />;
```

### Current Behavior
[#babel](https://babeljs.io/repl#?code_lz=G4QwTgBApgHiC2AHANlA3AKA7BKoQF4IAeAUTiVQgGcAXAT1QIAMBjAe2XbAC4IAjZAFd0zCAHoAfJiA&presets=react) (terrible error message btw)
```
JSX value should be either an expression or a quoted JSX text (3:25)

  1 | var example;
  2 |
> 3 | example = <Example style=`color: blue;` />;
    |                          ^
  4 |
```

### Possible Solution
Either give an explicit compile error or support the users' *stupidity*.
I suspect that the compiler currently checks if there are any interpolations instead of checking which quotes apply. Could be this was introduced when tripple-double-quotes (and tripple-single-quotes) output was changed to output as multi-line string literals.

### Context
Gave me a compile error later in the toolchain than necessary.
Seeing as the workaround is so easy (using curlies or single double-quotes) i consider this quite low priority.

### Environment
* CoffeeScript: v2.5.1
* Node.js: v14.16.1
* Babel: v6.23.0
* "@babel/plugin-transform-react-jsx": "7.13.12"

helixbass:
Fixes #5352 

Fixes issue where triple-quote (non-interpolated) JSX attribute values weren't getting transpiled correctly into a JSX expression container wrapping the template literal (eg ``<div a={`...`} />``)


GeoffreyBooth:
Thanks! Sorry for the delay.

PR #5358, opened 2021-07-12
Add type definitions for TypeScript consumers

4086606:
I don't think this requires CI, but someone more familiar with the AST and parser should check this


coffeescript.d.ts
```diff
@@ -0,0 +1,247 @@
...
+export interface CoffeeScriptASTBody {
+  classBody: boolean;
+  expressions: CoffeeScriptASTNode[] | [];
+  locationData: CoffeeScriptASTNodeLocationData;
+}
+
+/**
+ * List of precompiled CoffeeScript file extensions.
+ */
+export let FILE_EXTENSIONS: string[];
```
GeoffreyBooth, line 92: This is a known list: `.coffee`, `.coffee.md`, `.litcoffee`.

coffeescript.d.ts
```diff
@@ -0,0 +1,247 @@
...
+ */
+export let VERSION: string;
+
+/**
+ * Helpers used internally to compile CoffeeScript.
+ *
+ * @deprecated for external use and lacking TypeScript definitions.
+ */
+export interface helpers {
+  [handle: string]: (...args: any[]) => any;
```
GeoffreyBooth, line 105: Why not define each helper?

coffeescript.d.ts
4086606: Sure, this list seems highly stable. Let's hardcode that!

coffeescript.d.ts
4086606: 😱 29 helpers, not all documented. I couldn't automate the process and couldn't do it all by hand

4086606:
I'll add dtslint and helper definitions today. It might be worth adding the helper JSDOC from the definitions back into helpers.coffee, I can do that as well

coffeescript.d.ts
4086606: @GeoffreyBooth done in bd73064

4086606:
@aurium @helixbass

package.json
```diff
@@ -45,12 +47,15 @@
...
     "highlight.js": "~9.18.1",
     "jison": "~0.4.18",
     "markdown-it": "~10.0.0",
     "puppeteer": "~3.1.0",
     "underscore": "~1.9.2",
     "webpack": "~4.42.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@types/babel-core": "^6.25.7"
```
GeoffreyBooth, line 59: It’s a goal of the project to ship with no dependencies.

4086606: @GeoffreyBooth removed the dependency in d15ab8c

4086606:
@GeoffreyBooth is there anything else I should address?

phil294:
@4086606 Maybe I am doing something wrong, but I don't think your PR is actually working? When I do `npm i 'jashkenas/coffeescript#pull/5358/head'` and then import CoffeeScript in a TS or JS file, there are no typings available. The `/types` directory is missing, there is still only `/bin` and `/lib`.

4086606:
Try again

phil294:
> Try again

No, that did not help, unfortunately. I have got your latest changes, but there are no types included:
pic
I guess you need to add it to the `"files"` array (?)

4086606:
# 😩

types/index.d.ts
```diff
@@ -0,0 +1,784 @@
...
+ * @param {boolean} [options.ast=false] If true, output an abstract syntax tree of the input CoffeeScript source code.
+ * @param {boolean} [options.bare=false] If true, omit a top-level IIFE safety wrapper.
+ * @param {string} [options.filename=index.js] File name to compile.
+ * @param {boolean} [options.header=false] If true, output the `Generated by CoffeeScript` header.
+ * @param {boolean} [options.inlineMap=false] If true, output the source map as a base64-encoded string in a comment at the bottom.
+ * @param {boolean} [options.sourceMap=false] If true, output a source map object with the code.
+ * @param {babel.TransformOptions} [options.transpile={}] Babel transpilation options - see `babel.TransformOptions`.
+ * @returns {string} Compiled and unevaluated JavaScript code.
+ */
+export function compile(code: string, options?: Options): string;
```
phil294, line 664: One more issue I found, [according to the docs](https://coffeescript.org/#nodejs-usage), this can also return a `{js, v3SourceMap, sourceMap}` if `options: sourceMap` is `true`

phil294 (changes requested):

types/index.d.ts
4086606: Resolved in e55967d @phil294

helixbass:
Is it a better pattern to bundle manually-written type definitions in a package like this vs use [DefinitelyTyped](https://definitelytyped.org/) (ie `@types/*` packages)? I don't know what's conventional, off the top of my head it seems like publishing separately via DefinitelyTyped would give more flexibility because then you don't need to push out a new Coffeescript release in order to update the type definitions

4086606:
Oh god that sounds like hell on earth.. sure I'll move things there. Thankfully dtslint is passing already and they should merge very quickly. It also won't be necessary to register the typings as a dependency here.

GeoffreyBooth:
> publishing separately via DefinitelyTyped would give more flexibility because then you don’t need to push out a new Coffeescript release in order to update the type definitions

Yes, and that way you can define type definitions for past releases of CoffeeScript, since I can’t go back and modify those now.

Issue #5356, opened 2021-06-21
support for import.meta


johndeighan:
bug report - it's not listed under **Unsupported ECMAScript Features**

### Input Code

```coffee
import { dirname } from 'path';
import { fileURLToPath } from 'url';

__dirname = dirname(fileURLToPath(import.meta.url));
```

### Expected Behavior

This should give us the equivalent to __dirname in common js

### Current Behavior

Error message for the period after **import** in the last line

### Possible Solution

Should work as it does in ECMAScript, or at least be listed as not compatible

### Context

As a workaround, I put the **import.meta.url** inside backquotes so coffeescript uses it as is.
But I'm not sure yet whether that always works.

### Environment

* CoffeeScript version: CoffeeScript version 2.5.1
* Node.js version: v16.2.0

wdavidw:
This is the first time in 10 years that I am using the backquote behavior, thanks for the temporary solution.

edemaine:
It looks like this bug was fixed by #5324 (June 3, 2020) but that CoffeeScript hasn't released since then (v2.5.1 was released on NPM on January 31, 2020). Probably worth a new release?

GeoffreyBooth:
Yeah. There’s also top-level `await` that we need to support.

GeoffreyBooth:
Resolved in #5374.

PR #5371, opened 2021-09-13
Support top-level await

### LINKED ISSUE #5354: deno can await outside function, how to remove " error: await can only occur inside functions   "
deno can await outside function, how to remove " error: await can only occur inside functions   "

edemaine:
@GeoffreyBooth [mentioned](https://github.com/jashkenas/coffeescript/issues/5356#issuecomment-917525888) that CoffeeScript is missing support for [top-level await](https://v8.dev/features/top-level-await). This PR adds that support.

Mostly this involves doing fewer checks for invalidity.  The only hard part is when the code is wrapped in a top-level IFFE, where we need to detect whether that wrapper should be marked `async`.

Fixes #5354 (request from Deno side).

The REPL already supported top-level `await` by explicit wrapping in an `async function` closure. I think it's best to keep that code, because older versions of Node don't support top-level `await`.


test/async.coffee
```diff
@@ -204,3 +204,21 @@ test "#3199: await multiline implicit object", ->
...
     eq undefined, y
+
+test "top-level await", ->
+  eq CoffeeScript.compile('await null'), '''
+    (async function() {
+      await null;
+
+    }).call(this);
+
+  '''
```
GeoffreyBooth, line 215: We have the `eqJS` helper that’s more concise, and does `compile` with `bare`:

```suggestion
  eqJS ‘await null’, ‘await null;’
```

test/async.coffee
```diff
@@ -204,3 +204,21 @@ test "#3199: await multiline implicit object", ->
...
+  eq CoffeeScript.compile('-> await null'), '''
+    (function() {
+      (async function() {
+        return (await null);
+      });
+
+    }).call(this);
+
+  '''
```
GeoffreyBooth, line 224: I would change this to be a bit more realistic:

```suggestion
  eqJS 'do -> await null', '''
    (async function() {
      return (await null);
    })();
  '''
```

But it doesn’t belong in a “top level await” section, as this `await` isn’t top-level.

GeoffreyBooth (commented):
Thank you for doing this! A few minor notes and then happy to land.

test/async.coffee
edemaine: Ah, thanks, I figured you must have a helper like that.

I think there's value in testing both the bare and non-bare case, given that I had to write code to make the wrapper `async`.  Do you agree?  If so, should I include both tests like this, or make/modify helper to support nonbare?

test/async.coffee
edemaine: What I meant to test here is that the wrapper is *not* marked `async` (in non-bare mode). I can give this test a better (and separate) name though.

test/async.coffee
GeoffreyBooth: Let’s make separate tests for the non-bare wrapper, I think that’s worth calling out. I had no idea that’s what you were testing here.

I don’t think we need to modify the helper just for this one case. You can just use `CoffeeScript.compile` to test the wrapper.

edemaine:
I've updated the tests to

* use `eqJS`
* extend `eqJS` to allow specifying compile options
* label tests better as to what's being tested

Happy to make further changes.

edemaine:
Further revised the tests to check what they actually care about, which is a `startsWith` prefix.  `eqJS` is back to the original.

GeoffreyBooth (approved):

GeoffreyBooth:
Thanks. I want to leave this open a little bit in case anyone else wants to review (@helixbass?) and then will land and cut a new release.

I added https://github.com/jashkenas/coffeescript/pull/5372 to add Node 16 support to CI but the tests fail there, so that should also probably be addressed before we land this. cc @helixbass on that too as it looks like it has to do with AST stuff.

GeoffreyBooth:
Got CI fixed. Can you please rebase off of `master` or merge `master` into your branch? That should trigger a new test run that’ll include Node 16.

FYI besides rebuilding the regular compiler we also need to build the browser compiler (`cake build:browser`). We should add a CI step that rebuilds everything and then checks that there’s nothing different from the last git commit (which, if there was a difference, would mean that the author forgot to rebuild and commit one of the outputs).

PR #5373, opened 2021-09-16
ci: Check that build steps didn’t change any files

edemaine:
@GeoffreyBooth wrote:

> We should add a CI step that rebuilds everything and then checks that there’s nothing different from the last git commit (which, if there was a difference, would mean that the author forgot to rebuild and commit one of the outputs).

This PR adds this CI step, which turned out to be easy, because all the building was already getting tested.

For some reason, the browser build is different on Node 6.x and 8.x (in my CI testing). So we don't run the diff in those cases, which seems OK to me.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -43,3 +43,6 @@ jobs:
     - run: node ./bin/cake build:parser
     - run: node ./bin/cake build:full
     - run: node ./bin/cake build:browser
+
+    - run: git diff --exit-code
+      if: matrix.node-version != '6.x' && matrix.node-version != '8.x'


GeoffreyBooth:
Awesome! Yeah, Node 6 and 8 (and even 10) have been EOL for so long now that maybe we should take them out. I was thinking of getting the next release out first though before considering that.

Have you tested this? Like by pushing a branch (or opening a PR, if pushing a branch isn’t enough) where you change something that breaks tests, and don’t run `cake:build` or `cake build:browser`, to see if this catches it?

edemaine:
I tested in the following way: before I excluded Node 6 and 8, they generated different code from `cake build:browser`, and CI correctly failed.  Maybe you can see the workflow run here: https://github.com/edemaine/coffeescript/actions/runs/1239714992
If not, here's what it looks like:

image

image

Maybe not the best error messages (diffs for the minified browser compiles don't look good -- it's possible we should change diff formats) but at least it correctly failed.  I can test more if you'd like (or feel free to push a change, e.g. to a `src/*.coffee` file that doesn't correctly end up in the compile versions, and then revert).

> where you change something that breaks tests, and don’t run `cake:build` or `cake build:browser`, to see if this catches it?

I think you're right that, if tests fail, then `cake:build` and `cake build:browser` and `git diff --exit-code` won't run.  But presumably in that scenario there would be another push that fixes the tests, and then everything will rerun?  Or am I missing a use-case for running the build and diffs even when the tests fail?

GeoffreyBooth:
> Or am I missing a use-case for running the build and diffs even when the tests fail?

I’m trying to prevent what happened with https://github.com/jashkenas/coffeescript/pull/5324, where we merged in without running `cake build:browser` and so when I added Node 16 to CI, the tests for 16 were failing because they used the browser build which hadn’t been updated.

I think what you have is probably fine; maybe it’s not actually necessary because the tests themselves will catch it? But yeah _somehow_ we were able to get a passing CI in https://github.com/jashkenas/coffeescript/pull/5324 even though the browser output hadn’t been rebuilt. In https://github.com/jashkenas/coffeescript/pull/5372 I both added Node 16 and rebuilt the browser compiler, and it was the latter that got CI passing again.

edemaine:
Yes, this PR is designed to fix exactly this issue.  Before this PR, CI ran `cake build:browser` to check that it *could successfully built*, but then threw away any generated files.  After this PR, CI subsequently checks *whether any built files have changed* (via `git diff`, so anything checked into Git), and fails if so.

So if you forget to check in the result of a build process, the test will see that the new code successfully builds, but produces different results, and fail overall.

GeoffreyBooth:
Awesome, thanks!

PR #5375, opened 2021-09-20
Improve documentation of `jumps` in comments

edemaine:
This extracts the comment fixes from #5348, plus some additional rewording, to make clear that `x = (throw 1)` is allowed, while `x = (break)` isn't.  A small thing, but at least I can get something useful out of #5348. 🙂

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -392,11 +392,12 @@ exports.Base = class Base
   # has special awareness of how to handle comments within its output.
   includeCommentFragments: NO
 
-  # `jumps` tells you if an expression, or an internal part of an expression
-  # has a flow control construct (like `break`, or `continue`, or `return`,
-  # or `throw`) that jumps out of the normal flow of control and can’t be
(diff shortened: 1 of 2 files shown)


src/nodes.coffee
```diff
@@ -392,11 +392,12 @@ exports.Base = class Base
...
-  # has a flow control construct (like `break`, or `continue`, or `return`,
-  # or `throw`) that jumps out of the normal flow of control and can’t be
-  # used as a value. This is important because things like this make no sense;
-  # we have to disallow them.
+  # `jumps` tells you if an expression, or an internal part of an expression,
+  # has a flow control construct (like `break`, `continue`, or `return`)
+  # that jumps out of the normal flow of control and can’t be used as a value.
+  # (Note that `throw` is not considered a flow control construct.)
+  # This is important because flow control in the middle of an expression
+  # make no sense; we have to disallow them.
```
GeoffreyBooth, line 400:
```suggestion
  # makes no sense; we have to disallow it.
```

PR #5380, opened 2021-10-02
Fix #5378, stack overflow on large files

### LINKED ISSUE #5378: coffee -c fails on moderately large .coffee files in v2.6.0 (worked in 2.5 and earlier)
Compilation (via `coffee -c`) in v2.6.0 seems to fail on large(-ish) .coffee files that worked in v2.5.1 and previous releases. (And that are well within a range one could reasonably expect to work IMO.)

I have an electron app implemented in CoffeeScript. As part of my build process I concatenate a large number of .coffee files into a single renderer.coffee, and then compile with `coffee -c renderer.coffee` to generate `renderer.js`, which is the file loaded by my root HTML page in electron.

This process has worked well for a couple of years at least. However when I attempted to upgrade to CoffeeScript v2.6 I found that `coffee -c renderer.coffee` is now failing with an error like the following

```
RangeError: Maximum call stack size exceeded
    at Root.compileNode ([...]/node_modules/coffeescript/lib/coffeescript/nodes.js:740:15)
    at Root.compileToFragments ([...]/node_modules/coffeescript/lib/coffeescript/nodes.js:143:74)
    at Object.<anonymous> ([...]/node_modules/coffeescript/lib/coffeescript/coffeescript.js:166:23)
    at Object.<anonymous> ([...]/node_modules/coffeescript/lib/coffeescript/coffeescript.js:54:19)
    at Object.CoffeeScript.compile ([...]/node_modules/coffeescript/lib/coffeescript/index.js:45:29)
    at compileScript ([...]/node_modules/coffeescript/lib/coffeescript/command.js:293:33)
    at compilePath ([...]/node_modules/coffeescript/lib/coffeescript/command.js:237:14)
    at Object.exports.run ([...]/node_modules/coffeescript/lib/coffeescript/command.js:158:20)
    at Object.<anonymous> ([...]/node_modules/coffeescript/bin/coffee:22:45)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    [...]
```

I believe I've discovered a relatively simple example that reproduces this problem at just under 7000 lines.

Take the following script:

```coffeescript
# FILE: large-coffee-generator.coffee
max = 6969
console.log "i = 0"
for i in [0...max]
  console.log "console.log \"Line \#{i++}\""
```

Running that program (e.g., `coffee large-coffee-generator.coffee > foo.coffee`) will generate a file that looks something like:

```coffeescript
# FILE: foo.coffee
i = 0
console.log "Line #{i++}"
# ...etc....
console.log "Line #{i++}"
```

where that last line is repeated `max` times for a total of `max + 1` lines.

Now try to compile that output via `coffee -c foo.coffee`

 * With CoffeeScript v2.5.1 this works (somewhat slowly) with `max` values of 40,000 or more.

 * With CoffeeScript v2.6.0 this works for `max = 6968` - running in about 8s in on my MacBook Pro laptop - but fails for `max = 6969`. I.e. it starts to fail with a "Maximum call stack size exceeded" error at precisely 6970 lines.

For context:

 - I'm running this on a modernish MacBook Pro: 2 GHz Quad-Core Intel Core i5 with 32 GB RAM

 - I'm using Node v14.16.0 (the latest stable release)

 - I'm using CoffeeScript v2.6.0 (the latest stable release)

GeoffreyBooth:
See https://github.com/jashkenas/coffeescript/issues/5378#issuecomment-932688592


edemaine:
It looks like the browser compiler needs to be rebuilt.  (Maybe I need to improve this error message...)

GeoffreyBooth:
@edemaine Done. Otherwise does this look good?

edemaine:
Yes, it looks good to me (see other thread for details).

PR #5387, opened 2021-11-30
Replace broken link in docs

benhalpern:
I noticed a broken link in the polyfill section of the docs that attempts to link to a post getting into Polyfills for general information, but that post redirects to the home page of Hackernoon, so seems to have been deleted.

I replaced it with a helpful and informative DEV post that seems to accomplish similar goals for the link. Whether or not this is _the perfect replacement_, it is a good post, and is definitely an upgrade from the current broken link.

Small fix, have a nice day!


GeoffreyBooth:
The original article appears to also be at https://medium.com/hackernoon/polyfills-everything-you-ever-wanted-to-know-or-maybe-a-bit-less-7c8de164e423. I’m also happy to include three links.

To be mergeable, you need to also run `cake doc:site` and commit the changed files.

Issue #5377, opened 2021-09-20
Enhancement: Output let and const for variable declarations, without breaking changes


GeoffreyBooth:
Migrating this from https://github.com/jashkenas/coffeescript/issues/5344#issuecomment-922667010:

Now that CoffeeScript 2 outputs modern ES6+ syntax, we can use `let` and `const` in our output. Leaving aside the question of supporting block-scoped variables in the CoffeeScript _input,_ (see #4985) there’s no reason that our generated output needs to use `var` when it could use `let` or `const` instead. We could therefore make two improvements in output code readability:

1. Whenever possible, variable declarations should be placed at the block scope (via `let`) rather than always at the top of the function scope (the current behavior with `var`).

1. Whenever possible, variable declarations should use `const`. This would be whenever a variable is never reassigned.

The key here is to _not_ cause any breaking changes. So when in doubt, we would keep the current “define at the top of the function scope” behavior. A [declaration/first assignment within a loop](https://github.com/jashkenas/coffeescript/issues/5344#issuecomment-922636634) is a example of such a case. A `let x` within a loop means that `x` is declared multiple times (once for each iteration of the loop) which might be a breaking change for existing code.

If you look at this example:

```coffee
if new Date().getHours() < 9
  breakfast = if new Date().getDay() is 6 then 'donuts' else 'coffee'
  alert "Time to make the #{breakfast}!"
else
  alert 'Time to get some work done.'
```

`breakfast` is only used within that `if` block, but it’s currently getting a `var breakfast;` line at the top of the output. The output instead could be this:

```js
if (new Date().getHours() < 9) {
  let breakfast;

  breakfast = new Date().getDay() === 6 ? 'donuts' : 'coffee';
  alert(`Time to make the ${breakfast}!`);
} else {
  alert('Time to get some work done.');
}
```

And then phase 2 would see that `breakfast` is never reassigned _and_ never referenced before its assignment (read about the `let`/`const` [“temporal dead zone”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#temporal_dead_zone_tdz)) and use `const` instead:

```js
if (new Date().getHours() < 9) {
  const breakfast = new Date().getDay() === 6 ? 'donuts' : 'coffee';
  alert(`Time to make the ${breakfast}!`);
} else {
  alert('Time to get some work done.');
}
```

I would try to achieve the two enhancements as separate PRs, probably with the block-scoping first.

[...]

PR #5395, opened 2022-01-24
Push var down to first variable use when possible

edemaine:
This is a portion of my `typescript` branch that I spent the last few days working on, implementing the plan laid out in https://github.com/jashkenas/coffeescript/issues/5377#issuecomment-1019537087 : **push the `var` declaration of a variable down to its first clean assignment** if there is one.  Examples of clean assignments:

* `x = 7` &rarr; `var x = 7`
* `[x, y] = array` &rarr; `var [x, y] = array`
* `{x, y} = point` &rarr; `var {x, y} = point`
* `[{x, y}, {a, b}] = [point, pair]` &rarr; `var [{x, y}, {a, b}] = [point, point2];`

Unexamples of unclean assignments (so `var` won't try to be inserted here; if this is the only assignment, `var` will remain at top):

* `[{x, y}, {x: x2, y: y2}] = [point, point2]`
* `{x = 5} = obj`
* `x += 5`
* `if match = re.exec string`
* `console.log x = 5`

Note for reviewer: "clean" is referred to as `isDeclarable` in the code.

Only the first assignment to a variable is modified.  The only exception is if there is a joint assignment that has *some* undeclared variables in addition to some already declared variables. Some examples:

<table>
<tr>
<th>CoffeeScript
<th>JavaScript output from this PR

<tr>
<td>

```coffee
x = 5
x = 10
```

<td>

```js
var x = 5;
x = 10;
```

<tr>
<td>

```coffee
x = 5
[x, y] = array
```

<td>

```js
var x = 5;
var [x, y] = array;
```

</table>

While the duplicate `var` isn't pretty, I think it's rare, and [it's allowed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#description). `eslint` would complain about the output, but `eslint` complains about all sorts of things in CoffeeScript's output; it's better to run `eslint` on the CS AST.

There's also special handling of `for` loops and ranges.  (This took the longest to get right.)  Examples:  (comments added for explanation)

<table>
<tr>
<th>CoffeeScript
<th>JavaScript output from this PR

<tr>
<td>

```coffee
for x from it
  console.log x
```

<td>

```js
for (var x of it) {
  console.log(x);
}
```

<tr>
<td>

```coffee
for x, y of object
  console.log x, y
```

<td>

```js
for (var x in object) {
  var y = object[x];
  console.log(x, y);
}
```

<tr>
<td>

```coffee
for x in [1..100]
  console.log x
```

<td>

```js
for (var i = 1, x = i; i <= 100; x = ++i) {
  console.log(x);
}
```

<tr>
<td>

```coffee
for x in array
  console.log x
```

<td>

```js
for (var i = 0, len = array.length; i < len; i++) {
  var x = array[i];
  console.log(x);
}
```

<tr>
<td>

```coffee
for x, i in array
  console.log x
```

<td>

```js
for (var j = 0, i = j, len = array.length; j < len; i = ++j) {
  var x = array[i];
  console.log(x);
}
```

<tr>
<td>

```coffee
squares =
  for x in array
    x ** 2
```

<td>

```js
var x;  // needs to stay here so visible outside loop

var squares = (function() {
  var results = [];
  for (var i = 0, len = array.length; i < len; i++) {
    x = array[i];
    results.push(x ** 2);
  }
  return results;
})();
```

<tr>
<td>

```coffee
powers =
  for x, i in array
    x ** i
```

<td>

```js
var i, x;  // needs to stay here so visible outside loop

var powers = (function() {
  var j, len;  // can't declare in for loop or would shadow i
  var results = [];
  for (j = 0, i = j, len = array.length; j < len; i = ++j) {
    x = array[i];
    results.push(x ** i);
  }
  return results;
})();
```

</table>

As you can see in the more complicated examples, there are some tricky issues where it's crucial to avoid shadowing by excess `var`.  I *think* I got them all right; it was very hard to pass all the tests.  But it's possible that more tests should be done/added.

This PR also refactors `Scope` to ensure all variables have `positions` mapping,
so we can do faster searching for variables, and a new `get` method for retrieving various data about a variable. This makes it easier to keep track of whether a variable has already had a `var` declaration during some assignment.


edemaine:
There are some browser tests failing, caused by [this bug in babel-plugin-minify-dead-code](https://github.com/babel/minify/issues/1021). I guess `var` is tricky to implement after all. 😦

src/nodes.coffee
```diff
@@ -2336,13 +2348,37 @@ exports.Range = class Range extends Base
...
     namedIndex = idxName and idxName isnt idx
-    varPart  =
-      if known and not namedIndex
-        "var #{idx} = #{@fromC}"
+    if o.declare  # from compileArray: need to declare all variables as local
+      declare = yes
+    else
+      # Can do var declaration only if all relevant variables are in this scope.
+      # If we declare @fromVar, @toVar, or @stepVar, they're guaranteed
+      # to be in scope (generated in @compileVariables), so don't check those.
```
GeoffreyBooth, line 2356: The comments become the annotated source, so please always use curly quotes and backticks, and proper English:

```suggestion
      # We can output `var` declarations only if all relevant variables are in this scope.
      # If we declare `@fromVar`, `@toVar`, or `@stepVar`, they’re guaranteed
      # to be in scope (generated in `@compileVariables`), so don’t check those.
```

And so on for all the comments. The logic you’re adding in this PR is quite difficult to follow (at least by me) and excessive documentation of what it’s doing would be really good to have.

src/nodes.coffee
```diff
@@ -2336,13 +2348,37 @@ exports.Range = class Range extends Base
...
+      # Want to do var declaration if at least one of those variables needs
+      # declaration. Also keep track of which variables need declaration.
+      if canDeclare
+        declare = yes if o.scope.laterVar idx
+        declare = yes if namedIndex and o.scope.laterVar idxName
+        if declare
+          o.scope.laterVar @fromVar if @fromC isnt @fromVar
+          o.scope.laterVar @toVar if @toC isnt @toVar
+          o.scope.laterVar @stepVar if @stepC isnt @stepVar
+    varPart = if declare then "var " else ''
```
GeoffreyBooth, line 2368:
```suggestion
    declarationPrefix = if declare then 'var ' else ''
```

In case this ever becomes `const`.

src/nodes.coffee
```diff
@@ -2336,13 +2348,37 @@ exports.Range = class Range extends Base
...
+    if namedIndex
+      if declare
+        varPart += ", #{idxName} = #{idx}"
       else
-        "#{idx} = #{@fromC}"
+        varPart = "#{idxName} = #{varPart}"
     varPart += ", #{@toC}" if @toC isnt @toVar
     varPart += ", #{@step}" if @step isnt @stepVar
+    #if known and not namedIndex or o.scope.laterVar idx
+    #  varPart = "var #{varPart}"
```
GeoffreyBooth, line 2381: Should this be removed?

src/nodes.coffee
```diff
@@ -3303,11 +3346,13 @@ exports.ExportDeclaration = class ExportDeclaration extends ModuleDeclaration
     code.push @makeCode "#{@tab}export "
     code.push @makeCode 'default ' if @ instanceof ExportDefaultDeclaration
 
-    if @ not instanceof ExportDefaultDeclaration and
+    exportDefault = @ instanceof ExportDefaultDeclaration
+    if not exportDefault and
```
GeoffreyBooth, line 3350: Why this change?

src/nodes.coffee
```diff
@@ -3547,14 +3593,23 @@ exports.Assign = class Assign extends Base
...
           commentsNode.comments = name.comments
           commentFragments = []
           @compileCommentFragments o, commentsNode, commentFragments
-          o.scope.comments[name.value] = commentFragments
+          o.scope.comment name.value, commentFragments
+
+  makeDeclaration: (o) ->
+    # Consider adding 'var' prefix to this assignment.  Only works at top level
+    # and if LHS is a valid declaration, and assignment is '=' not another op.
+    return unless o.level == LEVEL_TOP and not o.undeclarable and
```
GeoffreyBooth, line 3608:
```suggestion
    return unless o.level is LEVEL_TOP and not o.undeclarable and
```

src/nodes.coffee
```diff
@@ -3596,11 +3652,13 @@ exports.Assign = class Assign extends Base
       return compiledName.concat @makeCode(': '), val
 
     answer = compiledName.concat @makeCode(" #{ @context or '=' } "), val
-    # Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration,
+    if @declaration
+      [@makeCode("var "), ...answer]
```
GeoffreyBooth, line 3656:
```suggestion
      [@makeCode('var '), ...answer]
```

src/scope.litcoffee
```diff
@@ -16,7 +16,6 @@ that should be output as part of variable declarations.
 
       constructor: (@parent, @expressions, @method, @referencedVars) ->
         @variables = [{name: 'arguments', type: 'arguments'}]
-        @comments  = {}
```
GeoffreyBooth, line 19: Not that this is wrong, but why is this removed? How are we tracking comments now?

src/scope.litcoffee
```diff
@@ -78,11 +77,17 @@ Generate a temporary variable name at the given index.
         else
           "#{name}#{index or ''}"
 
-Gets the type of a variable.
+Gets a variable and its associated data from this scope (not ancestors),
+or `undefined` if it doesn't exist.
```
GeoffreyBooth, line 81: What’s its associated data? Its properties?

src/scope.litcoffee
```diff
@@ -78,11 +77,17 @@ Generate a temporary variable name at the given index.
...
+      get: (name) ->
+        @variables[@positions[name]] if Object::hasOwnProperty.call @positions, name
+
+Gets the type of a variable declared in this scope,
+or `undefined` if it doesn't exist.
 
       type: (name) ->
-        return v.type for v in @variables when v.name is name
-        null
+        @get(name)?.type
```
GeoffreyBooth, line 90: I know this isn’t code that you added, but what’s a variable’s “type”? Can we explain that in this comment?

GeoffreyBooth (commented):
I pulled down this branch and rebuilt the documentation and read those examples. The JavaScript output is much more readable! This is quite an improvement!

I reviewed this mostly for style; I almost don’t know how to review it more deeply than that. The scope and variable tracking parts of this codebase feel the most incomprehensible to me; I know they have comments explaining the logic at a surface level, but the level of documentation feels inadequate for the scale of the complexity here. Whatever extra detail that you can add now that you’ve dived deeply into it and understand it better than I do would be beneficial to future contributors and maintainers.

I merged in latest `main` and rebuilt, and pushed the new commits on your branch. The Node tests pass but I see 9 failures in the browser tests. I saw the note about the Babel plugin bug, but unless you want to push a PR to fix it for them we might need to find to find another solution. Maybe we can disable that plugin somehow while still minifying somehow, either still via Babel or perhaps via a different minification library?

src/scope.litcoffee
```diff
@@ -94,27 +99,61 @@ compiler-generated variable. `_var`, `_var2`, and so on...
...
+Return a list of names of variables declared in this scope.
+Optionally restrict to assigned or unassigned variables.
+
+      declaredVariables: (assigned) ->
+        (v.name for v in @variables when v.type is 'var' and not v.laterVar and
+          switch assigned
+            when true then v.assigned?
+            when false then not v.assigned?
+            else true
+        ).sort()
```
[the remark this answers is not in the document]
vendethiel: Isn’t that switch just `assigned == v.assigned?`?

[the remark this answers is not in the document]
edemaine: The `switch` deals with three cases: `assigned` can be

* `true` meaning "only give me assigned variables" i.e. `when v.assigned?`
* `false` meaning "only give me unassigned variables" i.e. `when not v.assigned?` 
* `undefined` (or absent argument) meaning "give me all variables" i.e. `when true`.

I forget whether the `undefined` case is actually used; if it isn't, we could definitely simplify. But I don't think either simplification above is correct according to the 3-case spec.

Thanks @GeoffreyBooth for the thorough review! I will work on responding and adding a bunch of comments.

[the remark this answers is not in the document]
GeoffreyBooth: It might be going a bit far to do this all in a one-liner. You could maybe have a one-liner that retrieves all variables, then something like “if give me only assigned variables, filter the collection to just those; else if only give me unassigned variables, filter the collection to just those.” Then sort whatever’s left in the collection at this point and return it. Making this a multiline function also gives you opportunities to add comments explaining each branch and why you might want each.

STRd6:
I'm also excited about this, anything I can do to help?

phil294:
I noticed the following translation, not mentioned above or in the test cases:

```coffee
[ x = '', y = '' ] = []
x = x.slice(0)
```
becomes
```js
var y;

[x = '', y = ''] = [];

var x = x.slice(0);
```
is this a bug or intended? If I didn't read you wrong, this should have been `var x, y` instead.

edemaine:
@phil294 I can see why that would happen. The code must be interpreting `var [x = '', y = ''] = [];` as unsafe, so it doesn't do `var` there, leaving `x` and `y` needing declaration elsewhere.  `x = x.splice(0)` is a valid place to put the `x` declaration and so it puts it there.  `y` doesn't have an assignment so it gets hoisted.  This is all valid thanks to automatic hoisting.

However, `var [x = '', y = ''] = [];` is a valid declaration, so I should tweak the test to just put the `var` there.  Some assignments are not valid places to put `var` though (for example, `if x = 5`); in those cases, the next assignment to `x` will be where the `var` ends up.  We could alternatively decide that, if the first assignment isn't a valid place for `var`, then the `var` always gets hoisted to the top -- but that's not how this branch works currently.

phil294:
@edemaine Thanks for the clarification!

---

As you know, this PR branch has been in use in [CoffeeSense](https://github.com/phil294/coffeesense) for several months now, and indeed successfully so.

Today however, I have found an actual bug for the first time (I think). It does not happen with the normal exports, but only with the browser compiler version. Here's reproduction code, which I ran with Deno because Node/modules gave some cryptic errors which I couldn't be bothered with.

```javascript
import { compile } from "./node_modules/coffeescript/lib/coffeescript-browser-compiler-modern/coffeescript.js"
let coffee = 'a = 1\n[1..a]'
let response = compile(coffee, { sourceMap: true, bare: true })
console.log(response.js)
```
output:
```javascript

var a = 1;

(function() {
  var results = [];
  for (var i = 1; 1 <= a ? i <= a : i >= a; 1 <= a ? i++ : i--){ results.push(i); }
  return results;
}).apply(thisundefined);
```
note the `thisundefined` at the end.

Please, there is no need to investigate, this is a rare case with low priority for me or any user, I think. I just thought it might be helpful for developing this branch, should you ever get back to it - maybe it's also already covered above.

Thank you again for your efforts.

PR #5396, opened 2022-01-25
CI: Ensure browser tests are up-to-date

edemaine:
Today I learned about the cake rule `doc:test` which builds the browser tests from the main tests (similar to `build:browser` but for the tests). This PR ensures that those tests are up-to-date before CI can succeed, similar to the other build operations we do.

As a side note, I find the name `doc:test` rather confusing, as it's about building tests not documenting or running them. But that's less important.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -310,7 +310,7 @@ buildDocTests = (watch = no) ->
   # Helpers
   testsInScriptBlocks = ->
     output = ''
-    for filename in fs.readdirSync testsSourceFolder
+    for filename in fs.readdirSync(testsSourceFolder).sort()
       if filename.indexOf('.coffee') isnt -1
         type = 'coffeescript'
       else if filename.indexOf('.litcoffee') isnt -1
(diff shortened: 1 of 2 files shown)


GeoffreyBooth:
I think you also need to update the output in order to get _this_ PR’s CI to pass. `cake release` is the one that updates _everything._

Cakefile
```diff
@@ -310,7 +310,7 @@ buildDocTests = (watch = no) ->
   # Helpers
   testsInScriptBlocks = ->
     output = ''
-    for filename in fs.readdirSync testsSourceFolder
+    for filename in fs.readdirSync(testsSourceFolder).sort()
```
GeoffreyBooth, line 313: Why this change?

edemaine:
It turns out the issue wasn't that the output wasn't updated, it's that the order that the tests were combined was inconsistent, dependent on the filesystem. Sorting the directory seems to produce the original results. I'm guessing that some file systems automatically sort `listdir`, but Windows didn't, and apparently CI (ubuntu I think?) didn't.

GeoffreyBooth:
> the order that the tests were combined was inconsistent

Why does it matter? Do some tests depend on others?

edemaine:
The `cake doc:test` script concatenates a bunch of files into one file. If the file order is inconsistent, so are the results. The result needs to be consistent for CI to check that it was built properly. (It will also make for smaller diffs, if they get built on a nonalphabetical file system.)

Issue #5330, opened 2020-07-29
Bug: Implicit object literals in array split improperly


Inve1951:
[#try](http://coffeescript.org/#try:a%20%3D%20%5B%0A%20%20b1%3A%201%0A%20%20b2%3A%202%0A%2C%20c1%3A%201%0A%20%20c2%3A%202%0A%5D%0A)

### Input Code

```coffee
a = [
  b1: 1
  b2: 2
, c1: 1
  c2: 2
]
```

### Expected Behavior

```js
var a;

a = [
  {
    b1: 1,
    b2: 2
  },
  {
    c1: 1,
    c2: 2
  }
];
```

### Current Behavior

```js
var a;

a = [
  {
    b1: 1,
    b2: 2
  },
  {
    c1: 1
  },
  {
    c2: 2
  }
];
```
### Context

Fixing this should provide a reasonable alternative to #4952.

Side note: I ran into this issue writing the initial AST tests but never got around to [file it](https://github.com/GeoffreyBooth/coffeescript/blob/90d15176cb5b839b93f2989ec8fd63868d521458/test/abstract_syntax_tree.coffee#L426) (until now).
I felt like having read about this syntax/semantics in an official source like coffeescript.org but was unable to find the reference again.

### Environment

* CoffeeScript version: v2.5.1

PR #5397, opened 2022-01-25
Fix continuing first line of explicit [/{ literal

### LINKED ISSUE #5368: Bug: Destructuring Assignment edge case error
### Input Code

```coffee
{ a
  b: { c }
} = {}
```

### Expected Behavior
Should compile to
```javascript
var a, c;

({
  a,
  b: {c}
} = {});
```

### Current Behavior
```
[stdin]:2:12: error: unexpected }
  b: { c } } = {}
           ^
```

### Work-around
New line after opening brace:

```coffee
{      # New line added here
  a
  b: { c }
} = {}
```

It also works to swap a and b:
```coffee
{ b: { c }
  a
} = {}
```

### Context
This came up for me in the context of passing behavior functions around in a non-trivial object structure.

### Environment

* CoffeeScript version: 2.5.1
* Node.js version: 16.6.2

edemaine:
This PR makes two changes to fix continuing a nonempty first line of an explicit `[` array or `{` object literal:

1. Do not allow implicit call with an implicit object body from first line of `[` array or `{` object literal (as hinted at in comments, but not previously implemented), as in:

   ```coffee
   [x
    y: 2
   ]
   {x
    y: 2
   }
   ```
   Previously, these were treated as function calls where `x` is a function.

2. Support continuing an explicit `{` object literal when the first line has an item on it, as in:

   ```coffee
   {x: 1
    y: 2
   }
   ```
   
   Previously, this was treated as `{x: 1, {y: 2}}` which was invalid.

Fixes #5368


edemaine:
FYI, I [tried](https://github.com/edemaine/coffeescript/tree/continuing-first-line-2) extending this to handle #5330 also, but that seems substantially harder. 😕

src/rewriter.coffee
```diff
@@ -183,6 +183,7 @@ exports.Rewriter = class Rewriter
       [prevTag] = prevToken = if i > 0 then tokens[i - 1] else []
       [nextTag] = nextToken = if i < tokens.length - 1 then tokens[i + 1] else []
       stackTop  = -> stack[stack.length - 1]
+      stackTop2 = -> stack[stack.length - 2]
```
GeoffreyBooth, line 186: A name like `stackTop2` implies to me that it’s returning two elements, not that it’s referring to the second highest item in the stack. Maybe `stackSecond`?

Or maybe do without the helper function? It’s only called in one place, we could just put `stack[stack.length - 2]` there.

edemaine:
Thanks @GeoffreyBooth for the review!

Issue #5216, opened 2019-05-24
Bug: Stack traces from CoffeeScript.eval have JS line numbers


edemaine:
In what I believe is a bug, line numbers in stack traces from errors thrown by CoffeeScript code executed via `CoffeeScript.eval()` (in NodeJS) give the JavaScript line numbers instead of the CoffeeScript line numbers.

### Simple example

```coffee
CoffeeScript = require 'coffeescript'
CoffeeScript.eval '''
  if undefined.foo
    weird = true # cause var hoist to shift lines
'''
```

The resulting stack trace and initial message uses a line number of 3 (**current behavior**) instead of 1 (**expected behavior**):

```
evalmachine.<anonymous>:3
if ((void 0).foo) {
             ^

TypeError: Cannot read property 'foo' of undefined
    at evalmachine.<anonymous>:3:14
    at Script.runInThisContext (vm.js:91:20)
    at Object.runInThisContext (vm.js:298:38)
    at Object.CoffeeScript.eval (...\node_modules\coffeescript\lib\coffeescript\index.js:127:17)
    at repl:2:28
    at repl:3:3
...
```

### More real example with filename
In my application, the eval'd string actually comes from a `.coffee` file, and I can get the **filename** to appear correctly with enough options to `CoffeeScript.eval` (though admittedly I don't understand why so many arguments are needed for this to happen), yet the **line numbers** remain the same.

```coffee
CoffeeScript.eval 'if undefined.foo then weird = true',
  filename: 'test.coffee'
  sourceFiles: ['test.coffee']
  inlineMap: true
```
results in:
```
test.coffee:3
if ((void 0).foo) {
             ^

TypeError: Cannot read property 'foo' of undefined
    at evalmachine.<anonymous>:3:14
```

### Workaround

In my application, I constructed [a workaround](https://github.com/edemaine/svgtiler/blob/c44edcf5fdf5a7f30711732e6a30fb9df08c4c4f/src/svgtiler.coffee#L327-L347) that corrects the line number in the initial message by using `CoffeeScript.compile` to get a source map, looking up/mapping the line number, and modifying the error's stack trace.

### Proposed Solution

[...]

PR #5403, opened 2022-04-01
Don’t patch `Error.prepareStackTrace` if `--enable-source-maps` is used.

### LINKED ISSUE #5382: Bug: Requiring "coffeescript" breaks Node.js built-in source map support globally
Node.js has built in support for source maps via the `--enable-source-maps` command-line option.
If any library at any point `require`s "coffeescript", that support gets completely broken (ie. Node.js stack traces stop processing source maps for eg. TypeScript).

I am *guessing* this is due to the entry point in `coffee-script.js` unconditionally overriding `Error.prepareStackTrace`.

### Input Code

`test.ts`, compiled to `test.js` (with sourcemaps).
`package.json` has `"type": "module"` (probably irrelevant).

```typescript
import 'coffeescript';

throw new Error("test");
```

Run with `node --enable-source-maps test.js`.

### Expected Behavior
Stack trace refers to `test.ts`.

### Current Behavior
Stack trace refers to `test.js`.

### Context

One of the libraries I use `require`s "coffeescript" and now all my stack traces are broken.

### Environment

* CoffeeScript version: 2.6.0
* Node.js version: 16.10

STRd6:
Fixes #5382 Don't monkey patch `Error.prepareStackTrace` if `--enable-source-maps` is used.

This ended up being kind of a chonker but I'm pretty happy with the overall structure. Also a little terrified with how much this touches, as well as a little relieved that there are tons of existing tests to catch regressions 😅 

In summary:

- Calling `require('coffeescript')` no longer patches `Error.prepareStackTrace`. People must `require('coffeescript/register')` to opt-in. `CoffeeScript.run` already does this which will maintain the existing `coffee` cli experience.
- `Error.prepareStackTrace` is not patched if another library has already patched it even when calling `require('coffeescript/register')`.
- Move `Error.prepareStackTrace` patching in `coffeescript.coffee` is now exported rather than applied automatically. `register.coffee` imports it. The browser docs also call it directly.
- Moved caching of source maps into `sourcemap.coffee` (not sure if this is the ideal place).
- Tracking anonymously compiled modules by unique number rather than the brittle array walking of all files under the same `<anonymous>` name.
  - This makes `getSourceMap` and `registerCompiled` much simpler.
  - Still exporting `registerCompiled` so other tools can cache source maps and hook into CoffeeScript stack remapping (CoffeeCoverage, etc.)
  - Needed to change `helpers.syntaxErrorToString` to print `[stdin]` for `<anonymous.*` files as well as files with undefined file name to match the existing tests.
- In `index.coffee` `CoffeeScript.run` set `inlineMap` to be true and set the base `option.filename` to match the `mainModule` filename.
 
TODO
-------

- [x] `CoffeeScript.compile` always caches source maps if present. This makes the call to `registerCompiled` in `register.coffee` redundant. Need to do some thinking about how to untangle this.
- [x] Test case for `./bin/coffee --nodejs --enable-source-maps test/importing/error.coffee` (Looks like file path is currently incorrect: `C:\Users\duder\Documents\GitHub\coffeescript\test\..test\importing\error.coffee:3:9`)
- [x] Browser testing. I believe that modern browsers correctly handle inline source maps but it would be a good idea to have tests for the specific cases and to make sure we're adding them where appropriate. We should always add inline source maps when running / executing scripts and should respect the options given when used as a tool for compiling.


src/coffeescript.coffee
```diff
@@ -271,111 +246,141 @@ parser.yy.parseError = (message, {token}) ->
...
-  fileLocation = ''
-
-  if frame.isNative()
-    fileLocation = "native"
-  else
-    if frame.isEval()
-      filename = frame.getScriptNameOrSourceURL()
-      fileLocation = "#{frame.getEvalOrigin()}, " unless filename
+if process.execArgv.includes('--enable-source-maps') or process.env.NODE_OPTIONS?.includes('--enable-source-maps')
+  registerCompiled = ->
```
GeoffreyBooth, line 250: Don’t we want to only not monkey-patch `Error.prepareStackTrace`, not remove all of `registerCompiled`, when `--enable-source-maps` is set? `registerCompiled` does other stuff beyond source maps, it looks like.

STRd6: I've only found it used in relation to source maps though the comment seems to indicate otherwise.

STRd6:
It looks like `node 16.14.x` breaks the dynamic import test which worked in `16.13.x`. Not sure what the correct resolution for that is.

STRd6:
Tests also pass on `node 17.8.0`

src/coffeescript.coffee
GeoffreyBooth: The `sources` object seems to be used as a local cache of compiled output, so that CoffeeScript doesn’t recompile the same file. And there’s the comment “This is exported to enable an external module to implement caching of compilation results” which I’m not sure about, but seems to imply that this cache both needs to be present and exported for other libraries to use.

STRd6: I've used `CoffeeScript.registerCompiled` in my fork of CoffeeCoverage to cache source maps. I can't find anywhere in the source where it uses the cached files outside of source mapping.

I believe that previously there was some "cleverness" where it would try to cache files to avoid recompiling but I think that was removed and the comment wasn't updated.

STRd6:
Node `16.14.x` tests are fixed and updated here: https://github.com/jashkenas/coffeescript/pull/5404

src/coffeescript.coffee
GeoffreyBooth: > I believe that previously there was some “cleverness” where it would try to cache files to avoid recompiling but I think that was removed and the comment wasn’t updated.

Well, or the caching happens in an external package, like a build tool or build tool plugin that imports the `coffeescript` package. Which I think is the point, since the cache is exported?

Or put another way, is there any reason not to scope this PR as narrowly as possible? As in, why shouldn’t we use the `--enable-source-maps` check to decide whether to patch `Error.prepareStackTrace`, and limit the effect of `--enable-source-maps` to only that?

STRd6: From what I could see `registerCompiled` is the only relevant export that I could find. The `sources` and `sourceMaps` caches are not exported.

My primary reason for not doing any caching with the feature disabled is that since it only seems to be used for the source mapping it would be an unnecessary and possibly large amount of resources being used for no visible effect. Some files are hundreds of kb and some projects have hundreds of files so it could add up. I'm happy to reverse if I've misunderstood or we find out some other use case where this has an impact.

One other thing that I noticed is that we may also want to always add the inline map when `--enable-source-maps` is present so that the native node source mapping will work. I'll need to do some more testing running it with different environment configurations to make sure that all the parts fit together properly. Ideally I'll even add a test that checks the proper line number 😎

GeoffreyBooth: > From what I could see `registerCompiled` is the only relevant export that I could find. The `sources` and `sourceMaps` caches are not exported.

Okay, fair enough. And though this diff is huge, I assume you didn’t change anything about the “normal” `registerCompiled` other than indent it?

> One other thing that I noticed is that we may also want to always add the inline map when `--enable-source-maps` is present so that the native node source mapping will work. I’ll need to do some more testing running it with different environment configurations to make sure that all the parts fit together properly. Ideally I’ll even add a test that checks the proper line number 😎

Yes please! There have been various issues and PRs opened over the years regarding source maps line numbers. I feel like every time I fix one bug related to line numbers I create another.

Another thing to consider is that Node 12 reaches end-of-life at the end of this month, and `--enable-source-maps` was added in 12.12 / 14.18. So CoffeeScript could drop support for Node < 14, and if there’s anything in Node’s internal source maps support that we could use, we could use that instead of doing it ourselves. Like patching `Error.prepareStackTrace` might never be necessary, for example, because we could just update our docs to tell users to run `--enable-source-maps` for source maps in Node, including correct stack traces, rather than us adding all that. I don’t know if there’s anything else in https://nodejs.org/dist/latest/docs/api/module.html#source-map-v3-support that would let us remove any more of our source maps-related code, but it might be worth a look.

GeoffreyBooth:
> Node `16.14.x` tests are fixed and updated here: #5404

If you don’t mind updating with the latest `main`, this should be good to land now.

src/coffeescript.coffee
```diff
@@ -271,111 +249,145 @@ parser.yy.parseError = (message, {token}) ->
...
-  if frame.isNative()
-    fileLocation = "native"
-  else
-    if frame.isEval()
-      filename = frame.getScriptNameOrSourceURL()
-      fileLocation = "#{frame.getEvalOrigin()}, " unless filename
+# Use native source maps rather than monkey patching `Error.prepareStackTrace`
+# if native source maps are enabled. Do not patch `Error.prepareStackTrace`
+# if another library has patched it since that would break them.
+nativeSourceMaps = process?.execArgv.includes('--enable-source-maps') or process?.env.NODE_OPTIONS?.includes('--enable-source-maps')
```
GeoffreyBooth, line 255: Shouldn’t you define this variable above its first reference? I see it read on line 64 (inside the `compile` function), but defined on line 255. Assuming there are no calls to `compile` before line 255, this will work, since `nativeSourceMaps` will get defined before it’s read; but this feels brittle, because any calls to `compile` above line 255 will cause this to break (I think?).

Anyway this is a one-liner, perhaps we should just move it and its comment to near the top of the file?

test/sourcemap.coffee
```diff
@@ -63,3 +64,55 @@ test "#3075: v3 source map fields", ->
   v3SourceMap = JSON.parse v3SourceMap
   arrayEq v3SourceMap.sources, ['tempus_fugit.coffee']
   eq v3SourceMap.sourceRoot, './www_root/coffee/'
+
+# Source maps aren't accurate on Node v12 ??
```
GeoffreyBooth, line 68: Care to explain?

test/sourcemap.coffee
```diff
@@ -63,3 +64,55 @@ test "#3075: v3 source map fields", ->
...
+if process.version.split(".")[0] != "v12"
+  test "native source maps", ->
+    new Promise (resolve, reject) ->
+      proc = spawn "node", [
```
GeoffreyBooth, line 72: You can use `child_process.fork` to specifically spawn Node child processes: https://nodejs.org/api/child_process.html#:~:text=The%20child_process.fork()%20method%20is%20a%20special%20case%20of%20child_process.spawn()%20used%20specifically%20to%20spawn%20new%20Node.js%20processes.

test/sourcemap.coffee
```diff
@@ -63,3 +64,55 @@ test "#3075: v3 source map fields", ->
...
+        "--enable-source-maps"
+        "-r", "./register.js"
+        "-r", "./test/integration/error.coffee"
```
GeoffreyBooth, line 75: All of our test “fixture” files are in `test/importing`. Do you mind moving `error.coffee` there?

```suggestion
        '--require', './register.js'
        '--require', './test/importing/error.coffee'
```

And some style notes: I prefer to spell out command-line flags so that people know what I’m using without looking it up. And I haven’t been going back and fixing this for old code, but for new code we’ve been trying to follow the rule of single-quote strings unless there’s interpolation, or when following an established pattern (like the argument to `test`); and no single-character variable names.

test/sourcemap.coffee
```diff
@@ -63,3 +64,55 @@ test "#3075: v3 source map fields", ->
...
+          resolve()
+        catch e
+          reject(e)
+
+test "don't change stack traces if another library has patched `Error.prepareStackTrace`", ->
+  new Promise (resolve, reject) ->
+    proc = spawn "node", [
+      "-r", "./test/integration/prepare_stack_trace.js",
+      "-r", "./register.js",
+      "-r", "./test/integration/error.coffee",
```
GeoffreyBooth, line 98: I would think that you could test this more directly, without needing to throw an error or create a fixture?

```suggestion
      '--eval', '''
        const patchedPrepareStackTrace = function() {};
        require('./register.js');
        console.log(Error.prepareStackTrace === patchedPrepareStackTrace);
      '''
```

And then check if `stdout` contains `'true'`.

test/sourcemap.coffee
STRd6: Node v12 reports the line number as `4` where every other version after correctly reports it as `3`.

STRd6: On further investigation this is only for the error message, the stack is correct. Future versions of node also corrected the line in the error messages as well.

src/coffeescript.coffee
STRd6: 👍

STRd6:
I'm pretty happy with how this is functionally. There's some more clean up and testing that can be followed up with but I'm fairly confident that this will maintain the existing functionality while fixing the two cases where patching `Error.prepareStackTrace` breaks things. We now will no longer patch if `Error.prepareStackTrace` has been changed by another tool (such as `source-map-support`) and we also no longer patch if `--enable-source-maps` is set since that would apply an additional remapping making the stack trace incorrect there as well. The default behavior is still to patch `Error.prepareStackTrace` so that the basic 'out of the box' experience for people using CoffeeScript will still be correct stack traces.

STRd6:
There's actually one other case that this doesn't address correctly yet: when `--enable-source-maps` is present in code that calls `CoffeeScript.compile` (think webpack or a similar bundler) then as it stands right now it will always add an inline map even if that option wasn't specified to the compile function.

The fix is to move the logic into `register` and pass the correct options to `compile` based on the environment settings. I'll continue working in that direction. The downside is that it is a larger change overall but it should end up with simpler code that is clearer and more direct which will also clean up some of the historical weirdness around source maps.

src/sourcemap.litcoffee
```diff
@@ -179,9 +179,23 @@ Regular Base64 Encoding
...
 
-    module.exports = SourceMap
+    sourceMaps = Object.create {}
 
+    registerCompiled = (filename, source, sourcemap) ->
+      if sourcemap?
+        sourceMaps[filename] = sourcemap
 
+    getSourceMap = (filename) ->
+      sourceMaps[filename]
```
GeoffreyBooth, line 191: Why not just make these as additional methods on the class? And you can create the cache in the constructor, something like `@cache = Object.create {}`, if this class is a singleton; or define it at the top of the file if there are multiple instances of `SourceMap` instantiated but they should all share the same cache.

Cakefile
```diff
@@ -400,7 +400,8 @@ task 'bench', 'quick benchmark of compilation time', ->
 
 
 # Run the CoffeeScript test suite.
-runTests = (CoffeeScript) ->
+runTests = (CoffeeScript, pattern="") ->
+  pattern = new RegExp pattern
```
GeoffreyBooth, line 404: This feels like a separate PR. Also, why is it needed? For me on `main`, `cake test` takes about 6 seconds. Are the new tests in this PR slowing the tests significantly, with all the child processes being created?

STRd6: I'm happy to split this out. The main issue isn't the speed difference but the ability to isolate the output for a few tests when debugging. When the stack traces changed it caused almost all the error message tests to fail. When adding detailed `console.log` debugging sometimes the output would scroll beyond the capacity of the terminal. Being able to run a single test or subset of tests can make it easier to see the specific error.

Any thoughts on using Mocha as a test runner? (As a separate PR of cources)

GeoffreyBooth: > Any thoughts on using Mocha as a test runner?

Part of the ethos of this project has been that it has no dependencies (at least, other than dev dependencies) and that the CoffeeScript repo itself should serve as an example of how to build a robust application using CoffeeScript. Along those lines, I think there would be many who would object to adding a large dependency like that for something as fundamental as running tests; notice how we don’t have a library for generating the static website, either.

I could be wrong, though, if the benefits were significant. Personally the refactoring required to add Mocha feels like time that could be better spent on other enhancements.

GeoffreyBooth:
Please let me know when you’re done working on this branch and I’ll review it again. Please don’t forget the style notes (single-quoted strings unless interpolating, avoid single-character variables).

STRd6:
I'll get a few more updates in and summarize the changes as well over the next few days.

src/stacktrace.coffee
```diff
@@ -0,0 +1,78 @@
...
+      # Don't display stack frames deeper than the stack root `CoffeeScript.run`
+      # for example.
+      break if stackRoot and (frame.getFunction() is stackRoot)
+      "    at #{formatSourcePosition frame, getSourceMapping}"
+
+    "#{err.toString()}\n#{frames.join '\n'}\n"
+
+module.exports = {
+  attach
+}
```
GeoffreyBooth, line 78: Please don’t split this out into a new file. There hasn’t been a new file created out of the main core in many years, probably at least five years, and I don’t want to start now. We’d have to dig into how the build scripts assemble everything, like how certain parts do and don’t go into the browser build, and I don’t want to deal with that now. That’s out of scope for what we’re trying to fix in this PR.

STRd6: It looks like all that is needed for the build is the change in `Cakefile` adding to the browser build list but I also agree that having fewer files can keep things simpler. In that case I think this `attach` will go back to `coffeescript.coffee` and be used from both `register` and the browser docs code. That will also keep it closer to the original implementation.

This is getting very close. It also fixes another bug I found in browser based Try CoffeeScript where the stack trace line numbers are only right for the first time the code runs. Press "Run" twice [here](https://coffeescript.org/#try:fn%20%3D%20-%3E%0A%20%20throw%20new%20Error%20'heyy'%0Afn()%0A) to see the error. The last remaining piece is to hide the inline map from the compiled source in the JS preview editor. The browser needs at least the `//# sourceURL` comment in the executed code to be able to link up to the proper filename for the source map.

GeoffreyBooth: Thanks. I’m sure it’s probably not that big a deal to refactor the build, but that should at least be its own PR. It would also break all open PRs for the old code that moved. We’ve been toying with the idea of splitting up `nodes.coffee` for years now, moving each class into its own file, but it’s never quite seemed worth the effort in terms of lost history and broken PRs. Basically at this point I consider most of the project architecture decisions “fixed,” in that the cost of changing them is too high when the motivation/availability of contributors is as low as it is. Many of those decisions I would certainly make differently today, like having `lib/` version-controlled, but the project is well-enough designed that I don’t consider anything to be of sufficient hindrance that I feel like I _need_ to make a big change in order to be productive.

STRd6:
@GeoffreyBooth this is ready for review now. It's a bit larger than I originally anticipated. Please ask any questions you have so I can better document exactly what changed and how things work.

I'm not sure why the CI is flaky on windows but I should look into that before this is merged.

test/sourcemap.coffee
```diff
@@ -63,3 +64,119 @@ test "#3075: v3 source map fields", ->
...
+    ]
+
+    out = ""
+    proc.stdout.setEncoding('utf8')
+    proc.stdout.on 'data', (s) -> out += s
+
+    proc.on        'exit', (status) ->
+      try
+        equal status, 0
+        equal out, "undefined\n"
```
GeoffreyBooth, line 178:
```suggestion
        process.stdout.write(Error.prepareStackTrace === undefined ? 'unused' : 'defined');
      """
    ]

    out = ""
    proc.stdout.setEncoding('utf8')
    proc.stdout.on 'data', (s) -> out += s

    proc.on        'exit', (status) ->
      try
        equal status, 0
        equal out, 'unused'
```

Same as above.

GeoffreyBooth (commented):
This is very good! A few notes here and there and this should be ready to land.

Does this fix #5216 by chance? From a quick search of open issues, that appears to be the only other open issue related to source maps.

src/coffeescript.coffee
```diff
@@ -271,111 +247,77 @@ parser.yy.parseError = (message, {token}) ->
...
     else
-      "#{typeName}.#{methodName or '<anonymous>'} (#{fileLocation})"
-  else if isConstructor
-    "new #{functionName or '<anonymous>'} (#{fileLocation})"
-  else if functionName
-    "#{functionName} (#{fileLocation})"
-  else
-    fileLocation
-
-getSourceMap = (filename, line, column) ->
```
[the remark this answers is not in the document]
STRd6: That's correct. Now that we track each compilation with a unique identifier we don't need to hold on to the extra sources to recompile or walk through the array of `<anonymous>` sources guessing and checking.

src/helpers.coffee
```diff
@@ -188,6 +188,13 @@ exports.locationDataToString = (obj) ->
   else
     "No location data"
 
+# Generate a unique anonymous file name so we can distinguish source map cache
+# entries for any number of anonymous scripts.
+exports.anonymousFileName = do ->
+  n = 0
+  ->
+    "<anonymous-#{n++}>"
```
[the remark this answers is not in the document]
STRd6: It was previously `<anonymous>` which matches the "filename" Node's `eval`.

src/coffeescript.coffee
```diff
@@ -43,31 +46,6 @@ withPrettyErrors = (fn) ->
...
-# For each compiled file, save its source in memory in case we need to
-# recompile it later. We might need to recompile if the first compilation
-# didn’t create a source map (faster) but something went wrong and we need
-# a stack trace. Assuming that most of the time, code isn’t throwing
-# exceptions, it’s probably more efficient to compile twice only when we
-# need a stack trace, rather than always generating a source map even when
-# it’s not likely to be used. Save in form of `filename`: [`(source)`]
-sources = {}
-# Also save source maps if generated, in form of `(source)`: [`(source map)`].
-sourceMaps = {}
```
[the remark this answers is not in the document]
STRd6: 👍 I've updated the export comment with the accurate parts.

STRd6:
> Does this fix #5216 by chance? From a quick search of open issues, that appears to be the only other open issue related to source maps.

It doesn't fix #5216 but I think it lays some groundwork that could lead to a fix.

test/sourcemap.coffee
```diff
@@ -84,21 +84,21 @@ test "node --enable-source-map built in stack trace mapping", ->
         equal column, 9 # Mapped source column
 
         resolve()
-      catch e
-        reject(e)
+      catch exception
+        reject exception
 
-unless process.version.slice(1, 3) is "12"
+if Number(process.version.slice(1, 3)) >= 14
```
GeoffreyBooth, line 90:
```suggestion
if Number(process.versions.node.split('.')[0]) >= 14
```

I know your version won’t break until Node v100, but why not make it handle any version? And handle if they ever remove the leading `v`.

test/sourcemap.coffee
```diff
@@ -133,50 +130,50 @@ unless process.version.slice(1, 3) is "12"
...
-        catch e
-          reject e
+        catch exception
+          reject exception
 
 test "don't change stack traces if another library has patched `Error.prepareStackTrace`", ->
   new Promise (resolve, reject) ->
-    proc = spawn "node", [
-      "--eval", """
+    proc = spawn 'node', [
```
GeoffreyBooth, line 138: Should this one be `fork`?

test/sourcemap.coffee
```diff
@@ -133,50 +130,50 @@ unless process.version.slice(1, 3) is "12"
...
-      catch e
-        reject e
+      catch exception
+        reject exception
 
 test "requiring 'CoffeeScript' doesn't change `Error.prepareStackTrace`", ->
   new Promise (resolve, reject) ->
-    proc = spawn "node", [
-      "--eval", """
+    proc = spawn 'node', [
```
GeoffreyBooth, line 161: And this one.

test/sourcemap.coffee
STRd6: I don't think these can easily be switched to fork since it is using `--eval` and there is no `modulePath`. Though I could move these into `importing` and require them but I think that would obscure what is happening.

GeoffreyBooth: Ah okay. Maybe just add a comment then that explains that we can't use `fork` with `--eval`?

src/helpers.coffee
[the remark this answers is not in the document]
GeoffreyBooth: > which matches the “filename” of Node’s `eval`.

That’s fine; do you mind sharing where you see that in Node?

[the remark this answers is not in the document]
STRd6:
```plain
node
Welcome to Node.js v17.9.0.
Type ".help" for more information.
> eval("throw new Error")
Uncaught Error
    at eval (eval at <anonymous> (REPL1:1:1), <anonymous>:1:7)
>
```

STRd6:
One thing to note is that though this does fix a bug/unexpected behavior it could be considered a breaking change to how `require('coffeescript')` works. The good news is that it is very simple to restore the previous behavior if people want by calling `patchStackTrace` explicitly.

```js
CoffeeScript = require('coffeescript')
CoffeeScript.patchStackTrace()
```

GeoffreyBooth:
> The good news is that it is very simple to restore the previous behavior if people want by calling `patchStackTrace` explicitly.

I thought the intent was to keep the current behavior _unless_ `--enable-source-maps` was passed (or another plugin already called `Error.prepareStackTrace`).

STRd6:
I think that the correct long term solution is to not do anything drastic to people's environments when `require('coffeescript')` is called, similarly to how `require.extensions` patching was moved to `/register`. Requiring `coffeescript/register` still patches as before (excepting the cases when `--enable-source-maps` is provided or `Error.prepareStackTrace` has already been set) but requiring `coffeescript` does not as currently implemented in this PR.

If that is too breaking a change to go out soon then I can update the PR to also have `coffeescript` continue to patch until the next breaking release for compatibility.

GeoffreyBooth:
I thought the current behavior was not to patch unless `register` was called; are we patching when just `require`ing?

If so, yeah, we should restrict the patching to `register` (and within `register`, don’t patch if `--enable-source-maps` is set).

STRd6:
👍 

The current behavior in `2.6.1` patches no matter what when required: https://github.com/jashkenas/coffeescript/blob/76cf769abfd9ae7935c0e47c4137d6303e4482dc/src/coffeescript.coffee#L368

This PR restricts the patching to `register` and doesn't patch in those cases mentioned. People can use the `CoffeeScript.patchStackTrace()` if they need to for compatibility reasons.

src/coffeescript.coffee
```diff
@@ -186,7 +165,7 @@ exports.compile = compile = withPrettyErrors (code, options = {}) ->
   if options.inlineMap
     encoded = base64encode JSON.stringify v3SourceMap
     sourceMapDataURI = "//# sourceMappingURL=data:application/json;base64,#{encoded}"
-    sourceURL = "//# sourceURL=#{options.filename ? 'coffeescript'}"
+    sourceURL = "//# sourceURL=#{filename}"
     js = "#{js}\n#{sourceMapDataURI}\n#{sourceURL}"
```
GeoffreyBooth, line 169: FYI, not as part of this PR but related: https://github.com/jashkenas/coffeescript/pull/5237

GeoffreyBooth:
Thanks @STRd6 and congrats! We’re overdue for a release, are there any other PRs you’d like to land soon or should I prepare a release for later this week?

STRd6:
@GeoffreyBooth thanks for the review! I don't have anything else ready to land so release when ready 👍

PR #5404, opened 2022-04-01
Fix CI for Node 16.14.x

STRd6:
Guarding dynamic import tests based on Node version
These import assertions weren't actually running on older versions of
Node. They failed to parse so failed to run at all. Node 16.14.x added
support to parse the import assertion syntax but instead failed with
`Invalid module "data:application/json,..." has an unsupported MIME type "application/json"`

Node 17.8.x supports the syntax and supports the `application/json` mime
as well.

As a cleanup item also updated caniuse-lite.

### DIFF
--- a/docs/v2/test.html
+++ b/docs/v2/test.html
@@ -25716,22 +25716,32 @@ test "the `baseFileName` helper returns the file name to write to", ->
 <script type="text/x-coffeescript" class="test" id="import_assertions">
 # This file is running in CommonJS (in Node) or as a classic Script (in the browser tests) so it can use import() within an async function, but not at the top level; and we can’t use static import.
 test "dynamic import assertion", ->
-  if typeof process is "undefined" or process.version?.startsWith("v17")
+  try
     { default: secret } = await import('data:application/json,{"ofLife":42}', { assert: { type: 'json' } })
     eq secret.ofLife, 42
+  catch e
+    # This parses on Node 16.14.x but throws an error async so the `skipUnless` feature detection won't catch it
+    # rethrow any other erorr
+    unless e.message is 'Invalid module "data:application/json,{"ofLife":42}" has an unsupported MIME type "application/json"'
+      throw e
 
 test "assert keyword", ->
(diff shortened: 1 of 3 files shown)


GeoffreyBooth:
We’ve been using feature detection for this rather than hard-coding Node versions. See https://github.com/jashkenas/coffeescript/blob/3d39d205d8e00cf46b766cc455cfa38119101e34/Cakefile#L465-L479

Do you mind updating the last line there, about import assertions, such that that resolves the issue? So that we don’t need to put guards in the test file itself. Some of the tests don’t parse at all in certain versions of Node because of syntax, and so it’s important not to load the file at all (rather than loading it and skipping most or all of its contents based on a condition).

STRd6:
This is the same test referenced in the last `skipUnless`. The problem is that it now parses in node `16.4.x` but fails during the async execution of the test on `16.14.x`.

I'll poke around and see if I can extend `skipUnless` to detect async failures.

GeoffreyBooth:
Now that I think about it more, I think `skipUnless` is meant just for parsing errors, where we need to avoid loading that file at all; but there _are_ sometimes guards within tests based on missing functionality. But those guards I think are still feature detections, not version checks. Maybe you could do a feature detection at the top of the test file somehow that an `import` of `application/json` is supported? Not sure how, if that needs to be async.

GeoffreyBooth:
I dug into this a bit: https://github.com/nodejs/node/pull/41736#issuecomment-1086738670

Basically, Node 16.14.0 included support for the import assertions syntax, so that syntax starts parsing as of 16.14.0; but JSON modules still require the `--experimental-json-modules` flag on 16.x. That flag was removed in 17.5.0, but that change hasn’t been backported to 16 yet; it should land as part of 16.15.0, whenever that comes out. So we’re in a weird transition state where the syntax parses but the module is unrecognized, which will only be the case for the brief life of Node 16.14.x.

So I guess we can land this try/catch for now and just remove it once 16.15.0 is released, in order to unblock your other PRs.

PR #5406, opened 2022-04-04
Remove always true `if` check

STRd6:
This check has always been true since Node 0.3.0. It was added here: https://github.com/jashkenas/coffeescript/commit/c4a3e170e230d7bf0a2513c1224d71f685dafb43 and seems to rely on the behavior of `Module#_compile` from before `require.extensions`. If `require.extensions` is ever removed (it is deprecated after all) then `CoffeeScript.run` will no longer work. It will pass CoffeeScript source code to `Module#_compile` which is doomed to fail.

Browser code doesn't touch this path at all and is unaffected.

### DIFF
--- a/lib/coffeescript/index.js
+++ b/lib/coffeescript/index.js
@@ -60,10 +60,8 @@
     // Save the options for compiling child imports.
     mainModule.options = options;
     // Compile.
-    if (!helpers.isCoffee(mainModule.filename) || require.extensions) {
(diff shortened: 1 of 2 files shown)


STRd6:
In digging through the source mapping nitty gritty I found this small oddity. I'd prefer to split off small self contained cleanup PRs when possible so that the eventual source mapping PR isn't totally massive and coupled with indirectly related cleanup.

PR #5279, opened 2019-12-30
Allow leading logical operators

helixbass:
@GeoffreyBooth re-opening against `master`

See original PR: #5068

PR #5407, opened 2022-04-05
Allow leading logical operators

STRd6:
Rebased https://github.com/jashkenas/coffeescript/pull/5279


src/rewriter.coffee
```diff
@@ -753,6 +754,30 @@ exports.Rewriter = class Rewriter
       @detectEnd i + 1, condition, action
       return 1
 
+  # Returns `yes` if standing in front of what will become a LEADING_AND or
+  # LEADING_OR, ie TERMINATOR followed by && or ||
+  isLeadingLogical: (i) ->
+    @tag(i) is 'TERMINATOR' and @tag(i + 1) in LEADING_LOGICAL
```
GeoffreyBooth, line 760: I would think that you shouldn’t need to use the rewriter for this. The regular lexer often looks multiple tokens ahead; whenever it sees an and/or token followed by a terminator, it could create it as `LEADING_AND` or `LEADING_OR` in the first place rather than needing the rewriter to make another pass.

In general everything in `rewriter.coffee` is hacks that are unfortunate workarounds, and we want to avoid adding to the rewriter if at all possible. Each rewrite pass adds a performance hit to compilation.

PR #5409, opened 2022-04-17
Mark Generated Files

STRd6:
In theory setting these `.gitattributes` should prevent generated files from showing up in diffs by default.

https://github.com/github/linguist/tree/47b109be3657fe3d5933fba6840a6ceea0f94498#generated-file-detection

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
-lib/ linguist-generated=true
-docs/ linguist-generated=true
+lib/** linguist-generated=true
+docs/** linguist-generated=true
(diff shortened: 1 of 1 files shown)


GeoffreyBooth:
I think that’s already happening. If you look at https://github.com/jashkenas/coffeescript/pull/5403/files, all the `.js` files have a “Load diff” link, and the diff is hidden by default. Does adding this setting change anything?

We can always add this file and find out, as this doesn’t hurt anything, but I expect it won’t make a difference.

STRd6:
I think you're right, this doesn't seem like it would improve anything.

PR #5411, opened 2022-04-19
Add `patchStackTrace` export; test that all named exports are detected by Node

GeoffreyBooth:
Follow-up to #5403; this adds the new `patchStackTrace` export, so that `import { patchStackTrace } from 'coffeescript'` works in Node’s native ESM. I noticed this while writing up the changelog.

I also added a test so that we don’t miss these in the future. cc @STRd6


STRd6:
Thanks!

STRd6 (approved):

PR #5413, opened 2022-04-20
Fixes #5410

### LINKED ISSUE #5410: Unnecessary `ref` in `for x from` and `for await x from` loops
```coffee
->
  for await item from fn()
    console.log item
  return
```

```js
// Generated by CoffeeScript 2.6.1
(async function() {
  var item;
  var ref = fn();
  for await (item of ref) {
    console.log(item);
  }
});
```

It looks like CoffeeScript is adding an unnecessary `ref` during `for from` loop creation. This happens in both the async and regular kinds.

STRd6:
Fixes #5410 by removing unnecessary ref creation in `for from` loops.

I believe that the ref was originally created for iterating using `for x in ...` where caching the source and length makes sense. In `for from` loops (compile to `for of` in JS) the loop receives an iterator. JS handles creating the local var and the iterator manages the state so there is nothing to cache.

```coffee
fn = ->
  console.log "called"
  [1, 2, 3]

for x from fn()
  console.log x
```

With this change:
```js
// Generated by CoffeeScript 2.6.1
var fn, x;

fn = function() {
  console.log("called");
  return [1, 2, 3];
};

for (x of fn()) {
  console.log(x);
}
```

Previously:
```js
// Generated by CoffeeScript 2.6.1
var fn, ref, x;

fn = function() {
  console.log("called");
  return [1, 2, 3];
};

ref = fn();
for (x of ref) {
  console.log(x);
}
```


edemaine (commented):
Nice to see that this was so easy to fix! LGTM on quick glance, though maybe worth adding a test to make sure that `for...in` loops *do* generate `ref`s?  Although it's probably implicitly in other tests.

PR #5422, opened 2022-08-01
Fix #5421, docs code snippets not running

### LINKED ISSUE #5421: Site issue: examples don't run
The website at <https://coffeescript.org> is currently degraded in that some of the examples don't execute correctly when hitting the play button.

Most seem to run into undefined symbols (ReferenceError).

GeoffreyBooth:
I’m not sure how this broke, but somehow the code snippets aren’t running on the docs site currently. They work if you edit the CoffeeScript first, which sets a `js` property on the CodeMirror editor; but not if you don’t edit first (which used to work). I very slightly amended the “run example” logic to check for that `js` property specifically, which I guess was previously always present but somehow no longer is unless the user edits the CoffeeScript. Our version of CodeMirror is pinned and hasn’t changed, so I’m not sure how this could have broken, unless it has something to do with the latest version of popular browsers. cc @Inve1951


Inve1951:
i was just about to test it :smile:

GeoffreyBooth:
Well please let me know if anything’s still broken. Since the live site was already broken I figured I was unlikely to make it worse 😄

PR #5435, opened 2022-11-26
Heregex Comment

### LINKED ISSUE #5428: Bug: Heregex escaped space before comment includes comment in transpiled regex
### Input Code

```coffee
///
a\ # this is a bug
///
```

### Expected Behavior

```coffee
// this is a bug
/a /;
```

### Current Behavior

```coffee
// this is a bug
/a #thisisabug/;
```

* CoffeeScript version: 2.7.0

STRd6:
Fixes #5428

This fix is simple enough but it causes a substantial change in behavior for Heregexes.

`#` inside of a character class shouldn't be considered a comment. By treating `#` without whitespace in front as non-comments we could have slight compatability with Python.

There are other places in the CoffeeScript source where people avoided escaping the `#` by keeping it next to non-whitespace characters even outside of a character class. This is different than how Python does it and is probably a bug but maybe it is too late with the de facto CoffeeScript2 behavior.

I'm not sure if this should be merged in since it changes the behavior quite a lot. Maybe another one for the CoffeeScript3/Civet bucket.

Refs
---
Python Docs: https://docs.python.org/3/library/re.html#re.X


STRd6:
GitHub auto-closed when I rebased so re-opening.

STRd6:
The interpolation test that was changed is interesting...

It's inserting a string containing a single backslash into the RegExp ahead of an escaped `#`. The current behavior is to keep the single backslash which is escaping the `#` in the output which then combines with another single backslash from the interpolation which causes the output to be an escaped backslash. There's no comment about the intent other than `TODO improve heregex interpolation tests` so I'm not sure if this existing behavior is desired.

Issue #5448, opened 2023-05-11
Site issue: code blocks twitch on hover


Inve1951:
There's a new visual issue with https://coffeescript.org.

The first time you hover some code block CodeMirror kicks in. When it does that it incorrectly overrides a CSS rule.

```css
/* before */
.CodeMirror pre {
    line-height: 1.4em;
}

/* after */
.CodeMirror pre.CodeMirror-line {
    line-height: inherit;
}
```

On my device this reduces the effective line height from `19.6px` to `17.5px`, shrinking the vertical size of the code block.

But there's something else wrong too. Even with that rule disabled, most of the time CodeMirror kicks in, the container grows `1px` in height.

Tested on Firefox.

GeoffreyBooth:
Okay, can you suggest a fix?

PR #5449, opened 2023-05-11
Fix line-height change from hover

edemaine:
Partial solution to #5448, fixing the line height change when switching to CodeMirror mode.  But there are still tiny discrepancies in overall size which might be due to a single `<pre>` vs. multiple line `<div>`s (or something else; I'm not certain). Still, this fixes most of the issue.


GeoffreyBooth:
Thank you!

PR #5452, opened 2023-09-03
Optimize split in checkShebangLine

Connormiha:
A little optimization. String `split` has optional second argument to reduce elements count in result array and stop parsing in advance.


GeoffreyBooth:
Thanks for this! Can you please update the build output and include that too?

Connormiha:
> Thanks for this! Can you please update the build output and include that too?

Done

PR #5454, opened 2023-09-12
Update flatten helper without recursion

Connormiha:
I've rewritten the `flatten` algorithm. Instead of a recursive approach, i'm using a stack. This way, we avoid unnecessary function calls and the creation of intermediate arrays. `flattened` in each call and new array after each `concat`.
In cases of deep recursion or with a large number of arrays, this algorithm is more efficient and generates less work for the garbage collector.


src/helpers.coffee
```diff
@@ -48,11 +48,14 @@ extend = exports.extend = (object, properties) ->
...
-    else
-      flattened.push element
+  stack = [array]
+  while stack.length
+    current = stack.pop()
+    for element in current
+      if '[object Array]' is Object::toString.call element
+        stack.push(element)
+      else
+        flattened.push(element)
```
GeoffreyBooth, line 58: Perhaps instead we should use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat? I don’t think we even need to feature-detect it, support goes very far back.

Connormiha: Yes, we can use this if old browser/Node.js support is not required.
```coffeescript
exports.flatten = flatten = (array) ->
  array.flat(10)
```
10 - is max depth

GeoffreyBooth: > Yes, we can use this if old browser/Node.js support is not required.

You can look in our GitHub Actions to see our support matrix, but basically everywhere we support can run `.flat`. You can use `.flat(Infinity)` to go as far down as necessary.

Connormiha: I changed it and built.
```bash
npx cake build
npx cake build:browser
```

GeoffreyBooth: CI is failing. Try `cake release`?

Connormiha: Done

PR #5464, opened 2024-03-22
Also declare async function, if an  await was found in function's body

onepointerT:
**Solves the error** '_Unexpected reserved word_ "await"':
````
      dir = (await opendir(this.path));
             ^^^^^

SyntaxError: Unexpected reserved word
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:152:18)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:298:14)

Node.js v18.19.1````

### DIFF
--- a/src/nodes.coffee
+++ b/src/nodes.coffee
@@ -11,7 +11,7 @@ Error.stackTraceLimit = Infinity
 # Import the helpers we plan to use.
 {compact, flatten, extend, merge, del, starts, ends, some,
 addDataToNode, attachCommentsToNode, locationDataToString,
-throwSyntaxError, replaceUnicodeCodePointEscapes,
+throwSyntaxError, replaceUnicodeCodePointEscapes, count,
 isFunction, isPlainObject, isNumber, parseNumber} = require './helpers'
 
 # Functions required by parser.
@@ -3911,6 +3911,8 @@ exports.Code = class Code extends Base
       if (node instanceof Op and node.isAwait()) or node instanceof AwaitReturn
         @isAsync = yes
       if node instanceof For and node.isAwait()
+        @isAsync = yes
+      if node instanceof Function and count @body, 'await' > 0
         @isAsync = yes
 
     @propagateLhs()

### CODE REVIEW

**Comment by GeoffreyBooth** [comment]
This needs tests, and you need to commit the compiled output.

**Comment by onepointerT** [comment]
I'm not from yesterday, I know that and what testing is. I've tested it on the same testing script. The output is omitted, because no further output on that functionality. After using the corrected commit I've got output on a completly other topic, import statements in ES scripts, with `"type": "module"` in `package.json`. The problem with the asyncronous `await` is patched and delivers no further output, I manually controlled the function, it is now an `async function`, before the commit it was a simple function.

**Comment by edemaine** [comment]
What's the input that causes the output error you mention?

I'm worried that this code will add `async` to the outer function in this case (but I didn't test):

```coffee
->
  -> await undefined
```

onepointerT:
```
export class File extends Path
    constructor: (path) ->
        super path
    
    readSync: () ->
        fd = await open @path 'r'
        return fd.createReadStream()
    
    writeSync: (data) ->
        fd = await open @path 'rw'
        return fd.createWriteStream data
```
Somehow an `await undefined` sounds a little bit silly, since loops that iterate unless the stack is empty are implemented already. This feature request is required on every async function use inside of a function, except the already included async functionality with For and Of.

> What's the input that causes the output error you mention?
> 
> I'm worried that this code will add `async` to the outer function in this case (but I didn't test):
> 
> ```coffee
> ->
>   -> await undefined
> ```

edemaine:
Your code [already compiles to](https://coffeescript.org/#try:export%20class%20File%20extends%20Path%0A%20%20%20%20constructor%3A%20(path)%20-%3E%0A%20%20%20%20%20%20%20%20super%20path%0A%20%20%20%20%0A%20%20%20%20readSync%3A%20()%20-%3E%0A%20%20%20%20%20%20%20%20fd%20%3D%20await%20open%20%40path%20'r'%0A%20%20%20%20%20%20%20%20return%20fd.createReadStream()%0A%20%20%20%20%0A%20%20%20%20writeSync%3A%20(data)%20-%3E%0A%20%20%20%20%20%20%20%20fd%20%3D%20await%20open%20%40path%20'rw'%0A%20%20%20%20%20%20%20%20return%20fd.createWriteStream%20data)

```js
export var File = class File extends Path {
  constructor(path) {
    super(path);
  }

  async readSync() {
    var fd;
    fd = (await open(this.path('r')));
    return fd.createReadStream();
  }

  async writeSync(data) {
    var fd;
    fd = (await open(this.path('rw')));
    return fd.createWriteStream(data);
  }

};
```

It seems that `async` is already where it needs to be. So I'm not sure what this PR is aiming to fix. It would help if you added a test that is broken with the current CoffeeScript.

Issue #5471, opened 2024-11-05
Proposal: bind a name to a destructurable place with @


cosmicexplorer:
This is a **feature request for new syntax**. I believe this would be **entirely backwards-compatible**.

# Problem
Consider the function:
```coffee
f = ({x}) ->
   # if key name 'x' is changed,
   # this silently starts retrieving x from outer scope:
  {a, b} = x
  otherFunc x, a, b

# only way to write this e.g. in repl:
f = (o) -> otherFunc o.x, o.x.a, o.x.b

# saving the result of a function call cannot be done while destructuring:
x = g()
[a, b] = x
```
Because `otherFunc` requires both the parent object `x` as well as some of its fields `a` and `b`, we have to perform an additional destructuring step. This can be error-prone, as the name `x` must be duplicated for the required boilerplate. It also makes it impossible to write this logic in one line without avoiding destructuring entirely!

This syntactic limitation applies equally to all forms of destructuring assignment. The proposed solution frames the existing "destructuring" mechanics in terms of what I'm calling "place expressions", expounding on the second half of "destructuring *assignment*".

# Proposal: `@` to bind a name to a place
*Abstractly:*
- Any expression you can destructure (`place-expr`) can also be given a name before applying the requested destructuring:
    - `place-expr = <name>@<destructuring-expr>` *(except for object keys which use `@:`)*
    - The choice of `@` was motivated by [the `@` operator from rust](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#-bindings), where `let` and `match` introduce similarly "destructurable" place expressions.
- This staged destructuring desugars directly into a series of destructuring assignments.

*Concretely:*
This new syntax should be entirely backwards-compatible, and will show up in four places:
1. top-level assignment: `<name>@<destructuring-expr> = <value-expr>`
2. function arguments: `(<name>@<destructuring-expr>) -> <value-expr>`
3. destructured object key: `{<name>@: <destructuring-expr>, ...}`
4. destructured array element: `[<name>@<destructuring-expr>, ...]`

More concretely:
```coffee
# 1. top-level assignment
x@[a, b] = f()
# desugars to:
# x = f()
# [a, b] = x
assert.deepEqual x[...2], [a, b]

# 2. function arguments
f = (x@{a}) ->
  assert.deepEqual x.a, a
# desugars to:
# f = (x) ->
#   {a} = x
#   assert.deepEqual x.a, a

# 3. destructured object key
[...]

PR #5473, opened 2024-11-15
console output helper in Cakefile for test case selection and parser generation statistics

cosmicexplorer:
*This PR purely modifies the `Cakefile` in this repo and should not affect any users of the language.*

# Problem
I was playing around with changes to the parser and language semantics in e.g. https://github.com/jashkenas/coffeescript/issues/5471#issuecomment-2466690339. During the process I found it somewhat difficult to test my changes:
1. It's not possible to select a subset of tests to run with `cake test`.
  - *Some of them fail on main for example--I'll fix that in a separate change, but it's why I wanted to select specific tests.*
2. It takes a while to build the parser, and it's not easy to distinguish from a hang.
3. There's no mechanism to select what output should be logged when performing a console task with `cake`.

# Solution
- Add `-f/--file`, `-d/--desc` flags to select which tests to run using regex.
  - There are also `--negFile` and `--negDesc` flags to filter out matches.
- Add `-l/--level` flag to select which console outputs to display.
  - This requires subclassing `console.Console`, which requires some hijinx.
- Using the new log levels to make it ignorable, add some console output for the parser build process:
```shell
# By default, we still print nothing:
> cake build
# With 'info', we print more, including parser build statistics.
> cake -l info build
parser created (206 symbols, 107 terminals, 406 productions)
loading grammar: 11127.603219999999 ms
parser generation: 27.266960000000836 ms
full parser build time: 11154.87018 ms
# This runs tests, ignoring ones that are currently failing:
> cake --negFile sourcemap --negDesc 'assert keyword' --negDesc 'dynamic import assertion' --negDesc 4558 --negDesc 4418 test
test '#4418: stack traces for compiled files reference the correct line number' was filtered out by patterns
test '#4418: stack traces for compiled strings reference the correct line number' was filtered out by patterns
test '#4558: compiling a string inside a script doesn’t screw up stack trace line number' was filtered out by patterns
test 'dynamic import assertion' was filtered out by patterns
test 'assert keyword' was filtered out by patterns
file 'sourcemap.coffee' was filtered out by patterns
passed 1460 tests in 4.51 seconds 
# At level 'info', each passing test is also printed out:
> cake -l info --negFile sourcemap --negDesc 'assert keyword' --negDesc 'dynamic import assertion' --negDesc 4558 --negDesc 4418 test
# (lots of output omitted...)
passed: tagged template literal with backticks in test/tagged_template_literals.coffee
passed: tagged template literal with escaped backticks in test/tagged_template_literals.coffee
passed: tagged template literal with unnecessarily escaped backticks in test/tagged_template_literals.coffee
passed: tagged template literal with ES interpolation in test/tagged_template_literals.coffee
passed: tagged template literal with both ES and CoffeeScript interpolation in test/tagged_template_literals.coffee
passed: tagged template literal with escaped ES interpolation in test/tagged_template_literals.coffee
passed: tagged template literal with unnecessarily escaped ES interpolation in test/tagged_template_literals.coffee
passed: tagged template literal special escaping in test/tagged_template_literals.coffee
passed: #4467: tagged template literal call recognized as a callable function in test/tagged_template_literals.coffee
passed 1460 tests in 4.64 seconds 
```


cosmicexplorer:
Proposed a small and focused fix to jison at https://github.com/zaach/jison/pull/408 which improves parser build time by 3x!

PR #5476, opened 2024-11-24
make jison compiles go faster

cosmicexplorer:
# Problem
Jison compiles `parser.js` very slowly:
```shell
> time cake build:parser
cake build:parser  11.65s user 0.24s system 107% cpu 11.044 total
```

Since the parser is also compiled whenever we run `cake build`, this means any change to the compiler takes ~12 seconds to reload (or almost 30 seconds if you run `cake build` twice!).

# Solution
- I optimized one method in Jison: zaach/jison#408.

# Result
The generated parser seems to be byte-for-byte the same, while being generated much faster:
```shell
> time cake build:parser
cake build:parser  4.63s user 0.49s system 123% cpu 4.147 total
```

We can go even faster (https://github.com/cosmicexplorer/jison/tree/performance-work-for-coffeescript), but the additional changes in that branch are quite extensive, and don't produce a significant performance improvement. This change is easy to audit and does most of the work.

# TODO
- [ ] The `package.json` entry points to my own github user repository. It's pinned to a commit, so it can't change, but if we're going to use a git reference, it should probably be owned by a maintainer of the repo. 
  - We could also vendor Jison (since it doesn't seem to be actively maintained).


shreeve:
Take a look at the solar.coffee parser generator in #5495

I originally was based on Jison, but totally rewrote it from scratch.

I now generate the parser.js file for CoffeeScript in less than 100ms, down from 12.5 seconds.

cosmicexplorer:
The reason I spent the time to fix jison instead of rewriting it was the hope that it would be easier for maintainers to review a change. Rewriting it is the easy part.

PR #5479, opened 2024-11-26
Update deps to remove warning messages

cosmicexplorer:
Split out from #5478 so that PR can focus solely on fixing CI.

# Problem
Currently we get the following output when running `cake test`:
```shell
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
``` 

Additionally, executing `npm` commands currently produces the following output:
```shell
9 vulnerabilities (7 high, 2 critical)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force
```

# Solution
- Execute `npx update-browserslist-db@latest`.
  - This is what it tells you to run after executing `npx browserslist@latest --update-db` as in the warning message.
- Execute `npm audit fix --force`.

# Result
This changes our `puppeter` and `webpack` constraints in `package.json` to `^` instead of `~`. We may instead prefer to stick with `~`.

There is also still an additional message produced when executing e.g. `npm ci`:
```shell
npm warn deprecated nomnom@1.5.2: Package no longer supported. Contact support@npmjs.com for more info.
```

This does not seem to be in our direct dependencies, but would probably be good to resolve at some point in the future.


cosmicexplorer:
Looks like `nomnom` is from `jison`: https://github.com/zaach/jison/pull/367. Is there a point of contact for the jison project or is it effectively unmaintained?

PR #5501, opened 2026-06-03
Fixes #5480: clear  when re-inverting an 'in' Op

### LINKED ISSUE #5480: Bug: using "not in" operator in switch works as "in" instead
```coffee
switch
  when thing in arr
    console.log 'this'
  when thing not in arr
    console.log 'that'
```

### Expected Behavior
The above should probably compile to something along the lines of this:
```js
var indexOf = [].indexOf;

switch (false) {
  case indexOf.call(arr, thing) < 0:
    console.log('this');
    break;
  case indexOf.call(arr, thing) >= 0:
    console.log('that');
}
``` 

### Current Behavior
It however compiles to this:
```js
var indexOf = [].indexOf;

switch (false) {
  case indexOf.call(arr, thing) < 0:
    console.log('this');
    break;
  case indexOf.call(arr, thing) < 0:
    console.log('that');
}
```
### Environment

* CoffeeScript version: 2.7.0
* Node.js version: some version of 14 for legacy reasons
But the thing happens even on the "try coffeescript" page  on the website too

### Other context
This does not happen in coffee v1 nor in civet with coffee compat

rdeforest:
fixes #5480

Please note that tests other than the one I added are failing and were already failing. I will address those in a future PR if nobody else gets to it first.

In `Op::invert` in `nodes.coffee`, there was no reverse mechanism, so `op.invert().invert() !== op`. I added a check for already being inverted and removed the inversion flag. It feels like a bad code smell to use a string as a boolean, but I'm not heart to change the code style. I made the simplest, most obvious edit I could. Code style is probably a v3 thing, if such an effort were ever to come about.

This fix seems to only be relevant to switch case expressions, so I used that in the test I added. The test checks all eight permutations of subject included, value is actually in the list and expression is inverted. The bug was not triggered when the switch expression included a subject because the case expressions don't get inverted then. I didn't quite understand the logic behind inverting the case expressions, but since that part was working I didn't touch it. I separated each case into its own switch expression so that the fallthrough on match wouldn't hide any permutations.

The only changed files are nodes.coffee for the four-line fix, nodes.js after the rebuild and test/operations.coffee to add the test for this bug.

### DIFF
--- a/test/operators.coffee
+++ b/test/operators.coffee
@@ -466,11 +466,11 @@ test "#5480: 'when a not in b'", ->
   switch
     when needle not in needleStack then ok false
 
-  switch "with subject"
+  switch "with subject" and true
     when needle     in    hayStack then ok false
-  switch "with subject"
+  switch "with subject" and true
     when needle not in    hayStack then ok true
-  switch "with subject"
+  switch "with subject" and true
     when needle     in needleStack then ok true
-  switch "with subject"
+  switch "with subject" and true
     when needle not in needleStack then ok false
--- a/test/operators.coffee
+++ b/test/operators.coffee
@@ -451,3 +451,26 @@ test "'new' target", ->
 
   # classes must be called with `new`. In this case `new` applies to `get` only
   throws -> new get()()
+
+test "#5480: 'when a not in b'", ->
+  needle      = "needle"
+  hayStack    = []
+  needleStack = [needle]
+
+  switch
+    when needle     in    hayStack then ok false
+  switch
+    when needle not in    hayStack then ok true
+  switch
+    when needle     in needleStack then ok true
+  switch
+    when needle not in needleStack then ok false
+
+  switch "with subject"
+    when needle     in    hayStack then ok false
+  switch "with subject"
+    when needle not in    hayStack then ok true
+  switch "with subject"
+    when needle     in needleStack then ok true
+  switch "with subject"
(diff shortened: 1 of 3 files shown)


rdeforest:
The build error is "Error: Unable to find Node version '12.x' for platform darwin and architecture arm64." I'm guessing the rest were canceled because the first failed. I know we officially support Node back to v6, but the [NodeJS project doesn't support anything earliert than v22](https://nodejs.org/en/about/previous-releases). I know earlier releases are still available because NVM was able to find them when I tried to do my own testing.

Do I need to dig into how to obtain v12 for MacOS in the GitHub actions side of things? I'm pretty sure I don't have the access needed to do anything about fixing them, so I'd just be researching to help out.

test/operators.coffee
```diff
@@ -451,3 +451,26 @@ test "'new' target", ->
...
+    when needle not in needleStack then ok false
+
+  switch "with subject"
+    when needle     in    hayStack then ok false
+  switch "with subject"
+    when needle not in    hayStack then ok true
+  switch "with subject"
+    when needle     in needleStack then ok true
+  switch "with subject"
+    when needle not in needleStack then ok false
```
Inve1951: these switch constructs always hit the default branch because the string is never a boolean

rdeforest: Good catch! I updated the tests by adding "and true" after all the strings.

